PR #6, opened 2015-08-24
Make RecyclerViewDemoActivity.ChatHolder public 

htcbug:
Fix the bug that ChatHolder is not public so it could not be accessed, which causes crash.

@puf Can you review? Thanks!


htcbug:
I signed it!

PR #7, opened 2015-08-25
Add pull-to-loadmore more to RecyclerViewDemoActivity.

htcbug:
@puf 

Frank, please review this change which adds a pull-to-loadmore feature to RecyclerViewDemoActivity. It's implemented by updating the query passed in when user scrolls up.

The key change is in FirebaseArray.java. In the onChildAdded() method, I will first test whether the snapshot is already loaded. If it's already loaded we won't load it twice.


puf:
Hi there friendly Googlebot. I did indeed author these commits. :-)

PR #26, opened 2015-11-09
Scroll to load more

marcorei:
@puf as promised, my approach at adding scroll-to-load-more functionality to this lib. In my opinion it is still problematic that in order to add headers and/or footers you have to override onCreateViewHolder. Please let me know what you think.

### DIFF
--- a/library/src/main/java/com/firebase/ui/FirebaseArray.java
+++ b/library/src/main/java/com/firebase/ui/FirebaseArray.java
@@ -28,32 +28,73 @@
 
 package com.firebase.ui;
 
+import android.support.annotation.NonNull;
+
 import com.firebase.client.*;
 
 import java.util.ArrayList;
-import java.util.Collections;
 
 /**
  * This class implements an array-like collection on top of a Firebase location.
  */
-class FirebaseArray implements ChildEventListener {
+class FirebaseArray implements
+        ChildEventListener,
+        ValueEventListener
+{
     public interface OnChangedListener {
-        enum EventType { Added, Changed, Removed, Moved }
+        enum EventType { Added, Changed, Removed, Moved, Reset }
         void onChanged(EventType type, int index, int oldIndex);
     }
 
-    private Query mQuery;
-    private OnChangedListener mListener;
-    private ArrayList<DataSnapshot> mSnapshots;
+    public interface OnErrorListener {
+        void onError(FirebaseError firebaseError);
+    }
 
-    public FirebaseArray(Query ref) {
-        mQuery = ref;
+    public interface OnSyncStatusChangedListener {
+        enum EventType { UnSynced, Synced }
+        void onChanged(EventType type);
+    }
+
+    private Query mOriginalQuery;
+    private Query mQuery;
+    private OnChangedListener mOnChangedListener;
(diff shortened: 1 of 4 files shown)


marcorei:
@googlebot done.

puf:
Hey @marcorei , thanks for the awesome PR. Agreed that the API is still un-ideal, but this is definitely the cleanest infinite scroll I've seen on FirebaseUI. I have some comments, which I'll add in the code. 

I also don't think we're quite ready to add this functionality to FirebaseUI at this stage, so will keep this PR open for now.

library/src/main/java/com/firebase/ui/FirebaseArray.java
```diff
@@ -64,6 +107,24 @@ public DataSnapshot getItem(int index) {
...
+        }
+        else if(mOrderASC == true){
+            mQuery = mOriginalQuery.limitToFirst(mCurrentSize);
+        }
+        else {
+            mQuery = mOriginalQuery.limitToLast(mCurrentSize);
+        }
+        setSyncing(true);
+        mQuery.addChildEventListener(this);
+        mQuery.addListenerForSingleValueEvent(this);
```
puf: As you scroll, this keeps reloading the pervious items for each next page. This is a bandwidth concern, which unfortunately I think can only be solved by loading each "page" separately and then merging them into a single array "under the hood".

marcorei:
Thank you for your feedback! Are you referring to the stage of the library or the stage of the PR?

library/src/main/java/com/firebase/ui/FirebaseArray.java
marcorei: I agree that the simplicity of this approach comes at the cost of bandwidth. Loading pages separately could result in a much more complex system (Managing listeners in order to receive updates for previously loaded pages or using startAt() in conjunction with moveable children in Queries could become issues). I'll have to think about how to solve this properly!

puf:
Hey Marco. I meant the stage of the library, but the PR indeed would also need some iteration. Plus: we'd need to implement it in FirebaseUI-iOS too, which becomes a lot easier once we figure out the best API and implementation here first.

samtstern:
@marcorei this is an old PR and the conversation seems to have died out so I am going to close it.  We appreciate your contribution here and it's very likely that a feature like this could be useful in the future.  If you're still interested, please open an issue and we can discuss.

harshadura:
@puff @samtstern : wouldn't it be possible to release an external release of the new API after merging this code. IMO It would be very useful for amateur programmers who are using firebase and face this load-more issue.

PS: kudos @marcorei for the PR.

puf:
We will only release infinite scroll/pagination support in FirebaseUI when
it's available on both Android and iOS.

But since this is an open-source project, you can fork the library and add
it. This old PR may serve as a great basis for that.

harshadura:
Hello puff, like when you guys planning to support pagination for android.
Thanks

puf:
No timelines. If you need it, I recommend forking the repo and rolling your
own.

harshadura:
@ThunderRoid : Could you tell me how you have used this PR to enable scroll to load more feature. Thanks!

Issue #34, opened 2015-12-02
Login with google hangs with spinner


mihai-chiorean:
Tried this in the emulator. Used firebase UI. Fired up the dialog, clicked the login with google button, it just hangs on a spinner. Didn't hit a breakpoint in either onXXX handlers. Just hangs. Can't find any logs from the firebase library either.

puf:
I'm having the same problem in the emulator. It works fine on a physical device, but my main activity is showing a blank screen after adding Google auth. I'll let you know if we discover what causes this problem.

mihai-chiorean:
thank you!!

mihai-chiorean:
An update here: Tried on my phone. It worked here, but as soon as I selected a google account, I got this:
`failed: User closed login dialog.`

puf:
I'll consider that progress. :-)   I've tested on various emulators, but all are failing with similar symptoms. I would've given up, except that the sample app that comes with FirebaseUI-Android opens the login dialog without problems.

I'll give it a try on a physical device tomorrow. I know multiple people tested that flow, but that may all have been with the sample app.

frto:
Hi, I tested RecyclerViewDemoActivity on a Samsung S4 running Android 5.0.1. Google login and hereafter reading from my firebase app works fine the first time. Logout and again login results in a hanging spinner. Restarting the app and we can again login/logout for one time.

puf:
OK. Just tried this on some more "devices":

1. on the Android emulator with a Nexus 5X image I always get a blank popup, even with just the password provider. Interestingly enough the controls seem to be in there, as I can click and tab. So it's likely a rendering issue.
2. on the Genymotion emulator with a Nexus 5X image I get a popup for the password provider. But when I enable the google provider, I get an exception. Most likely my device is missing Google Play Services.
3. on a physical Nexus 7, I get the popup but am having some other issues. I'll keep digging.

I'm going to re-assign this issue to @abehaskins for fixing the layout issue. Let's treat the others as separate issues. That includes the one you just submitted a PR for @frto .

frto:
Hi all, tested again on Samsung S4 and issue is indeed solved now. Thanks for this!
I removed my pull request cause the proper solution is indeed commit 625cc85.

puf:
Thanks for testing Frederik! The fix will be in the next release. We're
still working on some more fixes that need to go in there.

troy21688:
[...]

PR #37, opened 2015-12-06
Solve Google logout/login issue.

frto:
Solves issue #34 

Logging out with Google account causes the mGoogleApiClient to be disconnected.
But this caused a blocking spinner, and reconnect of mGoogleApiClient  doesnt happen anymore.

After testing it seems that starting the signInIntent is not a problem when the mGoogleApiClient is disconnected, solving issue #34


frto:
I signed it!

PR #44, opened 2015-12-20
Add link to Android Config file generation

jhoon:
For seasoned developers, knowing where to get the `google-services.json` is basic knowledge. 
However, for people who are just getting started, it is simpler to send them there with a link, 
just like the [FirebaseUI for iOS does]
(https://github.com/firebase/firebaseui-ios#firebasegoogleauthprovider).

cc. @puf @mcdonamp

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,8 @@ If you're using Google authentication, add the following to your `<application>`
  ```
 
 **Note:** If you're using Google Sign-in you'll also need to ensure that your `google-services.json` file is created
-and placed in your app folder.
+and placed in your app folder. To get one for your project, you can do so [on this link]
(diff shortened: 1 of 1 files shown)


puf:
Great idea!

We're working on not requiring google-services.json for 0.3.1. But if that fails, we'll definitely add a link.

Thanks!

samtstern:
@jhoon this is an old PR and has merge conflicts so I am going to close.  With the way the library is structured now I believe it is clear that users need to get a `google-services.json` from the Firebase console.  However if you want to submit a PR to the README now, I'd be happy to look at it.

Thanks for taking the time to contribute!

Issue #20, opened 2015-10-15
Allow FirebaseListAdapter class to iterate object keys


ShamariFeaster:
I have a scenario where I want to be able to quickly check for key existence and because of that it was desirable to use object keys as the value (like a hash table).Essentially this means I needed to be able to extend `FirebaseListAdapter` using a `mModelClass` class of type `String`.

To support this functionality, I modified `FirebaseListAdapter` with a function `parseNode`

```
private T parseNode(DataSnapshot dataSnapshot){
                String key = dataSnapshot.getKey();
                T newModel = null;
                if(!(FirebaseListAdapter.this.mModelClass == String.class)){
                    newModel = dataSnapshot.getValue(FirebaseListAdapter.this.mModelClass);
                }else{
                    newModel = (T) key;
                }
                return newModel;
            }
```

Now in all of the `ChildEventListener` callbacks the `model` variable gets assigned the output of `parseNode` and the adapter is able to handle iterating over object keys. I just figured it's a pretty minor change to the class and it helped with my case so maybe it would be useful for other people.

ShamariFeaster:
thanks!

Issue #9, opened 2015-08-26
Add support for DataSnapshot as valid a model object


mcdonamp:
Currently we can't return a DataSnapshot in either Adapter, but we should be able to (for feature parity between clients as well as because some people already have custom DataSnapshot to model object conversion in place).

PR #47, opened 2015-12-30
Puf kill switches

puf:
Fixes https://github.com/firebase/FirebaseUI-Android/issues/20. But
also changes https://github.com/firebase/FirebaseUI-Android/issues/9 to
needing to override both `populdateView()` and `parseSnapshot()`.


library/src/main/java/com/firebase/ui/auth/core/AuthProviderType.java
```diff
@@ -0,0 +1,61 @@
...
+import com.firebase.ui.R;
+import com.firebase.ui.auth.facebook.FacebookAuthProvider;
+import com.firebase.ui.auth.google.GoogleAuthProvider;
+import com.firebase.ui.auth.password.PasswordAuthProvider;
+import com.firebase.ui.auth.twitter.TwitterAuthProvider;
+
+import java.lang.reflect.InvocationTargetException;
+
+public enum AuthProviderType {
+    GOOGLE  ("google",   "com.firebase.ui.auth.google.GoogleAuthProvider",     R.id.google_button),
```
abehaskins: Can we not do just "google.GoogleAuthProvider" and assume the "com.firebase.ui.auth" in the constructor? I feel like that might read better

library/src/main/java/com/firebase/ui/auth/core/AuthProviderType.java
```diff
@@ -0,0 +1,61 @@
...
+    public String getName() {
+        return mName;
+    }
+    public int getButtonId() {
+        return mButtonId;
+    }
+
+    public FirebaseAuthProvider createProvider(Context context, Firebase ref, TokenAuthHandler handler) {
+        try {
+            Class<? extends FirebaseAuthProvider> clazz = (Class<? extends FirebaseAuthProvider>) Class.forName(mProviderName);
```
abehaskins: `clazz` makes me kinda cringe, is this something people do? In Javascript we'd save class_ but I have no idea what the standard in other languages is.

library/src/main/java/com/firebase/ui/auth/google/GoogleAuthProvider.java
```diff
@@ -93,7 +82,7 @@ public void onResult(Status status) {
     public void login() {
         if (mGoogleApiClient.isConnected()) {
             Intent signInIntent = Auth.GoogleSignInApi.getSignInIntent(mGoogleApiClient);
-            mActivity.startActivityForResult(signInIntent, GoogleActions.SIGN_IN);
+            ((Activity)getContext()).startActivityForResult(signInIntent, GoogleActions.SIGN_IN);
```
abehaskins: Curious about the reasoning for the change here, I'm assuming this functioned before, so why the change?

abehaskins:
Overall lgtm - thanks for cleaning up a bit of my mess :)

library/src/main/java/com/firebase/ui/auth/core/AuthProviderType.java
puf: Fixed

library/src/main/java/com/firebase/ui/auth/core/AuthProviderType.java
puf: There is definitely a [convention to call these clazz](http://stackoverflow.com/questions/2529974/why-do-java-programmers-like-to-name-a-variable-clazz), but I'm open to any other name.

library/src/main/java/com/firebase/ui/auth/google/GoogleAuthProvider.java
puf: I pulled the mActivity member into the base class and gave that a
getContext() accessor method. That might have been in a previous refactor
already, with this change just getting rid of the mActivity member.

I am a bit torn on getContext(). Since we seem to most often want to use it
as an Activity, we might just keep it as an Activity and prevent the need
to downcast.

PR #61, opened 2016-01-23
Update Readme Enable Authentication Providers 

joseph-ortiz:
Enable Authentication Providers

In the code example to Enable Authentication Providers.
instead of calling AuthProvider.FACEBOOK, I had to use SocialProvider.facebook
AuthProvider was giving me a syntax error when trying to use AuthProvier


puf:
That probably means that you're using release 0.3.0. We just released 0.3.1 last week, which now uses uppercase constants for these values.

PR #67, opened 2016-01-30
use support DialogFragment and getSupportFragmentManager

Goddchen:
Just to be more compatible...


Goddchen:
Already signed a CLA, updated it with my Github name.

mickamy:
Hi, @Goddchen

 I think it is a nice contribute but maybe we should change AlertDialog into v7 one too.
There is no way to avoid using SupportLibrary's class unless some specific reasons exist.
And it is only the change of import sentence!
So, why don't we?

Actually I was doing the same thing, but I think you are the one to get an honor for this contribution. Would you update your PR or I will.
Thanks

Goddchen:
@mickamy you're absolutely right, changes added!

fanky10:
just curious: why is this not merged yet?

samtstern:
@Goddchen thanks for taking the time to contribute this.  As you probably can tell this is targeting an old and conflicting version of FBUI and the new library should not have this issue.  I am going to close this PR, if you have similar changes you'd like to make to the new code I'd be happy to take a look at them.

PR #73, opened 2016-02-14
Changed setListAdapter to setAdapter

PeEllAvaj:
setListAdapter only exists for ListActivities, but this example seems to be using ListViews


PeEllAvaj:
CC @puf Tiny documentation correction/improvement.

puf:
Good catch. Merging!

PR #76, opened 2016-02-15
Make sure that the click listener gets attached to the login buttons

UD4Y:
Login button in password provider was not responding to clicks


puf:
Good catch. The broken click handling was unreleased so far, so great to get a fix before that. Thanks!

Issue #69, opened 2016-02-03
Is there method available to request facebook email 


Shajeel-Afzal:
Hi. 

I am storing users information against their Email on Firebase. I am using this library to show Firebase Login Dialog. 

However in case of Facebook i am not getting user email. I want to know if there is any helper method available in the SDK that can set permissions on the Facebook Login. If not then will it be available in future?

Thanks.

abehaskins:
Not currently, no! 

However you should be able to easily modify the [FacebookAuthProvider](https://github.com/firebase/FirebaseUI-Android/blob/master/library/src/main/java/com/firebase/ui/auth/facebook/FacebookAuthProvider.java) class to request email scope and retrieve it that way while we look for a better solution.

samtstern:
FirebaseUI now gets email from Facebook, closing this issue.  Please re-open if the new solution is not working for you.

PR #81, opened 2016-02-18
Fix issue #69

UD4Y:
Clients will be able to request for email as part of Facebook authentication by adding a boolean `"com.firebase.ui.auth.facebook.RequestEmail"` metadata in AndroidManifest.

`<meta-data android:name="com.firebase.ui.auth.facebook.RequestEmail" android:value="true" />`


UD4Y:
Not everyone will want to request for email as part of Facebook authentication, so wanted it to be configurable. Didn't find a cleaner way than metadata flag to do this.

abehaskins:
Hey @UD4Y, thanks a ton for the PR! 

I think you're definitely in the right direction with adding this. I'm worried it's generic enough to support all use-cases, I think we should add a way to add any scope for any OAuth provider and not just add Facebook email specifically.

I'm all ears for ideas on an implementation that will is a little more open-ended.

Thanks again!
abe

UD4Y:
Agreed, thought of that limitation too, but was not sure if the Firebase team wanted that kind of functionality. Will work a better approach.

PR #83, opened 2016-02-20
Firebase/pr/login config

UD4Y:
Will help fixing issue #69


samtstern:
@UD4Y this PR is targeting a very old version of the code and would not be mergeable at this point.  Additionally I think the issue you mentioned (#69) is no longer an issue in the new FirebaseUI as we do request email from Facebook.

Closing this PR, thank you for taking the time to contribute!

PR #88, opened 2016-03-01
Added instructions for RecyclerAdapter cleanup & corrected method names

roughike:
Somebody that wants to jump right in and use the `RecyclerAdapter`, might just skip through the `ListAdapter` example and forget to call `cleanup()` on it.

Also, the method name seemed to be incorrect in the instructions but right in the code: instead of `release()` the correct method is `cleanup()` just like above.

### DIFF
--- a/README.md
+++ b/README.md
@@ -352,7 +352,7 @@ correct `TextView` controls from the `view`. The code is a bit verbose, but hey.
 
 ### Clean up When the Activity is Destroyed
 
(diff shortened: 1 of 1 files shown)


puf:
Thanks for the PR and good catches. I'm not sure how that `release()` snuck in there, because the method has always been called `cleanup()`. Well.... fixed now. Thanks again!

roughike:
@puf 
No probs, man! Thank you for the awesome product!

PR #89, opened 2016-03-01
Exposing FirebaseArray

rmarau:
Making it accessible from the lib.

### DIFF
--- a/library/src/main/java/com/firebase/ui/FirebaseArray.java
+++ b/library/src/main/java/com/firebase/ui/FirebaseArray.java
@@ -38,7 +38,7 @@ import java.util.ArrayList;
 /**
  * This class implements an array-like collection on top of a Firebase location.
  */
-class FirebaseArray implements ChildEventListener {
+public class FirebaseArray implements ChildEventListener {
     public interface OnChangedListener {
         enum EventType { Added, Changed, Removed, Moved }
         void onChanged(EventType type, int index, int oldIndex);


puf:
Thanks for the PR.

When we designed the public API for FirebaseUI, we decided against exposing
FirebaseArray publicly. The expectation was that many people would start
treating this as a normal array/collection on Firebase and be surprised by
the realtime updates to it.

So we are intentionally *not* exposing the FirebaseArray. You're free to
take the FirebaseArray from this project and use it as they see fit of
course, but it'll stay out of the public API for the moment. Let us know
what you use it for though, we'd love to learn more about that so that we
can potentially re-evaluate our decision in the future.

rmarau:
Thanks for replying,.

I understand your worries. 
IMO, it could be tempting to use it as a normal array/collection if you had a FirebaseArray<T>. As it is, we still have to parse DataSnapshot to our Class and there's not a setItem call. 

I'm using it because of the logic with the indexes. It's much easier to handle FirebaseArray .OnChangedListener than ChildEventListener. For me it's a convenient way to mirror and feed my own array. 
I know I'm duplicating data. By making FirebaseArray public I'm trying to not duplicate code in the project as well.

Thanks anyway

crrobinson14:
@puf Does the reasoning behind this decision still apply in light of the fact that `FirebaseUI-iOS` exposes `FUIArray`? It's an incredibly useful helper class. In particular, right now it's a lot of manual work to write a FragmentPagerAdapter to Firebase that the FirebaseArray would do.

PR #109, opened 2016-04-25
Signup feature

runia1:
I have finished the signup feature I promised.  Please review and merge if happy!  If you have any questions you can contact me anytime: runiam@sou.edu


samtstern:
@runia1 I am cleaning up some old Pull Requests here.  Looks like this one targets the old SDK, which is so different from the new one that I'm not sure this PR could be saved without a huge amount of work.

Thank you for your contribution, we really appreciate when people take the time to send PRs.  if you want to pursue this going forward, start the conversation in an issue and we can decide how to proceed.

PR #140, opened 2016-05-24
Travis CI Build Status

samtstern:
Just a basic sanity check to see if it's possible to build all library artifacts.  Also centralized the build tools version so that travis won't fail for that reason.

@puf or @iainmcgin please squash and merge.


puf:
lgtm (shipit)

PR #149, opened 2016-05-27
Fix for: Wrong result code after password reset flow

amandle:
Cancelling the password reset flow, or sending the password reset email no longer returns with the incorrect RESULT_OK status.  It now returns the user to the sign in form.


samtstern:
Merging.

PR #151, opened 2016-05-28
Update Android Gradle plug-in to 2.1.0

friederbluemle:


samtstern:
Thanks again Frieder!

PR #152, opened 2016-05-28
Check mEmail for empty String to prevent "credential identifier cannot be empty" error

### LINKED ISSUE #141: Facebook login error
I have gotten this error when using the new FirebaseUI with Facebook login:

```
java.lang.IllegalArgumentException: credential identifier cannot be empty
at com.firebase.ui.auth.ui.account_link.SaveCredentialsActivity.onConnected(SaveCredentialsActivity.java:120)
```

After inserting credentials and successfully logging in the app crashes and this is the error.

Do you know what might be going wrong?

stba42:
made code more robust, should fix #141

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/ui/account_link/SaveCredentialsActivity.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/account_link/SaveCredentialsActivity.java
@@ -89,7 +89,7 @@ public class SaveCredentialsActivity extends AppCompatBase
 
     @Override
     public void onConnected(@Nullable Bundle bundle) {
-        if (mEmail == null) {
+        if (mEmail == null || mEmail.equals("")) {
             Log.e(TAG, "Unable to save null credential!");
             finish(RESULT_FIRST_USER, getIntent());
             return;


auth/src/main/java/com/firebase/ui/auth/ui/account_link/SaveCredentialsActivity.java
```diff
@@ -89,7 +89,7 @@ protected void onStart() {
 
     @Override
     public void onConnected(@Nullable Bundle bundle) {
-        if (mEmail == null) {
+        if (mEmail == null || mEmail.equals("")) {
```
samtstern: This line can be replaced with `TextUtils.isEmpty(mEmail)`

samtstern:
@iainmcgin please review.

auth/src/main/java/com/firebase/ui/auth/ui/account_link/SaveCredentialsActivity.java
iainmcgin: We should investigate under what circumstances a null / empty email address ever arrives here. I agree with sam however that `if (TextUtils.isEmpty(mEmail)` is a more descriptive check here.

abidsoft: Empty email issue happens due to privacy setting of some users, where you get name and profile image but email is null

SUPERCILEX:
This PR can be closed now: fixed in #357.

samtstern:
#357 merged.

PR #163, opened 2016-06-03
resolve lint error for register_email_layout.xml

wuman:
Currently lint fails because an extra xmlns:android declaration is found in the register_email_layout.xml.  This pull request removes the extra declaration.


samtstern:
@wuman nice catch, thanks!

PR #167, opened 2016-06-03
Use the account selected from the hint picker when signing in with Google

amandle:


auth/src/main/java/com/firebase/ui/auth/provider/GoogleProvider.java
```diff
@@ -40,15 +41,19 @@
...
         GoogleSignInOptions googleSignInOptions;
 
-        googleSignInOptions = new GoogleSignInOptions.Builder(GoogleSignInOptions.DEFAULT_SIGN_IN)
+        GoogleSignInOptions.Builder builder = new GoogleSignInOptions.Builder(GoogleSignInOptions
+                .DEFAULT_SIGN_IN)
                 .requestEmail()
-                .requestIdToken(mClientId)
-                .build();
+                .requestIdToken(mClientId);
+        if (email != null) {
```
samtstern: We had another issue where email was coming through as "".  Maybe use TextUtils.isEmpty here.

samtstern:
One small comment.  Other than that LGTM and feel free to merge when you fix that/

PR #168, opened 2016-06-04
use a consistent and latest version for support library

wuman:
Support libraries are using inconsistent versions all over the place, which makes it difficult for applications to manage their dependencies.  This pull request makes all support library versions consistent.


samtstern:
@wuman this is a good improvement, thanks for the contribution!

PR #175, opened 2016-06-07
add fix to java 8 break in gradle build

jirungaray:


build.gradle
```diff
@@ -27,6 +27,15 @@ allprojects {
  */
 allprojects { project ->
 
+    //Skip Javadoc generation for Java 1.8 as it breaks build
+    if (JavaVersion.current().isJava8Compatible()) {
+        allprojects {
```
samtstern: Since this references allprojects, I think it should be outside of the outer allprojects block rather than nested.

jirungaray:
I signed it!

build.gradle
jirungaray: Agreed. Any convention/position in  particular this check should be placed?

samtstern:
@jirungaray thanks, merging!

PR #179, opened 2016-06-08
debug log not error log

wuman:
A log is falsely displayed as an error even though it is just a debugging log entry.  This pull request corrects it.


samtstern:
Nice catch, thanks!

PR #191, opened 2016-06-18
Use resource strings for login buttons.

Gloix:
Login buttons weren't translatable, so I made them point to their corresponding resource strings.


Gloix:
I signed it!

samtstern:
@Gloix sorry for the very, very long delay here but this change LGTM and I will merge it.  This should make the next release.

PR #194, opened 2016-06-19
added a FragmentStatePagerAdapter class

dabele91:
Added a FragmentStatePagerAdapter class, which uses the FirebaseArray class.


samtstern:
@dabele91 first of all thank you for this incredibly complete pull request (docs and everything!).  We love to see this kind of involvement from the community.

While I totally see the utility of this feature and I think your implementation is good, I have to close this Pull Request as it's out of scope for this library.  There are many adapters in Android that could be implemented on top of FirebaseArray (Fragments, Spinners, ViewPager, etc) but we'd like to keep the surface of the library small and only have the most popular features so that we can effectively support the library and keep it updated in the future.

For this feature I'd suggest making your own library that has a dependency on Firebase UI.  If you've never published an Android library before it's very easy these days with Bintray and jCenter!

Also in the future if you plan to implement a new feature it's always good to submit an Issue first to see what the reaction to a Pull Request would be before putting in all the hard work to implement.

dabele91:
@samtstern thanks for the great feedback :-)
It wasn't much work because i've needed the class anyway.
Next time i'll create an issue, to check, if the community needs such an class.

I'm thinking about it, to publish the class in an separate library the next days.
Thanks for the tips :-)

PR #201, opened 2016-06-28
upgrade gms & facebook sdk

b95505017:


wscheffer:
Please can we have this

samtstern:
@b95505017 thank you for your contribution.  I have just released version `0.4.2` which upgrades the underlying Google/Firebase dependencies to `9.2.0`.  As for Facebook, we will have to do some more testing on the new SDK version so I am going to close this pull request and let some of the FirebaseUI testers evaluate that switch.

PR #205, opened 2016-06-29
Have FirebaseArray accessible from subclasses

gufux:
Problem:
Currently, I think, It's not possible to remove items from the FirebaseListAdapter without removing it also on firebase db. 

Proposal:
I think (not sure) that having access directly to FirebaseArray would allow to create in the subclass custom adapter a method to remove items **only** on the adpater and not in db.

### DIFF
--- a/database/src/main/java/com/firebase/ui/database/FirebaseListAdapter.java
+++ b/database/src/main/java/com/firebase/ui/database/FirebaseListAdapter.java
@@ -52,7 +52,7 @@ public abstract class FirebaseListAdapter<T> extends BaseAdapter {
     private final Class<T> mModelClass;
     protected int mLayout;
     protected Activity mActivity;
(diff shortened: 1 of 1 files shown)


gufux:
I signed it

gufux:
I signed it

2016-06-29 18:15 GMT+02:00 googlebot <notifications@github.com>:

puf:
Thanks for the PR. We had some good reasons behind not making the FirebaseArray accessible to sub-classes. 

Can you open an issue (that's in general recommended before opening a PR) and give some more examples of how you'd use the array in your subclasses?

gufux:
Hi @puf 
I created the issue  [FirebaseArray in FirebaseListAdapter is not accessible from subclasses #206 ](https://github.com/firebase/FirebaseUI-Android/issues/206)

Issue #197, opened 2016-06-23
Facebook permissions need to be dynamic "public_profile", "email" are not enough


Ibrahimhassan1:
The latest Facebook Graph API doesn't return the list of user_friends who are playing the game, it used to return them with current library permissions ("public_profile", "email") but this is returning empty array now. So there should to be a way to add more permissions without touching the library core files.

samtstern:
@Ibrahimhassan1 thanks for the suggestion, I agree this would be useful and we'll look into it.

talosdev:
I needed custom permissions in my app too, so I created a fork and made the facebook permissions configurable.
I have created a pull request, my first one here, so let me know if there's something missing.

@puf 
@mcdonamp

talosdev:
Created a new pull request, correcting the error with the CLA.

samtstern:
This was fixed in #224.  Can now be  overridden in a string array resource.

tuxtux59:
Hi,
I'm using also the facebook connect using Facebook login using FirebaseUI and the app need to list friends. Where do I can add the users permissions?

Thanks :)

SUPERCILEX:
@tuxtux59 Please read the [docs](https://github.com/firebase/FirebaseUI-Android/tree/master/auth#facebook) or create a new issue if they don't cover your use case.

tuxtux59:
Thanks I found help using the IDE autocompletion ;)

PR #224, opened 2016-07-26
Facebook permissions are taken from an array resource

talosdev:
Instead of using the hardcoded permissions "public_profile" and "email", this pull request allows client apps to define any permissions they want, as requested in issue #197.

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/provider/FacebookProvider.java
+++ b/auth/src/main/java/com/firebase/ui/auth/provider/FacebookProvider.java
@@ -77,8 +77,11 @@ public class FacebookProvider implements IDPProvider, FacebookCallback<LoginResu
         mCallbackManager = CallbackManager.Factory.create();
         LoginManager loginManager = LoginManager.getInstance();
         loginManager.registerCallback(mCallbackManager, this);
+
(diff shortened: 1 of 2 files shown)


samtstern:
@talosdev thanks for the contribution.  So for a developer to customize the permissions in this case they would have to fork the repo, right?  Or would this work if they override the string array resource in their own `res` files?

talosdev:
@samtstern devs would have to fork the repo only if you don't actually accept the pull request...
The idea of the PR is exactly what you say, that they can override the resource in their own resource files (exactly the same way that the are required to do with the `facebook_application_id` property).

samtstern:
@talosdev ok this sounds good to me.  Since we only implement features when we have parity, I do have to put this on hold until someone can implement something similar on iOS.  I will raise the issue over there now.

Thanks again!

juanlet:
Thank you very much for this, I need it for my app as well since I need additional permissions using UI and Facebook......Do you know when you will be adding this to gradle for android?. Thank you very much

samtstern:
@talosdev it looks like iOS already had this ability so we can merge it in here.  It will be rolled into the next release, thanks!

Issue #242, opened 2016-08-07
Remove duplicates strings in auth/src/main/res/values/strings.xml


peterhav:
To make the live of translators easier, I would recommend to remove duplicates from the strings.xml:

Next (2x)
Sign In (5x)
Password (2x)
Ok -> replace with android.R.string.ok

Is it ok to add translations to other languages already?

samtstern:
@peterhav which languages were you considering adding translations for?  This is something we'd have to coordinate across platforms.

However a PR to reduce duplication in string resources would be welcome if you'd like to submit it.

peterhav:
The PR has been created. I would be able to add Dutch translations for the auth lib strings.xml

samtstern:
Thanks for the PR!  As for the dutch translation, let's hold off for now.  I think we have some plans to get all translations done at once for this library (and the iOS version) and I don't want to have you put in the effort now only to see it deleted later.  We do appreciate the offer though :+1:

samtstern:
The PR has been merged so I consider this issue fixed.  Thanks again!

PR #245, opened 2016-08-09
Cleanup strings.xml: removed duplicates, Google sign in button now us…

peterhav:
Changes as discussed in issue  #242, please verify @samtstern

### DIFF
--- a/auth/src/main/res/layout/idp_button_google.xml
+++ b/auth/src/main/res/layout/idp_button_google.xml
@@ -1,6 +1,6 @@
 <Button
     xmlns:android="http://schemas.android.com/apk/res/android"
     style="@style/FirebaseUI.Button.AccountChooser.GoogleButton"
-    android:text="@string/sign_in_with_google"
+    android:text="@string/common_signin_button_text_long"
     android:id="@+id/google_button"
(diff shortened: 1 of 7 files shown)


peterhav:
2016-08-09 21:52 GMT+02:00 googlebot <notifications@github.com>:

> I signed it!

auth/src/main/res/layout/idp_button_google.xml
```diff
@@ -1,6 +1,6 @@
 <Button
     xmlns:android="http://schemas.android.com/apk/res/android"
     style="@style/FirebaseUI.Button.AccountChooser.GoogleButton"
-    android:text="@string/sign_in_with_google"
+    android:text="@string/common_signin_button_text_long"
```
samtstern: Where is this defined, in Google Play Services?  I don't see it in `strings.xml`.  If it's defined in a dependency let's re-define it here so we don't get changing text over time, the string resources of Google Play Services are not considered a public API and may change or go away.

samtstern:
@peterhav one comment, otherwise LGTM.

samtstern:
LGTM, thanks!

PR #248, opened 2016-08-10
Re-organize sample and clear memory leaks

### LINKED ISSUE #244: Activity leak when using FirebaseRecyclerAdapter
I am noticing an activity leak with LeakCanary when using a FirebaseRecyclerAdapter.

Here's the trace from LeakCanary:

leak

My adapter does not hold a reference to a context.

The leak seems to go away if I explicitly remove the adapter from the recycler in the `onDestroy` method.

    @Override
    public void onDestroy() {
        super.onDestroy();
        recycler.setAdapter(null);
    }

Any ideas if this is normal behaviour?

  * Google Play Services version: 9.4.0
  * Firebase/Play Services SDK version: 9.4.0
  * FirebaseUI version: 0.4.4

samtstern:
Fixes #244

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -20,7 +20,7 @@
 
         <!-- Chat demo -->
         <activity
-            android:name=".ChatActivity"
+            android:name=".database.ChatActivity"
             android:label="@string/name_chat" />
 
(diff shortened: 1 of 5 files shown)


samtstern:
Build is green, simple change so merging myself.

PR #249, opened 2016-08-10
Update SDK, allow Facebook to use Chrome Custom Tabs

### LINKED ISSUE #214: Add option to enable Chrome Custom Tabs for Facebook signin
Facebook signin in Android has an option to [enable Chrome Custom Tabs](https://developers.facebook.com/docs/facebook-login/android#chrome_custom_tabs) instead of the webview for the login dialog when the Facebook app is not installed. There should be an option to enable this.

This has two main advantages:

1. Users don't have to re-enter their credentials each time since they're already logged into their Facebook accounts in Chrome most of the times.
2. Better security since Chrome Custom Tabs is not a part of the app so the users don't have to worry about the app spying on their passwords.

samtstern:
Fixes #214 

@amandle please review

### DIFF
--- a/auth/src/main/res/values/config.xml
+++ b/auth/src/main/res/values/config.xml
@@ -9,6 +9,16 @@
     -->
     <string name="facebook_application_id" translatable="false">CHANGE-ME</string>
 
+    <!--
+     To enable Chrome Custom Tabs for Facebook Login on devices without the Facebook app
+     installed, change this value to your Facebook App ID prefixed with 'fb'. This should be
+     'fb' + the value of the facebook_application_id string above
+
+     See:
+     https://developers.facebook.com/docs/facebook-login/android#chrome_custom_tabs
+    -->
+    <string name="facebook_login_protocol_scheme" translatable="false">fbYOUR_APP_ID</string>
+
 
     <!--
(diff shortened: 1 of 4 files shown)


samtstern:
PTAL

auth/src/main/res/values/config.xml
```diff
@@ -9,6 +9,16 @@
     -->
     <string name="facebook_application_id" translatable="false">CHANGE-ME</string>
 
+    <!--
+     To enable Chrome Custom Tabs for Facebook Login on devices without the Facebook app
```
amandle: I'm a little unclear here.  If they don't want to enable Chrome Custom Tabs can they do nothing and have it work?  Or will it try to use fbYOUR_APP_ID as the facebook_login_protocol_scheme and break.

samtstern: In my testing when you have a misconfigured custom tabs setup (scheme = `fbYOUR_APP_ID`) it acts as if you had no custom tabs setup and falls back onto the normal webview flow.

Likely the FB SDK asks the system to handle the URL with your App ID, the system says there are no registered intent handlers, and it falls back.

amandle:
LGTM

PR #252, opened 2016-08-15
Allow users to set a custom toolbar title.

PaoloRotolo:
FirebaseUI right now uses the app's name to set the Toolbar title in ```AuthMethodPickerActivity``` and ```EmailHintContainerActivity```.

This pull allows users to **set their own custom title** using
``` <string name="toolbar_title">"Custom title"</string> ```
in their app's ```strings.xml```.

*Result example:*
screenshot_1471277900

### DIFF
--- a/auth/src/main/AndroidManifest.xml
+++ b/auth/src/main/AndroidManifest.xml
@@ -3,36 +3,57 @@
     package="com.firebase.ui.auth">
 
     <application>
-        <activity android:theme="@style/FirebaseUI.Dialog"
-                  android:name="com.firebase.ui.auth.ui.email.ConfirmRecoverPasswordActivity">
-        </activity>
-        <activity android:theme="@style/FirebaseUI.Translucent" android:name="com.firebase.ui.auth.ui.email.EmailHintContainerActivity" >
-        </activity>
-        <activity android:theme="@style/FirebaseUI" android:name="com.firebase.ui.auth.ui.email.RecoverPasswordActivity" >
-        </activity>
-        <activity android:theme="@style/FirebaseUI" android:name="com.firebase.ui.auth.ui.email.RegisterEmailActivity" >
-        </activity>
-        <activity android:theme="@style/FirebaseUI" android:name="com.firebase.ui.auth.ui.email.SignInNoPasswordActivity" >
-        </activity>
-        <activity android:theme="@style/FirebaseUI" android:name="com.firebase.ui.auth.ui.email.SignInActivity" >
-        </activity>
-        <activity android:theme="@style/FirebaseUI.Translucent"
-                  android:name="com.firebase.ui.auth.ui.account_link.SaveCredentialsActivity">
-        </activity>
-        <activity android:theme="@style/FirebaseUI" android:name="com.firebase.ui.auth.ui.account_link.AccountLinkInitActivity" >
-        </activity>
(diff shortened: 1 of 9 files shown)


samtstern:
@PaoloRotolo thanks for the contribution.  A few things:

  1. It's always better to file an Issue before a PR so we can discuss the details.
  2. This pull request modifies two of the Activities in the sign in flow, but there are many others.  You can see them all in `auth/src/main/AndroidManifest.xml`.  What is your intention for the titles of those other activities?

PaoloRotolo:
Hi @samtstern, thanks for your reply.
1. I didn't know that, sorry.
2. Other activities have consistent names (for example *Sign in with email* or *Recover password*). You can change as well the titles of these activities declaring the corresponding string resource in your own ``strings.xml`` file as well.

Another example: ```RecoverPasswordActivity``` has a method ```setTitle(R.string.recover_password_title);```. The user can edit the title declaring
```<string name="recover_password_title">"Custom title"</string>``` in the app ```strings.xml```.

```AuthMethodPickerActivity``` instead takes it from the ```app_name``` string and there's no way for the user to customize that.

Please let me know if there are more activities that I missed that adopt the same behavior.

samtstern:
Ok I went through each Activity and listed the explicit title they have, if any:

- ChooseAccountActivity - None
- IDPSignInContainerActivity - None
- AuthMethodPickerActivity - None
- WelcomeBackPasswordPrompt - `R.string.sign_in`
- SaveCredentialsActivity - None
- SignInActivity - `R.string.sign_in`
- SignInNoPasswordActivity - `R.string.sign_in_with_email`
- WelcomeBackIDPPrompt - `R.string.sign_in`
- RegisterEmailActivity - `R.string.create_account_title`
- RecoverPasswordActivity - `R.string.recover_password_title`
- EmailHintContainerActivity - None
- ConfirmRecoverPasswordActivity - `R.string.check_your_email`

I think to solve this once and for all we should go a little farther than your initial proposal and do the following:
  * Create a unique string resource for the title of each Activity
  * Create a new string reference for the default title (you called it `toolbar_title`, I'd suggest `default_toolbar_title`).  Have the default value be `@string/app_name`.
  * Set the title in the `<activity>` tag XML, so it's easier to discover.  Remove all `setTitle()` calls.
  * For titles previously set by `setTitle` have the string resource point to the previous title.  So for example `<string name="title_welcome_back_password_prompt">@string/sign_in</string>`
  * For titles previously defaulting to `app_name`, make the string resource point to `default_toolbar_title`, for example `<string name="title_choose_account_activity">@string/default_toolbar_title</string>`.

I think this will make it easy for anyone to override any title, while making the titles in general more predictable and discoverable for someone reading the code.

If that sounds tedious, I am happy to write the PR myself.  But if you still want to contribute, update this PR and I will review the change.

PaoloRotolo:
Sounds good. I'm on it!

samtstern:
@PaoloRotolo awesome!  Also apologies for the merge conflicts, hope it's not too hairy.

PaoloRotolo:
@samtstern Done. I also ran Android Studio's code reformat on ```AndroidManifest.xml``` since it was difficult to read before.

samtstern:
@PaoloRotolo this looks great, thank you!

Ali-Ghali:
great is it available now for web ui as well

PR #253, opened 2016-08-15
Add option to disable SmartLock

### LINKED ISSUE #251: Create option to disable SmartLock flow
Sometimes in development it is more convenient not to have SmartLock changing the flow of FirebaseUI-Auth.  We could add a flag to disable SmartLock so that it's easier to get started with the API.

samtstern:
Fixes #251

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/ui/account_link/SaveCredentialsActivity.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/account_link/SaveCredentialsActivity.java
@@ -40,6 +40,7 @@ import com.google.android.gms.common.api.GoogleApiClient;
 import com.google.android.gms.common.api.ResultCallback;
 import com.google.android.gms.common.api.Status;
 import com.google.firebase.auth.FacebookAuthProvider;
+import com.google.firebase.auth.FirebaseUser;
 import com.google.firebase.auth.GoogleAuthProvider;
 
 public class SaveCredentialsActivity extends AppCompatBase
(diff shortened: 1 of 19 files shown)


auth/src/main/java/com/firebase/ui/auth/util/EmailFlowUtil.java
```diff
@@ -0,0 +1,27 @@
...
+import com.firebase.ui.auth.ui.email.EmailHintContainerActivity;
+import com.firebase.ui.auth.ui.email.SignInNoPasswordActivity;
+
+/**
+ * Helper class to kick off the Email/Password sign-in flow.
+ */
+public class EmailFlowUtil {
+
+    /**
+     * Return and intent for either {@link EmailHintContainerActivity} or
```
amandle: nit: typo here, should be "an"

amandle:
LGTM

samtstern:
@amandle fixed typo and also made sure credentials are not saved when smartlock is off (previous commit only dealt with the reading case).  PTAL to make sure you're ok with the additional code.

auth/src/main/java/com/firebase/ui/auth/ui/account_link/SaveCredentialsActivity.java
```diff
@@ -60,11 +60,19 @@
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setContentView(R.layout.save_credentials_layout);
+
+        // If SmartLock is not enabled, finish immediately
+        if (!getIntent().getBooleanExtra(ExtraConstants.EXTRA_SMARTLOCK_ENABLED, true)) {
```
amandle: I think it would be cleaner to just not start this activity if smartlock is not enabled, even though there are a few entry points that would need updating.  Thoughts?

samtstern: Agreed.  Now that I'm looking at it there is no reason for SaveCredentialsActivity to be an Activity at all besides encapsulation.  It has no UI, I am going to refactor it.

samtstern: Ah I see the activity is used to start resolutions ... OK I have to think of something else

samtstern:
Ok I refactored the logic to start the save Activity in all callers so now it does not launch.

amandle:
Left a couple comments, but LGTM

PR #260, opened 2016-08-17
Make tests run/pass

samtstern:
@amandle PTAL

Embarassingly we haven't been running our tests at all, so much so that I only now realized that Robolectric does not support API 24.

I made some changes to make all the tests pass again, but it requires a downgrade of the compileSdkVersion (and therefore a downgrade of the support libraries).

Since we have no Android N features I don't **think** this should cause any problems, and developers can always override our support lib dependency if they are using N features heavily.

WDYT?  I think having green tests on Travis is worth it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ android:
   components:
     - platform-tools
     - tools
-    - build-tools-23.4.0
+    - build-tools-23.0.3
     - android-23
     - extra-google-google_play_services
     - extra-google-m2repository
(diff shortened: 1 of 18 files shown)


samtstern:
cc @morganchen12, getting things test-y

samtstern:
@amandle any thoughts on this?

amandle:
LGTM, I can't think of any reasons why downgrading would cause issues and getting the tests running is pretty crucial.  Thanks for doing this!

Issue #157, opened 2016-05-30
Twitter Provider


raulperez:
Doesn't the last version include the twitter provider?

samtstern:
Firebase Auth does support Twitter login, however we have not yet implemented the UI flow in FirebaseUI.

If you want to use Twitter login with Firebase you can follow [this example](https://github.com/firebase/quickstart-android/blob/master/auth/app/src/main/java/com/google/firebase/quickstart/auth/TwitterLoginActivity.java) to implement the UI yourself.

raulperez:
Thank you Sam.

It will be implemented later on?

samtstern:
@raulperez yes we hope to support all providers eventually, but the priority now is fixing bugs in the Facebook/Google/Email implementation so that there is a solid base for that future work.

dazza5000:
:+1:

mockitoo:
![image](https://cloud.githubusercontent.com/assets/16718247/16340587/a73a2df2-3a31-11e6-9239-e9c3fa378115.png)

Chris-crisur:
@samtstern I have managed to implement Twitter login a while ago; would you like me to extend this library's UI flow to include it?
The recent(ish) Firebase upgrades have also meant Firebase Twitter auth works with Digits too now; but Firebase can give a `User has already been linked to the given provider` error if the user tries to login with Twitter AND Digits. One solution is to log them out of the old provider (e.g. Twitter) and auth again with the new provider (e.g. Digits).

Chris-crisur:
if this gets enough 'like's and the moderators (cough @samtstern cough) approve; I will work to implement it.
Moderators, would you like me to create a new issue with the proposed change (Twitter integration) as per your contribution steps?

samtstern:
I think amandle@ is already working on this, we know it's an important
feature for everyone and it's at the top of our list too.

Chris-crisur:
Awesome!
@amandle let me know if you need help.
Would also be keen to implement Digits, which would piggy back off Twitter.

dazza5000:
Looking forward to this!

dazza5000:
THANK YOU!

PR #268, opened 2016-08-26
Add Twitter as an IDP

amandle:
Add Twitter as a provider in response to issue #157 

I am not entirely sure how this will fit in with our feature parity plans with other platforms, we may have to wait for iOS to implement it before we can actually release.

Also, the Twitter Client Secret is stored in a resources file, which is not ideal but without writing a custom back-end server to coordinate with I'm not sure there is a better alternative (just obfuscation techniques) though I'm open to any suggestions.

### DIFF
--- a/app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
+++ b/app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
@@ -33,6 +33,8 @@ import com.firebase.uidemo.R;
 import com.google.firebase.auth.FirebaseAuth;
 
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 
 import butterknife.BindView;
 import butterknife.ButterKnife;
@@ -68,6 +70,9 @@ public class AuthUiActivity extends AppCompatActivity {
 
     @BindView(R.id.facebook_provider)
     CheckBox mUseFacebookProvider;
+
+    @BindView(R.id.twitter_provider)
+    CheckBox mUseTwitterProvider;
 
     @BindView(R.id.google_tos)
     RadioButton mUseGoogleTos;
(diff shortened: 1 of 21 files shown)


auth/src/main/AndroidManifest.xml
```diff
@@ -80,5 +80,9 @@
             </intent-filter>
         </activity>
 
+        <meta-data
+            android:name="io.fabric.ApiKey"
```
samtstern: Is this actually necessary?  In the Firebase Auth quickstart I was able to get Twitter Sign In to work with only the consumer key/secret and didn't need the app ID or any meta-data:

https://github.com/firebase/quickstart-android/blob/master/auth/app/src/main/AndroidManifest.xml

build.gradle
```diff
@@ -4,11 +4,12 @@ buildscript {
...
         jcenter()
         mavenLocal()
+        maven { url 'https://maven.fabric.io/public' }
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.1.2'
+        classpath 'com.android.tools.build:gradle:2.1.3'
         classpath 'com.google.gms:google-services:3.0.0'
-
+        classpath 'io.fabric.tools:gradle:1.+'
```
samtstern: nit: use explicit version numbers, '1.+' makes the build slower and less predictable.

samtstern:
@amandle this looks really good!  We do need to wait for iOS to release but I think we can still do work like this on a `vNext` branch or similar.  Can we move this PR to targeting a `version-1.0.0-dev` branch?

auth/src/main/AndroidManifest.xml
amandle: You're right, it doesn't appear to be necessary. Removed it.

build.gradle
amandle: Twitter actually recommends using the open-ended approach: https://fabric.io/kits/android/twitterkit/install 

They don't publish the latest version number anywhere, but some guess and check determined the latest is 1.21.7 so I've changed it to that.

app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
```diff
@@ -229,6 +245,17 @@ private boolean isFacebookConfigured() {
     }
 
     @MainThread
+    private boolean isTwitterConfigured() {
+        List<String> twitterConfigs = Arrays.asList(
+                getResources().getString(R.string.twitter_app_id),
```
samtstern: Missed one reference to twitter_app_id

app/src/main/res/values/config.xml
```diff
@@ -10,4 +10,22 @@
...
     -->
     <string name="facebook_login_protocol_scheme" translatable="false">fbYOUR_APP_ID</string>
+
+
+
+    <!--
+    Your Twitter app id. To use Twitter accounts with the demo application, register an application
+    with Twitter and provide the following three values.
+    -->
+    <string name="twitter_app_id" translatable="false">CHANGE-ME</string>
```
samtstern: Here too, just forgot to remove it from the sample/

build.gradle
samtstern: Yeah I think they do it to keep their docs always correct, but this will make our builds happier.  Thanks!

samtstern:
LGTM, thanks!

Issue #243, opened 2016-08-08
Allow retrieving data from a reference of keys in FirebaseArray


SUPERCILEX:
The documentation says to index data like so: https://firebase.google.com/docs/database/android/structure-data.

How would you go about giving Firebase UI a reference of your keys and a reference of the location of the data for those keys to populate a recyclerview? I've tried in the following diff, but it doesn't work and I think I'm going about it in the wrong way: https://github.com/SUPERCILEX/FirebaseUI-Android/commit/fb4e704f9a8d4ac1391ed28917b488aa0eed6d25

Could you guys implement this feature or give me some clues as to how to do it? Thanks!

puf:
See http://stackoverflow.com/questions/34559171/coupling-firebaserecyclerviewadapter-to-a-boolean-string-map-entry for one way to already join data from a different location when using FirebaseUI. We already have a feature request open for this, so I'm going to close this as a dupe of #12 .

SUPERCILEX:
Thanks! I got it working!

Issue #12, opened 2015-09-01
Support for joins


ginovva320:
Will there be any support here for joins across multiple Firebase paths, similar to what firebase-util offers for Javascript? All of my data is denormalized, so creating a ListView just based on one path would not be sufficient. If not, do you have any recommendations?

puf:
While it is technically possible (and an interesting challenge) to joins two lists together, it comes with all kinds of nasty edge cases and performance implications that are often better prevented by duplicating data. I'll leave your report open as an enhancement, but it is not on the list of features we intend to work on in the immediate feature. It might also fit better in a separate project, similar to how firebase-util is separate from AngularFire.

z-nut:
Check out my fork. I have tweaked the code to accept multiple firebase paths and then load them pretty much seamlessly into the same view holder. This is of course, a work in progress. 

https://github.com/z-nut/FirebaseUI-Android

samtstern:
I believe this was fixed a while ago in #276 but this issue was never closed.  Closing it now, please open a new issue if the new implementation does not meet your needs.

PR #276, opened 2016-09-01
Support for joins

SUPERCILEX:
- If this has been discussed in an issue, make sure to mention the issue number here.  If not, go file an issue about this to make sure this is a desirable change.
  - Issues #243 and #12 
- If this is a new feature please co-ordinate with someone on [FirebaseUI-iOS](https://github.com/firebase/firebaseui-ios) to make sure that we can implement this on both platforms and maintain feature parity.
  - I don't know objective C or swift. Anyone want to do this?

TODO:
- [x] tests


giautm:
I'm waiting for this deloy. 👍

puf:
Hey Alex. Thanks for the PR. Looks great, definitely a very clean implementation of attaching and detaching the listeners. I also like that it uses the data model from the Firebase documentation as its basis.

I don't think we should add this code to the basic FirebaseXxxAdapters though. Scanning your code structure I see many `if (condition) oldBehavior() else newBehavior()`, which would be cleaner to catch in a subclass. That would also keep the current classes as simple as we intentionally kept them and allows us to later add support for other join structures without breaking backwards compatibility of the basic adapters.

Can you see if you can isolate the behavior in separate sub-adapters?

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
```diff
@@ -47,76 +41,95 @@ public void cleanup() {
...
     public int getCount() {
         return mSnapshots.size();
-
     }
+
     public DataSnapshot getItem(int index) {
         return mSnapshots.get(index);
     }
 
+    protected DataSnapshot internalGetItem(int index) {
```
SUPERCILEX: This is completely gross IMO, but I don't know how to get rid of it. Any feedback is welcome!

SUPERCILEX:
Should I update the app example to index the data? I don't think it's necessary, but it would put people on the right track for scalability.

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
```diff
@@ -47,76 +41,95 @@ public void cleanup() {
...
+
     private int getIndexForKey(String key) {
         int index = 0;
         for (DataSnapshot snapshot : mSnapshots) {
             if (snapshot.getKey().equals(key)) {
                 return index;
-            } else {
-                index++;
             }
+            index++;
```
puf: Revert this change please.

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
```diff
@@ -47,76 +41,95 @@ public void cleanup() {
...
     }
-    
+
     protected void notifyCancelledListeners(DatabaseError databaseError) {
         if (mListener != null) {
             mListener.onCancelled(databaseError);
         }
     }
+
+    public interface OnChangedListener {
```
puf: Move the interface back to the top of the class please.

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
puf: Agreed. I'll brood on a better way to do this.

puf:
Thanks for the updates Alex. Great progress of an already impressive proposal.

I'm prefer to also have separate **adapters** for the indexed data, so that using them becomes a matter of using a different class, instead of using a different constructor/overload: `adapter = new FirebaseIndexListAdapter(...)`. Does that make sense?

Keep in mind btw that we'll still need to also get this feature in FirebaseUI-iOS, before we can merge. That will likely take time, since the iOS library is going through lots of changes already. So I hope you're in this for the long haul. Rest assured: we do really your work on this and would love to somehow get this functionality released in FirebaseUI at some point.

SUPERCILEX:
Hi Puf,
Yeah, I didn't understand that you also wanted the adapters changed. Making separate adapters definitely makes it clearer to the user!

Check out `FirebaseRecyclerAdapter`, and `FirebaseIndexRecyclerAdapter`, I think this is what you are looking for. I haven't changed list adapter yet because I want to make sure I'm on the right track.

database/src/main/java/com/firebase/ui/database/FirebaseRecyclerAdapter.java
```diff
@@ -63,32 +62,23 @@
...
-    /**
-     * @param modelClass Firebase will marshall the data at a location into an instance of a class that you provide
-     * @param modelLayout This is the layout used to represent a single item in the list. You will be responsible for populating an
-     *                    instance of the corresponding view with the data from an instance of modelClass.
-     * @param viewHolderClass The class that hold references to all sub-views in an instance modelLayout.
-     * @param ref        The Firebase location to watch for data changes. Can also be a slice of a location, using some
-     *                   combination of <code>limit()</code>, <code>startAt()</code>, and <code>endAt()</code>
-     */
-    public FirebaseRecyclerAdapter(Class<T> modelClass, int modelLayout, Class<VH> viewHolderClass, Query ref) {
+    FirebaseRecyclerAdapter(Class<T> modelClass, int modelLayout, Class<VH> viewHolderClass, FirebaseArray snapshots) {
```
giautm: Please public both classes: `FirebaseArray` and `FirebaseIndexArray`. So we can inherit and extends from outside of package. Example: I can extends FirebaseArray to provide filter at local.

SUPERCILEX: That was intentional, see https://github.com/firebase/FirebaseUI-Android/pull/89#issuecomment-191113404.

SUPERCILEX: @puf With this implementation, should `FirebaseRecyclerAdapter(Class<T> modelClass, int modelLayout, Class<VH> viewHolderClass, FirebaseArray snapshots)` be protected or public? I'm keeping it private to the library because it's more a private API to support `IndexFirebaseArray` than anything else. Any thoughts?

giautm: Make `FirebaseArray` implement an interface like `FirebaseArrayInterface`. And using interface in constructor instead of `FirebaseArray`. That will help we to implement a `has-a` relationship. Following is a example using:
```
public interface FirebaseArrayInterface {
    interface OnChangedListener {
        void onChanged(EventType type, int index, int oldIndex);

        void onFiltered();

        void onCancelled(DatabaseError databaseError);

        enum EventType {Added, Changed, Removed, Moved}
    }

    DataSnapshot getItem(int index);

    int getCount();

    void cleanup();

    void setOnChangedListener(OnChangedListener listener);
}
```

Filterable Array
```
public abstract class FirebaseFilterableArray implements FirebaseArrayInterface
        , FirebaseArrayInterface.OnChangedListener
        , Filterable {
    private FirebaseFilter mFilter;
    private List<DataSnapshot> mFiltered = new ArrayList<>();
    private FirebaseArrayInterface mSnapshots;
    private CharSequence mConstraint;
    private OnChangedListener mListener;

    public FirebaseFilterableArray(FirebaseArrayInterface snapshots) {
        mFilter = new FirebaseFilter();
        mSnapshots = snapshots;
        mSnapshots.setOnChangedListener(this);
    }

    @Override
    public int getCount() {
        return mFiltered.size();
    }

    @Override
    public DataSnapshot getItem(int index) {
        return mFiltered.get(index);
    }

    @Override
    public void cleanup() {
        mSnapshots.cleanup();
    }

    @Override
    public void setOnChangedListener(OnChangedListener listener) {
        mListener = listener;
    }

    @Override
    public void onChanged(EventType type, int index, int oldIndex) {
        if (mListener != null) {
            mListener.onChanged(type, index, oldIndex);
        }
        getFilter().filter(mConstraint);
    }

    @Override
    public void onFiltered() {
        if (mListener != null) {
            mListener.onFiltered();
        }
    }

    @Override
    public void onCancelled(DatabaseError databaseError) {
        if (mListener != null) {
            mListener.onCancelled(databaseError);
        }
    }

    @Override
    public Filter getFilter() {
        return mFilter;
    }

    private class FirebaseFilter extends Filter {
        @Override
        protected FilterResults performFiltering(CharSequence constraint) {
            mConstraint = constraint;

            List<DataSnapshot> result = new ArrayList<>();
            for (int i = 0; i < mSnapshots.getCount(); i++) {
                DataSnapshot item = mSnapshots.getItem(i);
                if (match(item, constraint)) {
                    result.add(item);
                }
            }

            FilterResults filterResults = new FilterResults();
            filterResults.values = result;
            return filterResults;
        }

        @SuppressWarnings("unchecked")
        @Override
        protected void publishResults(CharSequence constraint, FilterResults results) {
            mFiltered = (List<DataSnapshot>) results.values;
            if (mListener != null) {
                mListener.onFiltered();
            }
        }
    }

    protected abstract boolean match(DataSnapshot dataSnapshot, CharSequence constraint);
}
```

Example using
```
            mArrayFilter = new FirebaseFilterableArray(new FirebaseIndexArray(mListQuery,
                    mRootRef.child(BuildConfig.FIREBASE_LAST_MILE_ORDERS))) {
                @Override
                protected boolean match(DataSnapshot dataSnapshot, CharSequence constraint) {
                    if (constraint == null || constraint.length() == 0) {
                        return true;
                    } else {
                        LastMile lastMile = dataSnapshot.getValue(LastMile.class);
                        return lastMile.getDeliveryAddress().contains(constraint)
                                || lastMile.getRecipientName().contains(constraint);
                    }
                }
            };
            mAdapter = new FirebaseRecyclerAdapter<LastMile, LastMileRecentsViewHolder>(LastMile.class,
                    R.layout.fragment_last_mile_recents_item, LastMileRecentsViewHolder.class, mArrayFilter) {

                @Override
                protected void populateViewHolder(LastMileRecentsViewHolder viewHolder, LastMile lastMile, int position) {
                }
            };
```

Sorry, my English isn't good. :(

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
giautm: I found a bug here. When you remove the last element of the list, remove trigger events before and list have no element. So it would make an exception:
```
    java.lang.IndexOutOfBoundsException: Invalid index 0, size is 0
    at java.util.ArrayList.throwIndexOutOfBoundsException(ArrayList.java:255)
    at java.util.ArrayList.get(ArrayList.java:308)
    at com.firebase.ui.database.FirebaseArray.internalGetItem(FirebaseArray.java:61)
    at com.firebase.ui.database.FirebaseIndexArray.notifyChangedListeners(FirebaseIndexArray.java:119)
    at com.firebase.ui.database.FirebaseArray.notifyChangedListeners(FirebaseArray.java:119)
    at com.firebase.ui.database.FirebaseArray.onChildRemoved(FirebaseArray.java:100)
    at com.firebase.ui.database.FirebaseIndexArray.onChildRemoved(FirebaseIndexArray.java:29)
    at com.google.android.gms.internal.zzahe.zza(Unknown Source)
    at com.google.android.gms.internal.zzaje.zzcta(Unknown Source)
    at com.google.android.gms.internal.zzajh$1.run(Unknown Source)
    at android.os.Handler.handleCallback(Handler.java:739)
    at android.os.Handler.dispatchMessage(Handler.java:95)
    at android.os.Looper.loop(Looper.java:155)
    at android.app.ActivityThread.main(ActivityThread.java:5727)
    at java.lang.reflect.Method.invoke(Native Method)
    at java.lang.reflect.Method.invoke(Method.java:372)
    at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1029)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:824)
```

because `mSnapshots.remove(index)` was call before `notifyChangedListeners`
```
    @Override
    public void onChildRemoved(DataSnapshot snapshot) {
        int index = getIndexForKey(snapshot.getKey());
        mSnapshots.remove(index);
        notifyChangedListeners(OnChangedListener.EventType.Removed, index);
    }
```

SUPERCILEX: Thanks for catching that!!! I completely messed up removing an item: I wasn't even deleting the snapshot! Check it out now: https://github.com/firebase/FirebaseUI-Android/pull/276/files#diff-c7d9ac7c32439f20359555e3db253268R119

database/src/main/java/com/firebase/ui/database/IndexFirebaseArray.java
```diff
@@ -0,0 +1,138 @@
...
+                    mValueEventListeners.put(ref, valueEventListener);
+                    break;
+                case Changed:
+                    // We don't care because the key's value shouldn't matter.
+                    break;
+                case Removed:
+                    DatabaseReference rmRef = mRef.child(mSnapshots.get(index).getKey());
+                    rmRef.removeEventListener(mValueEventListeners.get(rmRef));
+                    mValueEventListeners.remove(rmRef);
+                    mSnapshots.remove(index);
```
giautm: No. The problem is still there. Because, `FirebaseArray.onChildRemoved()` really to call `mSnapshots.remove(index);`. After that, it call `Firebase.Array.notifyChangedListeners()` was override in `IndexFirebaseArray`.

To fix, override `onChildRemoved` on `IndexFirebaseArray`
```
    @Override
    public void onChildRemoved(DataSnapshot snapshot) {
        int index = getIndexForKey(snapshot.getKey());
        DatabaseReference rmRef = mRef.child(internalGetItem(index).getKey());
        rmRef.removeEventListener(mValueEventListeners.get(rmRef));
        mValueEventListeners.remove(rmRef);

        super.onChildRemoved(snapshot);
    }
```
and change following lines to:
```
                case Removed:
                    mListener.onChanged(OnChangedListener.EventType.Removed, index, -1);
                    break;
```

giautm:
@SUPERCILEX Nice! (y)

SUPERCILEX:
I noticed you deleted your previous comment, I think you got confused with `mSnapshots` being a list in the `FirebaseArrays` and the object in the adapters :). The item is first removed in `FirebaseArray`'s `mSnapshots`, then I get the key from `IndexFirebaseArray`'s `mSnapshots` which hasn't changed yet. I know, it's pretty messy...

giautm:
Yup, i got confused. :D

database/src/main/java/com/firebase/ui/database/IndexFirebaseArray.java
```diff
@@ -0,0 +1,123 @@
...
+                        @Override
+                        public void onDataChange(DataSnapshot dataSnapshot) {
+                            if (dataSnapshot.getValue() != null) {
+                                if (isNewListener[0]) {
+                                    mSnapshots.add(index, dataSnapshot);
+                                    mListener.onChanged(OnChangedListener.EventType.Added, index, -1);
+
+                                    isNewListener[0] = false;
+                                } else {
+                                    int index = getIndexForKey(dataSnapshot.getKey());
```
SUPERCILEX: Here I'm using the snapshots from `FirebaseArray`, not `IndexFirebaseArray`. I can't think of how this could be problematic since the keys are the same for both implementations, but I just wanted to put it out there anyway.

Also, this could be more performant since the snapshots from `FirebaseArray` should, in theory, be much smaller objects than the ones in `IndexFirebaseArray`.

SUPERCILEX:
I rewrote `IndexFirebaseArray` because of a bug. The plus side is that `internalGetItem` is gone and the implementation is a little bit cleaner depending on how you look at it. `mSnapshots` from `FirebaseArray` is no longer used in this implementation which is weird.

The bug: [Line 77](https://github.com/firebase/FirebaseUI-Android/blob/c6edad5f3b129823784c77b553c9548d71c09b27/database/src/main/java/com/firebase/ui/database/IndexFirebaseArray.java#L77) would throw an `IndexOutOfBoundsException` if the value of the dataSnapshot was null (could be because the user is offline, or a mistake by the dev). I couldn't think of a good way to solve this since `FirebaseArray` would have more items then `IndexFirebaseArray`. I tried just adding a null item, but that sucked pretty bad.

Either way, The new `IndexFirebaseArray` cleans up the mess with notifying listeners, but duplicates code which is somewhat unfortunate.

SUPERCILEX:
Update: Nevermind, the new implementation doesn't fix it. Now, I get an IllegalArgumentException at [line 82](https://github.com/firebase/FirebaseUI-Android/pull/276/files#diff-c7d9ac7c32439f20359555e3db253268R82). I'll have to think about this... any help is welcome!

This means that both implementations of `IndexFirebaseArray` are usable, it depends on which one we prefer.

giautm:
@SUPERCILEX : Is there nothing like `FirebaseIndexListAdapter` for `FirebaseListAdapter`?

SUPERCILEX:
Don't worry, there will be. I'm just waiting for confirmation from @puf that FirebaseIndexRecyclerAdapter is what he was looking for.

SUPERCILEX:
Hi @puf and @samtstern,
The PR is ready for final review: I fixed all known bugs, extracted `FirebaseIndexListAdapter`, and did some final cleanup. Enjoy!

SUPERCILEX:
Any idea when my PR will be merged?

database/src/main/java/com/firebase/ui/database/FirebaseRecyclerAdapter.java
puf: Hey @giautm : definitely an interesting idea to have multiple array implementations. I had considered it before (since then the adapter can just be the UI logic). But it sounds like a different direction than where this PR is heading.

puf:
Hey @SUPERCILEX. Thanks for all the work on this PR. I haven't gotten round to reviewing it yet, since I've also been trying to align with the iOS team (especially @mcdonamp who has  all kinds of opinions about this). I promise you that this PR won't be forgotten in that discussion. :-)

SUPERCILEX:
Awesome, can't wait! (We can use github's new review feature 😀)

MaciejCiemiega:
@puf I've done a refactoring of @SUPERCILEX `IndexFirebaseArray` class and submitted a PR to @SUPERCILEX repo  https://github.com/SUPERCILEX/FirebaseUI-Android/pull/11
I don't know whether this is an appropriate flow though (?)

Improvements:
1. It makes use of original `mSnapshots` from `FirebaseArray`
2. There is only one `ValueEventListener` listener instead of a map
3. I got rid of `mNullableSnapshots`
4. It uses most of `super` implementation for `ChildEventListener`, with a trick to mute the original `OnChangedListener` listener from being notified when something with keys changed on `super` methods.

In addition I think that `IndexFirebaseArray` class should be renamed to `FirebaseIndexArray` to be more consistent with names like `FirebaseIndexListAdapter` or `FirebaseIndexRecyclerAdapter`. Or name of these adapters should be changed to have `Index` in front...

SUPERCILEX:
@MaciejCiemiega I'm looking into your PR and yeah, I agree about the name.

@puf thanks for catching that!

SUPERCILEX:
Woops. Bumped the close button 😭

SUPERCILEX:
Is there any way to move the adapters into their separate package while still keeping `FirebaseArray` protected?

database/README.md
```diff
@@ -270,3 +270,18 @@ recycler.setAdapter(mAdapter);
...
+If your data is [properly indexed](https://firebase.google.com/docs/database/android/structure-data#best_practices_for_data_structure), change your adapter initalization like so:
+
+For the `FirebaseRecyclerAdapter`:
+```java
+new FirebaseRecyclerAdapter<Chat, ChatHolder>(Chat.class, android.R.layout.two_line_list_item, ChatHolder.class, keyRef, dataRef)
+```
+
+And the `FirebaseListAdapter`;
+```java
+new FirebaseListAdapter<Chat>(this, Chat.class, android.R.layout.two_line_list_item, keyRef, dataRef)
```
SUPERCILEX: Done.

SUPERCILEX:
I'm in the process of resolving https://github.com/SUPERCILEX/FirebaseUI-Android/pull/11.

If anyone has questions about why I need `mNullableSnapshots` or would like to try and figure out how to remove it, I left a nice explanation there.

database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
```diff
@@ -0,0 +1,150 @@
...
+                notifyChangedListeners(OnChangedListener.EventType.Removed, index);
+            }
+        }
+    }
+
+    @Override
+    public void onChildChanged(DataSnapshot snapshot, String previousChildKey) {
+        super.setOnChangedListener(null);
+        super.onChildChanged(snapshot, previousChildKey);
+        super.setOnChangedListener(mListener);
```
SUPERCILEX: @puf any idea how to get rid of these? See discussion at https://github.com/SUPERCILEX/FirebaseUI-Android/pull/11.

MaciejCiemiega: I would move my comment here because it's more relevant here now:

@SUPERCILEX There is always alternative way by using a flag to temporary avoid notifying the listener:

```java
private boolean mDoNotNotifyChangeListeners;

...
@Override
public void onChildChanged(DataSnapshot snapshot, String previousChildKey) {
	mDoNotNotifyChangeListeners = true;
	super.onChildChanged(snapshot, previousChildKey);
	mDoNotNotifyChangeListeners = false;
}
...	

@Override
protected void notifyChangedListeners(OnChangedListener.EventType type, int index, int oldIndex) {
	if (!mDoNotNotifyChangeListeners){
		super.notifyChangedListeners(type, index, oldIndex);
	}
}
```

Quick facts:
- There are no `super` calls here (but they do not really bother me)
- We can remove local `mListener`
- Remove overriding of `setOnChangedListener`
- Now `notifyChangedListeners` is overriden
- It introduces an extra variable. with local `mListener` it uses only something that already existed in `super` implementation.
- The end result is the same

Since IMO this is a matter of elegance and feeling we should ask @puf which solution is nicer in his opinion.

SUPERCILEX: I just don't like the idea of checking some condition, calling super, and then reverting that condition. Anyway, we'll see if @puf or @samtstern have any ideas.

SUPERCILEX:
@MaciejCiemiega Could you sign the [cla](https://cla.developers.google.com/clas) and reply "I signed it!" to make @googlebot happy?

MaciejCiemiega:
I have it already signed.

SUPERCILEX:
Cool!

SUPERCILEX:
Please wait for #297 so that I can do some package reorganization and clean stuff up.

database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
```diff
@@ -0,0 +1,150 @@
...
+    private boolean isMatch(int index, String key) {
+        return index >= 0 && index < getCount() && getItem(index).getKey().equals(key);
+    }
+
+    @Override
+    public void onChildAdded(DataSnapshot keySnapshot, String previousChildKey) {
+        super.setOnChangedListener(null);
+        super.onChildAdded(keySnapshot, previousChildKey);
+        super.setOnChangedListener(mListener);
+        mRef.child(keySnapshot.getKey()).addValueEventListener(this);
```
samtstern: I believe this is relying on the superclass's error handling for this value event listener.  In the event of a bad request this could result in one error per child being bubbled up, which is likely not the desired behavior as they will be very hard to interpret.

SUPERCILEX: I don't think it would be that bad. If there is an error accessing the indices, that would be handled directly in `FirebaseArray`. If there is an error accessing the data at some index, there would one error per child also handled through `FirebaseArray`. Wouldn't that be expected behavior to get errors for both the indices and each child at an index?

samtstern: It will be hard for the developer to tell where the error occurred and to handle them properly, it could also just result in a huge number of errors bubbling up.

@morganchen12 is working on this problem for the iOS version, the most important thing is to agree on some consistent behavior.

SUPERCILEX: What are you suggesting? I'm not sure how to handle this without going back to a list based approach for keeping track of listeners which is somewhat gross. If I did go back to a list based approach, should I disable errors from the super class? I'm not sure what you are looking for...

morganchen12: Hey @SUPERCILEX, I'm not so worried about the implementation details, my goal is to have the index array behave reasonably if any of its queries/references fails for some reason. My approach thus far has been:

1. If the index query (`keyRef`) raises an error, the index array isn't really usable. On iOS the error is then passed up to the owner of the index array.
2. If one of the `dataRef`'s child queries fails, the element at that index will be `null` until it's changed/deleted/moved, and the error is passed up to the owner of the array along with the index and the query that failed.

SUPERCILEX: I think this is what you are looking for (see review below).

database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
```diff
@@ -0,0 +1,170 @@
...
+                if (isMatch(index, key)) {
+                    mDataSnapshots.remove(index);
+                    notifyChangedListeners(OnChangedListener.EventType.REMOVED, index);
+                }
+            }
+        }
+
+        @Override
+        public void onCancelled(DatabaseError error) {
+            notifyCancelledListeners(error);
```
SUPERCILEX: I'm notifying errors for each child.

database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
```diff
@@ -0,0 +1,170 @@
...
+            DataSnapshot snapshot = mDataSnapshots.remove(oldIndex);
+            int newIndex = getIndexForKey(key);
+            mDataSnapshots.add(newIndex, snapshot);
+            notifyChangedListeners(OnChangedListener.EventType.MOVED, newIndex, oldIndex);
+        }
+    }
+
+    @Override
+    public void onCancelled(DatabaseError error) {
+    }
```
SUPERCILEX: Here I'm completely ignoring errors from the indexed key refs

morganchen12:
While I'm here and proposing changes, please add tests :D :D :D ❤️

SUPERCILEX:
TODO:

- [ ] tests

database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
```diff
@@ -0,0 +1,172 @@
...
+            mDataSnapshots.add(newIndex, snapshot);
+            notifyChangedListeners(OnChangedListener.EventType.MOVED, newIndex, oldIndex);
+        }
+    }
+
+    @Override
+    public void onCancelled(DatabaseError error) {
+        Log.e(TAG, "A fatal error occur retrieving the necessary keys to poplulate your adapter");
+        super.onCancelled(error);
+    }
```
SUPERCILEX: Devs can override [this method](url) in which they could reinitialize their adapter with a new ref or replace their view with an "Oops, an unexpected error occurred." message. I think this gives devs nice options. What do you think of having a separate error method to override for the key failure vs the ref failure?

SUPERCILEX:
@morganchen12 @samtstern 
While working on tests, I discovered that you cannot change the priority of the `dataRefs` (well you can, but it won't do anything). Should I add in the documentation that the order of the list items is determined by the order they are received in from the `keyRef`? If you think about it, that's kinda obvious because you're going to have all the other entries in `dataRef` and having a bunch of haphazard priorities in `dataRef` would be weird... should I add it anyway?

SUPERCILEX:
We've got a problem. The database tests aren't running at all and haven't ever been. Is this on purpose?

puf:
I've run the database test locally, manually when we just started the project. They've never been run automatically afaik.

SUPERCILEX:
@puf So as long as the tests ran fine for me, we're good?

I got all greens so unless you guys have a few final comments, this PR is good to merge!

puf:
I wouldn't mind them running in CI. We just never had CI when we started.

SUPERCILEX:
I'm pretty sure we'd need an emulator with the current state of the tests which would significantly slow Travis down.

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
```diff
@@ -75,45 +77,48 @@ public void onChildAdded(DataSnapshot snapshot, String previousChildKey) {
...
         mSnapshots.remove(index);
         notifyChangedListeners(OnChangedListener.EventType.REMOVED, index);
     }
 
+    @Override
     public void onChildMoved(DataSnapshot snapshot, String previousChildKey) {
         int oldIndex = getIndexForKey(snapshot.getKey());
         mSnapshots.remove(oldIndex);
-        int newIndex = previousChildKey == null ? 0 : (getIndexForKey(previousChildKey) + 1);
+        int newIndex = previousChildKey == null ? 0 : getIndexForKey(previousChildKey) + 1;
```
puf: Please put back the parentheses around `(getIndexForKey(previousChildKey) + 1)`. The behavior is the same, but it makes the intent more explicit.

database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
```diff
@@ -0,0 +1,171 @@
...
+
+    @Override
+    public void setOnChangedListener(OnChangedListener listener) {
+        super.setOnChangedListener(listener);
+        mListener = listener;
+    }
+
+    private class DataRefSnapshot implements ValueEventListener {
```
puf: We should name this class to indicate that it's a listener.

puf:
I had a few nits, aside from that this looks good to me. Thanks for all the hard work on this!

database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
SUPERCILEX: I named it `DataRefListener`, what do you think?

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
SUPERCILEX: Done.

SUPERCILEX:
https://github.com/firebase/FirebaseUI-iOS/pull/164 was just merged, any idea when this PR will be merged?

morganchen12:
@SUPERCILEX can you edit this PR to target the 1.0 branch? Once that's done and tests pass/conflicts are resolved I'll merge

SUPERCILEX:
No conflicts!!! @morganchen12 This is awesome!!! I finally get to delete my `tmpfirebasedatabase` package!

samtstern:
@SUPERCILEX two months later it's in!  Thank you for your patience and all the other contributions you made while this one was pending.  Version 1.0.0 is getting very close, excited to see this live.

SUPERCILEX:
@puf @samstern thank you guys for your patience guiding me towards writing better code. I really appreciate it! I've learned a ton working on FirebaseUI and I hope to keep it up. Thanks!

I'm pretty excited for 1.0 too–there are so many new features!

PR #285, opened 2016-09-06
handle ACTIVITY_RESULT_OTHER_ACCOUNT

amandle:
The returned result when clicking "None of the above" has changed to make it consistent between the credential picker and the hint picker.  This changes ensures that we handle the new value correctly.


samtstern:
LGTM, this will roll into the `0.5.2` release which should be today.

PR #288, opened 2016-09-07
Quickfix documentation: add that smart lock is enabled by default

SUPERCILEX:
Quickfix documentation: add that smart lock is enabled by default


samtstern:
LGTM, thanks!

samtstern:
Note: squashed and merged,

Issue #287, opened 2016-09-06
Multiple Google Sign In / Sign Out attempts in same Activity causes crash


samtstern:
The problem is in `GoogleApiClientTaskHelper`

```java
GoogleApiClient client = mBuilder
                .addConnectionCallbacks(new GoogleApiClient.ConnectionCallbacks() {
                    @Override
                    public void onConnected(@Nullable Bundle bundle) {
                        source.setResult(gacReference.get());
                    }

                    @Override
                    public void onConnectionSuspended(int i) {}
                })
                .addOnConnectionFailedListener(new GoogleApiClient.OnConnectionFailedListener() {
                    @Override
                    public void onConnectionFailed(@NonNull ConnectionResult connectionResult) {
                        source.setException(new IOException(
                                "Failed to connect GoogleApiClient: "
                                        + connectionResult.getErrorMessage()));
                    }
                })
                .build();
```

The line `source.setResult(...)` can be called multiple times, causing:

```
java.lang.IllegalStateException: Task is already complete
```

samtstern:
This issue is fixed in version `0.5.3`

Issue #257, opened 2016-08-16
Audit uses of TaskFailureLogger


samtstern:
In many cases `TaskFailureLogger` is added as the only failure listener to a Task.  This means the error is thrown out besides logging. 

We should audit all uses of this class and determine in which cases we need to actually be taking action in the case of an error.

Example:
In #234 we should be catching the exception where a user does not exist.

samtstern:
Fixed in `0.5.3`

PR #289, opened 2016-09-07
Refactor error handling

samtstern:
Went through and looked at all error handling that uses `TaskFailureLogger` to ensure that each case has some sensible error handling beyond logging.  In most cases I split `onCompleteListener` into separate `onSuccessListener` and `onFailureListener`s.

Also fixed a crash that can come from repeated uses of `GoogleApiClientTaskHelper`.

This addresses the following issues:
  * #257 
  * #287


samtstern:
@amandle PTAL, this is a somewhat large refactor (although not much new logic, still room for errors)

amandle:
@samtstern LGTM!

Issue #290, opened 2016-09-08
Facebook login stop working after updating to 0.5.2 


cutiko:
Greetings

I havent being arround for a while, but I notice there is a new version. The problem is, Facebook login keep crashing.

```
09-07 21:10:40.107 28023-28023/cl.cutiko.flash E/AndroidRuntime: FATAL EXCEPTION: main
                                                                     Process: cl.cutiko.flash, PID: 28023
                                                                     java.lang.UnsupportedOperationException
                                                                         at java.util.AbstractList.add(AbstractList.java:404)
                                                                         at java.util.AbstractList.add(AbstractList.java:425)
                                                                         at com.firebase.ui.auth.provider.FacebookProvider.startLogin(FacebookProvider.java:90)
                                                                         at com.firebase.ui.auth.ui.idp.AuthMethodPickerActivity$1.onClick(AuthMethodPickerActivity.java:127)
                                                                         at android.view.View.performClick(View.java:5198)
                                                                         at android.view.View$PerformClick.run(View.java:21147)
                                                                         at android.os.Handler.handleCallback(Handler.java:739)
                                                                         at android.os.Handler.dispatchMessage(Handler.java:95)
                                                                         at android.os.Looper.loop(Looper.java:148)
                                                                         at android.app.ActivityThread.main(ActivityThread.java:5417)
                                                                         at java.lang.reflect.Method.invoke(Native Method)
                                                                         at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
                                                                         at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)
```

At first I thought it could be something related to the OS so switch from API 19 (Samsung galaxy note 3 neo) to API 21 (Sony Xperia), both physical devices.

Im doing exactly what is done in the [documentation] (https://github.com/firebase/FirebaseUI-Android/blob/master/auth/README.md)

```
startActivityForResult(
                    AuthUI.getInstance()
                            .createSignInIntentBuilder()
                            .setProviders(
                                    AuthUI.EMAIL_PROVIDER,
                                    AuthUI.GOOGLE_PROVIDER,
                                    AuthUI.FACEBOOK_PROVIDER)
                            .build(),
                    RC_SIGN_IN);
```

MobilefactoryAT:
I had to add the facebook permission xml than it worked

cutiko:
Greetings

[...]

PR #292, opened 2016-09-08
Fix facebook permissions

samtstern:
Well I didn't think I'd be learning something new about `java.util.List` today, but I did.  Apparently `List#add()` is an "optional" operation and when you do `Arrays.asList()` you get a `List` that does not support this operation.

This addresses issue #290 and will require a `0.5.3` release.


samtstern:
@puf could you review>

puf:
I didn't know that either. lgtm

PR #293, opened 2016-09-08
Version 0.5.3

samtstern:
Bump to Version 0.5.3 and improve testing

Found out why Robolectric was not loading our resources (`string-array` vs `array`) so changed the test to now use the real resource loading code, which would have caught #290 and will prevent any regressions there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,13 +28,13 @@ libraries.
 ```groovy
 dependencies {
     // Single target that includes all FirebaseUI libraries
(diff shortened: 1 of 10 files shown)


samtstern:
@puf PTAL

puf:
lgtm on a code level. So I'm fine including this in 0.5.3.

I don't like the approach though. In general I prefer seeing these things in code, where they are more discoverable and can be made to auto-complete (I'm not sure if that applies to scopes specifically, since those are inherently stringly typed).

Issue #206, opened 2016-06-30
FirebaseArray in FirebaseListAdapter is not accessible from subclasses


gufux:
# **Situation**
You have a mobile app with Firebase as backend 
you would like to view all the items in firebase location as cards in UI
**you would like to remove some items just from the UI**
  (imagine a UI like Tinder app where a user can remove a card from UI and see the other cards in   the list, where cards are items in Firebase db but they are not deleted when removed from UI)

# **Problem with FirebaseListAdapter**
I use the FirebaseListAdapter as explained in the tutorial and I can easily retrieve the items from firebase and display them in the UI. 
I don't know how to remove items from UI without removing it from Firebase too. The code below remove items from the Adapter but also from firebase.

```
Firebase itemRef = adapter.getRef(position);
itemRef.removeValue();
```
I think the only way to remove items from the adpater without removing the firebase reference is to have access to the FirebaseArray used in the FirebaseListAdapter.

**Proposed solution ?**
Have the FirebaseArray protected so that when implementing our CustomAdapter we can have access to the FirebaseArray and remove items from it.

puf:
In general we kept filtering data out of FirebaseUI until we're ready to implement that feature properly (see #15). The potential for problems is quite big if you start manipulating the FirebaseArray. For example: how would you deal with move operations or other operation that require looking up an item by its key? While these questions may not apply to your use-case, the library is made to work equally for all of them.

How would you use the array in your custom adapter subclass?

gufux:
Hi Frank 
sorry for the late answer. I understand your point and I then changed the UI of my app so that now I have a list of elements (with LinearLayoutManager) and they are always visible with the user scrolling down, so I don't need anymore to "hide" elements of the list.

Knowing better Firebase than before I think one possible solution of my original problem would be to keep a reference on Firebase through index (ex: "card already seen") and update the original list checking always against the boolean "card already seen".

pantos27:
@puf 
I would also love to see FirebaseArray exposed

I'm using a RecyclerView with a long list of items
I would use it to retrieve the row number of a certain item using a search query on one of it's fields

samtstern:
@pantos27 you should be able to track the row index of items by creating a Map<Integer,Object> and populating it whenever you get a call to `populateViewHolder`.  Then you're only tracking items the user has actually seen, which in most cases are the relevant parts anyway.

Also the iOS comparison isn't exactly fair since Objective C doesn't really have access modification like Java has :smile:

[...]

PR #297, opened 2016-09-10
Update database access modifiers

SUPERCILEX:
As I was going through the adapters for PR #276, I noticed that the access modifiers were all weird. Is there something I'm not getting or is this change ok?


database/src/main/java/com/firebase/ui/database/FirebaseArray.java
```diff
@@ -25,7 +25,7 @@
  * This class implements an array-like collection on top of a Firebase location.
  */
 class FirebaseArray implements ChildEventListener {
-    public interface OnChangedListener {
```
SUPERCILEX: This one should definitely be changed, I don't know why it's not at least protected.

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
```diff
@@ -25,7 +25,7 @@
  * This class implements an array-like collection on top of a Firebase location.
  */
 class FirebaseArray implements ChildEventListener {
```
SUPERCILEX: I changed everything below to be no modifier since the class itself shouldn't be accessed from anywhere outside of the package, I don't see why methods inside shouldn't follow along.

samtstern:
@puf what do you think?

database/src/main/java/com/firebase/ui/database/FirebaseListAdapter.java
```diff
@@ -51,10 +51,10 @@
  */
 public abstract class FirebaseListAdapter<T> extends BaseAdapter {
 
+    private Activity mActivity;
```
puf: I've had some cases in the past where I wanted to use the activity and layout in my concrete implementation. I'm not sure if these still apply, but would prefer to keep these protected for now.

database/src/main/java/com/firebase/ui/database/FirebaseListAdapter.java
```diff
@@ -51,10 +51,10 @@
...
     private final Class<T> mModelClass;
-    protected int mLayout;
-    protected Activity mActivity;
-    FirebaseArray mSnapshots;
+    private int mLayout;
+    private FirebaseArray mSnapshots;
```
puf: The snapshot array is accessed by the unit tests. For that reason it must be package private/default access.

database/src/main/java/com/firebase/ui/database/FirebaseListAdapter.java
```diff
@@ -50,10 +50,9 @@
  * @param <T> The class type to use as a model for the data contained in the children of the given Firebase location
  */
 public abstract class FirebaseListAdapter<T> extends BaseAdapter {
-
-    private final Class<T> mModelClass;
-    protected int mLayout;
-    protected Activity mActivity;
+    Activity mActivity;
+    final Class<T> mModelClass;
```
SUPERCILEX: Why is this final?

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
TheCraftKid: In terms of SOLID programming, shouldn't at least the class itself be at least public for it to be open to extension? Building another adapter as an add-on to FirebaseUI using FirebaseArray wouldn't be possible with this configuration.

SUPERCILEX: I didn't think about that. Thanks! I'll update this PR when I get home.

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
```diff
@@ -24,8 +24,8 @@
 /**
  * This class implements an array-like collection on top of a Firebase location.
  */
-class FirebaseArray implements ChildEventListener {
-    public interface OnChangedListener {
+public class FirebaseArray implements ChildEventListener {
```
SUPERCILEX: The class is now public so that people can extend it and everything is `protected` for the same reason.

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
```diff
@@ -41,15 +41,15 @@ public FirebaseArray(Query ref) {
         mQuery.addChildEventListener(this);
```
SUPERCILEX: `public FirebaseArray(Query ref) {`

Leaving this public means anyone can create new `FirebaseArray`s which can lead to interesting usages such as #337.

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
SUPERCILEX: Oh and I forgot to mention: abstraction from `ArrayList<DataSnapshot> mSnapshots` is still maintained with this new model.

SUPERCILEX:
@TheCraftKid Thanks for catching that! I thought these methods were no modifier originally. Anyway, they're back to normal again.

MaciejCiemiega:
I would also vote for it. It would allow developers to implement many usable sub classes that are otherwise not possible without copying the code, which is not an elegant solution...

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
```diff
@@ -24,7 +24,7 @@
 /**
  * This class implements an array-like collection on top of a Firebase location.
  */
-class FirebaseArray implements ChildEventListener {
+public class FirebaseArray implements ChildEventListener {
```
puf: I so far consider `FirebaseArray` to *not* be a part of the FirebaseUI public API. It's fine to discuss whether my consideration is valid, but that discussion is best had in an issue and not in a PR. See for example #206 and #337 (which I haven't responded to yet).

database/src/main/java/com/firebase/ui/database/FirebaseListAdapter.java
```diff
@@ -50,11 +50,10 @@
...
  */
 public abstract class FirebaseListAdapter<T> extends BaseAdapter {
-
-    private final Class<T> mModelClass;
-    protected int mLayout;
     protected Activity mActivity;
-    FirebaseArray mSnapshots;
+    protected final Class<T> mModelClass;
+    protected int mLayout;
+    protected FirebaseArray mSnapshots;
```
puf: The `mSnapshots` field should not be modified by subclasses, hence it should not be accessible. All the pertinent info of snapshots is already available via `getItem()`, `getRef()` and `getRef()`.

database/src/main/java/com/firebase/ui/database/FirebaseListAdapter.java
```diff
@@ -50,11 +50,10 @@
  * @param <T> The class type to use as a model for the data contained in the children of the given Firebase location
  */
 public abstract class FirebaseListAdapter<T> extends BaseAdapter {
-
-    private final Class<T> mModelClass;
-    protected int mLayout;
     protected Activity mActivity;
-    FirebaseArray mSnapshots;
+    protected final Class<T> mModelClass;
```
puf: I'm more inclined to make `mLayout` private than to make `mModelClass` protected. The logic is the same: subclasses should not be modifying these fields as it may lead to unexpected behavior.

database/src/main/java/com/firebase/ui/database/FirebaseRecyclerAdapter.java
```diff
@@ -70,11 +70,10 @@
...
  */
 public abstract class FirebaseRecyclerAdapter<T, VH extends RecyclerView.ViewHolder> extends RecyclerView.Adapter<VH> {
-
-    Class<T> mModelClass;
+    protected Class<T> mModelClass;
     protected int mModelLayout;
-    Class<VH> mViewHolderClass;
-    FirebaseArray mSnapshots;
+    protected Class<VH> mViewHolderClass;
+    protected FirebaseArray mSnapshots;
```
puf: Same as for the other adapter: the fields should not be accessible to subclasses, as those might otherwise modify them. The pertinent information from the fields should already be available through getters.

database/src/main/java/com/firebase/ui/database/FirebaseListAdapter.java
SUPERCILEX: Done.

database/src/main/java/com/firebase/ui/database/FirebaseListAdapter.java
SUPERCILEX: Done.

database/src/main/java/com/firebase/ui/database/FirebaseRecyclerAdapter.java
SUPERCILEX: And done.

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
SUPERCILEX: I reverted to making it no modifier. If this is the case, I think all of `FirebaseArray`'s methods should also be no modifier since `FirebaseArray` will never be used outside of the package.

database/src/main/java/com/firebase/ui/database/FirebaseRecyclerAdapter.java
TheCraftKid: This does leave me wondering how multiple ViewHolders and models could be used. Maybe I'll mess around with that today.

puf:
I would probably start with a new adapter class for that. The benefit of
reusing code that you get by inheriting from the existing
FirebaseRecyclerAdapter may not outweigh having to work around the
difference in implementation.

SUPERCILEX:
@puf I've updated this PR since #276 was merged. What do you think? (mostly, I just made stuff `protected`)

samtstern:
@puf this is a very old PR with a small footprint, looks like @SUPERCILEX even updated it recently after #276 was merged.  Could you two make a yes/no decision on this and either merge or close it?

puf:
lgtm, but apparently github has a special button for that now. :-)

SUPERCILEX:
Awesome!

Issue #96, opened 2016-03-17
FirebaseRecyclerAdapter has overly strong cast


ScottCypher:
`onCreateViewHolder` casts the inflated `View` to a `ViewGroup` which causes a crash if your layout is simple (e.g. a `TextView` with no `ViewGroup` around it).  The constructor for `RecyclerView.ViewHolder` only requires a `View`, so this cast seems unnecessary.

puf:
Thanks for entering Scott! As discussed at Droidcon today, we'll need to find a way that keeps the simple model we have for the most common use-case today, while adding support for this use-case too.

SUPERCILEX:
This issue should now be fixed in #307.

samtstern:
@SUPERCILEX good catch

PR #307, opened 2016-09-16
Removed casting of inflated layout to ViewGroup

MaciejCiemiega:
Hello,
in current implementation this cast prevents from inflating just simple `TextView` as a list item (such as `android.R.layout.simple_list_item_1` etc.).

### DIFF
--- a/database/src/main/java/com/firebase/ui/database/FirebaseRecyclerAdapter.java
+++ b/database/src/main/java/com/firebase/ui/database/FirebaseRecyclerAdapter.java
@@ -164,7 +164,7 @@ public abstract class FirebaseRecyclerAdapter<T, VH extends RecyclerView.ViewHol
 
     @Override
     public VH onCreateViewHolder(ViewGroup parent, int viewType) {
-        ViewGroup view = (ViewGroup) LayoutInflater.from(parent.getContext()).inflate(viewType, parent, false);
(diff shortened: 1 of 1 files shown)


MaciejCiemiega:
I signed it!

samtstern:
@puf I remember a while ago you explained why it had to be this way.  Is that still the case?  If so could you enlighten us?

puf:
@samtstern that was in #96 . But I'm wondering if that case wasn't different, since here the `View`/`Viewgroup` seem local to `onCreateViewHolder()`

MaciejCiemiega:
It looks like it was like that since the beginning:
https://github.com/firebase/FirebaseUI-Android/blob/2d342d2f9d316cf02dcc21e53ac8d1b73cd3b40d/library/src/main/java/com/firebase/ui/FirebaseRecyclerViewAdapter.java#L128

samtstern:
@puf I suggest merging this one, we immediately do this:
```
Constructor<VH> constructor = mViewHolderClass.getConstructor(View.class);
```

Which means we are passing the `ViewGroup` to a method that takes a `View`, and we don't use the `ViewGroup` for any other purpose so I see no reason not to relax this.

puf:
I've been starting at this one for a while now, but can't remember why it seemed a bad idea before. The ViewGroup doesn't even escape from the method. Maybe that was another change after all. @samtstern: can you merge?

Issue #286, opened 2016-09-06
Return access / ID token after successful sign-in with federated provider


samtstern:
After signing in with Google/Facebook/Twitter/etc the developer should have access to the relevant token, such as the Google ID token or the Facebook access token.

juanlet:
Great Suggestion, I was just searching how to do this when I realized the token was not coming.....

Issue #241, opened 2016-08-07
Show internet required dialog in case of no internet connection


peterhav:
Using 0.4.4 of this library (and earlier), when starting the Sign In intent without internet, the behavior is as follows:

For a mail sign in: after clicking next on adding the user mail, nothing happens
For Google sign in: after clicking the sign in button nothing happens

I would propose to show a 'internet connection required' dialog at the moment the Sign In intent is started without having an internet connection available.

samtstern:
@peterhav this is a good idea.  It's worth considering if FirebaseUI should show a dialog or if it should return an error code to the calling app and allow that app to decide on error UI.

BoogieMAN2K:
I would prefer an error messages, that way you can customize your response to the error. 
El 8 ago. 2016 10:46, Sam Stern <notifications@github.com> escribió:@peterhav this is a good idea.  It's worth considering if FirebaseUI should show a dialog or if it should return an error code to the calling app and allow that app to decide on error UI.

hijamoya:
I need this function, too!

PR #308, opened 2016-09-16
Check for null password or lack of network

samtstern:
Fixes for:
  * #241 
  * #286


samtstern:
@amandle please review.

samtstern:
@puf or @amandle PTAL

Issue #254, opened 2016-08-15
Refactor SaveCredentialsActivity


samtstern:
SaveCredentialsActivity has no UI and therefore does not need to be an Activity at all.  It could be embedded as a Fragment or refactored in some other way to avoid the janky UI of starting a blank Activity.

peterhav:
+1 for this!

samtstern:
It seems that `ChooseAccountActivity` is the same, it is a no-UI activity that only serves to host a `GoogleApiClient` and interact with the Credentials API.

SUPERCILEX:
`IDPSignInContainerActivity` is in the same boat.

SUPERCILEX:
`EmailHintContainerActivity` too

SUPERCILEX:
I think `WelcomeBackIDPPrompt` should be removed as well since the user only has the option to proceed. Also, I don't think it is necessary for them to know they already have an account: all a user would really care about is just signing in.

Removing this would prevent extra steps for #309 when the user links an account with smart lock enabled.

SUPERCILEX:
@samtstern @amandle Any objections to me removing `WelcomeBackIdpPrompt`? See https://github.com/firebase/FirebaseUI-Android/issues/254#issuecomment-252446320.

amandle:
@SUPERCILEX if the user chose to sign in with facebook, then signed in I think it would be surprising to see the Google Sign In picker pop up. I think we need to explain why they are seeing it.

SUPERCILEX:
@amandle what do you think of only showing the prompt if the provider isn't email or Google?

amandle:
@SUPERCILEX I still think it needs to be shown since otherwise it will behave surprisingly without an explanation.  I don't think email or Google are special cases here and will still need the prompt.

SUPERCILEX:
@amandle agreed. I'll have to take out some of the sign-in logic in #309.

Issue #228, opened 2016-07-31
Support for multiple accounts-Multiple providers login need it 


juanlet:
The login behaviour it's exclusive. If you login with facebook you can't login with mail and viceversa. Same happens with the other providers.

I'm doing an app in which I need to log in the user even if they don't have facebook but I need to have the option of adding a provider on top of the one which you used for logging in, cause in case the guy logs in with email I need to be able to connect to facebook even after logging in....Maybe I can already do that and it's not documented it, can someone clarify me that?

alfongj:
This is not currently supported through this library. 

You can manually link different credentials by using FirebaseAuth#linkWithCredential. Take a look at the guide [here](https://firebase.google.com/docs/auth/android/account-linking)

raoravik:
It will be really great to have this feature. 

Scenario: User has logged in using email/google where the email address is abc@abc.com. Now at a later stage if the user is logging with facebook and we get the same email abc@abc.com, it should automatically merge the accounts and login successfully.

kustra:
I see there is actually a screen for this feature, but it doesn't seem to handle the case when there are no password-based credentials for the user.

Steps to reproduce the issue:
1. authenticate with a user using Google login. Don't create e-mail/password based credentials for the user.
2. authenticate with Facebook, set up so that it provides the same e-mail address as the Google login.
3. a screen will pop up, saying that there is already a user with the given e-mail address, and I should type his password to merge accounts. But I have no password to type. If I type anything, the app just jumps back to the account chooser activity and logs this stack trace to the console:

<pre>09-09 13:35:11.901 9683-9683/my.app.id W/WelcomeBackPassword: Error signing in with email and password
  com.google.firebase.auth.FirebaseAuthInvalidCredentialsException: The password is invalid or the user does not have a password.
      at com.google.android.gms.internal.zzafd.zzes(Unknown Source)
      at com.google.android.gms.internal.zzafa$zzg.zza(Unknown Source)
      at com.google.android.gms.internal.zzafl.zzet(Unknown Source)
      at com.google.android.gms.internal.zzafl$zza.onFailure(Unknown Source)
      at com.google.android.gms.internal.zzafg$zza.onTransact(Unknown Source)
      at android.os.Binder.execTransact(Binder.java:453)
</pre>

I have the "One account per email address" feature enabled in Firebase Console.

alfongj:
@kustra Looks like a bug. We will investigate.

amandle:
[...]

Issue #123, opened 2016-05-19
Anonymous user and account conversion to permanent account


danielkummer:
Hi there

Is there a way in which AuthUI supports the conversion of a anonymous account to a permanent one (one of the supplied auth providers)
As stated in the [documentation](https://firebase.google.com/docs/auth/android/anonymous-auth#authenticate-with-firebase-anonymously) and as I understand it, one must use the `linkWithCredential` method instead of the "normal" `signInWith` flow.

Can this be done with the current AuthUI version somehow?

iainmcgin:
Hi Daniel,

We don't have any direct support in FirebaseUI for upgrading anonymous accounts to an account backed by an authentication method just yet. You can do it directly with the Firebase Auth APIs, see [here](https://firebase.google.com/docs/auth/android/anonymous-auth#authenticate-with-firebase-anonymously) in the section "Convert an anonymous account to a permanent account".

I'll leave this issue open as a feature enhancement request; we can likely handle this automatically within FirebaseUI if an existing anonymous user is present in `auth.getCurrentUser()` when sign in occurs during the authentication flow.

danielkummer:
Hi @iainmcgin - thanks for your reply (and considering it as a feature enhancement :) ) 

I've read through the documentation you mentioned - maybe you can answer me this: Is there a way to get the **AuthCredential** object from AuthUI which was used by the user to sign-in? As I'm reading the docs, without this object instantiated with the users credentials there's not way to link accounts...

iainmcgin:
Ah, we have a slight problem here after discussing this with some colleagues. There is deliberately no way to get back an AuthCredential from an existing user, to avoid potentially leaking credentials. So, the linking of the authentication method to the anonymous account would _have_ to occur within Firebase UI if you are using the library. I'll take a look at this next week to get you unblocked ASAP.

danielkummer:
Great - looking forward to hearing from you 👍  (And thanks a lot!)

guillermomuntaner:
Facing similar case. My app uses default anonymous login to provide a frictionless experience while being able to store user data. Log-in (non anonymously) is a secondary optional feature if the user wants to save his data between devices or access specific features. So my app doesn't really have a true logout status and instead rely on a anon/non-anon statuses. 

**1 - Simple scenario;** 
[...]

PR #309, opened 2016-09-17
Support account linking

SUPERCILEX:
If this has been discussed in an issue, make sure to mention the issue number here.  If not, go file an issue about this to make sure this is a desirable change.
- https://github.com/firebase/FirebaseUI-Android/issues/123
- https://github.com/firebase/FirebaseUI-Android/issues/228

If this is a new feature please co-ordinate with someone on [FirebaseUI-iOS](https://github.com/firebase/firebaseui-ios) to make sure that we can implement this on both platforms and maintain feature parity.
- Anyone?

TODO:
- [x] Make sure linking non-anonymous accounts works
- [x] Add documentation
- [x] Update sample
- [x] Wait for #254 to get merged
- [x] Add tests
- [x] Update documentation to `IdpResponse`
- [x] Fix profile and name not being added (turns out this is also caused by #409)


auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -309,6 +311,10 @@ public SignInIntentBuilder createSignInIntentBuilder() {
         return new SignInIntentBuilder();
     }
 
+    public void notifyOnMergeFailedListeners(String prevUid) {
```
SUPERCILEX: I'm pretty sure this is not okay, but I don't know how to get rid of it. How would I serialize the listener into `FlowParameters`?

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -417,6 +425,15 @@ public SignInIntentBuilder setIsSmartLockEnabled(boolean enabled) {
             return this;
         }
 
+        /**
+         * Links the current user to a new one.
```
SUPERCILEX: TODO: bad wording

auth/src/main/java/com/firebase/ui/auth/ui/account_link/WelcomeBackIDPPrompt.java
```diff
@@ -186,7 +194,33 @@ public void onComplete(@NonNull Task<AuthResult> task) {
...
                     .addOnFailureListener(
                             new TaskFailureLogger(TAG, "Error linking with credential"))
-                    .addOnCompleteListener(new FinishListener());
+                    .addOnCompleteListener(new OnCompleteListener<AuthResult>() {
+                        @Override
+                        public void onComplete(@NonNull Task<AuthResult> task) {
+                            FirebaseAuth auth = FirebaseAuth.getInstance();
+                            FirebaseUser user = auth.getCurrentUser();
+                            if (!task.isSuccessful() && user != null && mPrevCredential != null) {
+                                user.delete();
```
SUPERCILEX: This is terrible idea, but `user` is the new user if I call `delete()` in the `OnCompleteListener` because the variable is a reference.

auth/src/main/java/com/firebase/ui/auth/ui/account_link/WelcomeBackIDPPrompt.java
```diff
@@ -186,7 +194,33 @@ public void onComplete(@NonNull Task<AuthResult> task) {
...
+                                auth.signInWithCredential(mPrevCredential)
+                                        .addOnFailureListener(
+                                                new TaskFailureLogger(
+                                                        TAG, "Error linking with credential"))
+                                        .addOnCompleteListener(new OnCompleteListener<AuthResult>() {
+                                            @Override
+                                            public void onComplete(@NonNull Task<AuthResult> task) {
+                                                if (task.isSuccessful()) {
+                                                    AuthUI.getInstance()
+                                                            .notifyOnMergeFailedListeners(prevUid);
```
SUPERCILEX: This is where I want to call `user.delete()`.

auth/src/main/java/com/firebase/ui/auth/ui/account_link/WelcomeBackIDPPrompt.java
```diff
@@ -186,7 +194,33 @@ public void onComplete(@NonNull Task<AuthResult> task) {
...
                             new TaskFailureLogger(TAG, "Error linking with credential"))
-                    .addOnCompleteListener(new FinishListener());
+                    .addOnCompleteListener(new OnCompleteListener<AuthResult>() {
+                        @Override
+                        public void onComplete(@NonNull Task<AuthResult> task) {
+                            FirebaseAuth auth = FirebaseAuth.getInstance();
+                            FirebaseUser user = auth.getCurrentUser();
+                            if (!task.isSuccessful() && user != null && mPrevCredential != null) {
+                                user.delete();
+                                final String prevUid = user.getUid();
```
SUPERCILEX: Need this because of java references.

auth/src/test/java/com/firebase/ui/auth/test_helpers/TestHelper.java
```diff
@@ -57,7 +57,8 @@ public static FlowParameters getFlowParameters(Context context, List<String> pro
                 AuthUI.getDefaultTheme(),
                 AuthUI.NO_LOGO,
                 null,
-                true);
+                true,
+                false);
```
SUPERCILEX: We should probably make a test for this.

SUPERCILEX:
I have one bug that I know of that isn't fixed yet: when going from an anonymous account that doesn't have a name or profile picture to a Google account, the profile picture and name don't get added to the account. Only the email gets added. Does anyone know why this is happening? I could just add a `UserProfileChangeRequest` once the account is converted, but that complicates things.

samtstern:
@SUPERCILEX thank you (again!) for the big contribution, we really appreciate how involved you are with this repo lately.  I will take a look and leave comments, but as you know, we try to keep FirebaseUI-Android in sync with FirebaseUI-iOS, so before this feature could be released we'd have to get some commitment there.

That said if things look good we can always put this on a feature branch and merge it back into `master` or `version-1.0.0-dev` later.

app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
```diff
@@ -97,15 +105,18 @@
...
-            finish();
-        }
-
         setContentView(R.layout.auth_ui_layout);
         ButterKnife.bind(this);
 
+        FirebaseUser user = FirebaseAuth.getInstance().getCurrentUser();
+        if (user != null) {
+            if (!user.isAnonymous()) {
+                startActivity(SignedInActivity.createIntent(this));
```
samtstern: Let's discuss changes to the sample last, I want to focus on the changes to the library at first since that will have effects on how the sample is structured.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -417,6 +425,15 @@ public SignInIntentBuilder setIsSmartLockEnabled(boolean enabled) {
...
+         */
+        public SignInIntentBuilder linkWithCurrentUser(OnMergeFailedListener onMergeFailedListener) {
```
samtstern: I find the pattern of adding a listener to an `Intent` builder strange. The result of the action in the `Intent` should be returned via `onActivityResult` if the `Activity` that fired it did so for result.

auth/src/main/java/com/firebase/ui/auth/ui/account_link/WelcomeBackIDPPrompt.java
samtstern: Can you explain exactly what you're doing in this bit?  I'm not clear on what you are trying to achieve.

auth/src/main/java/com/firebase/ui/auth/ui/account_link/WelcomeBackIDPPrompt.java
samtstern: In this block you are notifying the merge failed listeners on successful sign in, why is that the case?

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailActivity.java
```diff
@@ -135,11 +137,30 @@ public void onClick(View view) {
...
+                                             }
+                                         }), name, password);
+        } else {
+            addListeners(firebaseAuth.createUserWithEmailAndPassword(email, password),
+                         name,
+                         password);
+        }
+    }
+
+    private void addListeners(Task<AuthResult> task, final String name, final String password) {
```
samtstern: IMO this is a confusing abstraction, why not just add the listeners directly as before?

auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
```diff
@@ -154,17 +155,28 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+        CredentialSignInHandler handler = new CredentialSignInHandler(
+                AuthMethodPickerActivity.this,
+                mActivityHelper,
+                RC_ACCOUNT_LINK,
+                RC_SAVE_CREDENTIAL,
+                response);
+
+        if (user != null && mActivityHelper.getFlowParams().shouldLinkUser) {
+            user
+                    .linkWithCredential(credential)
```
samtstern: style nit: prefer not to have a continuation for the first method call in a chain if it would fit on the previous line.  `user` is just kinda floating out there.

auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
```diff
@@ -154,17 +155,28 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+                    .addOnFailureListener(logger)
+                    .addOnCompleteListener(handler);
+        } else {
+            firebaseAuth
+                    .signInWithCredential(credential)
```
samtstern: Seems like in this `if/else` block you are attaching the same complete/failure listener but switching on the Task type.  Would be clearer to do:

```java
// Either sign the user in or link the credential
Task<AuthResult> authTask;
if (user != null && mActivityHelper.getFlowParams().shouldLinkUser) {
  authTask = user.linkWithCredential(credential);
} else {
  authTask = firebaseAuth.signInWithCredential(credential);
}

task.addOnFailureListener(logger);
task.addOnCompleteListener(handler);
```

auth/src/main/java/com/firebase/ui/auth/util/OnMergeFailedListener.java
```diff
@@ -0,0 +1,11 @@
+package com.firebase.ui.auth.util;
+
+public interface OnMergeFailedListener {
```
samtstern: Building on my previous comment, I don't think we need to use the listener pattern for merge failures.  The only information returned on a merge failure is the previous user ID, so why not just do it through Activity results? The developer could handle it like this?

```java
@Override
public void onActivityResult(int requestCode, int resultCode, Intent data) {

  if (requestCode == RC_SIGN_IN) {
    if (resultCode == RESULT_MERGE_FAILED) {
      String prevUid = data.getStringExtra(EXTRA_PREV_UID);
      // Handle merge failure ....
    }
  }
}
```

This also solves the issue you raised of needing to somehow store the listener reference in the FlowParams, which won't work (as you mentioned).

SUPERCILEX:
`SignInActivity` test failing at
```java
assertEquals(
                TestConstants.EMAIL,
                nextIntent.intent.getExtras().getString(ExtraConstants.EXTRA_EMAIL)
```

Not sure why...

auth/src/main/java/com/firebase/ui/auth/ui/account_link/WelcomeBackIDPPrompt.java
```diff
@@ -186,7 +194,32 @@ public void onComplete(@NonNull Task<AuthResult> task) {
             authResultTask
                     .addOnFailureListener(
                             new TaskFailureLogger(TAG, "Error linking with credential"))
-                    .addOnCompleteListener(new FinishListener());
+                    .addOnCompleteListener(new OnCompleteListener<AuthResult>() {
+                        @Override
+                        public void onComplete(@NonNull Task<AuthResult> task) {
+                            FirebaseAuth auth = FirebaseAuth.getInstance();
+                            if (!task.isSuccessful() && task.getException() instanceof FirebaseAuthUserCollisionException) {
+                                setIntent(new Intent().putExtras(mActivityHelper.getMergeFailedIntent()));
```
SUPERCILEX: The is the new way I'm managing sending the old uid.

SUPERCILEX: The new version is a little bit different, but I'm basically doing the same thing as before.
If the task isn't successful, it means the link failed so I need to delete the anonymous user and get the old uid. `delete()` also signs out the user so I then need to reauthenticate the user and pass on the intent containing the old uid to be manually merged. This scenario only happens when the Google/Facebook/Twitter etc. account already exists.

auth/src/main/java/com/firebase/ui/auth/ui/ChooseAccountActivity.java
```diff
@@ -219,14 +219,13 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
             case RC_IDP_SIGNIN:
             case RC_AUTH_METHOD_PICKER:
             case RC_EMAIL_FLOW:
-                finish(resultCode, new Intent());
+                finish(resultCode, data);
                 break;
             case RC_PLAY_SERVICES:
                 if (resultCode != RESULT_OK) {
-                    finish(resultCode, new Intent());
```
samtstern: You can use these intents to plumb data back through to the calling activities, I think you're doing the right thign.

app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
samtstern: Now that you've refactored your approach, could you update the sample to show how it would look?  Don't worry about doing a perfect job, I just want to see a usage example that compiles

app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
```diff
@@ -150,6 +161,42 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
     @MainThread
     private void handleSignInResponse(int resultCode, Intent data) {
         if (resultCode == RESULT_OK) {
+            final String prevUid = data.getStringExtra(ExtraConstants.EXTRA_MERGE_FAILED);
+            if (prevUid != null) {
+                FirebaseDatabase.getInstance()
```
samtstern: Can we remove all this?  There shouldn't be anything referencing the DB or the chat sample in the AuthUI sample.

app/src/main/java/com/firebase/uidemo/database/Chat.java
```diff
@@ -0,0 +1,28 @@
+package com.firebase.uidemo.database;
```
samtstern: nit: let's just leave chat where it was.  We don't need to access it outside of that context anyway.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -410,13 +412,23 @@ public SignInIntentBuilder setProviders(@NonNull String... providers) {
...
             mIsSmartLockEnabled = enabled;
             return this;
         }
 
+        /**
+         * Links the current user to one created in the sign in flow.
+         *
+         * <p>Linking is disabled by default.
+         */
+        public SignInIntentBuilder linkWithCurrentUser(boolean shouldLinkUser) {
```
samtstern: nit: name this `setShouldLinkUser` or something more consistent with `setIsSmartLockEnabled` and the others.

auth/src/main/java/com/firebase/ui/auth/ui/ActivityHelper.java
```diff
@@ -96,6 +96,16 @@ public FirebaseUser getCurrentUser() {
         return getFirebaseAuth().getCurrentUser();
     }
 
+    public Intent getMergeFailedIntent() {
+        FirebaseUser user = getCurrentUser();
+        if (mFlowParams.shouldLinkUser && user != null) {
+            if (user.isAnonymous()) user.delete();
```
samtstern: I don't think we want to delete the anonymous user.

auth/src/main/java/com/firebase/ui/auth/ui/ChooseAccountActivity.java
samtstern: Your change is correct, the `new Intent()` was in the absence of anything important to plumb down.

auth/src/main/java/com/firebase/ui/auth/ui/account_link/WelcomeBackIDPPrompt.java
samtstern: nit: any reason not to just drop the `new Intent()` piece?

auth/src/main/java/com/firebase/ui/auth/ui/account_link/WelcomeBackIDPPrompt.java
```diff
@@ -186,7 +194,32 @@ public void onComplete(@NonNull Task<AuthResult> task) {
             authResultTask
                     .addOnFailureListener(
                             new TaskFailureLogger(TAG, "Error linking with credential"))
-                    .addOnCompleteListener(new FinishListener());
+                    .addOnCompleteListener(new OnCompleteListener<AuthResult>() {
+                        @Override
+                        public void onComplete(@NonNull Task<AuthResult> task) {
+                            FirebaseAuth auth = FirebaseAuth.getInstance();
+                            if (!task.isSuccessful() && task.getException() instanceof FirebaseAuthUserCollisionException) {
```
samtstern: nit: prefer separate `onSuccess` and `onFailure` listeners here, rather than checking `isSuccessful()` in the `onCompleteListener`

auth/src/main/java/com/firebase/ui/auth/ui/idp/CredentialSignInHandler.java
```diff
@@ -77,15 +78,19 @@ public void onFailure(@NonNull Exception e) {
...
         } else {
             mActivityHelper.dismissDialog();
 
             FirebaseUser firebaseUser = task.getResult().getUser();
-            SmartlockUtil.saveCredentialOrFinish(mActivity, mSaveCredentialsResultCode,
-                    mActivityHelper.getFlowParams(), firebaseUser,
-                    null /* password */, mResponse.getProviderType());
+            SmartlockUtil.saveCredentialOrFinish(mActivity,
+                                                 mSaveCredentialsResultCode,
+                                                 new Intent(),
```
samtstern: nit: I'd prefer to pass `null /* intent */` here and have `saveCredentialOrFinish` handle the `null` case.

app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
SUPERCILEX: We can remove it, but I think it should be moved to the documentation so that beginners know what to do with the previous user id.

auth/src/main/java/com/firebase/ui/auth/ui/ActivityHelper.java
SUPERCILEX: I'm deleting the anonymous user here because the expected usage of `getMergeFailedIntent()` is that a non anonymous user will be signed into immediately after this method is called. If I don't delete the anonymous users, there will be a bunch of them hanging around that will never be user again. AFAIK, you can't log back into an existing anonymous user so not deleting them would make a messy user base in the Firebase console.

What I would really like to do is to delete the anonymous user once I'm signed in, but I can't figure out how to do that because I keep getting references to the newly signed in user.

If you believe we should play it safe and not delete the anonymous user, that's fine with me but I had one last idea: maybe I could delete the anonymous user and if signing into the new user fails, I just create a new anonymous user and pass down the old anonymous user's id through the Intent. What do you think? It would probably just be easier to not delete the anonymous user, but again, I take issue with the messy Firebase console. Do you know if Firebase automatically delete anonymous users if they haven't logged in for a while? That would solve the issue.

auth/src/main/java/com/firebase/ui/auth/util/SmartlockUtil.java
```diff
@@ -25,31 +24,35 @@
...
     public static void saveCredentialOrFinish(Activity activity,
                                               int requestCode,
+                                              @Nullable Intent data,
                                               FlowParameters parameters,
                                               FirebaseUser firebaseUser,
                                               @Nullable String password,
                                               @Nullable String provider) {
+        if (data != null) {
+            data = new Intent();
+        }
```
SUPERCILEX: Is this what you were looking for?

SUPERCILEX:
Any idea why the test is failing?

SUPERCILEX:
Figured out why the test was failing: it might be a bug in the Android framework.

This works:
```java
data.putExtras(SaveCredentialsActivity.createIntent(
    activity,
    parameters,
    firebaseUser,
    password,
    provider));
```
This doesn't:
```java
SaveCredentialsActivity.createIntent(
    activity,
    parameters,
    firebaseUser,
    password,
    provider).putExtras(data);
```
I have no idea why one works, but not the other. Any thoughts?

auth/src/main/java/com/firebase/ui/auth/util/SmartlockUtil.java
```diff
@@ -12,44 +12,54 @@
...
     /**
      * If SmartLock is enabled and Google Play Services is available, start the save credential
      * Activity. Otherwise, finish the calling Activity with RESULT_OK.
-     * @param activity the calling Activity.
-     * @param requestCode request code to use when starting the save operation.
-     * @param parameters calling Activity flow parameters.
+     *
+     * @param activity     the calling Activity.
+     * @param requestCode  request code to use when starting the save operation.
+     * @param data         An Intent to be merged. Must be a {@code new Intent().putExtras(...)}.
```
SUPERCILEX: I'm really not a fan of this: I had to switch everything to be a `new Intent()` to workaround the merge bug. Do you think we should just make a parameter `prevUid`?

app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
samtstern: Ok that's fine, documentation comes last anyway.  Let's save that for another time.

auth/src/main/java/com/firebase/ui/auth/util/SmartlockUtil.java
samtstern: Can you explain a little what happened when you did the merge the other way?  What's the buggy behavior you're seeing?

app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
SUPERCILEX: K

auth/src/main/java/com/firebase/ui/auth/ui/ActivityHelper.java
SUPERCILEX: Done!

auth/src/main/java/com/firebase/ui/auth/util/SmartlockUtil.java
SUPERCILEX: See pics:

The two differeent ways of merging intents:
image

The left intent merge up top produces the left version below and vice versa:
image

So basically, the left version works in terms of keeping the email alive, but the it doesn't in the sense that it says it comes from the class that it originated from eg. `SignInActivity` instead of `SaveCredentialsActivity` (the test expects `SaveCredentialsActivity`).

samtstern: Ok I just looked through some Android source and here's whats happening.  When you merge the extras of two bundles (say `dst.Extras(src)`) it basically does this (pseudocode):

```
for (String key : src.getExtras().getKeys()) {
  dst.putExtra(key, src.getExtra(key));
}
```

This means for a given `key` if `src.getExtra(key)` is `null` then it will erase the extra with the same key in `dst`.  I think that's why your email value is disappearing.

Maybe you want to create a utility method to merge extras only if they're non-null?  I don't want to reverse the order of operations here just to get around the bug, as you said it makes the result of the intent counter-intuitive.

samtstern:
@amandle can you take a look at this CL as well?  It's very complex and I would like a second set of eyes.

auth/src/main/java/com/firebase/ui/auth/util/SmartlockUtil.java
SUPERCILEX: Yeah, I fixed it in a newer diff.

auth/src/main/java/com/firebase/ui/auth/util/SmartlockUtil.java
samtstern: Oooh yeah that's not gonna be pretty, you're right.  Not sure there's any way around that.

Maybe let's find out why there's a null extra in this situation and just prevent it from happening in the first place?

SUPERCILEX: YESSSSS!!!! It took me like a half hour of debugging, but it finally dawned on me to actually look at what the test was doing:

Real life in `AcquireEmailHelper`:
```java
if (email != null && !email.isEmpty()) {
    // Only proceeds if email is not null
}
```

Test:
```java
Intent startIntent = SignInActivity.createIntent(
        RuntimeEnvironment.application,
        TestHelper.getFlowParameters(
                RuntimeEnvironment.application,
                Collections.<String>emptyList()),
        ***null***);
```

So anyway, I fixed the test and life is good again. The end.

samtstern: @SUPERCILEX :+1: :boom:

auth/src/main/java/com/firebase/ui/auth/ui/account_link/WelcomeBackIDPPrompt.java
```diff
@@ -186,7 +194,32 @@ public void onComplete(@NonNull Task<AuthResult> task) {
...
+                                auth.signInWithCredential(mPrevCredential != null ? mPrevCredential : newCredential)
+                                        .addOnFailureListener(
+                                                new TaskFailureLogger(
+                                                        TAG, "Error linking with credential"))
+                                        .addOnCompleteListener(new OnCompleteListener<AuthResult>() {
+                                            @Override
+                                            public void onComplete(@NonNull Task<AuthResult> task) {
+                                                if (task.isSuccessful()) {
+                                                    mActivityHelper.dismissDialog();
+                                                    finish(Activity.RESULT_OK, getIntent());
```
amandle: I'm actually not sure what's going on here. It looks like if we fail to link with a credential we try to sign in anyway.  If the sign in succeeds (but the linking did not) we say it was a success?

auth/src/main/java/com/firebase/ui/auth/ui/account_link/WelcomeBackPasswordPrompt.java
```diff
@@ -114,6 +113,7 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
 
     private void next(String email, final String password) {
         final FirebaseAuth firebaseAuth = mActivityHelper.getFirebaseAuth();
+        setIntent(getIntent().putExtras(mActivityHelper.getMergeFailedIntent()));
```
amandle: I think this might be clearer if this were closer to where it is used on line 145

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailActivity.java
```diff
@@ -138,15 +139,22 @@ public void onClick(View view) {
     private void registerUser(String email, final String name, final String password) {
         final FirebaseAuth firebaseAuth = mActivityHelper.getFirebaseAuth();
         // create the user
-        firebaseAuth.createUserWithEmailAndPassword(email, password)
-                .addOnFailureListener(new TaskFailureLogger(TAG, "Error creating user"))
+        FirebaseUser user = firebaseAuth.getCurrentUser();
+        Task<AuthResult> task;
+        setIntent(getIntent().putExtras(mActivityHelper.getMergeFailedIntent()));
```
amandle: this only needs to happen in the true case of the if statement on line 145

samtstern:
My "approval" is a hand off to @amandle who understands the logic best.  I just have the following final requests:
  * Undo the sample changes that integrate the DB, as mentioned in previous comments.
  * This should either be merged into a feature branch or the version-1.0.0-dev branch. Feature branch might be safest since iOS is not guaranteed to do this in 1.0.

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailActivity.java
SUPERCILEX: Woops! That's actually not supposed to happen at all since we're linking with a new credential (eg the account doesn't exist yet).

auth/src/main/java/com/firebase/ui/auth/ui/account_link/WelcomeBackPasswordPrompt.java
SUPERCILEX: It can't be used anywhere after `signInWithEmailAndPassword` because then I wouldn't be saving the old uid.

SUPERCILEX:
@amandle Because Sam was confused too, I added an explanation to what I'm doing in `WelcomeBackIDPPrompt.java`: https://github.com/firebase/FirebaseUI-Android/pull/309/files#diff-7c82ecab41019b31cff372a5f3480a02R212.

SUPERCILEX:
@samtstern Thanks for all your help so far!!!

 * I undid the changes in the sample
 * I created #344 because this PR comes from master and I wanted to make it clearer that master would be merged into v1.0.0 by merging this PR. Once #344 is merged, we can change the branch this PR is being merged into to v1.0.0 without the changes from master also showing up as changes from this PR.

SUPERCILEX:
The documentation and sample still have to be discussed.

SUPERCILEX:
TODO:

- [ ] Make sure linking non-anonymous account works
- [ ] Add documentation
- [ ] Update sample
- [ ] Add tests

auth/README.md
```diff
@@ -197,14 +197,27 @@ startActivityForResult(
     RC_SIGN_IN);
 ```
 
+If anonymous user account conversion to permanent is required:
```
SUPERCILEX: Thoughts on documentation?

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -470,13 +472,23 @@ public SignInIntentBuilder setProviders(@NonNull String... providers) {
...
             mIsSmartLockEnabled = enabled;
             return this;
         }
 
+        /**
+         * Links the current user to an account created in the sign-in flow.
+         *
+         * <p>Linking is disabled by default.
+         */
+        public SignInIntentBuilder setShouldLinkAccounts(boolean shouldLinkAccounts) {
```
SUPERCILEX: Thoughts on the new name? `setShouldLinkAccounts` vs `setShouldLinkUser`. I prefer the first but it's up to debate.

samtstern: I like the new name, since we call the feature "Account Linking"

auth/README.md
samtstern: I would expand this sentence as there are a lot of advanced concepts at play here.  Something like this:

"The default FirebaseUI sign-in flow shows UI to either create a new account or sign into an existing account.  If you are using anonymous authentication (link) in your application before calling FirebaseUI, you may want to link the anonymous account to the permanent account the user selects in the UI flow."

SUPERCILEX: Done

hijamoya:
any update of this feature? we need this function. Thanks!!

SUPERCILEX:
This waiting on #357

app/src/main/java/com/firebase/uidemo/auth/SignedInActivity.java
```diff
@@ -174,20 +177,33 @@ private void populateProfile() {
...
+            findViewById(R.id.idp_token_layout).setVisibility(View.GONE);
+        } else {
+            ((TextView) findViewById(R.id.idp_token)).setText(token);
+        }
+        if (secret == null) {
+            findViewById(R.id.idp_secret_layout).setVisibility(View.GONE);
+        } else {
+            ((TextView) findViewById(R.id.idp_secret)).setText(secret);
+        }
+    }
```
SUPERCILEX: I did some cleanup here so that if `IdpResponse` is `null`, we don't show the layouts.

app/src/main/java/com/firebase/uidemo/auth/SignedInActivity.java
```diff
@@ -174,20 +177,33 @@ private void populateProfile() {
...
+
+    private void populatePrevUid() {
+        IdpResponse idpResponse = getIntent().getParcelableExtra(ExtraConstants.EXTRA_IDP_RESPONSE);
+        String prevUid = null;
+        if (idpResponse != null) prevUid = idpResponse.getPrevUid();
+        if (prevUid == null) {
+            findViewById(R.id.prev_uid_layout).setVisibility(View.GONE);
+        } else {
+            ((TextView) findViewById(R.id.prev_uid)).setText(prevUid);
```
SUPERCILEX: Is this an appropriate update to the sample?

auth/src/main/java/com/firebase/ui/auth/IdpResponse.java
```diff
@@ -30,25 +30,35 @@
...
         mSecret = secret;
+        mPrevUid = prevUid;
+    }
+
+    public IdpResponse(IdpResponse response, String prevUid) {
+        this(response.getProviderType(),
+             response.getEmail(),
+             response.getIdpToken(),
+             response.getIdpSecret(),
+             prevUid);
```
SUPERCILEX: This is a mess, but I don't know how to get around it. I need to add the previous user id to an existing `IdpResponse` and I choose not to use a setter since this is public api. Any thoughts on a better way to do this? Should I be using a setter? Also, if we stick with this, what is the convention for this kind of constructor? (`response.getProviderType()` vs `response.mProviderId`)

auth/src/main/java/com/firebase/ui/auth/IdpResponse.java
```diff
@@ -92,11 +111,14 @@ public String getIdpSecret() {
     }
 
     /**
-     * Get the email used to sign in.
+     * For setShouldLinkAccounts(true) users only. Get the previous user id if a conflict occurred.
+     * See the <a href="https://github.com/SUPERCILEX/FirebaseUI-Android/blob/master/auth/README.md#handling-account-link-failures">README</a>
+     * for a much more detailed explanation.
+     *
+     * @see AuthUI.SignInIntentBuilder#setShouldLinkAccounts(boolean)
```
SUPERCILEX: Is this clear enough?

auth/src/main/java/com/firebase/ui/auth/ui/account_link/WelcomeBackPasswordPrompt.java
```diff
@@ -83,8 +83,6 @@ protected void onCreate(Bundle savedInstanceState) {
         TextView bodyTextView = ((TextView) findViewById(R.id.welcome_back_password_body));
         bodyTextView.setText(spannableStringBuilder);
 
-        ((TextInputLayout) findViewById(R.id.password_layout)).setPasswordVisibilityToggleEnabled(false);
```
SUPERCILEX: See the support lib [release notes](https://developer.android.com/topic/libraries/support-library/revisions.html). In 25.0.1, they set the default to false.

auth/src/main/java/com/firebase/ui/auth/ui/accountlink/WelcomeBackIdpPrompt.java
```diff
@@ -155,42 +156,62 @@ private void next(final IdpResponse newIdpResponse) {
...
+        FirebaseUser currentUser = mActivityHelper.getCurrentUser();
         if (currentUser == null) {
-            Task<AuthResult> authResultTask = firebaseAuth.signInWithCredential(newCredential);
-            authResultTask.addOnCompleteListener(new OnCompleteListener<AuthResult>() {
-                @Override
-                public void onComplete(@NonNull Task<AuthResult> task) {
-                    if (task.isSuccessful() && mPrevCredential != null) {
-                        FirebaseUser firebaseUser = task.getResult().getUser();
-                        firebaseUser.linkWithCredential(mPrevCredential);
-                        firebaseAuth.signOut();
```
SUPERCILEX: This is similar to what I removed in `WelcomeBackPasswordPrompt`.

auth/src/main/java/com/firebase/ui/auth/ui/idp/CredentialSignInHandler.java
```diff
@@ -42,19 +42,19 @@
...
             @Nullable SaveSmartLock smartLock,
-            int accountLinkResultCode,
+            int accountLinkRequestCode,
             IdpResponse response) {
         mActivity = activity;
         mHelper = helper;
         mSmartLock = smartLock;
         mResponse = response;
-        mAccountLinkResultCode = accountLinkResultCode;
+        mAccountLinkRequestCode = accountLinkRequestCode;
```
SUPERCILEX: Looks like a typo.

auth/src/main/java/com/firebase/ui/auth/IdpResponse.java
```diff
@@ -80,10 +68,12 @@ public static IdpResponse fromResultIntent(Intent resultIntent) {
         }
     }
 
+    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
```
SUPERCILEX: I ❤️ this annotation!!!

auth/src/main/java/com/firebase/ui/auth/ui/accountlink/AccountLinker.java
```diff
@@ -0,0 +1,149 @@
...
+import com.google.android.gms.tasks.OnFailureListener;
+import com.google.android.gms.tasks.OnSuccessListener;
+import com.google.android.gms.tasks.Task;
+import com.google.firebase.auth.AuthCredential;
+import com.google.firebase.auth.AuthResult;
+import com.google.firebase.auth.FirebaseAuthUserCollisionException;
+import com.google.firebase.auth.FirebaseUser;
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+public class AccountLinker implements OnSuccessListener<AuthResult>, OnFailureListener {
```
SUPERCILEX: Welcome to the bad boy... I'll try my best to answer your questions, but even I struggle to understand what's going on! (It works though)

auth/src/main/java/com/firebase/ui/auth/ui/idp/CredentialSignInHandler.java
```diff
@@ -82,37 +93,52 @@ public void onComplete(@NonNull Task<AuthResult> task) {
                             .addOnFailureListener(new OnFailureListener() {
                                 @Override
                                 public void onFailure(@NonNull Exception e) {
-                                    // TODO: What to do when signing in with Credential fails
-                                    // and we can't continue to Welcome back flow without
-                                    // knowing providers?
+                                    mHelper.finishActivity(
+                                            mActivity,
+                                            ResultCodes.CANCELED,
+                                            IdpResponse.getErrorCodeIntent(ErrorCodes.UNKNOWN_ERROR));
```
SUPERCILEX: I believe this is an adequate way to handle the failure.

auth/src/main/java/com/firebase/ui/auth/ui/idp/CredentialSignInHandler.java
```diff
@@ -82,37 +93,52 @@ public void onComplete(@NonNull Task<AuthResult> task) {
...
+                        mResponse,
+                        credential,
+                        null);
+                return;
+            }
+
             mHelper.dismissDialog();
 
-            String provider = result.getProviders().get(0);
+            String provider = getHighestPriorityProvider(result.getProviders());
```
SUPERCILEX: `get(0)` was returning Twitter even though google was available which is completely unacceptable.

auth/src/main/java/com/firebase/ui/auth/ui/accountlink/AccountLinker.java
SUPERCILEX: Ignore my previous statement, I was bothered by not understanding what was going on so I went into a long debugging session and added extensive documentation.

auth/src/main/java/com/firebase/ui/auth/ui/TaskFailureLogger.java
```diff
@@ -30,6 +31,8 @@ public TaskFailureLogger(String tag, String message) {
 
     @Override
     public void onFailure(@NonNull Exception e) {
-        Log.w(mTag, mMessage, e);
+        if (!(e instanceof FirebaseAuthUserCollisionException)) {
```
SUPERCILEX: @samtstern These exceptions produce a lot of noise (around 2-3 logs to link two accounts) and FirebaseUI now fully handles these so I opted to remove any logging associated with collisions.

app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
```diff
@@ -39,6 +39,7 @@
 import com.firebase.uidemo.R;
 import com.google.android.gms.common.Scopes;
 import com.google.firebase.auth.FirebaseAuth;
+import com.google.firebase.auth.FirebaseUser;
```
b11094: import  com.google.firebase.auth.FirebaseUser ;

SUPERCILEX:
@samtstern Do you think this can be merged for 2.0? It would certainly beef up the release! 😄

samtstern:
@SUPERCILEX this one can't go into 2.0 because of iOS feature parity concerns, and we have barely any iOS engineering resources so I think it could be a while.  I wish I had a better answer.

SUPERCILEX:
@samtstern Ok, that's really too bad. Thanks for the update though! 😄

SUPERCILEX:
@samtstern Actually, I thought of a counter proposal while I was biking home (#BikerThoughts 😆). The main issue blocking this PR is feature parity on the IOS side, right? Sooooooo, what if we worked on getting this PR merged except I remove all public references to account linking in the docs make the `AuthUI` builder methods private? I see several pros to doing it that way:
1. This PR finally gets merged so the cool stuff like `AccountLinker` can be used in the FirebaseUI codebase. I haven't had time to look at @amandle's reauth stuff in-depth, but if it does any sort of linking, I think `AccountLinker` will come in handy.
2. For those determined developers who know what they're doing, they can always use reflection to make the builder methods public and enable account linking. It wouldn't be ideal, but feature parity would _technically_ be maintained while still having a secret backdoor to the new account linking stuff. :ghost:

@samtstern What do you think?

auth/src/main/java/com/firebase/ui/auth/ui/idp/CredentialSignInHandler.java
```diff
@@ -121,8 +146,32 @@ public void onSuccess(@NonNull ProviderQueryResult result) {
                                         .setProvider(provider)
                                         .build(),
                                 mResponse
-                        ), mAccountLinkResultCode);
+                        ), mAccountLinkRequestCode);
+            }
+        }
+
+        private String getHighestPriorityProvider(List<String> providers) {
```
SUPERCILEX: I wrote this code a very long time ago, and this method seems a bit much to me. The simpler but less efficient solution would be to do this:
```java
if (providers.contains(GoogleAuthProvider.PROVIDER_ID)) {
    return GoogleAuthProvider.PROVIDER_ID;
} else if (providers.contains(FacebookAuthProvider.PROVIDER_ID)) {
    return FacebookAuthProvider.PROVIDER_ID;
} else if (providers.contains(TwitterAuthProvider.PROVIDER_ID)) {
    return TwitterAuthProvider.PROVIDER_ID;
} else {
    return providers.get(0);
}
```

Which one should I use?

samtstern:
@SUPERCILEX let me think about that one.  Either way I don't think it makes sense to push this for 2.0.0 since it would add a lot of untested code to the codebase without much current benefit (since Aaron's account linking stuff is not making 2.0.0 either0

SUPERCILEX:
@samtstern yeah, I totally agree with you that we probably shouldn't push this for `v2.0`, I was mostly just thinking of it as a way to get this PR merged in the future. I do think it would be a little bit funky to add code that isn't technically accessible, but the most important part (`AccountLinker`) will still be used in `WelcomeBackIdpPrompt` so all is not lost. 😄 

Also, as I was going through the PR again last night, I saw some more stuff that could be refactored out into separate PRs which should simplify this one.

samtstern:
@SUPERCILEX this PR (which is a fork in disguise haha) now takes a very long time to load due to the number of commits, even though the number of lines changed is not so bad (+1000/-500).

Have you considered doing a rebase to squash most of the history away and force pushing back to the branch?  Not sure if that would break anyone who's using this (if they're relying on specific SHAs) but could make it easier to look at.

SUPERCILEX:
@samtstern Lol, I've just been super lazy because the iOS side of things isn't progressing so I wasn't expecting anyone to look at this anytime soon. But yeah, once the build passes I'll squash everything into a single commit. 😄

SUPERCILEX:
Done!

samtstern:
@SUPERCILEX yeah unfortunately the iOS issue still hangs over this PR ... but I still definitely look at it from time to time and I know other devs are clearly learning from it!

auth/README.md
```diff
@@ -336,6 +354,78 @@ see the Firebase Auth documentation to
...
+                .getReference()
+                .child("chatIndices")
+                .child(uid);
+        for (DataSnapshot snapshot : mChatKeys) {
+            chatIndices.child(snapshot.getKey()).setValue(true);
+            DatabaseReference chat = FirebaseDatabase.getInstance()
+                    .getReference()
+                    .child("chats")
+                    .child(snapshot.getKey());
+            chat.child("uid").setValue(uid);
```
isaisachen: Hmm... If I understand it correctly, your database example isn't using Rules to enforce user based security?
https://firebase.google.com/docs/database/security/user-security

If really enforcing user based security, there are only two options to migrate the data:
1. Back up all data in memory and then copy to the new (non-anon) user session
2. Having two user session (two FirebaseDatabase instances w/ different users) side by side and do the copying.

SUPERCILEX: See https://github.com/firebase/FirebaseUI-Android/issues/123#issuecomment-319178534, I haven't updated the docs in a bit.

isaisachen: Thanks @SUPERCILEX. I see, it works for your data structure. But I'm afraid this "anon-user session only" model might not be sufficient for other database structures.

But I do like your assisted merging idea! Thanks for helping all developers here! 

Let's continue the discussion (offline) how to make this feature work. At high level, I think we can allow developers choose from below three options:
a) Replace: Throw away anonymous user (same as today)
b) Link if possible or fail (and give back Credential for developers to implement their own merging logic) 
c) Link or replace with assisted merging

c) might be difficult to implement if we need to make it work for all database structures.

SUPERCILEX: @isaisachen Sorry for the late reply! 😊

K, I just realized you're a FirebaseUI iOS dev—thanks for getting this going! Since I thought you were a user, I didn't really explain much so let me try and fix that.

The link I sent you is actually for the extreme case where the user is playing around with three different accounts, which I've documented in much more detail [here](https://github.com/SUPERCILEX/FirebaseUI-Android/blob/cfe884fa8b6ce516d9b2310cf9a09cfec33b6c24/auth/src/main/java/com/firebase/ui/auth/data/model/CyclicAccountLinkingException.java). For normal use cases, I do perform automatic linking using the Firebase APIs so the original user account isn't thrown away. To address another point, this merging isn't limited to anonymous accounts—it works for any merging of two or more accounts.

As for your database concerns, I don't see how I could make it any more flexible—I'm literally letting the developer execute any piece of code they so desire during that no man's land sign-in period! 😄 However, I really like your idea of giving the developer the old credential too so we can cover all the possible cases (I already give them the new one).

@isaisachen Does that answer some of your questions? How can I help you build this feature?

isaisachen: Hi @SUPERCILEX. Thanks for the reply. I'm not FirebaseUI iOS dev :) But @samtstern will reach out.

Borbarad13:
As I am not fully of the full release process here, when could this be expected time wise: Are we talking about days, weeks, month or years here? Don't get me wrong, I am not blaming - I am actually very thankful that such projects exist and that people contribute so much. I just need to know if if it makes sense to wait until this is released for my project or I have to move on to an alternative.

SUPERCILEX:
These APIs will use `register` for 1.0, right? https://github.com/gradle/kotlin-dsl/blob/0eccacca4c9f87be19dd88045c2007ae3ee58f2f/subprojects/provider/src/main/kotlin/org/gradle/kotlin/dsl/ProjectExtensions.kt#L118-L150

SUPERCILEX:
Please ignore my previous post (now deleted), was the wrong tab. 😳😊

@Borbarad13 because this depends on the other platforms (which I'm not familiar with) implementing this feature too, waiting on it is not a viable solution. However, lots of people have been using my fork in https://github.com/firebase/FirebaseUI-Android/issues/123#issuecomment-317343123 to handle every possible merge case. I'll post some new commit hashes tomorrow.

SUPERCILEX:
@Borbarad13 Done: https://github.com/firebase/FirebaseUI-Android/issues/123#issuecomment-412972347

FanciROS:
Is someone authorized to proceed, can resolve the conflicts and merge the pull request ? This feature is what I am looking for ! 
Thank you in advance for your work.

danijorda1:
Hi, any update here? Is this feature avialble already? Doesn't seem to be the case. Thanks

danijorda1:
@samtstern  can you proceed, and resolve the conflicts and merge the pull request? 
This feature will be very useful to use Auth UI for Account Linking! 
Thanks!!

samtstern:
@danijorda1 this code is from 2016. Even the legendary @SUPERCILEX could not possibly resolve the merge conflicts and make this PR functional again, the library has been significantly refactored over the years.

We already have some account linking functionality in FirebaseUI. If you need specific improvements to that functionality, please file a new feature request or bug report.

danijorda1:
Hi @samtstern  which  account linking functionality are you referring to? Couldn't find anything in the Documentation. Thanks

danijorda1:
As far as I can see only Upgrading anonymous users is supported currently while [309](https://github.com/firebase/FirebaseUI-Android/pull/309), [1823](https://github.com/firebase/FirebaseUI-Android/issues/1823), [894](https://github.com/firebase/FirebaseUI-Android/issues/894), [1235](https://github.com/firebase/FirebaseUI-Android/issues/1235) and [123](https://github.com/firebase/FirebaseUI-Android/issues/123) are all related to being able to use Firebase Auth UI to link accounts. That is, when a user is signed in, the user has the possibility to press a Link Account button where a new Sign in activity will be launched allowing the user to link accounts with the selected Provider.
image

As far as I can see all these feature requests/bug reports have been closed and no solution has been provided except for @SUPERCILEX  solution that was never merged.... Will Account linking (other than Anonymous upgrade) be supported at some point?

Thanks in advance.

samtstern:
https://github.com/firebase/FirebaseUI-Android/tree/master/auth#upgrading-anonymous-users

danijorda1:
Do you know who is working on this project so I can directly ask him/her?
Thanks

PR #328, opened 2016-09-25
storage: fix link syntax

meirf:
This is not super intuitive compared to other linking syntaxes.
Here are all the ways linking is supported:
https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#links
I could not find other instances of this error in the rest of the code base.


samtstern:
@meirf thanks for this!

Issue #77, opened 2016-02-16
Translations


akrmn:
The Login fragment has hardcoded strings, which makes it difficult to use it in non-English projects.

abehaskins:
Great point! We'll improve this for the next release.

cardenuto:
@akrmn  I have been working on code that uses this library I needed to override the fragment_firebase_login.xml file locally and that gave me the ability to use resource strings instead. I am not sure this will help you in the short run since it has been a while. 

Feel free to use the code for other features 
https://github.com/cardenuto/FirebaseLogin any suggestions are welcome. Currently I designed my code to be added manually to a project as more of a template ... instructions are included in the repo.

samtstern:
For the firebase.google.com we support the following languages, therefore I suggest we have the same target here:

- el
- en
- es-419
- id
- ja
- ko
- pt-br
- tr
- zh-cn

samtstern:
Well we finally did it!  As of `2.1.0` FirebaseUI has translations!  Thank you everyone for your patience, let us know what you think!

PR #333, opened 2016-09-30
Spanish translation (auth and app)

javmarina:
This pull request includes Spanish translations for strings in the Firebase UI project. I have translated strings from Firebase UI Auth and the example app. This way, developers don't need to worry about translating library strings (like I did before).

### DIFF
--- a/app/src/main/res/values-es/strings.xml
+++ b/app/src/main/res/values-es/strings.xml
@@ -1,0 +1,39 @@
+<resources>
+    <string name="desc_chat">Muestra cómo usar el FirebaseRecyclerAdapter para cargar datos desde Firebase Database en un RecyclerView para una aplicación de chat básica.</string>
+    <string name="desc_auth_ui">Muestra el flujo de Firebase Auth UI, con opciones de personalización.</string>
+    <string name="desc_image">Enseña cómo mostrar una imagen de Firebase Storage usando Glide.</string>
+
+    <!-- strings for Auth UI demo activities -->
+    <eat-comment />
+    <string name="sign_in">Empezar</string>
+    <string name="green_theme">Tema verde</string>
+    <string name="purple_theme">Tema morado</string>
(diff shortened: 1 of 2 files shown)


samtstern:
@javmarina thanks for this contribution!  Appreciate the effort here.  Since this is one of the languages we support on firebase.google.com (see #77) am inclined to merge, but we need a few things first:

@alfongj can you check out the Spanish here?
@morganchen12 how hard is it to port this to iOS?

danielpassos:
@samtstern It's pretty simple to _port_ this to iOS. I'm going to send a PR there based on this one and another one for [mime](https://github.com/firebase/FirebaseUI-Android/pull/384)

javmarina:
Any update on this??

samtstern:
@javmarina sorry for the lack of updates here, we put off translations while working on the 1.0 release and are now back to discussing how to do them best.

javmarina:
OK, thanks!

javmarina:
When do you plan to add translations? I currently use my own in my projects but it's better to include them inside the library, available for every developer. I'm willing to translate the other strings (some of them have been added/modified).

samtstern:
@javmarina I have decided to close translation pull requests.  They've been sitting around for a while and the silent, infinitely-open PR is not fair to you nor is it useful to us.  For now you should continue to maintain translations as part of your own app or as part of a library fork.

Eventually, we will get official translations for FirebaseUI for the major languages that Firebase supports.  When we go through that process, I will definitely provide the contents of this Pull Request as suggestions to the translators (there is a place in the system for that).

Thank you for taking the time to contribute!  We do appreciate it, even if we sometimes fail to demonstrate our appreciation.

javmarina:
Don't worry about that. I understand the situation and am glad you explained it to me. This PR should be closed if it interferes with your current work. I just wanted to help developers but it is not an issue to maintain those strings on my own.

Issue #323, opened 2016-09-23
Leak in EmailHintContainerActivity


javmarina:
### Step 2: Describe your environment

  * Android device: Sony Xperia Z (C6603)
  * Android OS version: 5.0.2
  * Google Play Services version: 9.6.83
  * Firebase/Play Services SDK version: 9.4.0
  * FirebaseUI version: 0.5.3
  
### Step 3: Describe the problem:

Leak Canary reports a leak in EmailHintContainerActivity.

#### Steps to reproduce:

The only thing you have to do is sign in with an account and wait.
  
#### Observed Results:

  * What happened?  This could be a description, `logcat` output, etc.

`In com.javapps.iSwim:1.0:1.` (my app)
```
* com.firebase.ui.auth.ui.email.EmailHintContainerActivity has leaked:
* GC ROOT android.app.LoadedApk$ServiceDispatcher$DeathMonitor.this$0
* references android.app.LoadedApk$ServiceDispatcher.mConnection
* references com.google.android.gms.common.internal.zzn$zzb$zza.yX
* references com.google.android.gms.common.internal.zzn$zzb.yT
* references java.util.HashSet.backingMap
* references java.util.HashMap.table
* references array java.util.HashMap$HashMapEntry[].[3]
* references java.util.HashMap$HashMapEntry.next
* references java.util.HashMap$HashMapEntry.key
* references com.google.android.gms.common.internal.zzd$zzh.xG
* references com.google.android.gms.auth.api.credentials.internal.zzg.mContext
* leaks com.firebase.ui.auth.ui.email.EmailHintContainerActivity instance

* Retaining: 34 KB.
* Reference Key: 9bc898ef-9d68-4785-b001-cdbe8fd5eb66
* Device: Sony Sony C6603 C6603
* Android Version: 5.0.2 API: 21 LeakCanary: 1.4 6b04880
* Durations: watch=5023ms, gc=508ms, heap dump=7965ms, analysis=89489ms

* Details:
* Instance of android.app.LoadedApk$ServiceDispatcher$DeathMonitor
|   mName = android.content.ComponentName@314728624 (0x12c260b0)
|   mService = android.os.BinderProxy@314588032 (0x12c03b80)
|   this$0 = android.app.LoadedApk$ServiceDispatcher@314980624 (0x12c63910)
* Instance of android.app.LoadedApk$ServiceDispatcher
|   mActiveConnections = android.util.ArrayMap@316031584 (0x12d64260)
[...]

PR #334, opened 2016-09-30
Fix memory leak, switch to auto managed clients and bug fix

amandle:
Fix memory leak (#323), switch to auto managed clients where possible and fix dialog hanging around on bad password.


samtstern:
I'm not sure what's going on in the diff here but I already did the fix to dismiss the dialog on bad password, see:

https://github.com/firebase/FirebaseUI-Android/blob/11bc4ceb99ea71b535ced2b9d2e7593db447884c/auth/src/main/java/com/firebase/ui/auth/ui/email/SignInActivity.java#L123

Other than figuring out why you're not seeing that, LGTM and feel free to merge when Travis goes green.

amandle:
Ah, I hadn't pulled recently enough.  Thanks!

Issue #19, opened 2015-10-13
Include a class that handles Firebase.ServerValues.TIMESTAMP


ceruleanotter:
A class that handles Firebase.ServerValues.TIMESTAMP would be helpful because currently developers need to jump through some hoops to create POJOs that can both serialize and deserialize ServerValues.TIMESTAMP. Thank you!

WildWeazel:
+1
Dealing with a map that turns into a long in a statically typed language is tricky for such a basic feature.

jdimond:
One trick is to use Object and then have a getter that returns ServerValues.TIMESTAMP, the setter accepting a Long, and then using an getter with an exclusion annotation to always return a long for local use (Using System.currentTimeMillis() when it's set to ServerValues.TIMESTAMP).

puf:
I'm pretty sure that's what @ceruleanotter had in mind with this feature
request. :-)

SUPERCILEX:
This issue should probably be closed now with the explanation from #355.

samtstern:
@SUPERCILEX thanks for following up, and for taking a shot at implementing (even though it didn't pan out)

PR #335, opened 2016-10-02
Add class that handles ServerValue

SUPERCILEX:
  * #19 
  * Anyone?

Example usage:
```java
// After calling ref.getValue(Team.class)

public class Team extends Timestamp {
    public Team() {
        // Example with no config
        getTimestamp();
    }

    // Example getter override
    @PropertyName("timestamp" /* Or any other name as long as you create your own
    getters and setters, but then there is no point in extending Timestamp */)
    public Map<String, String> getCustomTimestamp() {
        if (mShouldUpdateTimestamp) {
            return ServerValue.TIMESTAMP;
        } else {
            return null;
        }
    }
```

### DIFF
--- a/database/src/main/java/com/firebase/ui/database/Timestamp.java
+++ b/database/src/main/java/com/firebase/ui/database/Timestamp.java
@@ -10,7 +10,7 @@ public class Timestamp {
     private long mTimestamp;
 
     @PropertyName("timestamp")
-    private Map<String, String> getServerValue() {
+    protected Map<String, String> getServerValue() {
         return ServerValue.TIMESTAMP;
     }
 
@@ -19,7 +19,7 @@ public class Timestamp {
         return mTimestamp;
     }
 
-    private void setTimestamp(long time) {
+    protected void setTimestamp(long time) {
         mTimestamp = time;
     }
 }
--- a/database/src/main/java/com/firebase/ui/database/Timestamp.java
+++ b/database/src/main/java/com/firebase/ui/database/Timestamp.java
@@ -1,0 +1,25 @@
+package com.firebase.ui.database;
+
+import com.google.firebase.database.Exclude;
+import com.google.firebase.database.PropertyName;
+import com.google.firebase.database.ServerValue;
+
+import java.util.Map;
+
+public class Timestamp {
+    private long mTimestamp;
+
+    @PropertyName("timestamp")
+    private Map<String, String> getServerValue() {
+        return ServerValue.TIMESTAMP;
+    }
+
+    @Exclude
+    public long getTimestamp() {
+        return mTimestamp;
+    }
+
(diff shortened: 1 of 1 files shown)


puf:
ServerValue.TIMESTAMP also exists on iOS:
https://firebase.google.com/docs/database/ios/offline-capabilities#server-timestamps

database/src/main/java/com/firebase/ui/database/Timestamp.java
```diff
@@ -0,0 +1,25 @@
...
+
+    @PropertyName("timestamp")
+    private Map<String, String> getServerValue() {
+        return ServerValue.TIMESTAMP;
+    }
+
+    @Exclude
+    public long getTimestamp() {
+        return mTimestamp;
+    }
```
SUPERCILEX: Should I check to see if `mTimestamp` is 0 (null) here? If so, I would return the system time. Thoughts?

samtstern: I don't think we'd want to return system time, that's not the same as unset.  I prefer how it is now, it's the least opinionated and most likely to solve the original problem (serialization/deserialization of time stamps is annoying)

samtstern:
@SUPERCILEX I am a little confused by your intended usage.  When I first saw this one I imagined it as a class you'd use in an instance variable, something like:

```
public class MyObject {

   public String name;

   // Easier to use than plain Object
   public Timestamp timestamp;

}
```

But your PR description implies that the major usage method is **extending** `Timestamp`, which I find counter-intuitive.  Could you explain a little?

SUPERCILEX:
@samtstern I was extending just because it was more convenient for me. However, while trying to figure out how to make `public Timestamp timestamp` work, I ran into some issues. When extending, the data is single level, but when using `Timestamp` as a variable, I get nested data like so:
```
timestamp: {
    timestamp: 0000000
},
other data
```
Is it possible with Firebase to flatten an object into another object? (merge variables from one object into another)

puf:
Nope. The Object graph must exactly match the JSON tree with the default
JSON/POJO codec. Once you do your own mapping, you can of course do
whatever you want.

But for timestamp it'll always be tricky, since the value-type that is
written (JSON: `{ ".sv": "timestamp" }`) is different from the value-type
that you read (number).

SUPERCILEX:
So is it fine to have a nested timestamp like that? I really don't like it, but if you say there is no other way...

puf:
@samstern: can you think of a way to annotate a class to write one type and
read another?

samtstern:
@puf sorry for the delay here ... not that I know of.  Upon further thought I don't think this feature belongs in FirebaseUI because it will have to be some sort of compromise :neutral_face:

SUPERCILEX:
Closing because of the tradeoffs.

PR #336, opened 2016-10-02
Database style cleanup

SUPERCILEX:
Style nit: [style guide](https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names) says emums should be in caps.


puf:
I have no problem with the change itself. But keep in mind that this is a change to a public interface, so must happen pre-1.0 or in a major. :-/

samtstern:
@puf actually I don't believe this is a public interface change since `OnChangedListener` is a child of `FirebaseArray` which is package protected.  When I try to create a new instance of `OnChangedListener` in my class I get:

```
"com.firebase.ui.database.FirebaseArray" is not public in "com.firebase.ui.database".  Cannot be accessed from outside package.
```

samtstern:
@puf are you ok with this change given my previous comment?

puf:
Yup.

samtstern:
Merged, @SUPERCILEX thanks!

SUPERCILEX:
Thanks!

PR #342, opened 2016-10-06
Specify additonal scopes through code rather than config.xml

amandle:


samtstern:
For scopes-only this approach LGTM, but I think we can give ourselves more breathing room if we instead have `SignInIntentBuilder` take some sort of `IDPConfig<>` object which has fields, one of which is a `List<String>` of scopes.

Then instead of calls to `setProviders()` and `setAdditionalGoogleScopes()` you could do something like `enableProvider(GOOGLE_PROVIDER, mGoogleIdpConfig)` or even have one method for each IDP like `enableGoogleProvider(mGoogleIdpConfig)`.

This way if we want to add more customization in the future (IDPs can change underneath us) we don't need a breaking API change, we just add a new field to `IDPConfig`.  It also makes the availability of the options clearer if you pass them at the time of enabling the IDP.  This is like how in GmsCore with `GoogleApiClient` you do `.addApi(mApi, mApiOptions)`.

amandle:
I refactored quite heavily. There is now an `IdpConfig` class as you suggested, and `setProviders` takes a list of them.  So as not to have two classes doing pretty much the same thing, and since the new `IdpConfig` class needs to be publicly available I removed `IDPProviderParcel` and replaced it with the `IdpConfig` class.  This cleans things up a bit.  This is kind of a large change now, apologies about that.

auth/src/main/java/com/firebase/ui/auth/provider/FacebookProvider.java
```diff
@@ -47,26 +46,27 @@
...
     private static final String PUBLIC_PROFILE = "public_profile";
 
+    private final List<String> mScopes;
     private CallbackManager mCallbackManager;
     private IDPCallback mCallbackObject;
 
-    public FacebookProvider (Context appContext, IDPProviderParcel facebookParcel) {
+    public FacebookProvider (
+            Context appContext,
+            List<String> scopes) {
```
samtstern: Why not just pass through the IDPConfig here?  That way you only have to change this once.

auth/src/main/java/com/firebase/ui/auth/provider/GoogleProvider.java
```diff
@@ -46,14 +47,16 @@
...
     private static final String ERROR_KEY = "error";
-    private static final String CLIENT_ID_KEY = "client_id_key";
     private GoogleApiClient mGoogleApiClient;
     private Activity mActivity;
     private IDPCallback mIDPCallback;
 
-    public GoogleProvider(FragmentActivity activity, IDPProviderParcel parcel, @Nullable String email) {
+    public GoogleProvider(
+            FragmentActivity activity,
+            @Nullable String email,
```
samtstern: Same question here about just passing IDPConfig (and elsewhere, I'll stop repeating myself)

auth/src/main/java/com/firebase/ui/auth/provider/GoogleProvider.java
```diff
@@ -46,14 +47,16 @@
...
+            List<String> scopes) {
         mActivity = activity;
-        String mClientId = parcel.getProviderExtra().getString(CLIENT_ID_KEY);
+        String mClientId = activity.getString(R.string.default_web_client_id);
```
samtstern: Is this equivalent behavior before?  What if they want to pass another client ID?

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -403,17 +407,111 @@ public static AuthUI getInstance(FirebaseApp app) {
...
     }
 
+
+    /**
+     * Configuration for an identity provider.
+     *
+     * In the simplest case, you can supply the provider ID and build the config like this:
+     * {@code new IdpConfig.Builder(AuthUI.GOOGLE_PROVIDER).build()}
+     */
+    public static class IdpConfig implements Parcelable {
```
samtstern: Would it be better if each provider had a subclass of this?  so `new GoogleIdpConfig.Builder().build()`, etc.  This way you have more control over provider differences.  For example doesn't look like you're supporting Twitter scopes so that builder would not have `setScopes()`.  It would also hide the `GOOGLE_PROVIDER` and other Strings from the dev.

This one is up to you, I can see arguments both ways.

auth/src/main/java/com/firebase/ui/auth/provider/FacebookProvider.java
```diff
@@ -47,26 +46,27 @@
...
         mCallbackManager = CallbackManager.Factory.create();
-        String applicationId = facebookParcel.getProviderExtra().getString(APPLICATION_ID);
+        if (scopes == null) {
+            mScopes = new ArrayList<>();
+        } else {
+            mScopes = scopes;
+        }
+        String applicationId = appContext.getString(R.string.facebook_application_id);
```
samtstern: If you took my suggestion of IdpConfig subclasses, these strings could be part of the builders and we could eliminate the resource-based config entirely.

auth/src/main/java/com/firebase/ui/auth/provider/GoogleProvider.java
```diff
@@ -62,8 +65,7 @@ public GoogleProvider(FragmentActivity activity, IDPProviderParcel parcel, @Null
                 .requestIdToken(mClientId);
 
         // Add additional scopes
-        String[] extraScopes = mActivity.getResources().getStringArray(R.array.google_permissions);
```
samtstern: Bonus points: since we're removing these scopes as resources, maybe we could log a big warning message if we see them defined and they're non empty.  Something like "DEVELOPER_WARNING: you have defined R.array.google_permissions but that is no longer respected as of FirebaseUI 1.0.0.  See README"

samtstern:
Also I merged master into version-1.0.0-dev today, I forgot this PR was outstanding so I apologize if there are conflicts for you (they should be simple if there are any)

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
amandle: I tried this, but it makes parceling and passing a list of providers around much uglier.  I would prefer to keep it this way.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -403,17 +407,111 @@ public static AuthUI getInstance(FirebaseApp app) {
...
+            return mProviderId;
+        }
+
+        public List<String> getScopes() {
+            return mScopes;
+        }
+
+        protected IdpConfig(Parcel in) {
+            mProviderId = in.readString();
+            mScopes = in.createStringArrayList();
```
amandle: createStringArrayList actually does read the list, it just handles the creation of the list for you: https://developer.android.com/reference/android/os/Parcel.html#createStringArrayList()

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -403,17 +407,111 @@ public static AuthUI getInstance(FirebaseApp app) {
...
+
+            /**
+             * Specifies the additional permissions that the application will request for this
+             * identity provider.
+             *
+             * For Facebook permissions see:
+             * https://developers.facebook.com/docs/facebook-login/android
+             * https://developers.facebook.com/docs/facebook-login/permissions
+             *
+             * For Google permissions see:
```
amandle: Twitter scopes are configured only through the developer console. I will add that to the comment.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -403,17 +407,111 @@ public static AuthUI getInstance(FirebaseApp app) {
...
+    }
+
     /**
      * Builder for the intent to start the user authentication flow.
      */
     public final class SignInIntentBuilder {
         private int mLogo = NO_LOGO;
         private int mTheme = getDefaultTheme();
-        private List<String> mProviders = Collections.singletonList(EMAIL_PROVIDER);
+        private List<IdpConfig> mProviders = new ArrayList<>();
```
amandle: We do need set like properties, but we also need to retain the order.  It looks like LinkedHashSet will do what we want.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
samtstern: Ok that's a good reason, keeping it as it is SGTM.

auth/src/main/java/com/firebase/ui/auth/provider/FacebookProvider.java
amandle: Done

auth/src/main/java/com/firebase/ui/auth/provider/GoogleProvider.java
amandle: Good idea, added.

auth/src/main/java/com/firebase/ui/auth/provider/FacebookProvider.java
```diff
@@ -53,10 +54,17 @@
...
 
-    public FacebookProvider (
-            Context appContext,
-            List<String> scopes) {
+    public FacebookProvider(Context appContext, IdpConfig idpConfig) {
         mCallbackManager = CallbackManager.Factory.create();
+
+        if (appContext.getResources().getIdentifier(
+                "google_permissions", "array", appContext.getPackageName()) != 0) {
+            Log.w(TAG, "DEVELOPER WARNING: You have defined R.array.facebook_permissions but that"
```
samtstern: This says facebook_permissions in one place and google_permissions in another.

amandle: Oops, good catch.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -229,22 +234,22 @@
      * Provider identifier for email and password credentials, for use with
      * {@link SignInIntentBuilder#setProviders}.
      */
-    public static final String EMAIL_PROVIDER = "email";
+    public static final String EMAIL_PROVIDER = EmailAuthProvider.PROVIDER_ID;
```
samtstern: Agreed, seems like a sensible change.

samtstern:
@amandle FYI you can carry my approval through from before.  Merge when ready and the tests are green.

SUPERCILEX:
@amandle Travis failed with exit code 137. I've been having that issue too, just restart the build or push some whitespace to get it going again. It's really random, but apparently it has something to do with the Travis sandbox only having 3GB of RAM available and us exceeding that.

amandle:
@SUPERCILEX Thanks, I'll give that a try!

SUPERCILEX:
@amandle I did some more Googling: see this issue https://github.com/travis-ci/travis-ci/issues/5582 and this solution https://github.com/srsudar/MamasDelRioAndroid/commit/ec64465f2980b0c865b9d82f19918065a87d01aa.

SUPERCILEX:
See #358

SUPERCILEX:
Maybe undo https://github.com/firebase/FirebaseUI-Android/pull/342/commits/f40ce5aae80a16c756aae39ea5600e2fd3146515 to force a rerun?

samtstern:
Note: when a build fails there is a *Restart Build* button on the page you can use:
screen shot 2016-10-17 at 11 21 47 am

amandle:
@samtstern I still don't see that button :-(

SUPERCILEX:
@samtstern that button is only available to repository admins (the person who set up Travis).

SUPERCILEX:
@amandle Looks like a compile error: 
```
/home/travis/build/firebase/FirebaseUI-Android/auth/src/main/java/com/firebase/ui/auth/ui/idp/IDPSignInContainerActivity.java:67: error: incompatible types: String cannot be converted to IdpConfig
            mIDPProvider = new GoogleProvider(this, mEmail, providerConfig);
```

SUPERCILEX:
I'm sorry, I just realized that I merged the Travis fix into master instead of 1.0.0. @samtstern could you merge master into dev or @amandle you could just merge master into this PR directly since I think the Travis fix is the only difference between master and dev is the Travis fix.

samtstern:
@SUPERCILEX @amandle  I merged master into 1.0.0-dev, the only changes were `travis.yml` and your small change to `SaveCredentialsActivity`

SUPERCILEX:
Yes! Thanks @samtstern it worked! Also, look at that build time!!! 4mins vs the usual ~10mins!

PR #343, opened 2016-10-07
Refactor SaveCredentialsActivity

SUPERCILEX:
  * #254

Placeholder PR. This refactoring is huge and in no way ready yet (as in it won't even compile), but I thought I'd inform you guys that I'm working on this and ask a few questions.


auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -265,13 +265,13 @@ private AuthUI(FirebaseApp app) {
      * ({@code !result.isSuccess()}).
      */
     public Task<Void> signOut(@NonNull Activity activity) {
-        // Get helper for Google Sign In and Credentials API
+        // Get helper for Google Sign In and SmartLock API
```
SUPERCILEX: Ignore this weirdness: intellij make some mistakes.

auth/src/main/java/com/firebase/ui/auth/ui/account_link/WelcomeBackPasswordPrompt.java
```diff
@@ -137,10 +137,8 @@ public void onSuccess(AuthResult authResult) {
                                         new OnSuccessListener<AuthResult>() {
                                             @Override
                                             public void onSuccess(AuthResult authResult) {
-                                                mActivityHelper.dismissDialog();
-                                                SmartlockUtil.saveCredentialOrFinish(
+                                                mSmartLock = SmartLock.saveCredentialOrFinish(
```
SUPERCILEX: Any thoughts on how to do this better?

auth/src/main/java/com/firebase/ui/auth/ui/account_link/WelcomeBackPasswordPrompt.java
```diff
@@ -107,8 +107,8 @@ public void onClick(View view) {
     @Override
     protected void onActivityResult(int requestCode, int resultCode, Intent data) {
         super.onActivityResult(requestCode, resultCode, data);
-        if (requestCode == RC_CREDENTIAL_SAVE) {
-            finish(RESULT_OK, new Intent());
+        if (mSmartLock != null) {
+            mSmartLock.onActivityResult(requestCode, resultCode);
```
SUPERCILEX: I really don't like this, but I can't think of another way to do this without copying code between Activities. Thoughts?

auth/src/main/java/com/firebase/ui/auth/util/SmartLock.java
```diff
@@ -43,53 +38,51 @@
...
-
-        mName = getIntent().getStringExtra(ExtraConstants.EXTRA_NAME);
-        mEmail = getIntent().getStringExtra(ExtraConstants.EXTRA_EMAIL);
-        mPassword = getIntent().getStringExtra(ExtraConstants.EXTRA_PASSWORD);
-        mProvider = getIntent().getStringExtra(ExtraConstants.EXTRA_PROVIDER);
-        mProfilePictureUri = getIntent()
-                .getStringExtra(ExtraConstants.EXTRA_PROFILE_PICTURE_URI);
-
-        mCredentialsApiClient = new GoogleApiClient.Builder(this)
+    public SmartLock(AppCompatBase activity,
```
samtstern: If this class is going to be tied to an Activity, then one way to structure it would be a headless `Fragment`.  You'd call `SmartLock.getInstance(activity, ...)` which would attach a `Fragment` to the Activity.  Then that `Fragment` can actually launch activities for result and receive the results directly, rather than the constant `if (mSmartLock != null) { smartLock.onActivityResult(...) }`.

SUPERCILEX: Great idea! I didn't know you could do that. Things are way nicer now.

auth/src/main/java/com/firebase/ui/auth/util/SmartLock.java
```diff
@@ -133,77 +140,100 @@ public void onConnectionFailed(@NonNull ConnectionResult connectionResult) {
...
+     * If SmartLock is enabled and Google Play Services is available, save the credentials.
+     * Otherwise, finish the calling Activity with RESULT_OK.
+     *
+     * @param activity     the calling Activity.
+     * @param parameters   calling Activity flow parameters.
+     * @param firebaseUser Firebase user to save in Credential.
+     * @param password     (optional) password for email credential.
+     * @param provider     (optional) provider string for provider credential.
+     */
+    public static SmartLock getInstance(AppCompatBase activity,
```
samtstern: It's more conventional to have this `getInstance()` method actually do the Fragment transaction and add itself to the Activity (first checking if there is already an instance of `SmartLock` attached to the Activity).

You can attach a SmartLock fragment to each Activity that needs it in `onCreate`. Then `SmartLock` could expose methods to kick off the various flows, rather than starting on Fragment attach.

So `getInstance()` would look more like this:
```java
        SmartLock result;

        FragmentManager fm = activity.getSupportFragmentManager();
        FragmentTransaction ft = fm.beginTransaction();

        Fragment fragment = fm.findFragmentByTag(SMARTLOCK_TAG);
        if (fragment == null || (!fragment instanceof SmartLock)) {
            shouldAdd = true;
        } else {
            result = (SmartLock) fragment;
        }

        if (shouldAdd) {
            result = new SmartLock(...);
            ft.add(result, SMARTLOCK_TAG).disallowAddToBackStack().commit();
        }

        return result;
    }
```

auth/src/main/java/com/firebase/ui/auth/util/SmartLock.java
```diff
@@ -133,77 +140,100 @@ public void onConnectionFailed(@NonNull ConnectionResult connectionResult) {
...
+                                        FlowParameters parameters,
+                                        FirebaseUser firebaseUser,
+                                        @Nullable String password,
+                                        @Nullable String provider) {
+        SmartLock smartLock = new SmartLock();
+        smartLock.mActivity = activity;
```
samtstern: For parameters of the Fragment like this it's better to use `setArguments()` to attach them to the Fragment in a way that the Android system can keep track of.

SUPERCILEX: How would I pass in things like an activity though?

auth/src/main/java/com/firebase/ui/auth/util/SmartLock.java
SUPERCILEX: Ohhhhh. I think I'm getting it now. You can pass in things like Activities to a fragment because they are a more conventional Java object: you can just do method injection.

auth/src/main/java/com/firebase/ui/auth/util/SmartLock.java
samtstern: You can just do getActivity() in the fragment, and if you need methods on `AppCompatBase` you can just do:
```java
if (!(getActivity() instanceof AppCompatBase)) {
  throw new IllegalStateException("SmartLock can only be attached to AppCompatBase!");
} else {
  // Cast and call method
}
```

Realistically you can just do the `check --> throw` thing once in the `Fragment` `onAttach`.

auth/src/main/java/com/firebase/ui/auth/util/SmartLock.java
```diff
@@ -58,38 +60,13 @@
...
-                .enableAutoManage(this, this)
-                .build();
-    }
-
-    @Override
     public void onConnected(@Nullable Bundle bundle) {
         if (mEmail == null) {
             Log.e(TAG, "Unable to save null credential!");
-            finish(RESULT_FIRST_USER, getIntent());
+            mActivity.finish(RESULT_OK, mActivity.getIntent());
```
samtstern: Is there a reason we changed this to RESULT_OK here?

samtstern:
Looks like you're still working on the tests, but this is starting to look very close to done!  Thanks for all the work so far.

@amandle could you take a look through this as well?  Removes the headless smartlock activity in favor of a Fragment-based system.  Should be a no-op in terms of logic and functionality but will make the back stack nicer and the whole thing a little more modular.

auth/src/main/java/com/firebase/ui/auth/util/SmartLock.java
SUPERCILEX: Yes. When this was an activity we finished with result `RESULT_FIRST_USER`, but the calling activity didn't actually care what the result was (all the calling activities finished with `RESULT_OK`). Since the fragment it closing the activity for it, we just finish everything with `RESULT_OK`.

samtstern: SGTM, thanks for clarifying :+1:

SUPERCILEX:
I think the `SaveCredentialsActivity` refactor is good now.

Also, I'll create separate PRs for the other zombie activities in #254.

auth/src/main/java/com/firebase/ui/auth/ui/account_link/WelcomeBackPasswordPrompt.java
```diff
@@ -139,10 +132,8 @@ public void onSuccess(AuthResult authResult) {
                                         new OnSuccessListener<AuthResult>() {
                                             @Override
                                             public void onSuccess(AuthResult authResult) {
-                                                mActivityHelper.dismissDialog();
```
samtstern: Just noticed this, why are we removing the calls to dismissDialog()?

samtstern:
LGTM!  I'll let @amandle take a look though in case I missed anything.  This came out really clean in the end.

auth/src/main/java/com/firebase/ui/auth/ui/account_link/WelcomeBackPasswordPrompt.java
SUPERCILEX: I decided to get rid of them because when you close the activity, they get closed with it. Leaving them open means the loading dialog stays open during the credential save.

auth/src/test/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivityTest.java
```diff
@@ -146,18 +146,16 @@ public void testFacebookLoginFlow() {
         assertNotNull(facebookButton);
         facebookButton.performClick();
 
-        ShadowActivity.IntentForResult nextIntent =
-                Shadows.shadowOf(authMethodPickerActivity).getNextStartedActivityForResult();
-        verifySaveCredentialIntent(nextIntent, FacebookAuthProvider.PROVIDER_ID);
+        Intent smartLockIntent = SmartLock.getInstance(authMethodPickerActivity).getIntentForTest();
```
amandle: These tests verify the intent when the intent is only created for the purposes of the test. It would be nice to update the tests to reflect the changes in SmartLock, ideally checking that the SmartLock API is called with the correct credential.

auth/src/main/java/com/firebase/ui/auth/util/SmartLock.java
```diff
@@ -133,77 +113,149 @@ public void onConnectionFailed(@NonNull ConnectionResult connectionResult) {
...
+        mActivity = activity;
+        mCredentialsApiClient = new GoogleApiClient.Builder(activity)
+                .addConnectionCallbacks(this)
+                .addOnConnectionFailedListener(this)
+                .addApi(Auth.CREDENTIALS_API)
+                .enableAutoManage(activity, this)
+                .build();
+    }
+
+    public Intent getIntentForTest() {
```
amandle: This is returning an intent when the SmartLock doesn't actually use intents.  This makes the tests slightly misleading IMO.  See the comments on the tests below.

auth/src/main/java/com/firebase/ui/auth/util/SmartLock.java
```diff
@@ -107,13 +84,16 @@ public void onConnected(@Nullable Bundle bundle) {
...
         if (mName != null) {
             builder.setName(mName);
         }
+
         if (mProfilePictureUri != null) {
             builder.setProfilePictureUri(Uri.parse(mProfilePictureUri));
         }
-        mActivityHelper.getCredentialsApi()
+
+        Auth.CredentialsApi
```
amandle: This change makes it difficult to use a mock CredentialsApi in the tests.

auth/src/main/java/com/firebase/ui/auth/util/SmartLock.java
```diff
@@ -133,77 +113,149 @@ public void onConnectionFailed(@NonNull ConnectionResult connectionResult) {
...
-                finish(RESULT_FIRST_USER, getIntent());
+                mActivity.finish(RESULT_OK, new Intent());
             }
         } else if (requestCode == RC_UPDATE_SERVICE) {
             if (resultCode == RESULT_OK) {
-                Credential credential = new Credential.Builder(mEmail).setPassword(mPassword).build();
-                mActivityHelper.getCredentialsApi()
+                Credential credential = new Credential.Builder(mEmail).setPassword(mPassword)
+                        .build();
+                Auth.CredentialsApi
```
amandle: Same comment here about the mocking

auth/src/test/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivityTest.java
SUPERCILEX: I'll try that again, but I keep getting a null pointer exception at line 221 in smartlock.java.

SUPERCILEX: I figured that part out

SUPERCILEX: Do we need the tests? All they were doing before is checking to see if the intent was passed in correctly... Right now I'm stuck on how to wait for `SmartLock` to save the credentials.

auth/src/test/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivityTest.java
```diff
@@ -171,32 +171,27 @@ public void testGoogleLoginFlow() {
 
         assertNotNull(googleButton);
         googleButton.performClick();
-        ShadowActivity.IntentForResult nextIntent =
-                Shadows.shadowOf(authMethodPickerActivity).getNextStartedActivityForResult();
 
-        verifySaveCredentialIntent(nextIntent, GoogleAuthProvider.PROVIDER_ID);
+        assertEquals(
+                Shadows.shadowOf(authMethodPickerActivity).getResultCode(),
+                Activity.RESULT_OK);
```
SUPERCILEX: If I could get robolectric to wait for `SmartLock` to finish, `Shadows.shadowOf(authMethodPickerActivity).getResultCode()` would change from `Activity.RESULT_CANCELED` to `Activity.RESULT_OK` which would confirm that the something happened if not the credentials being saved. Any ideas?

SUPERCILEX:
Do you guys want to merge the refactor of `SaveCredentialsActivity` and `ChooseAccountActivity` separately? For the `ChooseAccountActivity` refactor, I pulled changes from this PR because I'm completely reworking the Smart Lock stuff.

SUPERCILEX:
Nevermind, I'm going to submit a separate PR because the changes are massive. I'll share a video of the old vs the new. It's mind blowing how fast and beautiful things are now!!! Pretty exciting! The one downside is that there are breaking changes to AuthUI, but I think it's definitely worth it.

SUPERCILEX:
This PR and #357 should be released together because I'm checking how configuration changes affect the smart lock stuff.

auth/src/test/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivityTest.java
SUPERCILEX: Noooooooooooo!!!!!! I just spent two hours figuring this out!!

I'm almost done and then I'll update the status of this PR.

SUPERCILEX:
@amandle I finished up the PR, what do you think of the tests? BTW, to make them work, I had to remove the `mSmartLock` initialization from `onCreate` to where `saveCredentialsOrFinish` is called because I'm using a sneeky version of injection to make the tests work. Feel free to ask me any questions!

amandle:
@SUPERCILEX Tests look good to me, thanks for doing that!

SUPERCILEX:
Awesome! You're welcome!

amandle:
@samstern, any objections to me merging this now?

Issue #347, opened 2016-10-10
Double password visibility toggles


nicbell:
When using `com.firebaseui:firebase-ui-auth:0.6.0` with Android Support Library, revision `24.2.0`. I get two icons to toggle password visibility because this feature is now built-in.

As temporary fix I have downgraded my Android Support Library to `24.1.0`.

samtstern:
@nicbell thanks for the report!  We're actually using `23.x` for all our testing for other reasons, so we hadn't seen this yet.

SUPERCILEX:
I noticed this issue before, but never bothered to do extensive research on it because I thought it was an isolated issue. I did however notice that if you don't set the eye drawable it still shows up which is weird. Since somebody else noticed this issue on the same version of the support lib that I'm using, I decided to check the release notes and indeed in v24.2.0 they added the password visibly toggle as the default. I'll submit a PR upgrading us to 24.2.1 unless somebody else wants to.

samtstern:
@SUPERCILEX the reason we're not on `24.x` is because Robolectric still only has support for API <= 23, so if we go to 24 we lose testing (or we're testing strange zombie code, which is just as bad).

SUPERCILEX:
Mockito 2.x came out. Doesn't it support api 24?

samtstern:
It's Robolectric, not Mockito.

We can try specifying `sdk = 23` in the `@Config` annotation in all of our tests and then upgrading to support `24.x.x` and SDK `24`, if that works I'd be OK with it.

SUPERCILEX:
My bad, I'll try getting Robolectric to play nice tonight.

samtstern:
Thanks to #349 by @SUPERCILEX I was able to push version `0.6.1` which has all the latest dependencies and should no longer have this issue!

PR #349, opened 2016-10-11
Password visibility toggle fix

SUPERCILEX:
This was a pain. In the end, I had to resort to just adding a workaround which sucks because the current implementation has a glitch where when you toggle visibility, the cursor position resets. Anyway, I created an AOSP issue for the desing lib: https://code.google.com/p/android/issues/detail?id=224914&thanks=224914&ts=1476159232

Back to the PR though, I upgraded gradle dependencies and removed appcompat-v7 which we weren't using. We're using the design lib. I also changed the style of two edittexts that should have been passwords.

#347

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,15 +5,13 @@ android:
   components:
     - platform-tools
     - tools
-    - build-tools-23.0.3
-    - android-23
+    - build-tools-24.0.3
+    - android-24
+
+    # Extras
     - extra-google-google_play_services
     - extra-google-m2repository
     - extra-android-m2repository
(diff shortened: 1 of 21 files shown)


samtstern:
This change LGTM.

@amandle what do you think?  Also I propose if we merge this we do a `0.6.1` release since the underlying Firebase dependency has changed.

amandle:
LGTM, a `0.6.1` release makes sense to me

samtstern:
0.6.1 released

SUPERCILEX:
Awesome!

PR #356, opened 2016-10-15
Fix bug in Smart Lock not restoring Twitter credentials

SUPERCILEX:
@amandle woops?


SUPERCILEX:
I don't get what's up with the Travis builds... My are working fine: https://travis-ci.org/SUPERCILEX/FirebaseUI-Android/builds/167829064 but the library ones keep failing because of some memory error (according to Google). It seems random which is what's so frustrating...

samtstern:
LGTM thanks for catching this, squashing and merging.

Issue #354, opened 2016-10-13
AuthMethodPickerActivity is started multiple times


Alex100:
- Start the firebase demo (version-0.6.0) 
- Rotate the device during the Smartlock account chooser dialog (called by the `ChooseAccountActivity` in `resolveSavedEmails(Activity activity)`)

![image](https://cloud.githubusercontent.com/assets/4115023/19347879/5f1d2a4e-914b-11e6-8c7d-08c0d5c2fc72.png)
- Press the back button multiple times until the `AuthMethodPickerActivity` is visible

![image](https://cloud.githubusercontent.com/assets/4115023/19348136/ab287276-914c-11e6-9960-e75795fbd7d8.png)
- Then the `AuthMethodPickerActivity` is multiple times on the activity stack

The reason for that behaviour is that Smartlock account chooser dialog activity is started every time you rotate the device. This results in multiple calls of `onActivityResult` in ChooseAccountActivity, which starts every time the `AuthMethodPickerActivity`.

SUPERCILEX:
This is fixed in #357.

samtstern:
The fix for this issue has been released in version `1.0.1`.

Issue #171, opened 2016-06-07
Smart Lock does not retrieve saved Facebook credential


wuman:
The current implementation (as of commit c9962e7) uses the `CredentialsAPI` class to request for saved credentials in Smart Lock.  However, it does not appear to retrieve any other account types other than Google.  The end result is that a saved Facebook credential cannot be retrieved for further auto sign-in.

This can seemingly be fixed by requiring the `CredentialsAPI` class depend on a list of account types and then requesting for the credentials of those account types in `CredentialsAPI#requestCredentials()`:

```
public void requestCredentials(final boolean shouldResolve, boolean onlyPasswords) {
    CredentialRequest.Builder crBuilder = new CredentialRequest.Builder().setPasswordLoginSupported(true);

    if (!onlyPasswords) {
        crBuilder.setAccountTypes(IdentityProviders.GOOGLE, IdentityProviders.FACEBOOK);
    }

    ...
}
```

I'd love to send a pull request for this patch, but would like to first raise it here to see if there is a reasoning behind retrieving only the Google account type that I have overlooked in the grand scheme of things.

samtstern:
@wuman a pull request is always welcome, thanks for the detailed issue report.
@amandle do you know why the code is currently this way?

SUPERCILEX:
Fixed in #357.

amandle:
@samtstern I don't know why it was written that wa. It's strange since we have had Facebook support since the beginning.  I will send a PR to fix this for this version of the code, and @SUPERCILEX's fix will land when #357 lands

PR #357, opened 2016-10-16
Refactor chooseAccountActivity

### LINKED ISSUE #182: onActivityResult not fired!
I'm using the new version 0.4.1 of the auth library (on a device with Android 6.0.1).

I'm trying to use the Google sign in provider, launching this intent:

```
startActivityForResult(
                AuthUI.getInstance()
                        .createSignInIntentBuilder()
                        .setProviders(
                                AuthUI.EMAIL_PROVIDER,
                                AuthUI.GOOGLE_PROVIDER)
                        .build(),
                RC_SIGN_IN);
```

When i login with email&password everythings works fine, but when i try to login with the google sign in button, i select the account and i can see a loader for about 3-4 seconds: then the loader disappears and nothing else happens, i still remain into the activity with the two login buttons. 
`onActivityResult` isn't fired and i don't know if login was successful or not.

Thanks.

SUPERCILEX:
#254

TODO @samtstern:
- [x] merge #343 first so that the diff is more readable

TODO me:
- [x] Check auto sign is still working
- [x] Check `hideProgress` is working with auto sign in
- [x] Fix tests
- [x] Fix #182 
- [x] Fix roadblock: on low memory devices, `AuthUI.AuthUIResult` will become null and a null pointer will ensue. How to get rid of `AuthUI.AuthUIResult`. Any ideas?
- [x] Check `SaveSmartLock` behavior to make sure configuration changes don't cause a null pointer exception.
- [x] check for differences in https://github.com/firebase/FirebaseUI-Android/blob/master/auth/src/main/java/com/firebase/ui/auth/ui/ChooseAccountActivity.java
- [x] Do another round of testing
- [x] Do a leak check

Bugs fixed in this PR:

- #152
- #171
- #354
- Part 2 of #254
- Fixed https://github.com/firebase/FirebaseUI-Android/issues/182#issuecomment-241212893 in the sense that there are seperate callbacks for failures and successes meaning that developers can rely on those instead. Activity lifecycle events and auth state listeners will still always take precedence though.


auth/src/main/java/com/firebase/ui/auth/util/SignInDelegate.java
```diff
@@ -0,0 +1,450 @@
...
+
+    private void resolveSavedEmails() {
+        if (mCredentialRequestResult == null || mCredentialRequestResult.getStatus() == null) {
+            return;
+        }
+        Status status = mCredentialRequestResult.getStatus();
+        if (status.getStatusCode() == CommonStatusCodes.RESOLUTION_REQUIRED) {
+            try {
+                // TODO check the mActivity stuff
+                status.startResolutionForResult(mActivity, RC_CREDENTIALS_READ);
```
SUPERCILEX: Here's where I'm stuck. There are two options: either people using the library would have to call `onActivityResult` for us, or I would have to change `SignInDelegate` to an `AppCompatFragment` which would mean people using the library have to pass in an `AppCompatActivity` instead of being able to pass in any activity. I'm in favor of the latter because 99.9% of people are probably already using `AppCompatActivity` anyway since it is necessary for the `Toolbar` (added in API v21).

SUPERCILEX: I decided to go ahead with the `AppCompatActivity` thing because it is a direct subclass of Activity: https://developer.android.com/reference/android/support/v7/app/AppCompatActivity.html so there isn't really a tradeoff to the developer if they are using `Activity` (aka they wouldn't lose anything by switching).

auth/src/main/java/com/firebase/ui/auth/util/smartlock/SignInDelegate.java
```diff
@@ -0,0 +1,428 @@
...
+            Auth.CredentialsApi
+                    .request(mGoogleApiClient,
+                             new CredentialRequest.Builder()
+                                     .setPasswordLoginSupported(true)
+                                     .setAccountTypes(IdentityProviders.GOOGLE)
+                                     .build())
+                    .setResultCallback(this);
+        } else {
+            startAuthMethodChoice();
+        }
```
SUPERCILEX: The speed improvements here are phenomenal: with smart lock disabled, loading the picker activity is basically instant.

Just to refresh your memory, the old way started a bunch of smart lock stuff and only when it had the credentials did it check to see whether or not we actually cared about smart lock. The new way is pretty awesome!

auth/src/main/java/com/firebase/ui/auth/util/smartlock/SignInDelegate.java
```diff
@@ -0,0 +1,403 @@
...
+    }
+
+    @Override
+    public void onResult(@NonNull CredentialRequestResult result) {
+        Status status = result.getStatus();
+
+        if (status.isSuccess()) {
```
SUPERCILEX: Here's where I removed the checks to see if play services was available and if we care about smart lock.

Also, I simplified the whole isAutoSignInAvailable isSignInResolutionNeeded thing since the only time their values ever changed was in onResult which is the only place this stuff gets called.

auth/src/main/java/com/firebase/ui/auth/util/smartlock/SignInDelegate.java
```diff
@@ -0,0 +1,403 @@
...
+    }
+
+    private void handleCredential(Credential credential) {
+        mCredential = credential;
+
+        if (IdentityProviders.GOOGLE.equals(credential.getAccountType())) {
+            // Google account, rebuild GoogleApiClient to set account name and then try
+            initGoogleApiClient(credential.getId());
+            // Try silent sign-in with Google Sign In API
+            Auth.GoogleSignInApi.silentSignIn(mGoogleApiClient);
```
SUPERCILEX: Before, `googleAutoSignIn` was called both in here and also in `onResult`. In here, it is only called if the condition above is true, but in `onResult` it was called no matter what. I'm pretty sure that was a mistake so I removed the call in `onResult` and I'm only doing it here now.

auth/src/main/java/com/firebase/ui/auth/util/smartlock/SignInDelegate.java
```diff
@@ -0,0 +1,403 @@
...
+                .Builder(GoogleSignInOptions.DEFAULT_SIGN_IN)
+                .requestEmail();
+
+        if (accountName != null) {
+            gsoBuilder.setAccountName(accountName);
+        }
+
+        if (mGoogleApiClient != null) {
+            mGoogleApiClient.disconnect();
+        }
```
SUPERCILEX: I removed the `disconnect` `connect` logic from `onStart` and `onStop` because the fragment doesn't care about configuration changes.

auth/src/main/java/com/firebase/ui/auth/util/smartlock/SignInDelegate.java
```diff
@@ -0,0 +1,403 @@
...
+        mGoogleApiClient.connect();
+    }
+
+    private void startAuthMethodChoice() {
+        List<IDPProviderParcel> providers = mFlowParams.providerInfo;
+
+        // If the only provider is Email, immediately launch the email flow. Otherwise, launch
+        // the auth method picker screen.
+        if (providers.size() == 1
+                && providers.get(0).getProviderType().equals(EmailAuthProvider.PROVIDER_ID)) {
```
SUPERCILEX: While refactoring `IDPSignInContainerActivity`, I'll change this to be if any provider is the only one selected we'll launch that directly. Right now, if only Google (for example) is selected, there will be one lonely button in `AuthMethodPickerActivity`. Is that a reasonable change or was there a reason for only caring about whether or not the email provider was the only one selected?

auth/src/main/java/com/firebase/ui/auth/util/smartlock/SignInDelegate.java
```diff
@@ -0,0 +1,403 @@
...
+    }
+
+    private void finish(int resultCode, Intent data) {
+        if (mGoogleApiClient != null) {
+            mGoogleApiClient.disconnect();
+        }
+        hideProgress();
+        mAuthUIResult.onResult(resultCode, data);
+        getActivity().getSupportFragmentManager().beginTransaction().remove(this).commit();
+    }
```
SUPERCILEX: Here's where I do cleanup. Did I miss anything?

SUPERCILEX:
TODO @samtstern:

- [ ] merge #343 first so that the diff is more readable

TODO me:

- [ ] Check auto sign is still working
- [ ] Check `hideProgress` is working with auto sign in
- [ ] Refactor IDPSignInContainerActivity
- [ ] Fix tests

app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
```diff
@@ -125,30 +125,18 @@ public void onCreate(Bundle savedInstanceState) {
...
+                .setProviders(getSelectedProviders())
+                .setTosUrl(getSelectedTosUrl())
+                .setIsSmartLockEnabled(mEnableSmartLock.isChecked())
+                .build(this, this);
     }
 
     @MainThread
-    private void handleSignInResponse(int resultCode, Intent data) {
+    @Override
+    public void onResult(int resultCode, Intent data) {
```
SUPERCILEX: This is what the public API looks like

SUPERCILEX:
I just had a thought! Would a broadcast sender/receiver contract work to replace `AuthUI.AuthUIResult`?

SUPERCILEX:
I also found this [Stack Overflow post](http://stackoverflow.com/a/15007656/4548500) that suggests forcing the Activity to implement `AuthUI.AuthUIResult` (eg anonymous inner classes won't work which could frustrate/confuse newbies (like me a few weeks ago) who don't get that  implementing an interface and passing in `this` is almost the same as passing in `new ...`).

I really prefer the Stack Overflow solution because it is way easier to set up for people (no fiddling with the manifest), but I don't really like either solution (a broadcast receiver seems rather gross but more flexible). Thoughts?

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -213,6 +213,9 @@
  * </ul>
  */
 public class AuthUI {
+    public interface AuthUIResult {
+        void onResult(int resultCode, Intent data);
+    }
```
SUPERCILEX: See https://github.com/firebase/FirebaseUI-Android/pull/357#issuecomment-254104643

samtstern:
@SUPERCILEX just wanted to say that I have seen this PR but there is a lot here (most of it great) so I'm going to just focus on #343 and then take this after, rather than stretch my brain looking at them in parallel.

SUPERCILEX:
Yeah, I figured. I'm still working on tests for #343, but it's almost done.

amandle:
@SUPERCILEX #343 has been merged now :smile:

SUPERCILEX:
@amandle Awesome!

SUPERCILEX:
I'll post a new comment when this PR is worth looking at. Right now, I'm still figuring out merge conflicts and mistakes.

auth/src/test/java/com/firebase/ui/auth/ui/ChooseAccountActivityTest.java
```diff
@@ -1,150 +0,0 @@
...
-                        new AutoCompleteTask<AuthResult>(
-                                new FakeAuthResult(mock(FirebaseUser.class)), true, null));
-
-        chooseAccountActivity.onCredentialsApiConnected(mCredentialsAPI, mActivityHelper);
-
-        verify(mFirebaseAuth).signInWithEmailAndPassword(
-                TestConstants.EMAIL,
-                TestConstants.PASSWORD);
-
-        assertTrue(Shadows.shadowOf(chooseAccountActivity).isFinishing());
```
SUPERCILEX: This test was checking that the email/password survived `signInWithEmailAndPassword()`.

auth/src/test/java/com/firebase/ui/auth/ui/ChooseAccountActivityTest.java
```diff
@@ -1,150 +0,0 @@
...
-                IdpSignInContainerActivity.class.getName(),
-                nextIntent.intent.getComponent().getClassName()
-        );
-        assertEquals(
-                TestConstants.EMAIL,
-                nextIntent.intent.getExtras().getString(ExtraConstants.EXTRA_EMAIL));
-        assertEquals(
-                GoogleAuthProvider.PROVIDER_ID,
-                nextIntent.intent.getExtras().getString(ExtraConstants.EXTRA_PROVIDER)
-        );
```
SUPERCILEX: I removed this test that was making sure the email/password survived the intent.

auth/src/main/java/com/firebase/ui/auth/util/smartlock/SignInDelegate.java
SUPERCILEX: Is disconnect appropriate here? If we are resetting `mGoogleApiClient`, I'm pretty sure we want to disconnect and then reconnect here but I could be wrong.

SUPERCILEX:
I did some more refactoring to make `ActivityHelper` nicer: it now extends `BaseHelper` which has stuff I can use in Fragments.

auth/src/main/java/com/firebase/ui/auth/util/smartlock/SignInDelegate.java
SUPERCILEX: Do you mean multiple activities being started if the device is rotated (#354)? If so, that is fixed here: https://github.com/firebase/FirebaseUI-Android/pull/357/files#diff-34733cf80aa499877942f0fae763af19R18

samtstern:
@SUPERCILEX since I'd love to get this into the next version (which will likely be pushed in the next few days) I am going to give this another review today (I've lost all mental context since you first submitted this).

I believe this should be reviewed before #374 based on the comments in that PR.  Looks like most of the TODOs are cleared here besides updating the docs? That one aside, what does the last TODO mean about comparing to ChooseAccountActivity?

SUPERCILEX:
@samtstern I've been working on #309 so I'll need to spend some time cleaning this one up a bit. The important bits are here: https://github.com/firebase/FirebaseUI-Android/pull/357#pullrequestreview-5343039.
Yes, you are correct #374 is based on this PR.
For the last todo, since git is not showing the diff between `ChooseAccountActivity` and `SignInDelegate`, I need to go through recent commits to `ChooseAccountActivity` and make sure I haven't accidentally introduced regressions (by forgetting to include updates to `ChooseAccountActivity` in `SignInDelegate`).

auth/src/main/java/com/firebase/ui/auth/util/smartlock/SignInDelegate.java
```diff
@@ -0,0 +1,427 @@
...
+        if (BuildConfig.DEBUG) {
+            Log.d(TAG, "onActivityResult:" + requestCode + ":" + resultCode + ":" + data);
+        }
+
+        switch (requestCode) {
+            case RC_CREDENTIALS_READ:
+                if (resultCode == RESULT_OK) {
+                    // credential selected from SmartLock, log in with that credential
+                    Credential credential = data.getParcelableExtra(Credential.EXTRA_KEY);
+                    handleCredential(credential);
```
samtstern: This seems similar to the logic in lines 134-145 above, should we consolidate the logic?

auth/src/main/java/com/firebase/ui/auth/util/smartlock/SignInDelegate.java
```diff
@@ -0,0 +1,427 @@
...
+    }
+
+    private void handleCredential(Credential credential) {
+        mCredential = credential;
+
+        if (IdentityProviders.GOOGLE.equals(credential.getAccountType())) {
+            // Google account, rebuild GoogleApiClient to set account name and then try
+            initGoogleApiClient(credential.getId());
+            // Try silent sign-in with Google Sign In API
+            Auth.GoogleSignInApi.silentSignIn(mGoogleApiClient);
```
samtstern: What happens to the result of this silentSignIn?

auth/src/main/java/com/firebase/ui/auth/util/smartlock/SignInDelegate.java
```diff
@@ -0,0 +1,427 @@
...
+        } else {
+            // Email/password account
+            String status = String.format("Signed in as %s", credential.getId());
+            Log.d(TAG, status);
```
samtstern: nit: We should avoid logging PII (the email).  Let's just say "Signed in with email"

auth/src/main/java/com/firebase/ui/auth/util/smartlock/SignInDelegate.java
samtstern: The disconnect seems appropriate here.  We just need to be sure to call it when we are done with this Fragment (even if the system destroys us) since otherwise we could leak a service connection.

auth/src/main/java/com/firebase/ui/auth/util/smartlock/SignInDelegate.java
```diff
@@ -0,0 +1,427 @@
...
+                            Log.w(TAG, "deleteCredential:failure", task.getException());
+                        }
+                        startAuthMethodChoice();
+                    }
+                });
+    }
+
+    private void redirectToIdpSignIn(String email, String accountType) {
+        Intent nextIntent;
+        switch (accountType) {
```
samtstern: nit: don't we have a "provider translate" helper funciton somewhere?  In which case all the non-default cases of this switch could be:
   `nextIntent = IdpSignInContainerActivity.createIntent(getContext(), mHelper.getFlowParams(), translateProvider(accountType), email)`

auth/src/main/java/com/firebase/ui/auth/util/smartlock/SignInDelegate.java
```diff
@@ -0,0 +1,427 @@
...
+
+    private void finish(int resultCode, Intent data) {
+        if (mGoogleApiClient != null) {
+            mGoogleApiClient.disconnect();
+        }
+        mHelper.dismissDialog();
+
+        try {
+            if (resultCode == RESULT_OK) {
+                ((AuthUI.SignInResult) getActivity()).onAuthUISuccess(data);
```
samtstern: So the `Activity` that implements `AuthUI.SignInResult` here is the host of the Fragment, which is an internal Activity right?  I'm fine with having interfaces like this internally but I think we can find a way to not expose this and therefore not make this breaking change to the public API.

auth/src/main/java/com/firebase/ui/auth/util/smartlock/SignInDelegate.java
```diff
@@ -0,0 +1,427 @@
...
+                nextIntent = AuthMethodPickerActivity.createIntent(
+                        getContext(),
+                        mHelper.getFlowParams());
+        }
+        startActivityForResult(nextIntent, RC_IDP_SIGNIN);
+    }
+
+    private void finish(int resultCode, Intent data) {
+        if (mGoogleApiClient != null) {
+            mGoogleApiClient.disconnect();
```
samtstern: Is there some callback we could listen to so that we make sure to clean up like this when the parent Activity is destroyed?  I know this fragment is retained so we don't care about rotation but there's still the case where the system (or user) kills the Activity for some reason and we want to disconnect the API client and dismiss dialogs.

auth/src/main/java/com/firebase/ui/auth/util/smartlock/SignInDelegate.java
```diff
@@ -0,0 +1,427 @@
...
+                && manager.getActiveNetworkInfo() != null
+                && manager.getActiveNetworkInfo().isConnectedOrConnecting();
+    }
+
+    public static void delegateSignIn(FragmentActivity activity, FlowParameters parameters) {
+        FragmentManager fm = activity.getSupportFragmentManager();
+        FragmentTransaction ft = fm.beginTransaction();
+
+        Fragment fragment = fm.findFragmentByTag(TAG);
+        if (fragment == null || !(fragment instanceof SignInDelegate)) {
```
samtstern: What about the case where the Fragment is not null?  Is that handled by the fact that this is retained?

auth/src/main/java/com/firebase/ui/auth/util/smartlock/SmartLock.java
```diff
@@ -17,11 +24,27 @@
...
-    private static final String TAG = "SmartLockUtil";
+public abstract class SmartLock<R extends Result> extends BaseFragment implements
+        GoogleApiClient.ConnectionCallbacks,
+        ResultCallback<R>,
+        GoogleApiClient.OnConnectionFailedListener {
+    private static final String TAG = "SmartLockBase";
+
+    @Override
+    public void onConnectionSuspended(int i) {
+        if (BuildConfig.DEBUG) {
```
samtstern: Same nit here, unfortunately this will never log (also a few lines below this)

auth/src/main/java/com/firebase/ui/auth/util/smartlock/SignInDelegate.java
```diff
@@ -0,0 +1,427 @@
...
+    }
+
+    @Override
+    public void onConnected(@Nullable Bundle bundle) {
+    }
+
+    @Override
+    public void onActivityResult(int requestCode, int resultCode, Intent data) {
+        super.onActivityResult(requestCode, resultCode, data);
+        if (BuildConfig.DEBUG) {
```
SUPERCILEX: I just copied most of this stuff from `ChooseAccountActivity` so I'll just remove this for now, but if you want to keep it just tell me and I'll add it back.

auth/src/main/java/com/firebase/ui/auth/util/smartlock/SignInDelegate.java
SUPERCILEX: Since this was a `Log.d`, I just removed it altogether.

auth/src/main/java/com/firebase/ui/auth/util/smartlock/SignInDelegate.java
SUPERCILEX: Yup, I replaced it with an if statement.

auth/src/main/java/com/firebase/ui/auth/util/smartlock/SignInDelegate.java
SUPERCILEX: Yup, `onDestroy()` is called if the parent activity dies so I added some cleanup to `SmartLock` because I noticed that we weren't doing anything for `SaveSmartLock` either.

auth/src/main/java/com/firebase/ui/auth/ui/account_link/WelcomeBackPasswordPrompt.java
```diff
@@ -83,8 +82,6 @@ protected void onCreate(Bundle savedInstanceState) {
         TextView bodyTextView = ((TextView) findViewById(R.id.welcome_back_password_body));
         bodyTextView.setText(spannableStringBuilder);
 
-        ((TextInputLayout) findViewById(R.id.password_layout)).setPasswordVisibilityToggleEnabled(false);
```
SUPERCILEX: BTW, I've removed these because as of support lib 25.0.1, `false` is the default

auth/src/main/java/com/firebase/ui/auth/util/smartlock/SignInDelegate.java
SUPERCILEX: Turns out we never used it and never needed it.

I'm going to make a video soon showing all the possible entry points for you guys to see visually what this code is doing. At this point, there's so much different stuff going on that even I'm getting confused.

auth/src/main/java/com/firebase/ui/auth/util/smartlock/SignInDelegate.java
SUPERCILEX: That's the big problem with this refactor: `SignInDelegate` is attached to the activity devs give us in the sign in builder. I Googled anything and everything, but couldn't find an alternative to using interfaces. Also, it seems to be the suggested way to go: https://developer.android.com/training/basics/fragments/communicating.html#Implement

If you have any suggestions, I would love to hear them!

SUPERCILEX:
Okay, here's the video I promised (all we need is to get rid of those container activities and then FirebaseUI auth will be king among sign in flows!):
https://goo.gl/photos/vMoCqdKxmtGP2QMj8

auth/src/main/java/com/firebase/ui/auth/ui/email/RecoverPasswordActivity.java
```diff
@@ -70,12 +71,20 @@ private void next(final String email) {
...
+                                        getString(R.string.confirm_recovery_body),
+                                        email))
+                                .setOnDismissListener(new DialogInterface.OnDismissListener() {
+                                    @Override
+                                    public void onDismiss(DialogInterface anInterface) {
+                                        finish(RESULT_OK, new Intent());
+                                    }
+                                })
+                                .setPositiveButton(android.R.string.ok, null)
+                                .show();
```
SUPERCILEX: I turned the activity into an `AlertDialog` which makes things so much cleaner.

samtstern:
In the spirit of avoiding the API change (which would require a major version bump in the library, among other things) what if the sign in process was all kicked off inside an "invisible" activity that is launched for result.  I did a quick proof of concept here:
https://gist.github.com/samtstern/2705b8c56c86622033d6c30478fdd878

That Activity could host the SignInDelegate and kick off everything else.  It won't be quite as fast as attaching the Fragment to the host Activity directly but it will avoid the UI jank of random container Activities and also give us more control over the flow/interface.

SUPERCILEX:
Sounds good, I'll so some testing with `KickoffActivity` once I get home, push those commits, and we should be ready to go. At some point, we should really try to remove ui-less activites like that.

samtstern:
@SUPERCILEX agree, we should eliminate them (as you've been doing).  However keeping one as a gateway may not be bad in the long run, and it's a pattern I've seen done elsewhere with success.

SUPERCILEX:
@samtstern true. I just realized that using a gateway activity will make my refactor of the container activities much easy. That PR is in a garbage state right now because I couldn't find a good way to only finish the container if it was called from `AuthMethodPickerActivity` and not finish it if the container was called from `SignInDelegate` (because this would finish the user's activity). Either way, a gateway will make my life much easier.

auth/src/main/java/com/firebase/ui/auth/ui/FragmentHelper.java
```diff
@@ -0,0 +1,31 @@
+package com.firebase.ui.auth.ui;
```
samtstern: It looks like we have two copies of this class, one in the `.auth.ui` package and one directly in `.auth`.  I imagine you meant to delete one?

auth/src/main/java/com/firebase/ui/auth/ui/email/RecoverPasswordActivity.java
samtstern: This is great.  Not sure why it was ever an Activity.  It may be better to use a DialogFragment rather than an AlertDialog (to survive rotations) but I am fine to leave that as a TODO.

auth/src/main/java/com/firebase/ui/auth/util/signincontainer/SmartLock.java
```diff
@@ -17,11 +26,64 @@
...
- */
-public class SmartLockUtil {
-    private static final String TAG = "SmartLockUtil";
+public abstract class SmartLock<R extends Result> extends BaseFragment implements
+        GoogleApiClient.ConnectionCallbacks,
+        ResultCallback<R>,
+        GoogleApiClient.OnConnectionFailedListener {
+    private static final String TAG = "SmartLockBase";
+
+    protected GoogleApiClient mGoogleApiClient;
```
samtstern: Is this GoogleApiClient used by this base class at all?

auth/src/main/java/com/firebase/ui/auth/util/signincontainer/SmartLock.java
```diff
@@ -17,11 +26,64 @@
 import java.util.Collections;
 import java.util.List;
 
-/**
- * Helper class to deal with Smartlock Flows.
- */
-public class SmartLockUtil {
-    private static final String TAG = "SmartLockUtil";
+public abstract class SmartLock<R extends Result> extends BaseFragment implements
```
samtstern: nit: I don't think "SmartLock" is the right name for this class.  This class is mostly just a Fragment class that cleans up its own dialogs and GoogleApiClients, there is no SmartLock/CredentialsAPI functionality here.

auth/src/main/java/com/firebase/ui/auth/util/signincontainer/SmartLock.java
SUPERCILEX: Yes below:
```java
    public void cleanup() {
        if (mGoogleApiClient != null) {
            mGoogleApiClient.disconnect();
        }
    }
```

auth/src/main/AndroidManifest.xml
```diff
@@ -16,16 +16,16 @@
             android:value="@string/twitter_consumer_secret"/>
 
         <activity
-            android:name="com.firebase.ui.auth.ui.email.ConfirmRecoverPasswordActivity"
-            android:label="@string/title_confirm_recover_password_activity"
+            android:name=".KickoffActivity"
+            android:label="@string/default_toolbar_title"
             android:exported="false"
-            android:theme="@style/FirebaseUI.Dialog"/>
+            android:theme="@style/Theme.AppCompat.NoActionBar"/>
```
SUPERCILEX: Neither your solution nor mine works. Everything I tried had no effect; even the actionbar is still there. I think something is going on with the style not being applied properly.

auth/src/main/java/com/firebase/ui/auth/ui/FragmentHelper.java
SUPERCILEX: Yup, fixed it.

auth/src/main/java/com/firebase/ui/auth/ui/email/RecoverPasswordActivity.java
SUPERCILEX: Added `RecoveryEmailSentDialog`

SUPERCILEX:
@samtstern Unless you have more comments, this should be ready to go.

I added `RecoveryEmailSentDialog` to fix the dialog being dismissed on rotation.

SUPERCILEX:
I just fixed a leak so now it should be ready to go. 😄

Do we want to keep LeakCanary or should I remove it? I'm pro keeping it since it would help prevent cases like #396 where leak canary could have helped us catch those issues earlier.

SUPERCILEX:
*STOPSHIP*: I'm pretty sure I should be requesting the permissions specified in IdpConfig here: FirebaseUI-Android/pull/357/files#diff-741502abb522ecd56e7c30231e015bdbR112

app/build.gradle
```diff
@@ -46,6 +46,9 @@ dependencies {
     compile 'pub.devrel:easypermissions:0.2.1'
     compile 'com.jakewharton:butterknife:8.4.0'
     apt 'com.jakewharton:butterknife-compiler:8.4.0'
+    debugCompile 'com.squareup.leakcanary:leakcanary-android:1.5'
```
samtstern: +1 to keeping LeakCanary around

samtstern:
@SUPERCILEX regarding your 'STOPSHIP' comment you only need to configure the `GoogleSignInOptions` with scopes if you are going to use that GoogleApiClient to use the `GoogleSignInApi`.  From what I can tell you're only using it for the Credentials API, which means you can actually omit the `addApi(GOOGLE_SIGN_IN_API)` and the `GoogleSignInOptions` altogether.

Other than that I agree, this is looking good to go!

app/build.gradle
SUPERCILEX: Awesome!

SUPERCILEX:
@samtstern I think we're ready to roll!

samtstern:
@SUPERCILEX I agree, I'll wait for Travis' green light.  This has been a marathon!  Thank you so much for seeing it through.

samtstern:
Hate to be the bearer of bad news but found one more error!  Did some manual testing and got a crash here:

**SignInDelegate**
```java
    private void redirectToIdpSignIn(String email, String accountType) {
        if (accountType.equals(IdentityProviders.GOOGLE)
                || accountType.equals(IdentityProviders.FACEBOOK)
                || accountType.equals(IdentityProviders.TWITTER)) {
            IdpSignInContainer.signIn(
```

When I select an email credential from SmartLock, accountType == null so we get a NullPointerException.

SUPERCILEX:
@samtstern I'm not seeing the crash. Also, the only way it could have happened is if the password from the email you selected in smart lock was null which would be weird:
```java
if (!TextUtils.isEmpty(email)) {
    if (TextUtils.isEmpty(password)) {
        // log in with id/provider
        redirectToIdpSignIn(email, getAccountTypeFromCredential());
    } else {
        // Sign in with the email/password retrieved from SmartLock
        signInWithEmailAndPassword(email, password);
    }
}
```

Do you think there is something else going on?

samtstern:
Here's the call stack:
```
Attempt to invoke virtual method 'boolean java.lang.String.equals(java.lang.Object)' on a null object reference
                                                       at com.firebase.ui.auth.util.signincontainer.SignInDelegate.redirectToIdpSignIn(SignInDelegate.java:332)
                                                       at com.firebase.ui.auth.util.signincontainer.SignInDelegate.handleCredential(SignInDelegate.java:241)
                                                       at com.firebase.ui.auth.util.signincontainer.SignInDelegate.onActivityResult(SignInDelegate.java:178)
```

So that would imply that the password is actually empty.  I believe this is a "hint" credential.  The SmartLock picker sometimes shows "hints" which contain email addresses that you have used before but no passwords, basically as an advanced sort of autofill.

samtstern:
I'm not sure exactly why that Credential was in there, but I think we should be defensive and handle null accountType in that code branch (just fall back to email sign in).

SUPERCILEX:
Ah, I get it. My commit fixes the crash, but I still need to figure out flow.

samtstern:
@SUPERCILEX thanks for the commit to fix it.  I am going to merge this PR now!  That seems like a separate issue.

So which issues can I mark as "fix-implemented" now?  And hopefully I will do a release with this issue on Monday (don't want to do it today in case something goes wrong, everyone is out for Thanksgiving)

samtstern:
Also 114 commits!  That has to be some kind of record.  This was a great effort, part of making this library a lot leaner.  Excited to build on this.

SUPERCILEX:
Whew! We did it! I'll look for the issues that can be closed and I also created #416 to fix the smart lock email flow.

PR #359, opened 2016-10-17
Prevent IllegalArgumentException in SaveCredentialsActivity

SUPERCILEX:
#141 

This doesn't fix the root problem, but it will prevent the crash.


samtstern:
Not a full solution but looks better than a crash, thanks!

PR #362, opened 2016-10-17
Return tokens on successful sign in with an IDP

### LINKED ISSUE #286: Return access / ID token after successful sign-in with federated provider

amandle:
This exposes the `IdpResponse` object to the user.  It can be retrieved from the result intent received in `onActivityResult` using `IdpResponse.fromResultIntent(data)`. Resolves #286.


auth/src/main/java/com/firebase/ui/auth/IdpResponse.java
```diff
@@ -0,0 +1,120 @@
...
+     */
+    @Nullable
+    public String getIdpSecret() {
+        return mSecret;
+    }
+
+    /**
+     * Get the email used to sign in.
+     */
+    public String getEmail() {
```
samtstern: NonNull?  Also is there any case this would not be `FirebaseAuth.getInstance().getCurrentUser().getEmail()`?  Would prefer not to duplicate if so.

amandle: It's actually nullable (because Twitter always has to be special). We need to record the email on the `IdpResponse` because we can be in the state where the user has signed in to the IDP but were unable to sign in to Firebase and we need to know the email to to find the existing account to link.

PR #363, opened 2016-10-18
Fix link in proguard-rules.pro

eduardb:
The link for the documentation on proguard seems to be an internal one. Swapped for the public one instead.


eduardb:
I signed it!

samtstern:
Ah thanks!  I now have to put $1 in the "posted staging link on Github" jar 😄

PR #366, opened 2016-10-19
Audit for missed instances of Twitter

amandle:
Audited for missed instances of Twitter.  None of these will actually effect the functionality of the library, but it is better to be consistent.


samtstern:
LGTM, assume travis will pass so just merge once it goes green.  Thanks!

Issue #190, opened 2016-06-18
Unable to change the color of Floating Edit Text in LOGIN WITH EMAIL


kumarkeviv:
Hi Firebase Team

I am trying to change the color of the FLOATING EDIT TEXT in the Firebase UI Login with Email, but its not reflecting the changes as per the style.xml.

XML code i am using to change the LOGIN THEME, whose parent is FIREBASEUI THEME

`<item name="android:textColorHint">@color/nliveo_white</item>
<item name="colorControlNormal">@color/nliveo_blue_alpha_colorPrimaryDark</item>
<item name="colorControlActivated">@color/nliveo_blue_colorPrimary</item>
<item name="colorControlHighlight">@color/authui_colorActivated</item>`

kumarkeviv:
Here is the screenshot of my app:
![screenshot_2016-06-18-22-09-02](https://cloud.githubusercontent.com/assets/17118240/16172271/d35e2c82-35a2-11e6-8bc5-72b6960a2d55.png)

davidgcalle:
Same here, only works if you have a light backgound but if have a dark one, it's really hard to read.

kumarkeviv:
Firebase Team, are we expecting the option for custom changes in element design in recent release? Because without customization it really look awkward and will lead to bad user experience.

ErnestMushinge:
I had the same problem but I solved it by creating a theme in styles.xml then assigning the right parent to that theme I created in styles.xml (parent = "FirebaseUI"). After assigning a parent, then customize the settings in the theme. Make sure you are able to access the FirebaseUI styles in values.xml [www.shungarts.com](http://www.shungarts.com)

`<style name="GreenTheme" parent="FirebaseUI">



        <item name="android:windowBackground">@drawable/bg_login</item>
        <item name="colorPrimary">@color/colorPrimary</item>
        <item name="colorPrimaryDark">@color/colorPrimaryDark</item>
        <item name="colorAccent">@color/colorAccent</item>

        <item name="android:textColor">@color/colorTitle</item>
        <item name="colorButtonNormal">@drawable/button_selector</item>
        <item name="android:buttonStyle">@style/FriendlyButtonStyle</item>
        <item name="android:textColorHint">@color/colorPrimary</item>
        <item name="colorControlNormal">@color/colorPrimary</item>
        <item name="colorControlActivated">@color/colorTitle</item>
        <item name="colorControlHighlight">@color/colorAccent</item>
[...]

PR #367, opened 2016-10-21
Add support for dark themes

amandle:
To solve issue #190 I had to change a bunch about the way the styles were set up. It now relies more heavily on the built-in attributes.  The buttons are more standard material design buttons. There is, however, an issue with the textColor not propagating to the BodyText elements in API < 21 which I haven't figured out yet.  If anyone has suggestions I'd love to get some feedback on this.


amandle:
I figured out the text color propagation issue, this should work on all APIs. Tested on KitKat and Nougat

samtstern:
LGTM.  Thanks for digging into the theme madness.

SUPERCILEX:
@amandle I'm seeing this bug now ("Firebase UI" is no longer centered in the action bar"):

screenshot_20161021-215303

PR #371, opened 2016-10-21
Better logging when IDP login fails.

amandle:
Don't silently fail if Google Sign In fails #182


amandle:
@samtstern PTAL when you have a sec, another small change

samtstern:
LGTM

PR #373, opened 2016-10-22
Fix code that doesn't compile in auth readme

SUPERCILEX:
You need to use `Arrays.asList()`.


amandle:
Thanks for catching that, lgtm.

PR #374, opened 2016-10-22
Refactor container activities

### LINKED ISSUE #348: WelcomeBackIDPPrompt cancels sign in prematurely
While working on #349, I ran into an interesting issue that I'm pretty sure affect all users of 0.6.0. Here was the config I ran with:
image
- Android device: 6P
- Android OS version: 7.0
- Google Play Services version: 9.6.83

Here's some debugging:
image
I haven't had time to look into this issue in great detail, but `mIdpProvider` ends up being null because of `providerParcel` disagreeing with `mProviderId` about what the provider is. BTW, tests are happy.

BTW: I disabled smart lock, started AuthUI with only the email provider, and here is my existing user: 
image

SUPERCILEX:
#254


SUPERCILEX:
Again, #357 should be merged first.

There's one major roadblock that's preventing me from proceeding: Twitter only lets you receive `onActivityResult`s from an activity so there are three options:

1. Make more breaking api changes 😞 that would require devs to send us the `onActivityResult` from their activity.
2. Make two `IdpSignInContainer`s: one that's an activity for `SignInDelegate` and one that's a fragment for `AuthMethodPickerActivity`. (This means that when people are restoring their phones they will see grossness because smart lock will probably be enabled, but other than that, normal sign-in will be pretty) **Prefered option?**
3. Don't refactor `IdpSignInContainerActivity` at all (probably a bad idea)

Thoughts?

SUPERCILEX:
Done refactoring `EmailHintContainerActivity`. It turned out kinda gross so any thoughts would be appreciated.

SUPERCILEX:
I just realized #418 replaces this PR.

PR #380, opened 2016-10-26
Various bugfixes for 1.0

amandle:
Did a manual run through of a bunch of test cases and uncovered some issues which have been fixed here. As well as addressed some internal feedback we received.

fixed:
- idp tokens were being lost after smart lock save
- progress dialog was not dismissed after an incorrect linking password (#228)
- set exported="false" everywhere exported is not needed

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/util/SmartLock.java
+++ b/auth/src/main/java/com/firebase/ui/auth/util/SmartLock.java
@@ -228,13 +228,12 @@ public class SmartLock extends Fragment
         if (response != null) {
             mProvider = response.getProviderType();
         } else {
-            mPassword = null;
+            mProvider = null;
         }
         mProfilePictureUri = firebaseUser.getPhotoUrl() != null ? firebaseUser.getPhotoUrl()
                 .toString() : null;
 
         // If SmartLock is disabled, finish the Activity
-        Intent returnIntent = new Intent().putExtra(ExtraConstants.EXTRA_IDP_RESPONSE, response);
         if (!helper.getFlowParams().smartLockEnabled) {
             finish();
             return;
--- a/auth/src/main/java/com/firebase/ui/auth/util/SmartLock.java
+++ b/auth/src/main/java/com/firebase/ui/auth/util/SmartLock.java
@@ -27,8 +27,10 @@ import android.support.v4.app.FragmentTransaction;
 import android.util.Log;
 
 import com.firebase.ui.auth.BuildConfig;
+import com.firebase.ui.auth.IdpResponse;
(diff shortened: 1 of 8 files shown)


auth/src/main/java/com/firebase/ui/auth/util/SmartLock.java
```diff
@@ -201,7 +205,11 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
     }
 
     private void finish(int resultCode) {
-        mActivity.finish(RESULT_OK, mActivity.getIntent());
+        mActivity.finish(resultCode, mActivity.getIntent());
+    }
+
+    private void finish(int resultCode, Intent data) {
+        mActivity.finish(resultCode, data);
```
SUPERCILEX: You'll need to do some more refactoring for this to work. The reason I chose not to pass down the `resultCode` is because `SmartLock` used to be an Activity called by other activities (such as `RegisterEmailActivity`). Those activities didn't care about what the result code was and always returned `Activity.RESULT_OK`. Say you start the email flow through `AuthMethodPickerActivity` and end up at `RegisterEmailActivity` which calls this smart lock code. If saving the credential fails,`RegisterEmailActivity` will be finished with `Activity.RESULT_CANCELED`. However, the user is signed in and [this line in `AuthMethodPickerActivity`](https://github.com/amandle/FirebaseUI-Android/blob/470147405e20aa52599544ff7000ef04ddc928e4/auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java#L146) will prevent the activity from finishing even though the user is singed in. I don't think this is expected behavior.

amandle: I see, but since the resultCode was being ignored `finish(RESULT_CANCELLED)` like on line 167 would result in `finish(RESULT_OK)` being called.  Maybe it would make sense to change line 167 and line 174 to call `finish(RESULT_OK)` directly instead?

SUPERCILEX: Since this is probably going to be a point of confusion in the future, I think we should just remove the option to set a result at all and always pass down RESULT_OK. What do you think?

amandle: Sounds good to me, done!

samtstern:
LGTM

auth/src/main/java/com/firebase/ui/auth/util/SmartLock.java
SUPERCILEX: Awesome!

PR #382, opened 2016-10-27
Bugfixes for older versions of Android

amandle:
I did some testing on an older android device (4.4) with lower memory and uncovered a couple issues.
- The SmartLock fragment has to be added in `onCreate` otherwise the activity may be destroyed before it attaches (like when it gets backgrounded by the Facebook sign in on a low-memory device)
- Moving the `SmartLock.getInstance` to onCreate required me to rework the SmartLock testing procedure, since it relied on being able to intercept the call after onCreate was finished.
- On the 4.4 device the `GoogleApiClient` was not connecting even though `enableAutoManage` was called. Explicitly calling `connect` fixed this.


auth/src/main/java/com/firebase/ui/auth/ui/ActivityHelper.java
```diff
@@ -106,4 +107,8 @@ public static Intent createBaseIntent(
                 .putExtra(ExtraConstants.EXTRA_FLOW_PARAMS,
                         checkNotNull(flowParams, "flowParams cannot be null"));
     }
+
+    public SmartLock getSmartLockInstance(AppCompatBase activity, String tag) {
+        return SmartLock.getInstance(activity, tag);
```
samtstern: Does this method do anything besides pass through?  Seems like we could just use the static method.

auth/src/main/java/com/firebase/ui/auth/util/SmartLock.java
```diff
@@ -252,12 +255,18 @@ public void saveCredentialsOrFinish(AppCompatBase activity,
...
+            return;
+        }
+
+        mCredentialsApiClient = new Builder(activity)
                 .addConnectionCallbacks(this)
                 .addOnConnectionFailedListener(this)
                 .addApi(Auth.CREDENTIALS_API)
                 .enableAutoManage(activity, this)
                 .build();
+        mCredentialsApiClient.connect();
```
samtstern: Nothing wrong with this but the fact that you had to do it is concerning.  If you can reliably reproduce a situation where this is critical could you file a bug and tell the Gms client folks about it?

auth/src/main/java/com/firebase/ui/auth/util/SmartLock.java
```diff
@@ -269,7 +278,11 @@ public static SmartLock getInstance(AppCompatBase activity, String tag) {
         Fragment fragment = fm.findFragmentByTag(tag);
         if (fragment == null || !(fragment instanceof SmartLock)) {
             result = new SmartLock();
-            ft.add(result, tag).disallowAddToBackStack().commit();
+            try {
+                ft.add(result, tag).disallowAddToBackStack().commit();
+            } catch (IllegalStateException e) {
+                Log.e(TAG, "Can not add fragment", e);
```
samtstern: In this case should we return null or propagate the error somehow?

auth/src/test/java/com/firebase/ui/auth/test_helpers/ActivityHelperShadow.java
```diff
@@ -45,4 +51,9 @@ public FirebaseAuth getFirebaseAuth() {
     public CredentialsApi getCredentialsApi() {
         return credentialsApi;
     }
+
+    @Implementation
```
samtstern: Ah I think I see why you implemented this method, is it just so you can shadow it easily?

auth/src/main/java/com/firebase/ui/auth/ui/ActivityHelper.java
amandle: It just passes through. As you noted below, this makes shadowing it possible.

auth/src/main/java/com/firebase/ui/auth/util/SmartLock.java
amandle: Changed this to return null, and updated the usage to check for null.

PR #386, opened 2016-11-03
Fix header alignment and remove unused layout

amandle:
Apparently `textViewStyle` gets applied to the ActionBar.  Specifying it wasn't getting us anything anyway since all TextViews have their style set explicitly except for the ones in `delete_credential.xml` which was unused.


samtstern:
LGTM

PR #391, opened 2016-11-04
Add instructions for adding the Fabric repo

amandle:


samtstern:
LGTM.

cutiko:
Sorry to be nosy but, this "maven { url 'https://maven.fabric.io/public' }" should at least be warn in the project main page arround here https://github.com/firebase/FirebaseUI-Android#installation I was updating and got stuck with that for some minutes. No need to add full instruction but if you could please add some warning indicating the dev to read the auth docs.

Issue #411, opened 2016-11-20
A question about the FirebaseIndexRecyclerAdapter


TheCraftKid:
* Android device: Nexus 6P
  * Android OS version: Nougat 7.1.1
  * Google Play Services version: 10.0.84
  * Firebase/Play Services SDK version: 9.8.0
  * FirebaseUI version: 1.0.0

The documentation for the [`FirebaseIndexRecyclerAdapter`](https://github.com/firebase/FirebaseUI-Android/blob/master/database/README.md#using-firebaseui-with-indexed-data) is pretty sparse, not even providing an extended example like the one provided for the regular `FirebaseRecyclerAdapter`.

I'm using FirebaseUI version 1.0.0 for the record.

I tried to implement a subclass of the adapter while updating my project's Firebase Database to fit the recommended way to index objects, and the items weren't populated like expected. @SUPERCILEX, since you authored the commits, could you give some guidance?

My simplified database:

    {
      shopItems: {
        randomKey1: {
          name: "Item 1",
          description: "Just an item"
        },
        randomKey2: {
          name: "Item 2",
          description: "Just an item"
        }
      },
      users: {
        randomUserId1: {
          shopItems: {
            randomKey1: true,
            randomKey2: true
          }
        }
      }
    }

My simplified implementation:

    FirebaseUser user = FirebaseAuth.getInstance().getCurrentUser();
    DatabaseReference keyRef = FirebaseDatabase.getInstance().getReference().child("users")
            .child(user.getUid()).child("shopitems");
    DatabaseReference dataRef = FirebaseDatabase.getInstance().getReference().child("shopItems");

    RecyclerView userShopItemList = (RecyclerView) findViewById(R.id.list_user_shop_item);
    userShopItemList.setAdapter(new FirebaseIndexRecyclerAdapter<ShopItem, ShopItemHolder>(ShopItem.class, R.layout.shop_item_viewholder, ShopItemHolder.class, keyRef, dataRef) {
        @Override
        public void populateViewHolder(ShopItemHolder viewHolder, ShopItem item, int position) {
            viewHolder.setName(item.getName());
            viewHolder.setDescription(item.getDescription());
        }
    });

The end result is no items being visibly populated.

If you're wondering, the `ShopItemHolder` is basically a `RecyclerView.ViewHolder` subclass with two methods to set an item's name and description.

SUPERCILEX:
I think you might have made a typo:
[...]

PR #412, opened 2016-11-21
Fix small mistakes in FirebaseIndexArray

SUPERCILEX:
Hi @samtstern, while working on #411, I noticed a few things that slipped past me so this is a super small PR to fix them.

### DIFF
--- a/database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
+++ b/database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
@@ -67,7 +67,7 @@ class FirebaseIndexArray extends FirebaseArray {
             String superKey = super.getItem(keyIndex).getKey();
             if (key.equals(superKey)) {
                 break;
-            } else if (getItem(index).getKey().equals(superKey)) {
+            } else if (mDataSnapshots.get(index).getKey().equals(superKey)) {
                 index++;
             }
         }
@@ -75,7 +75,7 @@ class FirebaseIndexArray extends FirebaseArray {
     }
 
     private boolean isMatch(int index, String key) {
-        return index >= 0 && index < getCount() && getItem(index).getKey().equals(key);
+        return index >= 0 && index < getCount() && mDataSnapshots.get(index).getKey().equals(key);
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
```diff
@@ -67,15 +67,15 @@ private int getIndexForKey(String key) {
...
+            } else if (mDataSnapshots.get(index).getKey().equals(superKey)) {
                 index++;
             }
         }
         return index;
     }
 
     private boolean isMatch(int index, String key) {
-        return index >= 0 && index < getCount() && getItem(index).getKey().equals(key);
+        return index >= 0 && index < getCount() && mDataSnapshots.get(index).getKey().equals(key);
```
SUPERCILEX: We should have never used `getItem()` here because it's public api and anyone overriding that method will break our implementation.

database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
```diff
@@ -155,10 +155,11 @@ public void onDataChange(DataSnapshot snapshot) {
                     notifyChangedListeners(OnChangedListener.EventType.CHANGED, index);
                 }
             } else {
-                Log.w(TAG, "Key not found at ref: " + snapshot.getRef());
                 if (isMatch(index, key)) {
                     mDataSnapshots.remove(index);
                     notifyChangedListeners(OnChangedListener.EventType.REMOVED, index);
+                } else {
+                    Log.w(TAG, "Key not found at ref: " + snapshot.getRef());
```
SUPERCILEX: Only log the warning if the data wasn't in our list. Example: if people delete the data ref before the index ref, the warning would have been shown which is kinda pointless.

database/src/main/java/com/firebase/ui/database/FirebaseIndexListAdapter.java
```diff
@@ -41,7 +42,7 @@
      */
     public FirebaseIndexListAdapter(Activity activity,
                                     Class<T> modelClass,
-                                    int modelLayout,
+                                    @LayoutRes int modelLayout,
```
SUPERCILEX: I also just noticed that we never had an `@Res` annotation on our layouts so I added that too.

samtstern:
This LGTM, thanks!

SUPERCILEX:
Awesome!

PR #416, opened 2016-11-23
Fix smart lock email flow

SUPERCILEX:


auth/src/main/java/com/firebase/ui/auth/util/signincontainer/SignInDelegate.java
```diff
@@ -329,9 +329,19 @@ public void onComplete(@NonNull Task<Status> task) {
     }
 
     private void redirectToIdpSignIn(String email, String accountType) {
-        if (accountType != null && (accountType.equals(IdentityProviders.GOOGLE)
+        if (TextUtils.isEmpty(accountType)) {
+            startActivityForResult(
+                    SignInNoPasswordActivity.createIntent(
```
SUPERCILEX: Should I directly start the register email activity or leave it at this?

samtstern: So we actually can't go straight to RegisterEmailActivity since we're not sure at this point if it's a new or existing account.  I think this is safest.

SUPERCILEX: That's what I was thinking, play it safe.

PR #417, opened 2016-11-24
Use support library password toggle

SUPERCILEX:


SUPERCILEX:
There are huge benefits to doing this:

- Removes a bunch of pngs
- Much cleaner code (182 lines deleted vs only 21 lines added)
- Fixes bug where cursor position is reset when toggling password visibility
- Uses the support library so bugs will be fixed and features added

samtstern:
This looks good to me, thanks for the simplification!

PR #418, opened 2016-11-24
Complete rewrite/refactor of email sign in flow

### LINKED ISSUE #129: Propagate display name from Smart Lock for Passwords to account creation screen
The hint received from Smart Lock for Passwords is currently only used for the email address. The display name should also be collected and propagated to the account creation screen, if an account does not already exist for the email address.

SUPERCILEX:
I'm really excited about this one!!! The performance/ui/code is massively improved: we've gone from jumping through 3 activities to just saying on `RegisterEmailActivity` and everything regarding email flow is so much simpler and easier to understand now! Here are some of the important changes:

- fixes #129
- lets the user update their email in the `RegisterEmailActivity`
- no more confusing activity jumping (except when they already have an account, but that's for another day)
- performance improvements
- less code, fewer tests, and more features = happiness


auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailActivity.java
```diff
@@ -129,13 +165,129 @@ private void setUpTermsOfService() {
...
+                                                        mActivityHelper.getApplicationContext(),
+                                                        mActivityHelper.getFlowParams(),
+                                                        provider,
+                                                        null,
+                                                        email);
+                                                mActivityHelper.startActivityForResult(intent,
+                                                                                       RC_WELCOME_BACK_IDP);
+                                            }
+                                        }
+                                    } // TODO: 11/23/2016 what happens if we fail?
```
SUPERCILEX: Thoughts? We didn't do anything before, but I think we should display a message or at least do something.

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailActivity.java
```diff
@@ -94,21 +106,45 @@ protected void onCreate(Bundle savedInstanceState) {
...
+        String email = getIntent().getStringExtra(ExtraConstants.EXTRA_EMAIL);
+        if (!TextUtils.isEmpty(email)) {
             mEmailEditText.setText(email);
-            mEmailEditText.setEnabled(false);
+            mNameEditText.requestFocus();
+            return;
+        }
+
+        if (mActivityHelper.getFlowParams().smartLockEnabled) {
+            showEmailAutoCompleteHint();
```
SUPERCILEX: Why are we doing this only if smart lock is enabled? I think users should get autocomplete whether or not devs have an aversion to smart lock.

auth/src/main/res/values/strings.xml
```diff
@@ -59,6 +59,7 @@
     <string name="progress_dialog_sending">Sending…</string>
     <string name="progress_dialog_signing_in">Signing in…</string>
     <string name="progress_dialog_signing_up">Signing up…</string>
+    <string name="progress_dialog_checking_accounts">Checking for existing accounts…</string>
```
SUPERCILEX: Does this sound good?

SUPERCILEX:
I've been off and on trying to figure out if a memory leak was caused by my changes in this PR and I finally figured it out! Turns out this PR wasn't the cause, but `RegisterEmailActivity` helped me figure it out because it creates an instance of `SaveSmartLock`.

#### Full explanation:

I've stubbed my toe several times now with `setRetainInstance(true)` and I think I've finally eradicated all memory leaks related to the smart lock stuff. (Yay! 😄) If you look at the [commit fixing the smart lock leak](https://github.com/firebase/FirebaseUI-Android/pull/418/commits/78e83ca2eff79281a2f2c44ad46d0eb8ee64755e), you'll notice that I was passing in `Fragment#getContext()` to `BaseHelper`'s constructor which stores the context in `mContext`. (`BaseHelper`'s subclass, `FragmentHelper` is initialized in `Fragment#onCreate` which only occurs once because of `setRetainInstance`). In this case, `mContext` is the activity associated with `SaveSmartLock`. This would work fine without `setRetainInstance(true)`, however, because the fragment lives longer than its activity, we leak `SaveSmartLock`'s activity aka `mContext`. A simple fix is to use `Fragment#getContext()#getApplicationContext()`. Unfortunately, `ProgressDialog`s don't work with `getApplicationContext()` so I'm using [`fragment.getContext()` only when needed](https://github.com/firebase/FirebaseUI-Android/pull/418/commits/78e83ca2eff79281a2f2c44ad46d0eb8ee64755e#diff-1cf438cbf31168e7379ca3db1c5a5477R25) (to show the dialog).

### TLDR
Keep an eye out for `getContext()` leaks in `setRetainInstance(true)` fragments!

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailActivity.java
```diff
@@ -73,24 +87,43 @@ protected void onCreate(Bundle savedInstanceState) {
...
             mEmailEditText.setText(email);
-            mEmailEditText.setEnabled(false);
+            mNameEditText.requestFocus();
+            return;
         }
-        setUpTermsOfService();
-        Button createButton = (Button) findViewById(R.id.button_create);
-        createButton.setOnClickListener(this);
+
+        showEmailAutoCompleteHint();
```
SUPERCILEX: Me previously: 
> Why are we doing this only if smart lock is enabled? I think users should get autocomplete whether or not devs have an aversion to smart lock.

I went through with that idea so unless anyone has any objections, I really think autocomplete should be available whether or not smart lock is enabled.

samtstern: Technically the email hint is provided by SmartLock, so this would make the API somewhat dishonest.  Although I agree with you that enabling the hint for everyone may be a superior UX, I want to respect the intent of the developer when they say "disable smartlock".

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailActivity.java
```diff
@@ -111,28 +144,141 @@ private void setUpTermsOfService() {
...
+            case RC_SIGN_IN:
+            case RC_WELCOME_BACK_IDP:
+                finish(resultCode, data);
+                break;
+        }
+    }
+
+    @Override
+    public void onFocusChange(View view, boolean hasFocus) {
+        if (hasFocus) return;
```
samtstern: Just a general style comment (first noticed it here) but let's add some more comments in meaty methods like this.  For instance in this method the first line could use a comment like `// Only consider fields losing focus`.

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailActivity.java
```diff
@@ -111,28 +144,141 @@ private void setUpTermsOfService() {
...
+                    .addOnFailureListener(
+                            new TaskFailureLogger(TAG, "Error fetching providers for email"))
+                    .addOnCompleteListener(
+                            new OnCompleteListener<ProviderQueryResult>() {
+                                @Override
+                                public void onComplete(@NonNull Task<ProviderQueryResult> task) {
+                                    if (task.isSuccessful()) {
+                                        List<String> providers = task.getResult().getProviders();
+                                        if (providers != null && !providers.isEmpty()) {
+                                            // Account does exist
```
samtstern: Recommended to expand comments to something like this:
```java
// There is an account tied to this email.  If only the email provider is associated with the account, direct the user
// to the email sign in flow.  Otherwise, direct the user to sign in with the IDP they previously selected.
```
```

File: auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailActivity.java [hunk 7]
```diff
@@ -111,28 +144,141 @@ private void setUpTermsOfService() {
...
+        if (!TextUtils.isEmpty(email)) {
+            mActivityHelper.getFirebaseAuth()
+                    .fetchProvidersForEmail(email)
+                    .addOnFailureListener(
+                            new TaskFailureLogger(TAG, "Error fetching providers for email"))
+                    .addOnCompleteListener(
+                            new OnCompleteListener<ProviderQueryResult>() {
+                                @Override
+                                public void onComplete(@NonNull Task<ProviderQueryResult> task) {
+                                    if (task.isSuccessful()) {
```
samtstern: nit: use separate success and failure listeners rather than checking `task.isSuccessful()`

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailActivity.java
samtstern: Yeah not sure what the best course of action is here ... since we didn't do anything before let's leave this for another time.

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailActivity.java
```diff
@@ -143,13 +289,13 @@ public void onComplete(@NonNull Task<Void> task) {
...
                                                 mSaveSmartLock,
-                                                firebaseUser,
+                                                user,
                                                 password);
                                     }
                                 });
                     }
                 })
-                .addOnFailureListener(new OnFailureListener() {
+                .addOnFailureListener(this, new OnFailureListener() {
```
samtstern: Any reason we're only tying the failure listener to the Activity and not the success listener?

auth/src/main/res/values/strings.xml
samtstern: SGTM.

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailActivity.java
SUPERCILEX: Here was my logic:

If the request succeeds, we to continue whether or not the original activity is still there because `mSaveSmartLock` is already initialized and not tied to the activity lifecycles. That way, the activity will finish and the sign in will proceed.

If it fails, the only thing we are doing in the failure listener is updating views. I'm not sure how that works, but I would think something bad will happen if we try to update a view from a previous instance of an activity. Am I wrong?

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailActivity.java
SUPERCILEX: Sounds fair, but still it makes me sad inside. 😄

app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
```diff
@@ -50,7 +50,7 @@
 public class AuthUiActivity extends AppCompatActivity {
     private static final String UNCHANGED_CONFIG_VALUE = "CHANGE-ME";
     private static final String GOOGLE_TOS_URL = "https://www.google.com/policies/terms/";
-    private static final String FIREBASE_TOS_URL = "https://www.firebase.com/terms/terms-of-service.html";
+    private static final String FIREBASE_TOS_URL = "https://firebase.google.com/terms/";
```
SUPERCILEX: Just noticed this was using the old Firebase.

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailActivity.java
```diff
@@ -111,28 +149,156 @@ private void setUpTermsOfService() {
...
+                TypedValue typedValue = new TypedValue();
+                getTheme().resolveAttribute(R.attr.colorPrimary, typedValue, true);
+                @ColorInt int color = typedValue.data;
+
+                new CustomTabsIntent.Builder()
+                        .setToolbarColor(color)
+                        .build()
+                        .launchUrl(
+                                RegisterEmailActivity.this,
+                                Uri.parse(mActivityHelper.getFlowParams().termsOfServiceUrl));
```
SUPERCILEX: Since Facebook forces us to use CCT, I don't see why we shouldn't modernize.

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailActivity.java
```diff
@@ -111,28 +149,156 @@ private void setUpTermsOfService() {
         agreementText.setOnClickListener(new View.OnClickListener() {
             @Override
             public void onClick(View view) {
-                Intent intent = new Intent(Intent.ACTION_VIEW).setData(Uri.parse
-                        (mActivityHelper.getFlowParams().termsOfServiceUrl));
-                startActivity(intent);
+                // Getting default color
+                TypedValue typedValue = new TypedValue();
+                getTheme().resolveAttribute(R.attr.colorPrimary, typedValue, true);
```
SUPERCILEX: [This works](https://youtu.be/SmCo8-371Ec), but what if they don't have a `colorPrimary`. (Actually is it even possible to not have a colorPrimary? I'm pretty sure `AppCompatActivity` uses that for all default theming)

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailActivity.java
```diff
@@ -168,6 +334,7 @@ public void onFailure(@NonNull Exception e) {
                         } else if (e instanceof FirebaseAuthUserCollisionException) {
                             // Collision with existing user email
                             emailInput.setError(getString(R.string.error_user_collision));
+                            checkAccountExists(mEmailEditText.getText().toString());
```
SUPERCILEX: Come to think of it, this could be our error handling if `fetchProvidersForEmail` fails: the user can't sign up if a user already exists and we will retry to loading that user.

app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
samtstern: Wow, good catch :-)

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailActivity.java
samtstern: I think this should be fine ... and if people see a crash I think saying "you must define colorPrimary" is a reasonable response so not too worried.

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailActivity.java
samtstern: Agreed, looks great in your screencast!

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailActivity.java
samtstern: SGTM.

SUPERCILEX:
@samtstern Sounds good, I changed the base for all my PRs and pulled from master.

PR #420, opened 2016-11-25
Add and enforce checkstyle, lint, findbugs, and pmd

SUPERCILEX:
I noticed two things that lead to this PR:

1. We never did a full gradle check
2. We had a checkstyle, but it was never enforced

I started trying to get checkstyle to work and ended up adding all the common CI analysis tools/checks.

I went through and added checkstyle, lint, findbugs, and pmd. PMD is the only check that isn't enforced because it still needs to be configured to ignore tests. (It doesn't like the underscores in the test method names)

In the process, I fixed a ton of style and code issues.


.travis.yml
```diff
@@ -11,12 +11,45 @@ cache:
...
     - tools
-    - build-tools-25.0.0
+    - build-tools-25.0.1
     - android-25
-
-    # Extras
-    - extra-google-m2repository
+    - platform-tools
     - extra-android-m2repository
-script: ./gradlew clean :library:testAll :library:prepareArtifacts
```
SUPERCILEX: What was `prepareArtifacts` for?

build.gradle
```diff
@@ -8,7 +8,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.2.2'
+        classpath 'com.android.tools.build:gradle:2.3.0-alpha1'
```
samtstern: What does 2.3 do for us?

.travis.yml
samtstern: That's the command I run before uploading to jCenter, I want to run it on each build to make sure that the build will be "releasable".

build.gradle
SUPERCILEX: 2.3 added new lint features including the baseline and a whole [bunch of other stuff](http://tools.android.com/tech-docs/lint-in-studio-2-3).

.travis.yml
SUPERCILEX: Sounds good, I added it as a dependency of gradle's `check`.

SUPERCILEX:
Will do!

samtstern:
See: https://github.com/firebase/FirebaseUI-Android/pull/418#issuecomment-265209250

SUPERCILEX:
@samtstern merged.

SUPERCILEX:
This is awesome! The build failed because I forgot to delete an unused string resource.

samtstern:
That is cool, also I love describing a build failure as "awesome" :+1:

SUPERCILEX:
Yeah, "useful" would probably be a better word 😄

SUPERCILEX:
That's one thing I wasn't sure about, do we want to enforce a max line length?

SUPERCILEX:
@samstern, good to go! See my comment above though.

samtstern:
@SUPERCILEX regarding max line length I don't think we need to enforce that, I generally try to keep my lines under 100 chars but I subscribe to the golang philosophy:

'Go has no line length limit. Don't worry about overflowing a punched card. If a line feels too long, wrap it and indent with an extra tab.'

samstern:
Hey, I think you have the wrong guy. Are you confusing me with @samtstern?

samtstern:
@samstern yes I think he was, cool name though :smile:

samstern:
It's the BEST name!!

SUPERCILEX:
Oops, sorry @samstern that was a typo 😄

PR #424, opened 2016-11-28
Make use of standard RecyclerView Holder optimizations.

CodingDoug:
An important optimization that RecyclerView offers is the ability to cache the results of (potentially expensive) calls to findViewById() inside a ViewHolder so that they don't have to be repeated every time data is bound to a recycled view.

### DIFF
--- a/app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
+++ b/app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
@@ -226,49 +226,50 @@ public class ChatActivity extends AppCompatActivity implements FirebaseAuth.Auth
     }
 
     public static class ChatHolder extends RecyclerView.ViewHolder {
-        View mView;
+        private final TextView mNameField;
+        private final TextView mTextField;
+        private final FrameLayout mLeftArrow;
+        private final FrameLayout mRightArrow;
+        private final RelativeLayout mMessageContainer;
+        private final LinearLayout mMessage;
 
         public ChatHolder(View itemView) {
             super(itemView);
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
```diff
@@ -226,49 +226,50 @@ public String getText() {
...
-            int color;
+        public void setIsSender(boolean isSender) {
+            final int color;
             if (isSender) {
-                color = ContextCompat.getColor(mView.getContext(), R.color.material_green_300);
-
-                left_arrow.setVisibility(View.GONE);
-                right_arrow.setVisibility(View.VISIBLE);
-                messageContainer.setGravity(Gravity.END);
+                color = ContextCompat.getColor(itemView.getContext(), R.color.material_green_300);
```
samtstern: While we're feeling cache-y, might as well cache these two colors on ViewHolder creation (or even at a higher level)

samtstern:
LGTM, made one comment for an additional improvement while I have you here.

app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
CodingDoug: Considered that, but those could actually theoretically change when the configuration changes as well.  If we want to assume they don't, then they're cachable as well.

samtstern: Since this is a very simple sample app let's assume they don't.

PR #376, opened 2016-10-23
Auth code cleanup

SUPERCILEX:
For the removed todo, see http://stackoverflow.com/questions/2643949/can-i-use-a-static-variable-from-a-java-class-in-res-values-styles-xml.

I added RTL support, a wee bit of accessibility, and removed unused resources.

For other stuff, see the review below:

PR #425, opened 2016-11-28
Update AcquireEmailHelper.java

pahdo:
Changed request codes to public from private.

This is a sign-in example from the readme, which seems to require these request codes to be public.

```java
startActivityForResult(
    // Get an instance of AuthUI based on the default app
    AuthUI.getInstance().createSignInIntentBuilder().build(),
    RC_SIGN_IN);
```

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/ui/email/AcquireEmailHelper.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/email/AcquireEmailHelper.java
@@ -33,10 +33,10 @@ import java.util.List;
 
 public class AcquireEmailHelper {
     private static final String TAG = "AcquireEmailHelper";
-    private static final int RC_REGISTER_ACCOUNT = 14;
-    private static final int RC_WELCOME_BACK_IDP = 15;
-    static final int RC_SIGN_IN = 16;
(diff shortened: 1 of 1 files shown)


samtstern:
@pahdo sorry I think the README is just not clear here.

You should define `RC_SIGN_IN` yourself (in your Activity) and then check for it in `onActivityResult`.  These internal values are not what you are looking for and should remain private.

I am going to close this PR, appreciate your taking the time to do it!  In general you should submit an issue before a PR so that things like this can be discussed (and maybe save you some time).

SUPERCILEX:
@samtstern Sorry about that, I was actually using the `RC_SIGN_IN` from `AcquireEmailHelper` too so there might be an uptake of confused people because of #376.

samtstern:
@SUPERCILEX doesn't look like #376 changed the README at all, what change do you think may have caused this confusion/

SUPERCILEX:
@samtstern It came from [here](https://github.com/firebase/FirebaseUI-Android/pull/376/files#diff-e86d7fc5698ebf408d03c9547303195bL36) and [this conversation](https://github.com/firebase/FirebaseUI-Android/pull/376/files/00a59cd0a9655a5f2f9b2e9ec0a2baddb931e55a#diff-1c771a2d188abf568a7f838ed24396d6).

Issue #405, opened 2016-11-14
Possibility to differentiate between canceled & failed sign in


Juchar:
There is currently no possibility to differentiate between a **canceled** sign in (e.g. the user pressed the back button to leave the auth method picker) and a **failed** sign in attempt.

As far as I can see this always results in `RESULT_CANCELED` as a return code.

By consequence it is difficult to show an according message to the user only if the login failed, as for a canceled login I do not want to show an error message (user decided to cancel the login).

Would it be an idea to introduce something like a `RESULT_SIGN_IN_FAILED` return code?
Or how is this inteded to be used in general?

SUPERCILEX:
#357 has that goal in mind: your activity would have to implement (I'm hoping to find a way around that) `AuthUI.SignInResult`:
```java
public interface SignInResult {
    void onAuthUISuccess(IdpResponse data);

    void onAuthUIFailure(int resultCode);
}
```

You could get different failure codes from the `resultCode`. There is still a lot more work to be done, but I think something like that will happen by 2.0.

Juchar:
Ok, thank you for the information.
And to break the code of conduct :) -> is there an ETA for the next release? Some milestone?

Juchar:
I have to say that this is currently really a blocker for me regarding usability. The way it is now I have to tell the user that the sign in failed even if he consciously pressed on the back button and cancelled the sign in process. 

Is there possibly at least some workaround or something similar?
Or do you possibly have a suggestion on how to handle this case from a UI point of view?

I am really a bit struggling (sorry if this is the wrong place to ask for help, please be so kind to redirect me if needed).

SUPERCILEX:
@Juchar Haleluya! This issue will probably be fixed in `v1.1.0`. Sorry about not getting it done in #357 like I had originally planned. In the end, we didn't change the public api so I decided to hold back on a `ResultCodes` refactor.

You can check out the progress of this refactor here: #426.

samtstern:
A fix for this issue has been released in version `1.1.0`

PR #426, opened 2016-11-28
Differentiate between user cancellation and actual errors

SUPERCILEX:
#405


app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
```diff
@@ -34,7 +34,7 @@
 import com.firebase.ui.auth.AuthUI;
 import com.firebase.ui.auth.AuthUI.IdpConfig;
 import com.firebase.ui.auth.IdpResponse;
-import com.firebase.ui.auth.ui.ResultCodes;
+import com.firebase.ui.auth.ResultCodes;
```
SUPERCILEX: The library version should probably be bumped to `1.1.0` since this is slight breakage.

auth/src/main/java/com/firebase/ui/auth/ResultCodes.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+public final class ResultCodes {
+    private ResultCodes() {
+        // We don't want people to initialize this class
+    }
+
+    /**
+     * Sign in succeeded
+     **/
+    public static final int OK = Activity.RESULT_OK;
```
SUPERCILEX: I decided to unify result codes and make `ResultCodes` a central place for those.

I decided to use the existing activity result codes to not break anything.

auth/src/main/java/com/firebase/ui/auth/ResultCodes.java
```diff
@@ -0,0 +1,33 @@
...
+
+    /**
+     * Sign in canceled by user
+     **/
+    public static final int CANCELED = Activity.RESULT_CANCELED;
+
+    /**
+     * Sign in failed due to lack of network connection
+     **/
+    public static final int NO_NETWORK = 10;
```
SUPERCILEX: I also removed the `RESULT_` bit since the class is named `ResultCodes` and it looks nicer without the extra `RESULT_`.

auth/src/main/java/com/firebase/ui/auth/ResultCodes.java
```diff
@@ -0,0 +1,33 @@
...
+
+    /**
+     * An unknown error has occurred
+     **/
+    public static final int UNKNOWN_ERROR = 20;
```
SUPERCILEX: Should we be more specific here? The errors finished with this code vary widely and we log each of them so I didn't think it was necessary. The only thing people want to know is if the user cancelled the sign in or an error occurred.

SUPERCILEX:
I also added an `isNewUser` method to `IdpResponse` in this PR because I thought it would have something to do with `ResultCodes`. Turns out it didn't and I'm not quite sure why I thought that in the first place, but here it is, crammed into this PR.

SUPERCILEX:
Turns out there's no way to know if an IDP is a new user so I changed the method name to `isNewEmailUser`. If I'm wrong, I'd love hear a solution!

app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
samtstern: I think this would technically require a major version (if you're going by semver) since old code would break when updated.

Could we keep the `.ui.ResultCodes` class but have it be all `@Deprecated` constants that are just symlinks to the new class?  Then it's fully backwards compatible.

auth/src/main/java/com/firebase/ui/auth/ResultCodes.java
samtstern: This is fine for now, it would be nice to be more expressive in the long term but I don't want to overwhelm people with 50 error codes.

auth/src/main/java/com/firebase/ui/auth/IdpResponse.java
```diff
@@ -99,6 +110,13 @@ public String getEmail() {
         return mEmail;
     }
 
+    /**
+     * @return true if the user was created in the sign in flow, false if it already existed.
+     */
+    public boolean isNewEmailUser() {
```
samtstern: I think we should delay this feature because with a little more effort (which would be out of scope here) we could do a more complete job with the "new or existing user" problem.

At the time when we acquire the users email (either directly or through an IDP response) we could do a query for providers from email and use that to determine if it's a new or existing user.  It comes at the cost of an extra request but at least it would be correct.

The other option is just to recommend that developers keep an index of known users in the Realtime Database, which is what ~90% of devs are doing  anyway (for various reasons).  We could even provide a FBUI helper function to manage this.

In any case, we should discuss it thoroughly.

auth/src/main/java/com/firebase/ui/auth/ResultCodes.java
SUPERCILEX: That's what I was thinking: we can expand later, but for now at least we have something to work with.

auth/src/main/java/com/firebase/ui/auth/IdpResponse.java
SUPERCILEX: I learned something new today! I thought `fetchProvidersForEmail` only worked for email/password accounts.

I agree, since this is going to be much more complicated, I'll save it for another PR.

app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
samtstern: We can remove deprecated things after a major version change (if we want to).  Or we can leave it for life.  This will be the first example of deprecation in this library so there's no precedent for us, and every library does it a bit differently.

app/src/main/java/com/firebase/uidemo/storage/ImageActivity.java
```diff
@@ -82,7 +83,7 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
         super.onActivityResult(requestCode, resultCode, data);
 
         if (requestCode == RC_CHOOSE_PHOTO) {
-            if (resultCode == RESULT_OK) {
+            if (resultCode == ResultCodes.OK) {
```
samtstern: This seems like an IntelliJ refactor bug, in this case I am not using AuthUI I want to see if the Camera returned Activity.RESULT_OK

auth/README.md
```diff
@@ -204,8 +204,8 @@ startActivityForResult(
 
 #####Response codes
 The authentication flow only provides three response codes:
```
samtstern: This sentence is no longer true ("only three response codes..."). Maybe we should just point to the ResponseCodes file here and rely on JavaDoc comments so we don't have to keep updating this.

app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
SUPERCILEX: Ok, thanks for the info! (I'm pretty sure we also deprecated `setProviders(String... providers)`)

auth/src/main/java/com/firebase/ui/auth/util/signincontainer/IdpSignInContainer.java
```diff
@@ -69,7 +69,7 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
 
         if (providerConfig == null) {
             // we don't have a provider to handle this
-            finish(Activity.RESULT_CANCELED, new Intent());
+            finish(ResultCodes.UNKNOWN_ERROR, new Intent());
```
samtstern: Changing all of the CANCELED to UNKNOWN_ERROR is still a breaking API change, as people checking for CANCELED in code will no longer see the same behavior.

This makes me think maybe we should simplify to three cases:
  * OK
  * CANCELED (with optional extras describing the error)
  * NO_NETWORK (just to keep backwards compat, otherwise I'd say we lump this into CANCELED)

samtstern: That seems like the right answer :+1:

auth/README.md
```diff
@@ -203,38 +203,41 @@ startActivityForResult(
...
+the user to sign in again later, or proceed with anonymous sign-in if supported.
 
 ```java
-    if (resultCode == RESULT_OK) {
-        // user is signed in!
-        startActivity(new Intent(this, WelcomeBackActivity.class));
-        finish();
+    if (data == null) {
```
samtstern: nit: let's check `if (requestCode != ResultCodes.OK && data == null)` here to more clearly indicate what happened.

app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
```diff
@@ -199,22 +200,33 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
 
     @MainThread
     private void handleSignInResponse(int resultCode, Intent data) {
-        if (resultCode == RESULT_OK) {
-            startActivity(SignedInActivity.createIntent(this, IdpResponse.fromResultIntent(data)));
-            finish();
-            return;
+        if (data != null) {
```
SUPERCILEX: What do you think of this way of handling the sign in response?

samtstern: There are a few things here:

- I'd like checking for RESULT_OK to be a top-level check.  That's our contract, not the nullity of `data`.
- When the resultCode is CANCELED we will have two quick snackbars right?  One because the IdpResponse contains UNKNOWN_ERROR and one because it was canceled.

auth/README.md
```diff
@@ -203,38 +203,46 @@ startActivityForResult(
...
     super.onActivityResult(requestCode, resultCode, data);
-    if (resultCode == RESULT_OK) {
-        // user is signed in!
-        startActivity(new Intent(this, WelcomeBackActivity.class));
-        finish();
-        return;
-    }
+    if (requestCode == RC_SIGN_IN) {
+        if (data != null) {
+            IdpResponse response = IdpResponse.fromResultIntent(data);
```
samtstern: How about:

```java
if (requestCode == RC_SIGN_IN) {
  IdpResponse response = (data == null) ? null : IdpResponse.fromResultIntent(data);
  if (resultCode == ResultCodes.OK) {
    startActivity(SignedInActivity.createIntent(this, response));
    finish();
    return
  }

  if (response != null) {
    // All the error handling ...
  }
```

auth/src/main/java/com/firebase/ui/auth/ErrorCodes.java
```diff
@@ -0,0 +1,20 @@
...
+ */
+public final class ErrorCodes {
+    private ErrorCodes() {
+        // We don't want people to initialize this class
+    }
+
+    /**
+     * Sign in failed due to lack of network connection
+     **/
+    public static final int NO_NETWORK = 10;
```
samtstern: Isn't this one actually a result code and not an error?  We still need to return this via onActivityResult in order to honor our previous API contract.

auth/README.md
SUPERCILEX: That's cleaner, but it doesn't work. Since we are now returning `ResultCodes.CANCELED` with an added error code, it's a little different. So how about [this](https://github.com/SUPERCILEX/FirebaseUI-Android/blob/d56058b85b02648ed2c4479c699ffe9f6259b6c7/app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java#L202)?

auth/src/main/java/com/firebase/ui/auth/ErrorCodes.java
SUPERCILEX: No, I don't think so. It should be an error code, but to keep backwards compatibility I'm [returning it in `SignInDelegate`](https://github.com/firebase/FirebaseUI-Android/pull/426/files#diff-741502abb522ecd56e7c30231e015bdbR83) which the [old `ResultCodes` points to](https://github.com/firebase/FirebaseUI-Android/pull/426/files#diff-314d061f90014b41124025139571c0e3R18).

samtstern: Ah ok i see what you did here.  So you're returning it in both places for SignInDelegate, with the recommendation being that from here on out people use the errorCode to make decisions when the result is not OK.  That makes sense!

samtstern:
Everything LGTM.  I need to figure out what the next milestone will look like before I merge this into master, we might want a dev branch instead.

SUPERCILEX:
@samtstern Sounds good to me!

SUPERCILEX:
@samtstern I [added 1 commit](https://github.com/firebase/FirebaseUI-Android/pull/426/commits/b3c4892327289d78cde969c8e1a6c3ab8339cd7d) since you last approved my PR.

SUPERCILEX:
@samtstern looks like we're good to go!

PR #431, opened 2016-12-01
Add a Filter to the FirebaseRecyclerAdapter

Ab-Appletic:
**Edit:** Oops, I did something wrong. I did this incorrectly (This is my first time) and I will try to delete this

### DIFF
--- a/app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
+++ b/app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
@@ -226,49 +226,54 @@ public class ChatActivity extends AppCompatActivity implements FirebaseAuth.Auth
     }
 
     public static class ChatHolder extends RecyclerView.ViewHolder {
-        View mView;
+        private final TextView mNameField;
+        private final TextView mTextField;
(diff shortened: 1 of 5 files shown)


SUPERCILEX:
Hey @Ab-Appletic, in case you need some help, here is a "beginner's guide" of sorts to get you going:

1. In your copy of FirebaseUI's directory, add FirebaseUI as a remote so you can directly check out the master branch: `git remote add firebase https://github.com/firebase/FirebaseUI-Android.git`
2. Go to Android Studio/Intellij and VCS->git->Fetch (or Ctrl+Shift+A and type fetch)
3. Now, bottom right, you can see your branches:
image
4. Here's the tricky part: if you already have a branch, you'll want to click on `firebase/master` and then "merge". You might have to resolve conflicts so I think the easiest thing to do would be to click on  `firebase/master` and then "Checkout as new local branch". Then, you can copy over your changes knowing you have the latest version of FirebaseUI. Once you've made a commit (Ctrl+K), you can push using (Ctrl+Shift+K). Here, you'll have to change the remote and push a new branch:
image
image

PR #432, opened 2016-12-02
Update gradlew scripts

SUPERCILEX:
I learned the other day that when you update the gradle wrapper version, you should also run `./gradlew wrapper` to update its scripts.


samtstern:
:+1:

Issue #434, opened 2016-12-02
Clarify RC_SIGN_IN in documentation


samtstern:
@CodingDoug reported this to me personally, so filing an issue to track:

When looking at the bits of code to start the auth flow, there is always a constant shown called RC_SIGN_IN to pass to startActivityToResult.  If you're familiar with how Android activities work, you might recognize this as convention for passing a "request code" to the next activity so you can receive it's result.  If you're not familiar with how this works, it's confusing, because RC_SIGN_IN doesn't have any explanation in the docs here.  One dev ended up looking for this and found com.firebase.ui.auth.ui.email.AcquireEmailHelper.RC_SIGN_IN, thinking that was the value you're supposed to pass in (but it's package private, so it doesn't work!).

To improve this situation, it's probably worth mentioning up front why we're telling people to use startActivityForResult(), and what RC_SIGN_IN means for them (and how they can define it and use it).  On top of that, the same code for handling the result doesn't itself check if the response code was RC_SIGN_IN before looking at the result code.  If you only have one response code RC_SIGN_IN, it doesn't matter practically speaking, but checking it there (and saying why it's checked there) may help drive home the significance of RC_SIGN_IN as discussed earlier.

samtstern:
@CodingDoug and just like that, @SUPERCILEX has fixed it.  Record time.

PR #435, opened 2016-12-02
Add startActivityForResult explanation

SUPERCILEX:
#434

### DIFF
--- a/auth/README.md
+++ b/auth/README.md
@@ -91,14 +91,15 @@ Twitter app as reported by the [Twitter application manager](https://apps.twitte
 </resources>
 ```
 
-In addition, if are using Smart Lock or require a user's email, you must enable the
+In addition, if you are using Smart Lock or require a user's email, you must enable the
 "Request email addresses from users" permission in the "Permissions" tab of your app.
 
 ## Using FirebaseUI for Authentication
 
 Before invoking the FirebaseUI authentication flow, your app should check
 whether a
-[user is already signed in](https://firebase.google.com/docs/auth/android/manage-users#get_the_currently_signed-in_user) from a previous session:
+[user is already signed in](https://firebase.google.com/docs/auth/android/manage-users#get_the_currently_signed-in_user)
+from a previous session:
 
 ```java
 FirebaseAuth auth = FirebaseAuth.getInstance();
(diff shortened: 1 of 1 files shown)


auth/README.md
```diff
@@ -91,14 +91,15 @@ Twitter app as reported by the [Twitter application manager](https://apps.twitte
 </resources>
 ```
 
-In addition, if are using Smart Lock or require a user's email, you must enable the
+In addition, if you are using Smart Lock or require a user's email, you must enable the
```
SUPERCILEX: Whoops

auth/README.md
```diff
@@ -147,6 +148,14 @@ startActivityForResult(
     RC_SIGN_IN);
 ```
 
+<a href="https://developer.android.com/reference/android/app/Activity.html#startActivityForResult(android.content.Intent, int)">`startActivityForResult`</a>
```
SUPERCILEX: This needs to be in HTML because of the parentheses in the link.

auth/README.md
```diff
@@ -147,6 +148,14 @@ startActivityForResult(
...
+is part of the Android framework and takes two parameters: an Intent to start an activity--in this case,
+the sign in builder--and a request code of type `int`. For example, `REQUEST_CODE_SIGN_IN` or `RC_SIGN_IN`
+for short is a constant defined in our activity: `private static final int RC_SIGN_IN = 100;`.
+The number you chose is arbitrary as long as it doesn't conflict with other
+`startActivityForResult` requests you might be making. In [response codes](#response-codes),
+you will see how we can use `RC_SIGN_IN` to differentiate between multiple `startActivityForResult` requests.
+
```
SUPERCILEX: This is really meaty, what do you think?

samtstern: I think that discusses the mechanics of startActivityForResult a little too much, how about something like this:

To kick off the FirebaseUI sign in flow, call `startActivityForResult(...)` on the sign in Intent you built.  The second parameter (`RC_SIGN_IN`) is a request code you define to identify the request when the result is returned to your app in `onActivityResult(...)`.  See the [response codes](#response-codes) section below for more details on receiving the results of the sign in flow.

auth/README.md
```diff
@@ -216,28 +225,30 @@ supported.
...
-        startActivity(new Intent(this, WelcomeBackActivity.class));
-        finish();
-        return;
-    }
-
-    // Sign in canceled
-    if (resultCode == RESULT_CANCELED) {
-        showSnackbar(R.string.sign_in_cancelled);
-        return;
+    if (requestCode == RC_SIGN_IN) {
```
samtstern: Add a small comment here something like `// RC_SIGN_IN is the request code you passed into startActivityForResult when starting the sign in flow`

auth/README.md
```diff
@@ -147,6 +148,14 @@ startActivityForResult(
     RC_SIGN_IN);
```
samtstern: We should add a line above this code snippet to be super clear:

```java

// Choose an arbitrary request code value
private static final int RC_SIGN_IN = 123;

// ...

  startActivityForResult(
```

samtstern:
Thanks for getting on this so fast!

auth/README.md
SUPERCILEX: Yup, I went a bit too far... 😄

SUPERCILEX:
Done!

PR #443, opened 2016-12-07
Remove unnecessary (?) SignInActivity

SUPERCILEX:
Hi @samtstern, this is the last PR in my "simplify email auth" series.

I'm confused about `SignInActivity`: why is it around? `WelcomeBackPasswordPrompt` basically does the same thing as `SignInActivity` but better. It doesn't let the user edit their email mid sign-in and gives them a friendly `Welcome back!` message. Was there a reason for keeping `SignInActivity` around?


samtstern:
I will take a deeper look at this tomorrow but I agree with your initial assessment, those two Activities seem to do basically the same job and SignInActivity can be removed.

After this goes in I am going to refactor the email flow a little further, I played with it a lot today and I found some cases where it seems to be jumping around.  But let's get through this change first, just a warning that if you do decide to do any more email-flow changes after this one give me a heads up since we'll almost certainly conflict.

auth/src/main/java/com/firebase/ui/auth/ui/accountlink/WelcomeBackPasswordPrompt.java
```diff
@@ -122,29 +128,32 @@ private void next(final String email, final String password) {
                 .addOnSuccessListener(new OnSuccessListener<AuthResult>() {
                     @Override
                     public void onSuccess(AuthResult authResult) {
-                        // Get the social AuthCredential from the IDPResponse object, link
-                        // it to the email/password account.
                         AuthCredential authCredential =
                                 AuthCredentialHelper.getAuthCredential(mIdpResponse);
-                        authResult.getUser().linkWithCredential(authCredential);
-                        firebaseAuth.signOut();
```
SUPERCILEX: @samtstern Do you know why we were doing this? It seems unnecessary and just using `linkWithCredential` works fine.

auth/src/main/res/values/strings.xml
```diff
@@ -32,9 +31,8 @@
         password.</string>
     <string name="error_email_does_not_exist">That email address doesn\'t match an existing account</string>
     <string name="welcome_back">Welcome back!</string>
-    <string name="next">Next</string>
     <string name="trouble_signing_in">Trouble signing in?</string>
-    <string name="sign_in">Sign in</string>
+    <string name="sign_in_default">Sign in</string>
```
SUPERCILEX: This was a nasty one! `WelcomeBackPasswordPrompt` kept using the word "Start" everywhere which was extremely confusing. Turns out our [sample was accidentally overriding](https://github.com/firebase/FirebaseUI-Android/blob/master/app/src/main/res/values/strings.xml#L15) this string resource. Since I think this could be a common occurrence, I've changed the name to `sign_in_default`.

SUPERCILEX:
@samtstern merged!

auth/src/main/java/com/firebase/ui/auth/ui/accountlink/WelcomeBackPasswordPrompt.java
samtstern: Looks like it was added in [this commit](https://github.com/firebase/FirebaseUI-Android/commit/bf9ef293a9202226b3ab834f8d8a4edb26d7cb45#diff-a21268aefa41be1f56b69905b4679cc3)

The only reason I can think to do this is that known issue about `FirebaseUser` changes not taking effect until signout/signin, but I believe that's only for explicit profile information change requests.

One thing I am seeing now is that the `linkWithCredential` call is done as fire-and-forget, which will almost certainly create some strange race conditions.

Seems like the original author should have been chaining the link-->signout-->sign in operations together.

SUPERCILEX: Humm, so should I revert and use a continuation or leave it at `linkWithCredential`?

auth/src/test/java/com/firebase/ui/auth/ui/email/WelcomeBackPasswordPromptTest.java
```diff
@@ -39,76 +40,47 @@
...
-                .buildActivity(SignInActivity.class)
+                .buildActivity(WelcomeBackPasswordPrompt.class)
                 .withIntent(startIntent)
                 .create()
                 .visible()
                 .get();
     }
 
     @Test
-    public void testSignInButton_validatesFields() {
```
samtstern: Shouldn't `WelcomeBackPasswordPrompt` validate fields now that we're removing `SignInActivity`?

auth/src/main/java/com/firebase/ui/auth/ui/accountlink/WelcomeBackPasswordPrompt.java
samtstern: I think we leave it with your new implementation.  I can't see a reason for the sign out thing.

samtstern:
This whole change LGTM, merging into 1.1.0-dev

PR #447, opened 2016-12-07
Refactor email flow into fragments

samtstern:
I was playing around with the new sign-in flow and the check email step felt very "jumpy" to me.  You'd enter an email and could just get throw into some other screen without much chance to fix it.  I think this is why the email screen was a separate step to begin with.

So I moved the email screen back to being its own step, but refactored things into Fragments so we keep all of the speed improvements implemented in the previous refactor.  Now RegisterEmailActivity is a controller for a few different states, and each Fragment/Activity is doing less which I like.


samtstern:
@SUPERCILEX I'd like your thoughts on this.  In a way it's a step backwards since what you did had less screens, but as I played around with things more I kept getting hung up on how the register screen would just jump to something else after the email field lost focus.

samtstern:
Also my tests pass locally ... can't figure out what Travis is angry about.

SUPERCILEX:
@samtstern For the tests, I don't think this is related to Travis because I tried to run them on my machine and all of them failed:
image

For the refactoring, what if we unified email sign in into a single `SignInActivity` god class of sorts that deals with registering new users and welcoming back existing ones? I'm thinking of using relative layouts and `View.GONE` to remove the name field and insert the welcome back message if the user already exists. I'll see if I can come up with a proof of concept today so you can see what I mean.

samtstern:
Would that be a fundamentally different UX flow than the fragment swap?  Or just avoiding the fragments?

SUPERCILEX:
I just tried your refactor, and no, it doesn't look like mine is different UX wise.

Having that extra step to register and preventing the user from changing their email once they are in the process of registering really perturbs me which is why I changed it in the first place, but I agree that sending the user to a new screen without any choice could be really frustrating.

What if we showed an Alertdialog instead of sending them directly to the welcome back activity? It could be something like "You already have an account with this email" CONTINUE, CHANGE EMAIL. What do you think of that? I feel like it's the best of both worlds.

samtstern:
I'll experiment with a dialog and see how that feels

SUPERCILEX:
Sounds good!

samtstern:
There's also the issue of presenting users who already have an account with the "create account" screen when they attempt to sign in.  My personal reaction would be to click "back" since I'd assume I was on the wrong screen for an existing user.

SUPERCILEX:
Oh, that didn't occur to me. In the end, I think we should proceed with your change: it keeps the performance improvements and reduces confusion. Authentication is hard! 😄

samtstern:
@SUPERCILEX yeah that didn't occur to me either when we were reviewing yours, sorry!

I'll keep playing with the UI and see if I can make the transitions nice.  And then I'll figure out what's going on with the tests

samtstern:
Added some nice animations:
https://drive.google.com/file/d/0B3CNQ9QwPTJ-YU41ckVGQXNpcmM/view

SUPERCILEX:
Oooh, that looks slick!

auth/src/main/AndroidManifest.xml
```diff
@@ -33,7 +33,6 @@
 
         <activity
             android:name=".ui.email.RegisterEmailActivity"
-            android:label="@string/title_register_email_activity"
```
SUPERCILEX: Shouldn't we replace this title with something?

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailActivity.java
```diff
@@ -14,339 +14,104 @@
...
-import com.google.firebase.auth.FirebaseUser;
-import com.google.firebase.auth.ProviderQueryResult;
-import com.google.firebase.auth.UserProfileChangeRequest;
-
-import java.util.List;
 
 /**
- * Activity displaying a form to create a new email/password account.
+ * Activity to control the entire email sign up flow. Plays host to {@link CheckEmailFragment}
+ * and {@link RegisterEmailFragment} and triggers {@link WelcomeBackPasswordPrompt}.
```
SUPERCILEX: `and triggers {@link WelcomeBackPasswordPrompt} and {@link WelcomeBackIdpPrompt}`

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailFragment.java
```diff
@@ -0,0 +1,253 @@
...
+
+        mEmailEditText.setOnFocusChangeListener(this);
+        mNameEditText.setOnFocusChangeListener(this);
+        mPasswordEditText.setOnFocusChangeListener(this);
+        v.findViewById(R.id.button_create).setOnClickListener(this);
+
+        // If email is passed in, fill in the field and move down to the name field.
+        String email = getArguments().getString(ExtraConstants.EXTRA_EMAIL);
+        if (!TextUtils.isEmpty(email)) {
+            mEmailEditText.setText(email);
```
SUPERCILEX: Don't we have to lock the email field again? We didn't have to do this before because it would check to see if the user had an existing account on every focus change. Now it looks like the user can change the email without us knowing if they already have an account. Also, shouldn't this condition always be true?

auth/src/main/res/anim/slide_in_right.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<set xmlns:android="http://schemas.android.com/apk/res/android">
+    <translate
+        android:duration="@android:integer/config_mediumAnimTime"
+        android:fromXDelta="100%p"
+        android:toXDelta="0%p" />
+    <!--<alpha-->
+        <!--android:duration="@android:integer/config_mediumAnimTime"-->
+        <!--android:fromAlpha="0.0"-->
+        <!--android:toAlpha="1.0" />-->
```
SUPERCILEX: Looks like these aren't being used.

auth/src/main/res/anim/slide_out_left.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<set xmlns:android="http://schemas.android.com/apk/res/android">
+    <translate
+        android:duration="@android:integer/config_mediumAnimTime"
+        android:fromXDelta="0%p"
+        android:toXDelta="-100%p" />
+    <!--<alpha-->
+        <!--android:duration="@android:integer/config_mediumAnimTime"-->
+        <!--android:fromAlpha="1.0"-->
+        <!--android:toAlpha="0.0" />-->
```
SUPERCILEX: Same comment as above.

auth/src/main/res/layout/register_email_layout.xml
```diff
@@ -14,23 +14,23 @@
             is showing we use a relative layout to position them absolutely
         -->
         <android.support.design.widget.TextInputLayout
-            xmlns:app="http://schemas.android.com/apk/res-auto"
             android:id="@+id/email_layout"
```
SUPERCILEX: See comment above, we could lock the email field from here: `android:enabled="false"`.

auth/src/main/java/com/firebase/ui/auth/ui/email/CheckEmailFragment.java
```diff
@@ -0,0 +1,220 @@
...
+            throw new IllegalStateException("Activity must implement CheckEmailListener");
+        }
+        this.mListener = (CheckEmailListener) getActivity();
+
+        // Check for email
+        String email = getArguments().getString(ExtraConstants.EXTRA_EMAIL);
+        if (!TextUtils.isEmpty(email)) {
+            mEmailEditText.setText(email);
+            validateAndProceed();
+        } if (mHelper.getFlowParams().smartLockEnabled) {
```
SUPERCILEX: This should probably go on a new line.

auth/src/main/AndroidManifest.xml
samtstern: It's supplied by the fragments, but I'll add a default

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailFragment.java
samtstern: I actually think we can leave the email field unlocked.  Here's basically the worst thing the user could do, and it's not so bad:

  * User enters `foo@bar.com` in the check email screen, gets sent to RegisterEmailFragment since it's a new account.
  * User then changes her mind and enters `baz@bar.com`, which is not a new account.
  * After entering name and password, the error message says "an account already exists" (thanks to your new error handling).

In that case the user will have to press back and try again, but since that's so unlikely I think it's ok to have a sub-optimal experience for those users, and now we have the editable email field (which you mentioned that you prefer)

samtstern:
I cannot for my life replicate the issues Travis CI is having with Robolectric ...

samtstern:
Related:
https://github.com/robolectric/robolectric/pull/2737

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailFragment.java
SUPERCILEX: That's great! The user not being able to edit their email really bothered me so now that it's no longer the case... This change is going to be really cool especially with those animations!  👍

SUPERCILEX:
Ok... @samtstern So it does have something to do with the manifest hack. I have a successful build here where I commented out the manifest stuff: https://travis-ci.org/SUPERCILEX/FirebaseUI-Android/builds/182411811.

samtstern:
Ok that is very good to know!  I will hunt based on that

samtstern:
Ok undid the manifest hack ... I actually don't need it anymore.  I did need it when I was trying to test the Fragments directly and I figured it couldn't hurt but apparently it **can** hurt haha.

SUPERCILEX:
Haha yeah. 😄 It looks like you set `sudo: false` which gives us the OOM exception again: https://github.com/firebase/FirebaseUI-Android/pull/447/commits/57ee6c22929b33c006a7aa54677444f3b4fcd860#diff-354f30a63fb0907d4ad57269548329e3R3

SUPERCILEX:
Fixed! 😄

samtstern:
@mdeck1 are you filling in for @amandle  yet?   If so could you PTAL at this?  It looks large but it's mostly just moving logic (rather than changing it).  Happy to discuss with you offline.

samtstern:
@amandle PTAL.  Like I said above I know it's large (in diff) so we can meet about this if you're not sure what's going on/

SUPERCILEX:
@samtstern STOPSHIP!!! I was working on a some features for your refactor and noticed that screen rotation is completely broken. Rotating the screen cancels the `RegisterEmailFragment` and starts a the autocomplete hint. Rotating the screen a second time crashes the app because of `enableAutoManage` already managing the id for the autocomplete hint. I'll see if I can get in a fix in my PR.

SUPERCILEX:
@samtstern I created https://github.com/samtstern/FirebaseUI-Android/pull/4 to address a slew of bugs:
image

samtstern:
Ok I am going to merge this into the `1.1.0-dev` branch so that we can continue working without so many conflicts.  This change is non-controversial in spirit (since it has no API changes) so any issues with this would be quality issues.  I am committed to running thorough QA before releasing `1.1.0` anyway.  

If this lingers any longer it will get even more out of hand and also prevent any additional work.

SUPERCILEX:
Thanks @samtstern!

Issue #441, opened 2016-12-06
Add a callback in FirebaseIndexArray when a Key doesn't exist in a Ref


DenisBronx:
In my case when the dataRef (The Firebase location to watch for data changes) gives a null snapshot value I wanted to trigger the parseSnapshot method in order to create the missing data instead of not displaying the row of the adapter.

Currently it only prints :

> Log.w(TAG, "Key not found at ref: " + snapshot.getRef());

samtstern:
@morganchen12 just curious what do we do in iOS?  Print a warning or callback?

DenisBronx:
Also it would be useful if we can get the DataSnapshots of the keyRef Query in case that we don't have only a set of keys in that location.
In this way we could merge two DataSnapshots in a single object.

morganchen12:
iOS has [a callback for handling errors per reference](https://github.com/firebase/FirebaseUI-iOS/blob/master/FirebaseDatabaseUI/FUIIndexArray.h#L56).

SUPERCILEX:
@samtstern I noticed this issue and the PR are in the "Needs decision" category, what's holding you back?

samtstern:
@SUPERCILEX thanks for reminding me about this one.  I really have no objection to the design of your PR, but I do have a higher-level objection.

One of the goals of all the `FirebaseArray` changes we made is to make it easier for developers to support complex use cases not easily supported by `FirebaseRecyclerAdapter` and other simple adapters.  In fact if we had done this a long time ago, we probably would have never added the indexed adapters to the core library at all.  Instead, you would have used `FirebaseArray` to create them in your own project or in a fork.

So following that same logic, I think it's ideal to stop expanding the surface of the adapters and encourage people to instead implement their own custom adapters for their use case.

In the case of this `JoinResolver` PR (which again, is a nice design) could we instead add a single overrideable method to the adapter like `onJoinFailed()` and then let developers take it from there?

SUPERCILEX:
@samtstern Ok, that makes sense. Let me look into updating my PR.

SUPERCILEX:
[...]

PR #448, opened 2016-12-09
Add a callback in FirebaseIndexArray when a Key doesn't exist in a ref

SUPERCILEX:
#441

@puf Please review.

- [x] #544


database/src/main/java/com/firebase/ui/database/OnIndexMismatchListener.java
```diff
@@ -0,0 +1,7 @@
+package com.firebase.ui.database;
+
+import com.google.firebase.database.DataSnapshot;
+
+interface OnIndexMismatchListener {
```
SUPERCILEX: @puf, so were you thinking of something like this:
```java
@Override
public void onChildAdded(DataSnapshot keySnapshot, String previousChildKey) {
    // ...
    Query ref = mJoinResolver.onJoin(keySnapshot); // Get the data ref
    mRefs.put(ref, ref.addValueEventListener(new DataRefListener()));
}

@Override
public void onChildRemoved(DataSnapshot keySnapshot) {
    Query removeQuery = mJoinResolver.onDisjoin(keySnapshot); // Get the remove ref
    removeQuery.removeEventListener(mRefs.remove(removeQuery));
    // ...
}
```

```java
private class DataRefListener implements ValueEventListener {
    @Override
    public void onDataChange(DataSnapshot snapshot) {
        if (snapshot.getValue() != null) {
            // ...
        } else {
            if (isMatch(index, key)) {
               // ...
            } else {
                mJoinResolver.onJoinFailed(index, snapshot); // notify failure
            }
        }
    }
}
```

puf: Yup. That first snippet looks like it. Not sure about the second snippet though. What would the `mJoinResolver` consist of there?

SUPERCILEX: @puf the second snippet is occurs when `dataRef` returns a null snapshot. The original purpose of this PR was to change that piece of code from a log (`Log.w(TAG, "Key not found at ref: " + snapshot.getRef());`) to a callback. Is it still okay to follow through with `mJoinResolver.onJoinFailed(index, snapshot);`?

database/src/main/java/com/firebase/ui/database/JoinResolver.java
```diff
@@ -0,0 +1,33 @@
+package com.firebase.ui.database;
+
+import com.google.firebase.database.DataSnapshot;
+import com.google.firebase.database.Query;
+
+public interface JoinResolver {
```
SUPERCILEX: Thoughts on the documentation?

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
```diff
@@ -26,69 +28,72 @@
...
 
-    public void cleanup() {
+    public void stopListening() {
         mQuery.removeEventListener(this);
+        mSnapshots.clear();
+        mIsListening = false;
     }
 
-    public int getCount() {
+    public int size() {
```
TheCraftKid: These look like pretty breaking changes, but I'm only just someone really curious about this project.
Also, was a consensus ever reached on using enums in the code? Shouldn't we be doing something like this:

    public interface ChangeListener {
        public static class EventType {
            public static final int ADDED = 1;
            public static final int CHANGED = 2;
            // ...
        }
    }

SUPERCILEX: @TheCraftKid Nope, these aren't breaking api changes. 🎉 We're really lucky @puf wanted to hold out on open sourcing `FirebaseArray` and `FirebaseIndexArray` or these would have been (You're thinking of `FirebaseRecyclerViewAdapter#getItemCount()`, but I didn't change that.)

As for enums, @puf decided against it and I'm fine with that because I found out proguard usually converts enums to `int`s anyway: http://stackoverflow.com/a/32236950/4548500. Cheers!

TheCraftKid: Well, then.
Wait, FirebaseArray is being open-sourced? How will that work? Will I finally be able to implement a proper FirebaseSpinnerAdapter?

SUPERCILEX: Woah woah woah! 😄 I was just saying that eventually @puf will have to open source it (technically he doesn't, but come on!). I think he was waiting for some additional refinement and this PR + another one I have ready will clean things up a bit, making it more likely that `FirebaseArray` will be open sourced. Then yes, you will finally be able to implement a proper `FirebaseSpinnerAdapter`!

puf:
Uhm.... I'm confused. FirebaseArray is just as open-source as the rest of
FirebaseUI-Android. If you're interested in how it works, you can admire
all 12 lines of it here:
https://github.com/firebase/FirebaseUI-Android/blob/master/database/src/main/java/com/firebase/ui/database/FirebaseArray.java
.

SUPERCILEX:
@puf I meant making `FirebaseArray` a public class so people can use it without having to copy it.

SUPERCILEX:
@samtstern I merged everything, but I'd rather #481 be merged first so that I can handle the conflicts. In addition, this PR doesn't actually let devs handle joining themselves yet because I'd like to submit a PR opening `FirebaseArray` to world by making it implement `List<DataSnapshot>` and use the start/stop logic from this PR.

SUPERCILEX:
@samtstern Should we add extra constructors in the index adapters? I feel like this is a niche feature and people who use it will know how to pass in a `FirebaseIndexArray` to the `FirebaseRecyclerAdapter`. What do you think?

SUPERCILEX:
@samtstern While we're on the merge rollercoaster, I think this one is the next PR to take a look at. 😄

samtstern:
@SUPERCILEX I agree, this one is definitely next.  I will take a look tomorrow.  A lot has changed in how we handle the IndexedArray (and really everything) since this was first sent out.

Btw we really appreciate that you always keep these PRs up to date and you're ready to discuss when we find the time.  It's above-and-beyond as a contributor and arguably we have not earned it since we go radio silent too often.

SUPERCILEX:
@samtstern No problem, I'm just happy whenever the ball moves forward a bit! 😄

Yeah, this was actually the PR that got me started on the whole "Let's redo `firebase-ui-database`" thing. It's not even a feature I use in my app which is funniest part. 😆

SUPERCILEX:
Going to close this PR since there is very little demand and it isn't on the FirebaseUI roadmap.

Issue #392, opened 2016-11-04
[Known Issue] Error: Failed to resolve: com.twitter.sdk.android:twitter:2.0.0 when syncing with Gradle


amandle:
If you are upgrading a project to version 1.0.0 you may encounter the error `Failed to resolve: com.twitter.sdk.android:twitter:2.0.0` when syncing your project with Gradle. Version 1.0.0 has added a new required configuration step. To resolve this issue you must add the Fabric repository to your repositories:

```groovy
repositories {
    // ...
    maven { url 'https://maven.fabric.io/public' }
}
```

kirtan403:
Adding this to readme will be better.

samtstern:
@kirtan403 we have it in the [auth README](https://github.com/firebase/FirebaseUI-Android/blob/master/auth/README.md).  You don't need this repository addition if you're only using database or storage.

kirtan403:
@samtstern  Ohh.. I didn't checked the auth one. I checked the main one. Thanks for pointing that out. 😄 Is it really needed even if I do not use twitter authentication?

samtstern:
@kirtan403 it's needed just to get gradle to be happy since the Twitter Android SDK is listed as a transitive dependency in the `pom.xml`.  If you use ProGuard on your application this should not have a large size impact.

We are looking for ways in later versions to remove this friction.

jay-thakur:
I have added fabric repository to build.gradle(App) as - 

```
repositories {
    mavenLocal()
    flatDir {
        dirs 'libs'
    }
    maven { url 'https://maven.fabric.io/public' }
}
```

Getting Error - 

> Error:Could not download gson.jar (com.google.code.gson:gson:2.6.1): No cached version available for offline mode

I added gson dependency but didn't work.
@amandle  @samtstern **Please Suggest.**

@kirtan403 did your issue resolved by adding fabric repo only ?

SUPERCILEX:
@jay-thakur see this stack overflow post on how to turn off offline mode: http://stackoverflow.com/q/28712025/4548500

And of course make sure you are online.

jay-thakur:
@SUPERCILEX thanks. Worked.

itpablo:
Even adding the reference the problem remains guys (same as reported in the issue title). I can compile only in the 0.6.2 version.

My gradle:

`
(...)
    repositories {
        jcenter()
        maven { url 'https://maven.fabric.io/public' }
    }
(...)
`

[...]

PR #449, opened 2016-12-09
add fabric maven url under allprojects

jdsingh:
Previously just writing repositories makes it confusing whether to add the url under `buildscript` or `allprojects`.

I was facing issue #392 because of this.


samtstern:
This seems like a good idea, thanks!

PR #453, opened 2016-12-10
Fix NPE in FacebookProvider

SUPERCILEX:
My bad, I forgot to test how other providers reacted the Facebook change. Not well apparently since users would get a NPE every time they tried to log in with a provider besides Facebook. Thankfully, this wasn't released. 😌


samtstern:
:+1:

PR #459, opened 2016-12-16
Use vectors instead of pngs!

SUPERCILEX:
It looks much better now IMO. 🎉

## Old:
screenshot_20161215-174354
## New:
screenshot_20161215-174322


auth/src/main/res/drawable/ic_facebook_white_22dp.xml
```diff
@@ -0,0 +1,13 @@
+<vector
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="22dp"
+    android:height="22dp"
```
SUPERCILEX: For some weird reason, the Facebook icon looked too big at `24dp`.

samtstern:
Looks nice!  Just curious, where did you get the vector assets?

SUPERCILEX:
@samtstern Oh crud, I forgot about copyright stuff!

I got the email asset from Android Studio's material icon asset generator and I got each provider's icon from their respective wikipedia pages:

- Google: https://en.wikipedia.org/wiki/File:Google_%22G%22_Logo.svg
- Facebook: https://commons.wikimedia.org/wiki/File:F_icon.svg
- Twitter: https://en.wikipedia.org/wiki/File:Twitter_bird_logo_2012.svg

I then imported all those into Android Studio.

👮 🚓 I'm hoping this isn't copyright infringement! 😄

puf:
We should do something better than "hoping" for that. The links to the
original source are a good start for doing so. And of course the same
concern on trademark status applies to the bitmaps these are replacing.

Good work btw!

SUPERCILEX:
@puf I don't know anything about legal stuff, do you know how we can check the copyright? I kind of assumed that anything from wikipedia could be used, but maybe I'm wrong.

puf:
The page links you provided have a license for each of the files. Also be
sure to check your CLA. That's unfortunately as close to legal advice as I
can get.

SUPERCILEX:
@puf Thanks! I didn't noticed the license info at the bottom. Since I don't want to be sued, I found the logo guidelines and I believe I am allowed to provide those logos.

The CLA I signed says the following:

> You represent that each of Your Contributions is Your original creation (see section 7 for submissions on behalf of others). **You represent that Your Contribution submissions include complete details of any third-party license or other restriction** (including, but not limited to, related patents and trademarks) of which you are personally aware and which are associated with any part of Your Contributions.

To "include complete details" here are the guidelines I found for each provider:
#### Facebook:

> _Do I need to request permission for each use of the Facebook brand assets?_
Please follow the general rules for using the Facebook brand, along with the rules for the specific asset you want to use. **Permission is required if you want to use Facebook brand assets on TV or Film.**

Thus, permission is not required for use online. There are extra guidelines for the "f" logo, but we're all in the green: https://facebookbrand.com/wp-content/uploads/2016/08/facebook_brandassetsguide.pdf.

#### Twitter:

> By using the Twitter trademarks in these Brand Guidelines, you agree to follow these Twitter Trademark Guidelines (the “Guidelines”) as well as our Terms of Service and all other Twitter rules and policies.

Since we are following the guidelines and the ToS, we are allowed to use their logo. More info here:
https://cdn.cms-twdigitalassets.com/content/dam/brand-twitter/asset-download-zip-files/Twitter_Brand_Guidelines.pdf

#### Google:
We're not violating the guidelines (https://developers.google.com/identity/branding-guidelines), so we can use the "g".

puf:
Thanks for looking into this @SUPERCILEX.

samtstern:
Thanks for checking the copyright stuff!  Can you confirm this works on API < 21?  I normally use `app:srcCompat` when using vector drawables but since these are placed using `app:drawableLeft` I am worried it may crash on lower API versions.

SUPERCILEX:
@samtstern I couldn't get an api 16 emulator with the latest play services, but I don't think that matters anyway since all the vector/material stuff was introduced in api 21. So anyway, it's working fine on api 19:
image

samtstern:
:+1: API 19 is good enough for me!

SUPERCILEX:
Awesome!

Issue #468, opened 2016-12-21
IdpResponse.fromResultIntent(data) return null


joknu1:
* Android device: Samsung Galaxy S4
  * Android OS version:5.0.1
  * Google Play Services version: 10.0.1
  * Firebase/Play Services SDK version:10.0.1
  * FirebaseUI version:1.0.1
  
### Step 3: Describe the problem:
`IdpResponse.fromResultIntent(data)` return `null` in `onActivityResult`

I launch Firebase UI with 

                AuthUI.SignInIntentBuilder signInBuilder = AuthUI.getInstance().createSignInIntentBuilder();
                List<AuthUI.IdpConfig> idpConfigs = new ArrayList<>();
                AuthUI.IdpConfig googleProvider = new AuthUI.IdpConfig.Builder(AuthUI.GOOGLE_PROVIDER).build();
                idpConfigs.add(googleProvider);

                signInBuilder.setProviders(idpConfigs);
                signInBuilder.setIsSmartLockEnabled(false);
                startActivityForResult(signInBuilder.build(), REQUEST_CODE_FIREBASE_UI_AUTH);

then in

    @Override
    public void onActivityResult(int requestCode, int resultCode, Intent data) {
        if (requestCode == REQUEST_CODE_FIREBASE_UI_AUTH) {
            if (resultCode == RESULT_OK) {
                IdpResponse idpResponse = IdpResponse.fromResultIntent(data);

`idpResponse` is null.

Debugger get me to `SaveSmartLock#idpResponse` where `smartLockEnabled` is false. Then  `finish` is called without `IdpResponse` parameter being set to the `SaveSmartLock` object. The `IdpResponse` parameter holds the correct values at this point according to debugger, but is not returned.

Traversing with debugger in `SaveSmartLock` gives me:

     private void finish() {
         **I end up here, with mResponse being null**
        Intent resultIntent = new Intent().putExtra(ExtraConstants.EXTRA_IDP_RESPONSE, mResponse);
        finish(RESULT_OK, resultIntent);
    }

    /**
     * If SmartLock is enabled and Google Play Services is available, save the credentials.
     * Otherwise, finish the calling Activity with RESULT_OK.
     *
     * Note: saveCredentialsOrFinish cannot be called immediately after getInstance because
     * onCreate has not yet been called.
     *
     * @param firebaseUser Firebase user to save in Credential.
     * @param password     (optional) password for email credential.
[...]

PR #469, opened 2016-12-22
Cleanup + make sure IdpResponse is never null if resultCode == ResultCodes.OK

SUPERCILEX:
#468

Note: this PR includes #447 to make my life easier. 😄

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/ErrorCodes.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ErrorCodes.java
@@ -4,10 +4,6 @@ package com.firebase.ui.auth;
  * Error codes retrieved from {@link IdpResponse#getErrorCode()}.
  */
 public final class ErrorCodes {
-    private ErrorCodes() {
-        // We don't want people to initialize this class
-    }
-
     /**
      * Sign in failed due to lack of network connection
      **/
@@ -17,4 +13,8 @@ public final class ErrorCodes {
      * An unknown error has occurred
      **/
     public static final int UNKNOWN_ERROR = 20;
+
+    private ErrorCodes() {
+        // We don't want people to initialize this class
+    }
 }
--- a/auth/src/main/java/com/firebase/ui/auth/ErrorCodes.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ErrorCodes.java
@@ -15,6 +15,6 @@ public final class ErrorCodes {
     public static final int UNKNOWN_ERROR = 20;
 
     private ErrorCodes() {
-        // We don't want people to initialize this class
+        // no instance
     }
 }
--- a/auth/src/main/java/com/firebase/ui/auth/ResultCodes.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ResultCodes.java
@@ -18,6 +18,6 @@ public final class ResultCodes {
     public static final int CANCELED = Activity.RESULT_CANCELED;
 
     private ResultCodes() {
-        // We don't want people to initialize this class
+        // no instance
     }
 }
--- a/auth/src/main/java/com/firebase/ui/auth/ResultCodes.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ResultCodes.java
@@ -7,10 +7,6 @@ import android.app.Activity;
  * {@code startActivityForResult}.
  */
 public final class ResultCodes {
-    private ResultCodes() {
-        // We don't want people to initialize this class
-    }
-
     /**
      * Sign in succeeded
      **/
@@ -20,4 +16,8 @@ public final class ResultCodes {
      * Sign in canceled by user
      **/
     public static final int CANCELED = Activity.RESULT_CANCELED;
+
+    private ResultCodes() {
+        // We don't want people to initialize this class
+    }
 }
--- a/auth/src/main/java/com/firebase/ui/auth/ui/accountlink/WelcomeBackIdpPrompt.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/accountlink/WelcomeBackIdpPrompt.java
@@ -40,6 +40,7 @@ import com.firebase.ui.auth.ui.BaseHelper;
 import com.firebase.ui.auth.ui.ExtraConstants;
 import com.firebase.ui.auth.ui.FlowParameters;
 import com.firebase.ui.auth.ui.TaskFailureLogger;
+import com.firebase.ui.auth.ui.User;
 import com.google.android.gms.tasks.OnCompleteListener;
 import com.google.android.gms.tasks.OnSuccessListener;
 import com.google.android.gms.tasks.Task;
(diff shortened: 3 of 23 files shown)


samtstern:
@SUPERCILEX I went ahead and merged #447 (commented there on reasoning) so you should be able to make the diff for this sane now.

branflake2267:
nice job!

auth/src/main/java/com/firebase/ui/auth/ui/accountlink/WelcomeBackIdpPrompt.java
```diff
@@ -127,83 +123,71 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
-                        FirebaseUser firebaseUser = task.getResult().getUser();
-                        firebaseUser.linkWithCredential(mPrevCredential);
-                        firebaseAuth.signOut();
-                        firebaseAuth
-                                .signInWithCredential(mPrevCredential)
-                                .addOnFailureListener(new TaskFailureLogger(
-                                        TAG, "Error signing in with previous credential"))
-                                .addOnCompleteListener(new FinishListener(newIdpResponse));
-                    } else {
-                        finish(ResultCodes.OK, IdpResponse.getIntent(newIdpResponse));
```
SUPERCILEX: BAD! (I think)
Here's the if: `if (task.isSuccessful() && mPrevCredential != null)`
What if the task fails!? We'll just finish with result ok? I'm pretty sure this was a mistake.

auth/src/main/java/com/firebase/ui/auth/ui/accountlink/WelcomeBackIdpPrompt.java
```diff
@@ -127,83 +123,71 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+                                        .addOnCompleteListener(new FinishListener(idpResponse));
+                            } else {
+                                finish(ResultCodes.OK, IdpResponse.getIntent(idpResponse));
+                            }
+                        }
+                    })
+                    .addOnFailureListener(new OnFailureListener() {
+                        @Override
+                        public void onFailure(@NonNull Exception e) {
+                            finishWithError();
```
SUPERCILEX: On failure, we fail.

auth/src/main/java/com/firebase/ui/auth/ui/accountlink/WelcomeBackPasswordPrompt.java
```diff
@@ -103,10 +103,10 @@ public void onClick(View view) {
             next(mEmail, mPasswordField.getText().toString());
         } else if (id == R.id.trouble_signing_in) {
             startActivity(RecoverPasswordActivity.createIntent(
-                    getApplicationContext(),
+                    this,
                     mActivityHelper.getFlowParams(),
                     mEmail));
-            finish(ResultCodes.OK, new Intent());
+            finish(ResultCodes.CANCELED, IdpResponse.getErrorCodeIntent(ErrorCodes.UNKNOWN_ERROR));
```
SUPERCILEX: I believe this was also a mistake, the user is not signed in so we shouldn't finish with OK.

auth/src/main/java/com/firebase/ui/auth/ErrorCodes.java
```diff
@@ -17,4 +13,8 @@ private ErrorCodes() {
      * An unknown error has occurred
      **/
     public static final int UNKNOWN_ERROR = 20;
+
+    private ErrorCodes() {
```
samtstern: nit: why do we need a constructor for this class at all?

auth/src/main/java/com/firebase/ui/auth/ResultCodes.java
```diff
@@ -20,4 +16,8 @@ private ResultCodes() {
      * Sign in canceled by user
      **/
     public static final int CANCELED = Activity.RESULT_CANCELED;
+
+    private ResultCodes() {
```
samtstern: Same comment as above.

auth/src/main/java/com/firebase/ui/auth/ui/accountlink/WelcomeBackIdpPrompt.java
```diff
@@ -89,21 +91,15 @@ protected void onCreate(Bundle savedInstanceState) {
...
-        if (mPrevIdpResponse != null) {
-            mPrevCredential = AuthCredentialHelper.getAuthCredential(mPrevIdpResponse);
-        }
-
         if (mIdpProvider == null) {
-            getIntent().putExtra(
-                    ExtraConstants.EXTRA_ERROR_MESSAGE,
-                    "Firebase login successful." +
-                            " Account linking failed due to provider not enabled by application");
+            Log.w(TAG, "Firebase login successful." +
```
samtstern: nit: this error message should include which provider was not enabled.

auth/src/main/java/com/firebase/ui/auth/ui/idp/CredentialSignInHandler.java
```diff
@@ -94,12 +95,6 @@ public void onFailure(@NonNull Exception e) {
     }
 
     private class StartWelcomeBackFlow implements OnSuccessListener<ProviderQueryResult> {
```
SUPERCILEX: This class isn't static so I'm not sure why we passed in the email. Seems like extra boilerplate to me.

auth/src/main/java/com/firebase/ui/auth/util/signincontainer/SaveSmartLock.java
```diff
@@ -169,17 +170,17 @@ private void finish() {
     public void saveCredentialsOrFinish(FirebaseUser firebaseUser,
                                         @Nullable String password,
                                         @Nullable IdpResponse response) {
+        mResponse = response;
+
         if (!mHelper.getFlowParams().smartLockEnabled
-                || !PlayServicesHelper.getInstance(getContext()).isPlayServicesAvailable()
-                || getActivity().isFinishing()) {
```
SUPERCILEX: This was pre `setRetainInstance(true)`, my bad.

auth/src/main/java/com/firebase/ui/auth/ui/accountlink/WelcomeBackIdpPrompt.java
```diff
@@ -127,83 +123,71 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+                        @Override
+                        public void onSuccess(AuthResult result) {
+                            if (mPrevCredential != null) {
+                                result.getUser()
+                                        .linkWithCredential(mPrevCredential)
+                                        .addOnFailureListener(new TaskFailureLogger(
+                                                TAG, "Error signing in with previous credential"))
+                                        .addOnCompleteListener(new FinishListener(idpResponse));
+                            } else {
+                                finish(ResultCodes.OK, IdpResponse.getIntent(idpResponse));
```
samtstern: This `finish()` was previously for failure of the `Task` but not it occurs when the task succeeds but `mPrevCredential == null`.  Is this a behavior change for developers?

auth/src/main/java/com/firebase/ui/auth/ErrorCodes.java
SUPERCILEX: We don't, this was just formalizing the class as a "`Constants`" class that shouldn't be instantiated.

auth/src/main/java/com/firebase/ui/auth/ui/accountlink/WelcomeBackIdpPrompt.java
SUPERCILEX: I believe this was the original intention: if we have a previous credential, link it, else, finish ok. Previously, if we either didn't have a previous credential or the task failed, we would finish ok.

Because we would finish ok if we didn't have a previous credential, this isn't a behavior change.

auth/src/main/java/com/firebase/ui/auth/ui/accountlink/WelcomeBackIdpPrompt.java
```diff
@@ -127,83 +123,71 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
     }
 
     @Override
-    public void onClick(View view) {
-        next(mPrevIdpResponse);
```
samtstern: Yeah it took me a little while to see what was going on here too.  Good find!  This stray listener could have bitten us down the road.

auth/src/main/java/com/firebase/ui/auth/ErrorCodes.java
samtstern: Got it.  Had not seen that technique before but I like it.

samtstern:
This LGTM!  Gonna squash and merge.

PR #470, opened 2016-12-23
Cleanup

SUPERCILEX:
This is an optional-ish PR (not that my other PRs were required! 😄) that has been waiting for #447.

I've noticed that most people (including @samtstern in his email refactor) put their static factory methods at the top of classes and Intellij keeps reformatting my code in that way so I thought I'd create a PR to fix that.

This PR looks big, but it's mostly just moving stuff around and letting Intellij do its thing.


auth/src/main/java/com/firebase/ui/auth/util/CredentialsApiHelper.java
```diff
@@ -34,80 +27,28 @@
...
 public class CredentialsApiHelper {
-
     @NonNull
     private final GoogleApiClientTaskHelper mClientHelper;
 
     private CredentialsApiHelper(GoogleApiClientTaskHelper gacHelper) {
         mClientHelper = gacHelper;
     }
 
-    public CredentialRequest createCredentialRequest(List<IdpConfig> providers) {
```
SUPERCILEX: This is the only non move stuff around change: a few of these methods weren't being used so I removed them.

auth/src/main/java/com/firebase/ui/auth/ui/email/CheckEmailFragment.java
```diff
@@ -180,44 +178,54 @@ public void checkAccountExists(@NonNull final String email) {
...
-                                mListener.onNewUser(new User.Builder(email)
-                                                            .setName(name)
-                                                            .setPhotoUri(photoUri)
-                                                            .build());
-                            } else if (EmailAuthProvider.PROVIDER_ID.equalsIgnoreCase(providers.get(0))) {
-                                mListener.onExistingEmailUser(new User.Builder(email).build());
-                            } else {
-                                mListener.onExistingIdpUser(new User.Builder(email)
-                                                                    .setProvider(providers.get(0))
+                    .addOnCompleteListener(getActivity(),
```
samtstern: nit: the formatting between this complete listener and the success listener that follows is inconsistent.  I prefer any style that minimizes indentation which means I vote for either the style before this change or the success listener style below.

auth/src/main/java/com/firebase/ui/auth/ui/email/CheckEmailFragment.java
```diff
@@ -180,44 +178,54 @@ public void checkAccountExists(@NonNull final String email) {
...
+                                }
+                            });
         }
     }
 
     private void showEmailAutoCompleteHint() {
         try {
-            startIntentSenderForResult(getEmailHintIntent().getIntentSender(), RC_HINT, null, 0, 0, 0, null);
+            startIntentSenderForResult(getEmailHintIntent().getIntentSender(),
+                                       RC_HINT,
```
samtstern: nit: we don't do this (argument alignment) throughout the project, any reason to start with this method?

samtstern:
Just a comment or two about style (which I only bothered to do since this PR is mostly about style in the first place)

auth/src/main/java/com/firebase/ui/auth/ui/email/CheckEmailFragment.java
SUPERCILEX: Done.

auth/src/main/java/com/firebase/ui/auth/ui/email/CheckEmailFragment.java
SUPERCILEX: Nope, we should have been using `mFragmentHelper.startIntentSenderForResult` anyway to get rid of all those zeros and nulls.

PR #475, opened 2016-12-27
Use native string formatter

SUPERCILEX:
`getString` does some extra locale stuff.


samtstern:
@SUPERCILEX thanks for this, this is a good idea.

PR #476, opened 2016-12-27
Rename BaseDialog and BaseFragment to be like other bases in FirebaseUI

SUPERCILEX:
I'm going off `AppCompatBase`


samtstern:
I see no problem with this rename, squashing and merging.

Issue #118, opened 2016-05-18
Handle empty state for recycleview


ismaelaarab10:
Is there a way for passing a different view if there is no data? Or does this require a second data check up outside de adapter

samtstern:
@ismaelaarab this is a good feature request and something that I have implemented in my own apps many times on top of FirebaseUI.  It would most likely take the form of a simplified AdapterDataObserver that works with Firebase.

I will put this on the Enhancements list and target it for a future release.  If you have a good solution, always feel free to send a PR.

zsweigart:
I am also running into needing this for my application.  I think a simple solution would be just to make an interface for onDataChanged and call it at the end of mSnapshot's data change listener

winstonma:
Agree on this one.

The LoaderManager.LoaderCallbacks.onLoadFinished listener would be very useful.

Sottti:
This is exactly what is holding me back right now. Basically there's a few scenarios that are common to a lot of my lists: Empty (size 0) and Error(impossible to get any data from network or cache.) I'd be really really happy to see this implemented. In my case I hide the recyclerView to show something else.

samtstern:
@Sottti @winstonma Here's a trick you could use in the meantime.  If you attach a `ValueEventListener` to the same `DatabaseReference` that you're passing into the RecyclerView you can use it as an indication of when the load is done.  You'll get the `onDataChange` callback in the `ValueEventListener` and this is guaranteed to happen **after** all of the child events have been propagated to the RecyclerView.  So if you get `onDataChange` and the RecyclerView still has 0 items you can show your "no items" message or do something similar while being confident that the initial load is complete.

winstonma:
Thanks @samtstern 

I saw many workarounds and it works. But I agree that it would be great if the features like `EmptyView` and `OnLoadFinished` is implemented so no one need to StackOverflow and make apps easier to build.

Thanks for the effort. In the meantime I will stick with your solution.

samtstern:
@winstonma agree that some sort of `OnLoadFinished` callback would be a good feature.  Hopefully we will get to it in a coming release.

samtstern:
A fix for this issue has been released in version `1.1.0`

winstonma:
Thanks appreciated

cutiko:
Greetings.

I was waiting for this, is great thanks! Can I suggest adding it in the docs, took me awhile to find it in the demo app.

Thanks.

dri94:
[...]

PR #477, opened 2016-12-27
Batch updates and empty lists

### LINKED ISSUE #458: Allow for batch updates
The FirebaseUI adapters surface an atomic event for each update to a child item. In some cases it may be more efficient to update the UI only after a batch up updates has been received.

I'd like it if the FirebaseArray and adapters would raise an additional event when a `value` event is received. Given that 'value' events are guaranteed to fire after the corresponding `child_` events, this would be the perfect way to implement such an event.

    public void onDataChanged()

Note that we don't pass in a DataSnapshot here. The idea is that this is a flow-event only, you're supposed to capture the data in the regular child events and just just `onDataChanged` to capture the completion moment.

puf:
Fixes #458 and #118. Ready for a first review, but don't merge yet.

### DIFF
--- a/app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
+++ b/app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
@@ -308,11 +308,12 @@ public class ChatActivity extends AppCompatActivity implements FirebaseAuth.Auth
             mTextField.setText(text);
         }
 
-        public void slideMessage(int x) { slideMessage(x, false); }
-        public void slideMessage(int x, boolean forceDisable) {
+        public boolean slideMessage(int x) { return slideMessage(x, false); }
+        public boolean slideMessage(int x, boolean forceDisable) {
             mMessageRow.setTranslationX(-x);
             mRemoveText.setTranslationX(mRemoveText.getWidth()-Math.min(x, mRemoveText.getWidth()));
             mRemoveText.setEnabled(x >= mRemoveText.getWidth() && !forceDisable);
+            return x >= mRemoveText.getWidth();
         }
         public void resetSlide() {
             // animate the message bubble and the remove label back to their original positions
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
```diff
@@ -74,23 +81,19 @@ protected void onCreate(Bundle savedInstanceState) {
...
-                            Log.e(TAG, "Failed to write message", databaseError.toException());
-                        }
-                    }
-                });
-
-                mMessageEdit.setText("");
+                sendCurrentMessage();
+            }
+        });
+        // Allow hitting the Send keyon the soft keyboard to send the message
```
morganchen12: nit: space between 'key' and 'on'

SUPERCILEX:
Hey @puf,
I really love this PR, but **`onReady` will be inconsistently called when offline**. If a new ref is created offline and subscribed to via a `ValueEventListener` **`Query`** (such as `.orderByValue().equalTo(1)`), `onDataChanged` will never be called while offline whereas `onChildAdded` will. Because of this, **I don't think this PR should be merged until that issue is addressed because it could leave users waiting indefinitely for an `onReady` call that will never come**. I've already contacted Firebase support, but since you work on the database, I was hoping you could speed up the fix.

A mcve with plain old `ValueEventListener`/`ChildEventListener`s: https://github.com/SUPERCILEX/issue-mcve/blob/46f41526ab047f24241ac548b8ba6dd04b98cc42/app/src/main/java/com/supercilex/myapplication/MainActivity.java#L32
@puf If you would like, I can create a mcve with your PR to show you that `onReady` will not always be called when offline.

puf:
Great find @SUPERCILEX . Agreed on holding off on merging until we figure this one out.

Unfortunately this is part of the behavior of the underlying SDK and not something we can change in FirebaseUI. It is not even something we can fix in the SDK, at least not without reworking a large part of the API surface. 

On the lower level we have no knowledge whether the ValueListener is listening for a single document (where we have a hard guarantee that we will never fire with incomplete results) or it is listening on a collection (where incomplete results are acceptable, as long as each individual child is complete).

Can you tell support to ping me on the case?

SUPERCILEX:
@puf Yes, I told support to ping you.

@puf That's a real bummer for the internal api, but I still have hope because I've been using a hack to get around this problem in my app. In my mvce, if you ask the `ValueEventListener` for the parent database instead of the query once while online, that entire ref is cached and queries will work offline. From what you explained above, caching the parent ref guarantees complete results which is why it works.

Here's what I mean:
```java
// 1. While online, add a listener to the root ref:
mRef.addValueEventListener(this);
// 2. Go back to querying and it will work offline and online:
mRef.orderByValue().equalTo(1).addValueEventListener(this);
```

@puf Couldn't you guys do some sort of cache hack where if a query is executed offline and returns no results, you guys query the parent ref and use a for loop with `getChildren` to filter the results and return the proper query? If that doesn't work for the native Firebase apis, couldn't we use a similar hack for FirebaseArray?

puf:
@SUPERCILEX: I'm against workarounds for behavior that is missing in the
Firebase SDK. Those types of workarounds tend to become a maintenance
problem as the SDK evolves.

SUPERCILEX:
@puf sounds reasonable so we just wait for you guys to fix it on the SDK side.

puf:
That is indeed my take. But this one likely won't be fixed soon, given that
it is inherent to the database's model: it's just a JSON tree. On the wire
we have no way to distinguish collections (where partial results are
acceptable and wanted) from "documents" (where we guarantee to never raise
partial results).

samtstern:

samtstern:
@puf going to close this.  Here are the changes in this PR, as I see it:

- Batch events in the adapters.
- Soft keyboard and swipe support in the sample app.

We've already implemented the former (with your help) and the latter is really cool but not critical to the library.  Since this has been open for many months and is currently not mergeable there's not much to do with it.  If you ever get some extra time (unlikely, I know) and want to get the app changes merge-able again then I'm happy to make these changes.

PR #481, opened 2016-12-29
Perf and fixups

CodingDoug:
Some optimizations - one in particular improves child change events to constant time from linear time (problematic for long lists).  Also modernized and fixed up the database tests.
@samtstern @puf

### DIFF
--- a/database/src/androidTest/java/com/firebase/ui/database/FirebaseArrayTest.java
+++ b/database/src/androidTest/java/com/firebase/ui/database/FirebaseArrayTest.java
@@ -14,9 +14,8 @@
 
 package com.firebase.ui.database;
 
+import android.support.test.InstrumentationRegistry;
 import android.support.test.runner.AndroidJUnit4;
-import android.test.InstrumentationTestCase;
-import android.test.suitebuilder.annotation.SmallTest;
 
 import com.google.firebase.FirebaseApp;
 import com.google.firebase.database.DatabaseReference;
@@ -32,28 +31,29 @@ import java.util.concurrent.Callable;
 import static com.firebase.ui.database.TestUtils.getAppInstance;
 import static com.firebase.ui.database.TestUtils.isValuesEqual;
 import static com.firebase.ui.database.TestUtils.runAndWaitUntil;
+import static org.junit.Assert.assertEquals;
 
 @RunWith(AndroidJUnit4.class)
-@SmallTest
-public class FirebaseArrayTest extends InstrumentationTestCase {
+public class FirebaseArrayTest {
+    private static final int INITIAL_SIZE = 3;
     private DatabaseReference mRef;
     private FirebaseArray mArray;
 
     @Before
     public void setUp() throws Exception {
-        FirebaseApp app = getAppInstance(getInstrumentation().getContext());
+        FirebaseApp app = getAppInstance(InstrumentationRegistry.getContext());
         mRef = FirebaseDatabase.getInstance(app).getReference().child("firebasearray");
         mArray = new FirebaseArray(mRef);
         mRef.removeValue();
         runAndWaitUntil(mArray, new Runnable() {
             public void run() {
-                for (int i = 1; i <= 3; i++) {
+                for (int i = 1; i <= INITIAL_SIZE; i++) {
                     mRef.push().setValue(i, i);
                 }
             }
         }, new Callable<Boolean>() {
             public Boolean call() throws Exception {
-                return mArray.getCount() == 3;
+                return mArray.getCount() == INITIAL_SIZE;
             }
         });
     }
@@ -70,13 +70,7 @@ public class FirebaseArrayTest extends InstrumentationTestCase {
     }
 
     @Test
-    public void testSize() throws Exception {
-        assertEquals(3, mArray.getCount());
-    }
-
-    @Test
     public void testPushIncreasesSize() throws Exception {
-        assertEquals(3, mArray.getCount());
         runAndWaitUntil(mArray, new Runnable() {
             public void run() {
                 mRef.push().setValue(4);
--- a/database/src/androidTest/java/com/firebase/ui/database/FirebaseArrayTest.java
+++ b/database/src/androidTest/java/com/firebase/ui/database/FirebaseArrayTest.java
@@ -112,7 +112,7 @@ public class FirebaseArrayTest {
     }
 
     @Test
-    public void testChangePriorities() throws Exception {
+    public void testChangePriorityBackToFront() throws Exception {
         runAndWaitUntil(mArray, new Runnable() {
             public void run() {
                 mArray.getItem(2).getRef().setPriority(0.5);
@@ -120,6 +120,19 @@ public class FirebaseArrayTest {
         }, new Callable<Boolean>() {
             public Boolean call() throws Exception {
                 return isValuesEqual(mArray, new int[]{3, 1, 2});
+            }
+        });
+    }
+
+    @Test
+    public void testChangePriorityFrontToBack() throws Exception {
+        runAndWaitUntil(mArray, new Runnable() {
+            public void run() {
+                mArray.getItem(0).getRef().setPriority(4);
+            }
+        }, new Callable<Boolean>() {
+            public Boolean call() throws Exception {
+                return isValuesEqual(mArray, new int[]{2, 3, 1});
             }
         });
     }
--- a/database/src/main/java/com/firebase/ui/database/FirebaseArray.java
+++ b/database/src/main/java/com/firebase/ui/database/FirebaseArray.java
@@ -21,7 +21,7 @@ import com.google.firebase.database.Query;
 import com.google.firebase.database.ValueEventListener;
 
 import java.util.ArrayList;
-import java.util.List;
+import java.util.HashMap;
 
 /**
  * This class implements an array-like collection on top of a Firebase location.
@@ -38,11 +38,12 @@ class FirebaseArray implements ChildEventListener, ValueEventListener {
 
     }
 
-    private Query mQuery;
+    private final Query mQuery;
+    private final ArrayList<DataSnapshot> mSnapshots = new ArrayList<>();
+    private final HashMap<String, Integer> mSnapshotMap = new HashMap<>();
     private OnChangedListener mListener;
-    private List<DataSnapshot> mSnapshots = new ArrayList<>();
 
-    public FirebaseArray(Query ref) {
+    FirebaseArray(Query ref) {
         mQuery = ref;
         mQuery.addChildEventListener(this);
         mQuery.addValueEventListener(this);
@@ -61,16 +62,13 @@ class FirebaseArray implements ChildEventListener, ValueEventListener {
         return mSnapshots.get(index);
     }
 
-    private int getIndexForKey(String key) {
-        int index = 0;
-        for (DataSnapshot snapshot : mSnapshots) {
-            if (snapshot.getKey().equals(key)) {
-                return index;
-            } else {
-                index++;
-            }
+    public int getIndexForKey(String key) {
+        final Integer index = mSnapshotMap.get(key);
+        if (index != null) {
+            return index;
+        } else {
+            throw new IllegalArgumentException("Key not found");
         }
-        throw new IllegalArgumentException("Key not found");
     }
 
     @Override
@@ -80,6 +78,7 @@ class FirebaseArray implements ChildEventListener, ValueEventListener {
             index = getIndexForKey(previousChildKey) + 1;
         }
         mSnapshots.add(index, snapshot);
+        mSnapshotMap.put(snapshot.getKey(), index);
         notifyChangedListeners(OnChangedListener.EventType.ADDED, index);
     }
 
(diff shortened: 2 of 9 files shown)


database/src/main/java/com/firebase/ui/database/FirebaseArray.java
```diff
@@ -20,25 +20,26 @@
...
 import java.util.ArrayList;
-import java.util.List;
+import java.util.HashMap;
 
 /**
  * This class implements an array-like collection on top of a Firebase location.
  */
 class FirebaseArray implements ChildEventListener {
-    public interface OnChangedListener {
+    interface OnChangedListener {
```
samtstern: @CodingDoug why change the visibility on this `interface` as well as the constructor below? I believe @puf just got done approving this particular visibility.

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
```diff
@@ -97,6 +97,7 @@ public void onChildMoved(DataSnapshot snapshot, String previousChildKey) {
         mSnapshots.remove(oldIndex);
         int newIndex = previousChildKey == null ? 0 : (getIndexForKey(previousChildKey) + 1);
         mSnapshots.add(newIndex, snapshot);
+        mSnapshotMap.put(snapshot.getKey(), newIndex);
```
samtstern: Wouldn't `onChildMoved` affect the indexes of every key?  I have a feeling that's why @puf didn't use a `Map` in the first place but I could be wrong.

database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
```diff
@@ -44,9 +41,8 @@ public FirebaseIndexArray(Query keyRef, Query dataRef) {
     @Override
     public void cleanup() {
         super.cleanup();
-        Set<Query> refs = new HashSet<>(mRefs.keySet());
-        for (Query ref : refs) {
-            ref.removeEventListener(mRefs.remove(ref));
+        for (Map.Entry<Query, ValueEventListener> entry : mRefs.entrySet()) {
+            entry.getKey().removeEventListener(entry.getValue());
```
samtstern: This is not equivalent to the previous behavior because there is no call to `mRefs.remove()`

samtstern:
Added some comments but want @puf to be the reviewer here as it's his code.

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
puf: Yup. This won't work.

puf:
I highly recommend doing a performance test before proposing this change. The performance of looking up keys is indeed quadratic, but when we measured it a few months ago that only started becoming significant once we added > 1000 items, which is far beyond what I'd recommend doing.

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
CodingDoug: Studio gives me a lint warning saying it's not necessary, which means it's not actually being used outside the package, just like the FirebaseArray class itself.

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
CodingDoug: All the tests passed on my end.  Do we need more tests?

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
samtstern: Yeah that's true but Studio is not really considering that this is a library for public consumption, so people may want to subclass etc.

common/constants.gradle
```diff
@@ -5,6 +5,6 @@ project.ext.submodules = ['database', 'auth', 'storage']
 project.ext.group = 'com.firebaseui'
 project.ext.version = '1.0.1'
 project.ext.pomdesc = 'Firebase UI Android'
-project.ext.buildTools = '25.0.2'
```
SUPERCILEX: @CodingDoug the `version-1.1.0-dev` branch has the latest everything (build tools, support lib, gradle build plugin, etc...). Changing the base branch of this PR to `version-1.1.0-dev` would minimize merge conflicts when merging it back into `master`.

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
samtstern: I don't think we have any tests that actually cover item re-order events, we should add one if we are going to change this functionality.

database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
CodingDoug: Updated

database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
```diff
@@ -41,8 +42,11 @@ public FirebaseIndexArray(Query keyRef, Query dataRef) {
     @Override
     public void cleanup() {
         super.cleanup();
-        for (Map.Entry<Query, ValueEventListener> entry : mRefs.entrySet()) {
+        final Iterator<Map.Entry<Query, ValueEventListener>> it = mRefs.entrySet().iterator();
+        while (it.hasNext()) {
+            final Map.Entry<Query, ValueEventListener> entry = it.next();
             entry.getKey().removeEventListener(entry.getValue());
+            it.remove();
```
samtstern: Does calling `Iterator#remove()` on an `Iterator` created from `.entrySet()` actually delete from the underlying map?  Maybe we should add a test that after `cleanup()` we verify `mRefs` is empty.

CodingDoug:
@puf The performance test in a simple situation may prove to be "OK", but the fact of the matter is that other work may end up being executed alongside this in the same "batch" of work on the main thread, and if that entire batch exceeds 16ms, Android starts dropping frames.  This is a preventative measure to make sure this adapter doesn't work unnecessarily against that 16ms barrier.

Alternately, if we don't think FirebaseRecyclerAdapter is a good choice in some known situations, we should probably document those situations so people aren't mislead at the time a choice is being made.

database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
CodingDoug: Yes, that case is specifically mentioned in the javadoc for Map.entrySet().  https://developer.android.com/reference/java/util/Map.html#entrySet()

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
CodingDoug: How will anyone subclass OnChangedListener if it's inside a package private class?  They won't be able to see it.  Are we expecting that people put their classes inside this package so they can make use of it?  Also, this interface isn't exposed anywhere in in the public API surface, so there would be no point in extending or implementing it.

If there is a use case here, maybe that needs some more elaboration?

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
CodingDoug: Note that the ordering implied by mSnapshots is not changed.  All that's really changed here is what getIndexForKey() does to fulfill its job.  There can be only one item with a particular key here, and that relationship is simply be expressed in the map now, so it doesn't have to scan the list for the unique key.

There is a test that checks for order changes called testChangePriorities().  It alters the priority of an item so that the array order changes from 1,2,3 to 3,1,2.  This test passes.

CodingDoug:
@samtstern Found that when an item moves front to back (rather than back to front, as was only being tested previously), there is an inconsistency with the recordkeeping of the new map.  Fixed with a new test to verify that.

samtstern:
@puf all my concerns are satisfied now, PTAL to see if yours are as well.

puf:
Before adding code for performance improvements, let's have a look at what the (order of) performance of cases is in the current release and will be after this PR.

1. Append item (at end)
2. Insert item (anywhere)
3. Move item
4. Remove item
5. Read N items (sequential)
6. Read N random items

Up front: #1 is by far the most common write operation, so its performance should not decrease to benefit #2 and #3. I think #5 is the most common read patter, but that is more based on web knowledge. In fact, I wonder if scrolling up doesn't lead to a 5b where we get the items in reverse.

puf:
I did some experiments with the performance of various write operations. My complete code is here: https://gist.github.com/puf/b2f2940188e11971afc688674d7a1cee

The current `getIndexForKey()` is an O(n^2) operation. If we make 2x more items to the list, looking up the index for a key takes 4x more time.

1. append is currently an O(log n) operation. It solely depends on `ArrayList.add` performance for append cases.
2. insert is currently an O(n^2) operation, because `getIndexForKey()` is O(n^2).
3. move is currently an O(n^2) operation. Strictly speaking 2*O(n^2), since we do two key lookups.
4. remove is currently an O(n^2) operation, since it relies on `getIndexForKey()`.

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
```diff
@@ -74,6 +83,7 @@ public void onChildAdded(DataSnapshot snapshot, String previousChildKey) {
             index = getIndexForKey(previousChildKey) + 1;
```
puf: Shouldn't this case be rebuilding the "key-to-index" map too?

CodingDoug: Perhaps! Could you write a test that should succeed, but instead fails?

puf:
In the [updated version](https://github.com/CodingDoug/FirebaseUI-Android/blob/0e04c7607cda6314d0e471e03908348a1aeaa2af/database/src/main/java/com/firebase/ui/database/FirebaseArray.java) `getIndexForKey()` is now an O(1) operation. The new `scanArrayForKey()` is an O(n^2) operation. Rebuilding the "key to index" map (in `onChildMoved`) is O(n) as far as I can tell (I didn't measure it yet).

1. append is O(log n) operation, but slightly slower now with an `add` to the end of an `ArrayList` and a `Map.put()`.
2. insert is O(log n) operation, since it uses `getIndexForKey()`
3. move is O(n^2), since it depends on `scanArrayForKey()`
4. remove is O(1) now, since it uses `getIndexForKey()`. But it likely should be O(n), since it should rebuild the "key to index" map.

CodingDoug:
@puf I believe scanArrayForKey is O(n) since it simply iterates a list at most once, looking for the matching key.

puf:
When I tested `getIndexForKey()` ([link](https://gist.github.com/puf/b2f2940188e11971afc688674d7a1cee)) it turned out to be O(n^2). Hence I also tested `scanArrayForKey()` and (not surprisingly at this point anymore) it behaved as O(n^2) too. I updated the code in the gist if you feel like testing it.

CodingDoug:
@puf Iteration of an array is definitely not quadratic.  Your test harness is making measurements that are inherently quadratic in nature, based on the value of RUN_SIZE.  It's performing RUN_SIZE number of lookups against an array of size RUN_SIZE, iterating through RUN_SIZE/2 elements each time.  This can only cause quadratic gains as RUN_SIZE increases.  If you limit the number of lookups to something constant instead of RUN_SIZE, and you cause some appropriate variance in the values you're looking for, you'll see linear gains at each step up in RUN_SIZE.

For example, instead of this:

```
for (int i=0; i < count; i++) {
    scanArrayForKey(list, i / 2);
}
```

You could say this to test the worst case scenario with predictability:

```
for (int i=0; i < 10000; i++) {
    scanArrayForKey(list, (int) count-1);
}
```
Or, even more accurate to real life, you could use random numbers for the value to look up and assume that a large number of iterations will average things out.

puf:
Ah... I get what you're saying: RUN_SIZE should only change the list size,
not the number of lookups. Good catch, I was already wondering why both
were quadratic.

For performing the append/insert I *should* loop over count items, but not
for timing the lookups. That definitely make a difference.

With that change, the lookups both `getIndexForKey` and `scanArrayForKey`
times grow sublinear with the increased number of items in the list. In
fact: they're almost at constant performance now.

puf:
Ah... I get what you're saying: RUN_SIZE should only change the list size,
not the number of lookups. Good catch, I was already wondering why both
were quadratic.

For performing the append/insert I *should* loop over count items, but not
for timing the lookups. That definitely make a difference.

With that change, the lookups both `getIndexForKey` and `scanArrayForKey`
times grow sublinear with the increased number of items in the list. In
fact: they're almost at constant performance now.

database/src/androidTest/java/com/firebase/ui/database/FirebaseArrayTest.java
```diff
@@ -129,4 +123,17 @@ public Boolean call() throws Exception {
             }
         });
     }
+
+    @Test
+    public void testChangePriorityFrontToBack() throws Exception {
```
samtstern: @puf FYI

samtstern:

SUPERCILEX:
@samtstern I see you added this to the "Needs decision". If we don't want to use `HashMap`s, I would really like to see the test fixups go in since we're currently using stuff that's deprecated and way out of date.

samtstern:
#579 has been merged so this PR will likely want a rebase.

samtstern:
Closing this since we haven't discussed this in ~months and it's not in a mergeable state anyway.  I believe we've incorporated many of the testing changes.  @CodingDoug if you ever convince @puf about all of this we can get back into it,

Issue #483, opened 2016-12-31
Where is my .setTosUrl ?


stasbar:
Hello.
Where is the Terms of Services url available to users during login process ?
`.setTosUrl("...")`

SUPERCILEX:
Only in the email sign up flow.

SUPERCILEX:
![screenshot_20161231-112446](https://cloud.githubusercontent.com/assets/9490724/21578982/c92e841c-cf4b-11e6-9f7c-223658511c9b.png)

samtstern:
Thanks @SUPERCILEX for explaining.

PR #484, opened 2016-12-31
Fix create account title not showing up (and cleanup strings)

SUPERCILEX:
The original purpose of this PR was to fix [this](https://github.com/firebase/FirebaseUI-Android/issues/483#issuecomment-269878767) still saying "Sign in", but the `strings.xml` file drove me crazy so I did a ton of cleanup there.


auth/src/main/res/values/strings.xml
```diff
@@ -1,68 +1,71 @@
...
+    <string name="sign_in_with_email">Sign in with email</string>
+
+    <!-- Email common-->
+    <string name="next_default">Next</string>
     <string name="email_hint">Email</string>
     <string name="password_hint">Password</string>
-    <string name="name_hint">First &amp; last name</string>
-    <string name="password_recovery_body">Get instructions sent to this email that explain how to
-        reset your password.</string>
-    <string name="create_account_title">Create account</string>
```
SUPERCILEX: I changed this to "Sign up" to be more consistent with our progress dialog.

auth/src/main/res/values/strings.xml
```diff
@@ -1,68 +1,71 @@
...
     <string name="required_field">You can\'t leave this empty.</string>
-    <plurals name="password_length">
-        <item quantity="one">Strong passwords have at least %1$d character and a mix of letters
-        and numbers</item>
-        <item quantity="other">Strong passwords have at least %1$d characters and a mix of letters
-        and numbers</item>
-    </plurals>
     <string name="invalid_email_address">That email address isn\'t correct</string>
     <string name="missing_email_address">Enter your email address to continue</string>
-    <string name="error_weak_password">Password not strong enough. Use at least 6 characters and a mix of letters and numbers.</string>
```
SUPERCILEX: Got rid of this; using the plural instead.

auth/src/main/res/values/strings.xml
```diff
@@ -1,68 +1,71 @@
...
     <string name="app_name">ui_flow</string>
 
+    <!-- Common -->
     <string name="default_toolbar_title">@string/app_name</string>
-    <string name="title_welcome_back_password_prompt">@string/sign_in_default</string>
-    <string name="title_welcome_back_idp_prompt">@string/sign_in_default</string>
-    <string name="title_check_email">@string/sign_in_default</string>
-    <string name="title_register_email_activity">@string/create_account_title</string>
-    <string name="title_recover_password_activity">@string/recover_password_title</string>
-    <string name="title_confirm_recover_password_activity">@string/check_your_email</string>
```
samtstern: While these `title_` strings are not part of the public API, I know there are some devs that do override them.  We seem to have changed `title_confirm_recover_password_activity` to `title_confirm_recover_password` here, can we undo that just to keep the resource name as long as we can?  Or is there a reason for the change?

SUPERCILEX: I meant to change it to `title_confirm_recover_password_dialog` because it's used in the dialog:
```java
return new AlertDialog.Builder(getContext(), R.style.FirebaseUI_Dialog)
                .setTitle(R.string.title_confirm_recover_password)
```

@samtstern Would you like me to change it back to `title_confirm_recover_password_activity` or to `title_confirm_recover_password_dialog`?

samtstern: Ah if  was never used as an Activity title it's fine as-is.

SUPERCILEX: It was an activity in the dark ages before https://github.com/firebase/FirebaseUI-Android/pull/357#discussion_r89051606. I just never changed the string name...

PR #485, opened 2017-01-01
Make Snackbar dismissable in sample

SUPERCILEX:
The diff got all messed up, the only change in `auth_ui_layout.xml` is being wrapped with a `CoordinatorLayout` and adding the `root` id.


SUPERCILEX:
@samtstern Every Travis build is going to fail now because `alpha1` is too old for the android gradle plugin. Because of that, I think the best course of action is to upgrade to `beta1` while disabling `RestrictedApi` and then wait for `beta2` which @tnorbye says fixes the `RestrictedApi` stuff. Hopefully `beta2` will come out in the first week of January.

I have a PR coming that changes the disabled checks so I'll remember to remove the hack you're concerned about that (we don't need to disable `InvalidPackage` for all projects, only the auth module; my PR fixes that).

app/src/main/res/layout/auth_ui_layout.xml
```diff
@@ -1,235 +1,242 @@
 <?xml version="1.0" encoding="utf-8"?>
-<ScrollView
+<android.support.design.widget.CoordinatorLayout
```
samtstern: Why use a CoordinatorLayout here with only one child?

SUPERCILEX: @samtstern I read [this great post](https://medium.com/google-developers/intercepting-everything-with-coordinatorlayout-behaviors-8c6adc140c26) from Ian Lake to get the PR to work. Turns out the FAB, Snackbar, and other design lib components rely on a Coordinator Layout to tell them when stuff is happening. Since the Snackbar wasn't dismissable, I took a guess that adding a Coordinator Layout would fix it, and it did!

Ian Lake:
> This ability to tie Views together is how much of the cooler functionality of the Design Library works — take, for example, the interaction between the FloatingActionButton and the Snackbar. The FAB’s Behavior depends on instances of the Snackbar being added to the CoordinatorLayout, then using the onDependentViewChanged() callback to translate the FAB upward to avoid overlapping the Snackbar.

<br>

PS: you made me curious so I decided to find out how the Snackbar really works:
`SwipeDismissBehavior<V extends View> extends `**`CoordinatorLayout.Behavior<V>`**
and
`Snackbar extends BaseTransientBottomBar<Snackbar>`
In `BaseTransientBottomBar`:
```java
behavior.setListener(new SwipeDismissBehavior.OnDismissListener() {
    @Override
    public void onDismiss(View view) {
        view.setVisibility(View.GONE); // <<<<<<
        dispatchDismiss(BaseCallback.DISMISS_EVENT_SWIPE); // <<<<<<
    }

    @Override
    public void onDragStateChanged(int state) {
        switch (state) {
            case SwipeDismissBehavior.STATE_DRAGGING:
            case SwipeDismissBehavior.STATE_SETTLING:
                // If the view is being dragged or settling, cancel the timeout
                SnackbarManager.getInstance().cancelTimeout(mManagerCallback);
                break;
            case SwipeDismissBehavior.STATE_IDLE:
                // If the view has been released and is idle, restore the timeout
                SnackbarManager.getInstance().restoreTimeout(mManagerCallback);
                break;
        }
    }
});
```

Pretty cool!

samtstern: Thanks for all that!  I realized I was confusing `CoordinatorLayout` and `ConstraintLayout` in my head!  First day back fuzz I guess.

SUPERCILEX: Oh, haha! 😄

PR #487, opened 2017-01-03
This change allows easy detection of the list being empty *and* respo…

### LINKED ISSUE #458: Allow for batch updates

puf:
Fixes #458 and #118. Ready for a final review, mostly to see how everyone feels about the name.


puf:
@SUPERCILEX Can you have a look at this PR and weigh in on the name of the new method?

database/src/main/java/com/firebase/ui/database/FirebaseRecyclerAdapter.java
```diff
@@ -110,9 +110,25 @@ public void onChanged(EventType type, int index, int oldIndex) {
             public void onCancelled(DatabaseError databaseError) {
                 FirebaseRecyclerAdapter.this.onCancelled(databaseError);
             }
+
+            @Override
+            public void onReady() {
+                FirebaseRecyclerAdapter.this.onDataChanged();
```
SUPERCILEX: Wouldn't we want to unify these names and pick either `onReady` or `onDataChanged`? Other than that, LGTM!

SUPERCILEX:
@puf, I found a stopship bug when offline https://github.com/firebase/FirebaseUI-Android/pull/477#issuecomment-270283875.

database/src/main/java/com/firebase/ui/database/FirebaseRecyclerAdapter.java
puf: Yeah, I wasn't sure. In the context of `FirebaseArray` the name `onDataChanged()` is too close to the existing `onChanged()` name. But I like the name as a whole.

We could consider renaming `onChanged()` to `onChildChanged()` (keep its signature aside from the name-change). Since the interface isn't part of the public API (yet), we could do that in a minor version.

SUPERCILEX: I think we should go for the second option (with `onChildChanged` letting us use `onDataChanged`). It would make the api more consistent with the native Firebase apis and fits well in the context of my other PR: https://github.com/firebase/FirebaseUI-Android/pull/488/files#diff-3620a1e9acaa17bea769cfbc3079f39aR18.

samtstern: Agree, might as well change the signature now while it's still private and get the name you actually want.

samtstern:
@puf I am ready to merge this, can you change the target branch to `version-1.1.0-dev` and then we'll merge?

SUPERCILEX:
@samtstern, I think @puf wanted to [hold off on merging](https://github.com/firebase/FirebaseUI-Android/pull/477#issuecomment-270287790) because of [this](https://github.com/firebase/FirebaseUI-Android/pull/477#issuecomment-270283875). I could be wrong, but I just wanted to make sure.

(#487 and #477 appear to be dulpicates.)

samtstern:
I interpreted the end of that discussion to mean "it will be fixed here when it's fixed in the SDK" but will let @puf confirm

puf:
@samtstern @SUPERCILEX there is no ongoing work to fix this in the SDK, but I don't want to implement a workaround in FirebaseUI. If that is the only thing blocking this PR, I suggest we merge and keep https://github.com/firebase/FirebaseUI-Android/pull/477#issuecomment-270283875 open.

puf:
@SUPERCILEX #487 is a small subset of #477. The latter includes a more elaborate demo, which we're (it's @samtstern for once) not ready to merge yet.

SUPERCILEX:
@puf, SGTM.

SUPERCILEX:
Ah, OK. Thanks for the clarification @puf.

app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
```diff
@@ -70,7 +71,10 @@ protected void onCreate(Bundle savedInstanceState) {
         mSendButton = (Button) findViewById(R.id.sendButton);
         mMessageEdit = (EditText) findViewById(R.id.messageEdit);
 
-        mChatRef = FirebaseDatabase.getInstance().getReference().child("chats");
+        mEmptyListView = findViewById(R.id.emptyTextView);
+
+        mRef = FirebaseDatabase.getInstance().getReference();
+        mChatRef = mRef.child("chats");
```
SUPERCILEX: @puf looks like the merge didn't work (`mChatRef` got split into a nonexistent `mRef` for some reason).

puf: Yeah, I split into two refs and apparently forgot to initialize one when resolving the conflict. It was my first time using the web conflict resolution editor and I must've borked something. I'll fix it tomorrow back at my desktop.

SUPERCILEX: Oh cool! I haven't been able to try that out yet. Github just needs to add checks to make sure your code compiles 😄 (oh, and why not add Intellij while we're at it and just write all of your code for you! 😄)

app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
```diff
@@ -153,6 +158,12 @@ public void populateViewHolder(ChatHolder chatView, Chat chat, int position) {
                     chatView.setIsSender(false);
                 }
             }
+
+            @Override
+            protected void onDataChanged() {
+                // if there are no chat messages, show a view that invites the user to add a message
+                mEmptyListView.setVisibility(mRecyclerViewAdapter.getItemCount() == 0?View.VISIBLE:View.INVISIBLE);
```
SUPERCILEX: Space around `:` and `?`.

app/src/main/res/layout/activity_chat.xml
```diff
@@ -6,6 +6,13 @@
     android:layout_height="match_parent"
     tools:context=".database.ChatActivity">
 
+    <TextView
+        android:id="@+id/emptyTextView"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:text="No messages. Start chatting at the bottom!"
```
SUPERCILEX: @puf Hmmm, this is kind of an experiment so we'll have to see how it goes. I would recommend running `./gradlew check` to make sure lint and checkstyle are happy, but in this case it wants you to create and `@string` resource. You can just ignore it with `tools:ignore="HardcodedText"` if you want.

samtstern:
Going to squash and merge this, thank you all!

Issue #15, opened 2015-09-29
Support for filtering and predicates


mcdonamp:
See https://github.com/firebase/FirebaseUI-iOS/pull/12

mcdonamp:
@zachariah25 This is all yours :)

puf:
It might be a good approach to add this as a `Filterable` implementation on Android.
See https://groups.google.com/forum/#!topic/firebase-talk/O1kJ4VLUg10

frangulyan:
I tried a very simple (and maybe not the most efficient) client-side filtering solution [here](https://gist.github.com/frangulyan/b80876d017643b768bb45777816bad33) (not yet fully tested). Any caveats that I don't see yet?

cha211:
@frangulyan how does your solution suppose to be implemented?

Ab-Appletic:
Any forks for this?

puf:
Not that I know off. Feel free to fork and create something yourself though.

Ab-Appletic:
Maybe some other time

Ab-Appletic:
@puf We've been begging for this feature for over a year. How come you still haven't added this feature? No one gives a crap about the fact that Android has to be in sync with iOS. Don't give your users **nothing** just for a stupid reason. Give some users what they want, then little by little it will be completed. You guys can't sit and do nothing.

All I want it to be able to search a RecyclerView. Too hard to implement? You're Firebase engineers and you have that title for a reason. We've been waiting for this feature for a year and not even a seed of progress has been made. At least give a workaround! I've never seen such a lacking library. If I've known there were better alternatives than Parse, I'd never move to Firebase. But now I'm stuck. Terrible service. It's a simple feature. No wonder I couldn't agree to upgrade my plan, because you don't deserve the money. It's a damn simple feature, instead of giving more big features, finish off the mandatory small ones first.

SUPERCILEX:
@Ab-Appletic How complex is your filtering? If it isn't too crazy, you can use Firebase's built in [filtering](https://firebase.google.com/docs/database/android/lists-of-data#filtering_data) and [ordering](https://firebase.google.com/docs/database/android/lists-of-data#sort_data) support. For example, I'm using:
```java
BaseHelper.getDatabase()
    .child(Constants.FIREBASE_SCOUT_INDICES)
    .child(BaseHelper.getUid())
    .orderByValue() // Since I'm using the push() method and a number in equalTo(), everything will be ordered by time
[...]

PR #488, opened 2017-01-04
Let devs intercept onChanged events and control when to update their adapter

SUPERCILEX:
Would address issues like #482 and see https://github.com/firebase/FirebaseUI-Android/issues/15#issuecomment-265644228.


app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
```diff
@@ -93,11 +94,7 @@ public void onComplete(DatabaseError databaseError, DatabaseReference reference)
         });
 
         mMessages = (RecyclerView) findViewById(R.id.messagesList);
-
         mManager = new LinearLayoutManager(this);
-        mManager.setReverseLayout(false);
-
-        mMessages.setHasFixedSize(false);
```
SUPERCILEX: Unnecessary, these are defaults.

app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
```diff
@@ -169,17 +165,25 @@ public void onItemRangeInserted(int positionStart, int itemCount) {
...
+                })
+                .addOnCompleteListener(new OnCompleteListener<AuthResult>() {
                     @Override
                     public void onComplete(@NonNull Task<AuthResult> task) {
-                        Log.d(TAG, "signInAnonymously:onComplete:" + task.isSuccessful());
                         if (task.isSuccessful()) {
-                            Toast.makeText(ChatActivity.this, "Signed In",
-                                    Toast.LENGTH_SHORT).show();
-                            attachRecyclerViewAdapter();
+                            Toast.makeText(getApplicationContext(),
```
SUPERCILEX: Show toast and continue logging in even is screen was rotated.

SUPERCILEX:
@puf PTAL, these changes shouldn't be anything major and can be merged in this branch since I'm targeting the dev branch. I'll merge your other PRs into this branch and resolve conflicts since I think there will be style issues causing the Travis build to fail. (FirebaseUI now has static analysis checks!)

app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
samtstern: Why did we remove the call to `attachRecyclerViewAdapter()` here?  Also if we scope the onComplete listener to the Activity context then we can avoid dead activity issues.

samtstern:
@puf could you take a look?  Basically this does two things:

1) Move the `OnChangedListener` interface out of `FirebaseArray` and make it first class.
2) Make a `protected onChanged()` function in the adapters so that some developers can change the behavior.

app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
SUPERCILEX: @samtstern I've extracted the toast logic into a static class to prevent activity leaks. As for the `attachRecyclerViewAdapter`, I moved that up to an onSuccess listener that is tied to the activity.

SUPERCILEX:
@puf Since `OnChangedListener` is now public would you be ok with me renaming things?

1. `OnChangedListener` -> **`ChangeEventListener`** to be more consistent with Firebase apis
2. `onChanged` -> **`onChildChanged`** as discussed in https://github.com/firebase/FirebaseUI-Android/pull/487#discussion_r94525452

@puf @samtstern What do you guys think?

app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
puf: There are valid reasons for having this code here, which we could debate here. But these reasons have nothing to do with letting devs intercept change events. In future PRs please keep functional changes separate from code cleanup.

app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
```diff
@@ -108,10 +106,10 @@ public void onStart() {
         // Default Database rules do not allow unauthenticated reads, so we need to
         // sign in before attaching the RecyclerView adapter otherwise the Adapter will
         // not be able to read any data from the Database.
-        if (!isSignedIn()) {
-            signInAnonymously();
-        } else {
+        if (isSignedIn()) {
             attachRecyclerViewAdapter();
+        } else {
```
puf: Same as before: this change seems unrelated to the topic of this PR and just expresses a personal preference. While that preference might well be better than the existing code, it should not pollute this PR.

app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
```diff
@@ -286,4 +276,21 @@ public void setText(String text) {
             mTextField.setText(text);
         }
     }
+
+    private static class SignInResultNotifier implements OnCompleteListener<AuthResult> {
```
puf: Nice addition! I'd appreciate some documentation on this class though.

database/src/main/java/com/firebase/ui/database/OnChangedListener.java
```diff
@@ -0,0 +1,41 @@
+package com.firebase.ui.database;
+
+import com.google.firebase.database.ChildEventListener;
+import com.google.firebase.database.DataSnapshot;
+import com.google.firebase.database.DatabaseError;
+
+public interface OnChangedListener {
```
puf: I'm fine pulling this up. But wouldn't it be better as an interface in FirebaseArray (which I see moving towards becoming an abstract base-class or interface itself)? Having this inside FirebaseArray makes it more explicitly scoped as "these operations apply to this FirebaseArray (or subclass) object".

SUPERCILEX: @puf, I would like that too, but you've chosen to keep `FirebaseArray` package private so people will get an error when overriding `onChanged` saying they can't access `FirebaseArray.OnChangedListener.EventType`. To put `OnChangedListener` in `FirebaseArray` we would have to go public and I don't think `FirebaseArray` is quite ready for that.

app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
SUPERCILEX: I removed this change, but I am curious, why are we setting default values and adding seemingly unnecessary code?

app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
SUPERCILEX: @puf I write my PRs with the philosophy that small improvements as we go along are better than constantly submitting side PRs for this kind of change. However, I see where you are coming from and removed these changes from this PR.

puf:
@SUPERCILEX: Not sure why we had these layout properties in there, but I
know for a fact I referred developers specifically to this part of the
example multiple times. It's the same as explicitly initializing a variable
to its default: you make it clear that it's not mistake and can more easily
refer to it to discuss it.

SUPERCILEX:
I see, it makes sense now. @puf Thanks for letting me know!

SUPERCILEX:
@puf since you pushed 21d5d44397a670d617b7ff2e41996d733ff122f1, I went ahead with the [rename](https://github.com/firebase/FirebaseUI-Android/pull/488#issuecomment-270735029).

samtstern:

SUPERCILEX:
@puf, any chance you could take another look at this? (I just merged everything)

database/src/main/java/com/firebase/ui/database/OnChangedListener.java
puf: But then isn't this essentially doing the same: making a part of the internal workings of FirebaseArray public?

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
```diff
@@ -80,21 +69,21 @@ public void onChildAdded(DataSnapshot snapshot, String previousChildKey) {
             index = getIndexForKey(previousChildKey) + 1;
         }
         mSnapshots.add(index, snapshot);
-        notifyChangedListeners(OnChangedListener.EventType.ADDED, index);
+        notifyChangedListeners(ChangeEventListener.EventType.ADDED, index);
```
puf: Definitely a better name! +1

database/src/main/java/com/firebase/ui/database/FirebaseListAdapter.java
```diff
@@ -51,7 +51,7 @@
  *            contained in the children of the given Firebase location
  */
 public abstract class FirebaseListAdapter<T> extends BaseAdapter {
-    private static final String TAG = FirebaseListAdapter.class.getSimpleName();
+    private static final String TAG = "FirebaseListAdapter";
```
puf: Can you explain why this change is needed/useful?

puf:
@SUPERCILEX I left a few comments. Most are very minor. The only real concern is that we'll end up exposing internal details of the `FirebaseArray` when those aren't sufficiently hardened yet. But I'm very willing to take a vote there, as the recent changes feel like a move in the right direction.

Of course we'll need the same ability to listen for changes in FirebaseUI-iOS, before we can merge. @morganchen12 did you already have a look at this?

database/src/main/java/com/firebase/ui/database/FirebaseListAdapter.java
SUPERCILEX: @puf This isn't so much needed in `FirebaseListAdapter`, but this is a massive no-no in the [`FirebaseRecyclerAdapter`](https://github.com/firebase/FirebaseUI-Android/pull/488/files/17a136d524ad4840f49bea54ba740016b7245b19#diff-724a30131c3284a3a9d723d6bb38561dR76) where the name is 23 characters long. I made this tag in one of my first PRs and I really regret it: if someone were to create a tag using `getSimpleName()` in `FirebaseIndexRecyclerAdapter`, `Log.anything` will [throw an exception](http://stackoverflow.com/questions/28168622/the-logging-tag-can-be-at-most-23-characters) because the tag name exceeds 24 characters. You should _**NEVER**_ use dynamic tag names for that reason.

@puf, since this is higher priority than #488, I've create #525.

SUPERCILEX:
@puf SGTM! I think you'll like my upcoming PRs for `FirebaseArray`... 😄

morganchen12:
I think this feature is added with the [transforms PR](https://github.com/firebase/FirebaseUI-iOS/pull/186) on iOS.

SUPERCILEX:
@puf @morganchen12 @samtstern This PR doesn't provide explicit support for joins because @puf [doesn't want that](https://github.com/firebase/FirebaseUI-Android/issues/15#issuecomment-264465745), but it does allow devs to do their own filtering as seen in [my crappy example](https://github.com/firebase/FirebaseUI-Android/issues/15#issuecomment-265644228). So, @morganchen12 @puf do you guys want to merge this PR and https://github.com/firebase/FirebaseUI-iOS/pull/186 and call it good because @puf doesn't want to support on device filerting? Or should we do something else?

morganchen12:
iOS doesn't add on-device filtering either, so no feature parity issues there.

SUPERCILEX:
@samtstern @puf @morganchen12 So are we good to merge?

samtstern:
Still fine with me, will let @puf pull the trigger.  I have just pushed a new `version-1.2.0-dev` branch off of `master` where we should put this feature.

morganchen12:
Fine with me, pending @puf's decision as well

SUPERCILEX:
@puf, could you please take another look at this? It's blocking progress on my other PRs... Thanks!

samtstern:
Ok after some team discussion it's time to merge this.  Squashing and merging into the `1.2.0` branch.

SUPERCILEX:
@samtstern Really glad to have this merged!

I started writing a post demoing some use cases so just in case anyone wants to see it, here you go: https://github.com/firebase/FirebaseUI-Android/issues/482#issuecomment-274956357.

Issue #490, opened 2017-01-04
Add `IdpProvider#startLogin(Fragment)`


goncalossilva:
As it stands, `IdpProvider#startLogin` expects an activity and nothing else. This is problematic when invoking it from a fragment, as the client needs to manually proxy `Activity#onActivityResult` to its own fragments' `onActivityResult` so that they can call `IdpProvider#onActivityResult`.

By using `Fragment#startActivityForResult` inside each provider, Android will take care of the proxying transparently, hence no leaking to the associated `Activity` is necessary.

samtstern:
@goncalossilva are you calling `IdpProvider#startLogin` directly from your own code?

SUPERCILEX:
@goncalossilva I don't think the `provider` package is public api (that's up to debate in #491), but if you want to know the reason behind that decision, here it is: it's all Twitter's fault as usual (sorry, I don't have high regards for Twitter's api). Their [`TwitterAuthClient#authorize`](https://docs.fabric.io/javadocs/twitter-core/2.3.0/index.html) method only supports activities. Internally, we've been using an activity and forwarding event to our Fragments like so: https://github.com/firebase/FirebaseUI-Android/blob/version-1.1.0-dev/auth/src/main/java/com/firebase/ui/auth/KickoffActivity.java#L79.

goncalossilva:
@samtstern Yes. We found the UIs and their flows on this library to be a tad too restrictive, but most of the things behind the scenes are _very_ useful. At the moment, these are the classes referenced in our codebase:
```
AuthUI
IdpResponse
provider.IdpProvider
provider.GoogleProvider
provider.FacebookProvider
util.GoogleApiConstants
```

@SUPERCILEX Huh, Twitter 😩  It's the only provider we're not using, hence I missed this when checking if there were any limitations to adding a `startLogin(Fragment)` method.

samtstern:
@goncalossilva thanks for the clarification.  As @SUPERCILEX we don't consider those classes to be part of the public API so we don't plan to add `Fragment` support there since we don't use it internally.

While I totally understand your use case, I am going to close this issue because it's working as intended from our perspective.  I think you may want to consider forking the library and modifying it to suit your needs if you find the current API surface too restrictive.  Thanks for the input!

Issue #408, opened 2016-11-17
Attempt to read from field 'int com.firebase.ui.auth.ui.FlowParameters.themeId' on a null object reference 


alssory:
### Step 2: Describe your environment

  * Android device: Galaxy S4
  * Android OS version: 5.0.1
  * Google Play Services version: 7.1.15
  * FirebaseUI version: 1
  
#### Steps to reproduce:

  1. Sign in and go to another activity(Activity1)
  2. Press on signOut button in (Activity1)
  3. The app crash
  
#### Observed Results:

java.lang.RuntimeException: Unable to start activity ComponentInfo{*****.****.firebaserules/com.firebase.ui.auth.ui.email.SignInActivity}: java.lang.NullPointerException: Attempt to read from field 'int com.firebase.ui.auth.ui.FlowParameters.themeId' on a null object reference

  
#### Expected Results:

signOut from account and go to signInActivity
  
#### Relevant Code:

  ```
  // TODO(you): code here to reproduce the problem
//onClick
public void signOut(View view) {
            AuthUI.getInstance()
                    .signOut(this)
                    .addOnCompleteListener(new OnCompleteListener<Void>() {
                        public void onComplete(@NonNull Task<Void> task) {
                            // user is now signed out
                            startActivity(new Intent(MainActivity.this, SignInActivity.class));
                            finish();
                        }
                    });
    }
  ```

versions:
 'com.firebaseui:firebase-ui:1.0.0'
'com.google.gms:google-services:3.0.0'

SUPERCILEX:
You can't just start the AuthUI flow manually, see the proper way to start a sign in flow: https://github.com/firebase/FirebaseUI-Android/blob/master/auth/README.md#sign-in

samtstern:
@alssory please see the advice above.

PR #491, opened 2017-01-04
Use lint to define what is public and what isn't

SUPERCILEX:
@samtstern There have been multiple instances of people using non-public apis (#408, #490, and probably more) so I think we should define what is public and what isn't here. [Firebase Job Dispatcher](https://github.com/firebase/firebase-jobdispatcher-android/tree/master/jobdispatcher/src/main/java/com/firebase/jobdispatcher) took the approach of putting everything in one package and making it all package private. I really don't think that's a good idea since it makes developing the library difficult and confusing. Instead, we can use the `@RestrictTo` annotation from the support lib folks to get around this problem.

I've left the `provider` package public because it could theoretically be used out of FirebaseUI, but all of this is up to debate. I've also left some helpers public for the same reason. @samtstern what are your thoughts on this?


samtstern:
Never knew about this!  That's pretty cool.  Can you add a screenshot showing what will happen in Android Studio if you try to use a restricted class?

SUPERCILEX:
@samtstern Sadly, there isn't an inspection for this in Android Studio [yet](https://code.google.com/p/android/issues/detail?id=231481). However, lint will blow up on you if use a class marked with `@RestrictTo`: I found out about the annotation when we encountered that bug in the android gradle plugin alpha3 (https://github.com/samtstern/FirebaseUI-Android/pull/3#discussion_r92259692). So it will look like this if we run lint manually:

`@RestrictTo`:
image


`@VisisbleForTesting`:
image

samtstern:
Looks great, thanks!

goncalossilva:
> I've left the `provider` package public because it could theoretically be used out of FirebaseUI

This has been immensely helpful for us at Doist. Our UI flows are a bit different from what Firebase proposes, and it has been a great experience to be able to leverage all the work in the providers without having to use the UI bits as well. We hope this remains in the public API 😊

samtstern:
@goncalossilva thanks for the feedback.  This change only affects lint so you can continue to ignore it, but it at least makes the division clear.  Do you have a public app that uses these modified flows?  I'd love to check it out.

goncalossilva:
@samtstern It's currently in review, should be out in about 2/3 weeks, and will A/B test two different approaches. We'll let you know once it's out.

pedrodanielcsantos:
@samtstern our work with the integration of FirebaseUI (the authentication module) where we're using the providers directly, as @goncalossilva said, was released yesterday, so in case you want to take a look the work we've done, you can check [Todoist](https://play.google.com/store/apps/details?id=com.todoist) 👍

samtstern:
@pedrodanielcsantos thanks for the pointer, this is awesome!  I love Todoist, will download the update and take a look.

PR #495, opened 2017-01-05
Make KickoffActivity truly transparent

samtstern:
Video in action:
https://drive.google.com/file/d/0B3CNQ9QwPTJ-Vi0yMlRQUXFIXzQ/view?usp=drivesdk


SUPERCILEX:
Woah!!! Thanks for this @samtstern, it looks amazing! I can't believe we never thought of the `configureTheme` thing!

samtstern:
Yeah the only downside here is the ProgressDialog in KickoffActivity is not themed with the user's custom theme but I think that's a worthwhile sacrifice given how much better it looks.

We didn't think of this because we were working on #357 at the time and had a lot on our minds :+1:

PR #496, opened 2017-01-05
Remove unnecessary theme attributes in AndroidManifest.xml

SUPERCILEX:
@samtstern I think the reason we were originally confused is because we have theme attributes in the manifest. However, these aren't used so they shouldn't be there to make it clearer that we use `configureTheme`.

Note: this change includes #495 to minimize merge conflicts so that should be merged first.

### DIFF
--- a/auth/build.gradle
+++ b/auth/build.gradle
@@ -53,6 +53,7 @@ dependencies {
     testCompile 'junit:junit:4.12'
     //noinspection NewerVersionAvailable, GradleDynamicVersion
     testCompile 'org.mockito:mockito-core:2.5.+'
+    //noinspection NewerVersionAvailable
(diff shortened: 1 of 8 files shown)


samtstern:
I just merged #495, so I'll review this once Github gets the real diff figured out.

SUPERCILEX:
SGTM, I'll merge in a bit.

SUPERCILEX:
@samtstern merged!

SUPERCILEX:
The default theme is used here if people don't set one: https://github.com/firebase/FirebaseUI-Android/blob/version-1.1.0-dev/auth/src/main/java/com/firebase/ui/auth/AuthUI.java#L516.

PR #500, opened 2017-01-07
[HIGH PRIORITY] Fix inflation errors

SUPERCILEX:
Hey @samtstern I started noticing inflation crashes in my app earlier today. It turns out there was a reason for the theme stuff in the manifest file: I didn't know this, but setting a theme doesn't actually overwrite the previous theme, it just adds on values (and overwrite identical ones). The theme I provide in my app doesn't extend the default FirebaseUI theme so some color stuff was nonexistent causing the crash.

Example error:
`Caused by android.view.InflateException: Binary XML file line #9: Binary XML file line #9: Error inflating class android.support.design.widget.TextInputLayout`


Phantomphyre:
overlay.rtl.dev dual pnp while turning off outline and on ltr touches,screened update screens can auto emulateand wont break.

samtstern:
@SUPERCILEX the more you know, thanks!

PR #501, opened 2017-01-09
Add getItems method to adapters

SUPERCILEX:
@puf I've been needing to get all the items in my adapter recently and there is no way to do that without writing a for loop each time so I created a simple `getItems` method for each adapter. @puf PTAL and confirm that this is a reasonable change.

### DIFF
--- a/auth/src/main/AndroidManifest.xml
+++ b/auth/src/main/AndroidManifest.xml
@@ -29,32 +29,27 @@
         <activity
             android:name=".ui.email.RecoverPasswordActivity"
             android:label="@string/title_recover_password_activity"
-            android:exported="false"
-            android:theme="@style/FirebaseUI"/>
+            android:exported="false"/>
 
         <activity
             android:name=".ui.email.RegisterEmailActivity"
             android:label="@string/sign_in_default"
-            android:exported="false"
-            android:theme="@style/FirebaseUI"/>
+            android:exported="false"/>
 
         <activity
             android:name=".ui.accountlink.WelcomeBackIdpPrompt"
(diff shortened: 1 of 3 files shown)


samtstern:

puf:
Please open an issue before submitting a PR, so that we can discuss the
use-case before discussing the implementation.

What is the use-case for needing access to all items? In general that is
going to be problematic as you have a longer list, so you should try to
avoid it.

SUPERCILEX:
@puf I have a multi item selector in my app with the ability to select all. Usually, developers have access to the full dataset in list form that they pass into the adapter, but here it's the other way around and we have to depend on the adapter to get data. I need to get a snapshot of all items to pass around my different classes and while I could pass the adapter around and use `getItems`, that adds more dependencies and just feels wrong to have to depend on your adapter to get data.

I know this isn't the best performance wise if people use it in a for loop (`for (DataSnapshot snapshot : mAdapter.getItems())`)  since they are basically iterating through the list twice, but I have PR written that I'll submit sometime soon which turns `FirebaseArray` into a list (`implements List<DataSnapshot>`) which will let us just safely return `mFirebaseArray`. If you're really bothered by the performance implications, I can wait and move this change into my aforementioned "turn `FirebaseArray` into a list" PR.

SUPERCILEX:
@puf in light of focusing your efforts, I'm going to close this PR for now. I might reopen it at a later time with more arguments as to why I think it should be added to the public api, but for now I would rather keep your brain time on #488.

puf:
Thanks Alex!

PR #507, opened 2017-01-10
Add string resource to puf-ondatachange

SUPERCILEX:
@puf this fixes the build issues in #487. I'm also going to update the contribution guidelines with the check stuff and we'll see how it goes.


puf:
Thanks for fixing. I often get lazy with string resources. You should really not let me work on the demo app. ;-)

SUPERCILEX:
Haha it's fine, you must have been tired! :smile: (See #512)

PR #511, opened 2017-01-11
Update dependencies and fix lint stuff

SUPERCILEX:
@samtstern I didn't forget! I removed the `disable 'RestrictedApi'` hack and updated jfrog.


app/build.gradle
```diff
@@ -43,7 +43,6 @@ dependencies {
     // The following dependencies are not required to use the Firebase UI library.
     // They are used to make some aspects of the demo app implementation simpler for
     // demonstrative purposes, and you may find them useful in your own apps; YMMV.
-    compile 'com.github.bumptech.glide:glide:3.7.0'
```
SUPERCILEX: > The following dependencies are not required to use the Firebase UI library.

That's no longer true: Glide is included in `firebase-ui-storage`.

samtstern:
Merged, thanks!

PR #515, opened 2017-01-12
Version 1.1.0 qa

samtstern:
Did some QA, things are looking mostly great.  I had to revert a PR regarding NightMode since it was messing up all of our non-nightmode styles and I would rather be safe than sorry.


samtstern:
@puf or @SUPERCILEX PTAL

auth/src/main/res/layout/auth_method_picker_layout.xml
```diff
@@ -1,36 +1,33 @@
 <?xml version="1.0" encoding="utf-8"?>
-<LinearLayout
-    xmlns:android="http://schemas.android.com/apk/res/android"
+<ScrollView xmlns:android="http://schemas.android.com/apk/res/android"
```
SUPERCILEX: This is awesome, thanks!

PR #518, opened 2017-01-13
Simplify .gitignore

SUPERCILEX:
Hey @samtstern,
This is a simple PR just reducing the amount of files in each module. The `.gitignore`s only excluded `/build`, but that can easily be done in the root `.gitignore` by removing the `/`.


samtstern:
LGTM!

Issue #521, opened 2017-01-13
Odd placement of Login buttons in AuthUI


the-dagger:
### Step 2: Describe your environment

  * Android device: OnePlus3
  * Android OS version: 7.0
  * Google Play Services version: 10.0.84
  * Firebase/Play Services SDK version: 10.0.1
  * FirebaseUI version: 1.1.0
  
### Step 3: Describe the problem:
Odd placement of login buttons in Authentication Screen
#### Steps to reproduce:
Try to login.
#### Observed Results:

 
![screenshot_20170113-200527](https://cloud.githubusercontent.com/assets/6565270/21933659/3aa09434-d9cc-11e6-805e-395156293db3.png)

  
#### Expected Results:

![screenshot_20170113-200842](https://cloud.githubusercontent.com/assets/6565270/21933660/3cef1a80-d9cc-11e6-9aa6-cce556626fcc.png)

samtstern:
This was caused by me in #515, and it can be worked around by setting a logo with `setLogo()`.  I realize now we neglected to test the layout without setLogo() which causes the buttons to float to the top in the absence of one.

I will probably release a `1.1.1` version to fix this issue.

samtstern:
This has been fixed by `1.1.1`

Issue #375, opened 2016-10-22
Landscape Layout in Authentication Method Picker


HusseinElFeky:
I have a small issue, but it can be very easily fixed. In the Authentication Method Picker activity, I put a logo and 3 providers (3 buttons). In portrait mode, everything looks neat and fine, but in landscape mode, the image is very small because there is no place for it above the 3 buttons. So I would suggest, in landscape mode, that if the developer decides to put a logo, the logo should be shown at the left, and the buttons should be shown at the right. This will be a much better approach.

SUPERCILEX:
We should probably use constraint layout to reduce nesting. Also, from what I understand, `View.GONE` rebuilds all the constraints so it would make supporting landscape layouts easier.

Akshit94:
I made a workaround for this issue. I don't know whether it's correct or not but it's working just fine. What I did was made a copy of the  "**auth_method_picker_layout.xml**" file in the library in my own "**layout-land**" folder and tweaked the layout a little bit to suit my needs. This layout overrides the original layout in the library. @HusseinElFeky you can try this approach if you want to.

HusseinElFeky:
@Akshit94 @SUPERCILEX Yes, that's definitely correct. That's the way it should be implemented. It will be much better if it is added in the next version though.

samtstern:
This issue is a duplicate of #522 and a fix for the auth method picker has been implemented in #562, look out for version `1.2.0`!

PR #523, opened 2017-01-13
Undo layout changes from #515

samtstern:


samtstern:
Fixes #521.

@puf  or @SUPERCILEX PTAL.  This will have to become `v1.1.1` since we broke people who don't use a logo (which means we broke the default).

samtstern:
First build errored out due a Travis issue, but it passed on restart:
https://travis-ci.org/samtstern/FirebaseUI-Android

SUPERCILEX:
@samtstern LGTM. Would you be ok with me trying to rewrite the auth picker layout using `ConstraintLayout`? It would let us use constraint biases meaning we don't need hard dp values making the layout more flexible. If not, I would still like to submit a PR making horizontal layouts usable by moving the buttons to the right of the logo.

PS: if you're cool with `ConstraintLayout`, I would like to slowly convert our codebase to use it thus minimizing nesting and improving performance. The original plan was to submit one massive PR converting everything to `ConstraintLayout`, but I don't think that's going to happen. Instead, we can go through it one layout at a time.

samtstern:
@SUPERCILEX I do think in general ConstraintLayout is a great solution to this problem.  However in the interest of un-breaking people quickly I think this simple revert as 1.1.1 is the right thing to do (surprisingly we never got complaints in the past about the layout).

How do you feel about targeting fluid layouts for 1.2.0?

SUPERCILEX:
@samtstern no no, I totally agree with you. Sorry for not making this clearer: I'm expecting the `ConstraintLayout` transition to be slow and drawn out so we can ensure quality and not mess stuff up. This PR should definitely be merged into the 1.1.1 release and the `ConstraintLayout` stuff my even be spread out into the 1.3.0 release depending on how quickly you want to get 1.2.0 out.

PS: I believe #375 is a complaint about landscape layout.

samtstern:
@SUPERCILEX :+1:

PR #527, opened 2017-01-18
Remove apply constants from module build.gradle

SUPERCILEX:
This is a really small change, but I got burned by it while working on a different project. `apply from: "../constants.gradle"` doesn't work for some reason and is unused. Instead, the constants are applied in the [root `build.gradle`](https://github.com/firebase/FirebaseUI-Android/blob/master/build.gradle#L1).


samtstern:
LGTM.  There was a time when this worked (or appeared to work) but don't need it anymore.

PR #529, opened 2017-01-18
Set android:supportsRtl="true" in manifest

ay1man4:
It conflicts with manifest of the application that uses this library which is defaulted to true. keeping it false is making error during gradle dependencies resolving which is not easy to be discovered by newbies.

Issue and solution discussed here:
https://github.com/firebase/FirebaseUI-Android/issues/392

### DIFF
--- a/auth/src/main/AndroidManifest.xml
+++ b/auth/src/main/AndroidManifest.xml
@@ -6,7 +6,7 @@
     <uses-permission android:name="android.permission.INTERNET"/>
 
     <application
-        android:supportsRtl="false">
+        android:supportsRtl="true">
 
         <meta-data
             android:name="com.google.android.gms.version"


samtstern:
So I agree the merging error is tough for new users, but on the other hand we **dont** support RTL right now.  Maybe we should just remove this property?

samtstern:
Yeah actually the more I think about it, it may not be appropriate for the library to declare this attribute at all.  Can you modify this PR to just remove it?

samtstern:
@ay1man4 could you sign the CLA?  Then we can merge #534 which includes your commit and gives you credit.

cc: @SUPERCILEX

ay1man4:
@SUPERCILEX Could you share the link where I have to sign/accept the CLA?

Sorry but this is the first time I contribute with project having CLA agreement.

ay1man4:
@SUPERCILEX I've signed the CLA.

samtstern:
@ay1man4 thanks!  See #534.

Issue #530, opened 2017-01-19
Disable sign-up if email not found.


xsorifc28:
Welcome to FirebaseUI and thanks for submitting an issue!

### Step 1: Are you in the right place?

  * For issues or feature requests related to the code **in this repository** file a GitHub issue.
  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/) tagged appropriately.
  * For general Firebase discussion, use the [firebase-talk google group](https://groups.google.com/forum/#!forum/firebase-talk)
  * For help troubleshooting your application that does not fall under one of the above categories, reach out to the [personalized Firebase support channel](https://firebase.google.com/support/contact/troubleshooting/)

### Step 2: Describe your environment

  * Android device: Nexus 6P
  * Android OS version: 7.1
  * Google Play Services version: 10.0.1
  * Firebase/Play Services SDK version: 10.0.1
  * FirebaseUI version: 1.1.1
  
#### Steps to reproduce:

  1. Sign in with an email not in firebase's authentication database
  
#### Observed Results:

   FirebaseUI will open sign-up page

  
#### Expected Results:

  Be given the option to disable sign up and only allow sign in if email pre-exists.
  
#### Relevant Code:

  None

samtstern:
@xsorifc28 we've seen this request before and I don't think we plan to support this in FirebaseUI.  I am going to close the issue for now, but can you explain why you don't want sign up in your app?

xsorifc28:
@samtstern We are building an app for a conference which should only allow users to login if they have pre-registered and paid using our web-app.

samtstern:
Ok thanks for explaining.  This was also asked in #385.  I think forking the library to do this should be fairly simple, but right now (as I said) it's not on our roadmap.

samtstern:
The fix for this issue has been released in version `1.2.0`.

lindsaymacvean:
Where is the configuration setting for this? I have looked through the documentation and I cannot find it.

ArtbyRere:
Trying to add,
import java.util.ArrayList;
11
<https://github.com/firebase/FirebaseUI-Android/blob/8bda799672d44fd2994d2498441859fae3ccb530/app/src/main/java/com/firebase/uidemo/util/ConfigurationUtils.java#L11>
import java.util.Arrays;
12
[...]

Issue #385, opened 2016-10-31
Disable account creation


JacobVanAssche:
Is it possible to only allow for users to login if they have an account and disable account creation? I want to be able to manually add accounts and allow users to login. I do not want them to be able to create accounts.

samtstern:
@JacobVanAssche I can see the use case here but I don't think this is something we will support in FirebaseUI, since the goal of the library is to implement *all* the auth flows in a simple way with as little configuration as possible.  The more "disable foo" or "enable bar" settings we add, the more complexity we get and the more chances we have for error.

Thanks for the suggestion!

samtstern:
See #530

PR #536, opened 2017-01-22
Add option in AuthUI.SignInIntentBuilder to disable account creation

xsorifc28:
As requested in #530 and #385, add option in AuthUI.SignInIntentBuilder to disable account creation if email does not pre-exist.
`gradlew check` passes.

### DIFF
--- a/app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
+++ b/app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
@@ -102,6 +102,9 @@ public class AuthUiActivity extends AppCompatActivity {
     @BindView(R.id.smartlock_enabled)
     CheckBox mEnableSmartLock;
 
+    @BindView(R.id.allow_new_email_accounts)
+    CheckBox mAllowNewEmailAccounts;
+
     @BindView(R.id.facebook_scopes_label)
     TextView mFacebookScopesLabel;
 
@@ -119,9 +122,6 @@ public class AuthUiActivity extends AppCompatActivity {
 
     @BindView(R.id.google_scope_games)
     CheckBox mGoogleScopeGames;
-
-    @BindView(R.id.allow_new_email_accounts)
-    CheckBox mAllowNewEmailAccounts;
 
     @Override
     public void onCreate(Bundle savedInstanceState) {
--- a/app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
+++ b/app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
@@ -120,6 +120,9 @@ public class AuthUiActivity extends AppCompatActivity {
     @BindView(R.id.google_scope_games)
     CheckBox mGoogleScopeGames;
 
+    @BindView(R.id.create_account_enabled)
+    CheckBox mCreateAccountEnabled;
+
     @Override
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
@@ -183,6 +186,7 @@ public class AuthUiActivity extends AppCompatActivity {
                         .setProviders(getSelectedProviders())
                         .setTosUrl(getSelectedTosUrl())
                         .setIsSmartLockEnabled(mEnableSmartLock.isChecked())
+                        .setCreateAccountIfEmailNotExists(mCreateAccountEnabled.isChecked())
                         .build(),
                 RC_SIGN_IN);
     }
@@ -206,6 +210,9 @@ public class AuthUiActivity extends AppCompatActivity {
         if (resultCode == ResultCodes.OK) {
             startActivity(SignedInActivity.createIntent(this, response));
             finish();
+            return;
+        } else if(resultCode == ResultCodes.NO_EMAIL_ACCOUNT) {
+            showSnackbar(R.string.error_email_does_not_exist);
             return;
         } else {
             // Sign in failed
--- a/app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
+++ b/app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
@@ -211,9 +211,6 @@ public class AuthUiActivity extends AppCompatActivity {
             startActivity(SignedInActivity.createIntent(this, response));
             finish();
             return;
-        } else if (resultCode == ResultCodes.NO_EMAIL_ACCOUNT) {
-            showSnackbar(R.string.error_email_does_not_exist);
-            return;
         } else {
             // Sign in failed
             if (response == null) {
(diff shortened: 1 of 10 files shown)


auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -632,7 +633,15 @@ public FlowParameters getFlowParams() {
                                       mTheme,
                                       mLogo,
                                       mTosUrl,
-                                      mIsSmartLockEnabled);
+                                      mIsSmartLockEnabled,
+                                      mCreateAccountIfEmailNotExists);
         }
+
+        public SignInIntentBuilder setCreateAccountIfEmailNotExists(boolean enabled) {
```
samtstern: I think we can pick a shorter and/or clearer name for this.  Suggestions:

- `setAllowNewEmailUsers()`
- `setAllowNewEmailAccounts()`
- `setEnableEmailAccountCreation()`
- `setDisableEmailAccountCreation()` --> in this case we'd have to invert the boolean throughout

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailActivity.java
```diff
@@ -118,17 +120,25 @@ public void onExistingIdpUser(User user) {
...
+            FragmentTransaction ft = getSupportFragmentManager().beginTransaction()
+                    .replace(R.id.fragment_register_email, fragment, RegisterEmailFragment.TAG);
 
-        ft.disallowAddToBackStack().commit();
+            View v = findViewById(R.id.email_layout);
+            if (v != null) ft.addSharedElement(v, "email_field");
+
+            ft.disallowAddToBackStack().commit();
+        } else {
+            finish(ResultCodes.NO_EMAIL_ACCOUNT, IdpResponse.getErrorCodeIntent(ErrorCodes.UNKNOWN_ERROR));
```
samtstern: I think calling `finish()` here is a little severe, since the user could reach this state as a result of a typo.  Maybe we should just add an error to the email text field saying `No account exists for this email address` or something similar.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
xsorifc28: I think `setEnableEmailAccountCreation()` is a better suggestion.

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailActivity.java
xsorifc28: Agreed. I didn't think of that case, just assumed email was entered correctly (given the SmartLock flow, I never even had to type an email :p ) . The implementation can then just customize the error text that shows up.

auth/src/main/java/com/firebase/ui/auth/ui/email/CheckEmailFragment.java
```diff
@@ -100,9 +101,11 @@ public View onCreateView(LayoutInflater inflater,
         View v = inflater.inflate(R.layout.check_email_layout, container, false);
 
         // Email field and validator
+        mEmailLayout = (TextInputLayout) v.findViewById(R.id.email_layout);
         mEmailEditText = (EditText) v.findViewById(R.id.email);
-        mEmailFieldValidator = new EmailFieldValidator(
-                (TextInputLayout) v.findViewById(R.id.email_layout));
+        mEmailFieldValidator = new EmailFieldValidator(mEmailLayout);
+        mEmailLayout.setOnClickListener(this);
```
samtstern: Why do we need click listeners on the layout and edittext?

xsorifc28:
@samtstern Applied the requested changes.

auth/src/main/java/com/firebase/ui/auth/ui/email/CheckEmailFragment.java
samtstern: Oh nevermind I see, when you click into the field the error disappears.  :+1:

xsorifc28: Because the error text does not clear once the EditText is clicked. I felt that if the user clicks this field, the error message should reset. Thus I implemented the click listener and set the error to null.

samtstern:
Ok @xsorifc28 this all looks great!  The only thing left is for me to ask a few other people what they think of adding this feature, but if the vote is in favor of adding we can merge this PR.

@SUPERCILEX @amandle @morganchen12 thoughts?

auth/src/main/java/com/firebase/ui/auth/ui/email/CheckEmailFragment.java
xsorifc28: @samtstern Right, I believe EmailFieldValidator doesn't have this behavior, the error message stays until the next button is clicked and validation is re-run.

app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
```diff
@@ -120,6 +120,9 @@
     @BindView(R.id.google_scope_games)
     CheckBox mGoogleScopeGames;
 
+    @BindView(R.id.allow_new_email_accounts)
+    CheckBox mAllowNewEmailAccounts;
```
SUPERCILEX: Small nit, this should be under `mEnableSmartLock`. To be honest, all of these should be reorganized in the order they appear in the sample so I guess it doesn't really matter.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -517,6 +517,7 @@ public IdpConfig build() {
         private LinkedHashSet<IdpConfig> mProviders = new LinkedHashSet<>();
         private String mTosUrl;
         private boolean mIsSmartLockEnabled = true;
+        private boolean mCreateAccountIfEmailNotExists = true;
```
SUPERCILEX: nit: shouldn't this be something like `mAllowNewEmailAccounts`?

xsorifc28: Yeah...missed that one.

app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
xsorifc28: Done

SUPERCILEX:
@xsorifc28

samtstern:
@xsorifc28 my bad, a recent PR merged caused a small strings conflict here.  Would you mind resolving?  Or I can do it via the web resolution editor.

xsorifc28:
@samtstern I believe its good to go.

samtstern:
Ok @morganchen12 is going to give this a shot on iOS so I am goign to squash/merge this into the version 1.2.0 branch.

auth/src/main/java/com/firebase/ui/auth/ui/email/CheckEmailFragment.java
hmuhammadazeem: Maybe, instead of text, we can use Hint attribute to show the error. This would eliminate the need for listener.

Issue #539, opened 2017-01-23
feat: FirebaseObservableMap


jpventura:
I was not sure if I should PR this feature or discuss here first.

#### ObservableMap as a facade to event listeners

Basically I would like to know if we could provide a feature for hide read and write operations behind an [`ObservableMap`](https://developer.android.com/reference/android/databinding/ObservableMap.html). So all we would have to do to access data would be:

    // Remote updates immediately send you your view
    map.addOnMapChangedCallback(mUserChangedCallback);

    // Non blocking operation to update database
    Task<User> task = map.create(user);

    // Local up-to-date cache
    User user = map.get(primaryKey);
  
#### Relevant Code:
A detailed description is available at [Stack Overflow](http://stackoverflow.com/questions/41767616/how-can-i-turn-firebase-listeners-into-observable-map-using-data-binding-library) and a working code at my [repository fork](https://github.com/jpventura/FirebaseUI-Android/commits/data_binding).

puf:
Where/how would you *use* such a `ObservableMap`?

jpventura:
@puf It is more a readability improvement the functionality.

We are using Firebase in a project and I was trying to prevent the developers of writing too many nested listeners:
  - Emulating `Map` interface, the read/write became more readable
  - Sequence os tasks became simpler to write

puf:
A map usually is a key-value collection with unspecified order semantics, where a Firebase DataSnapshot is an ordered key-value collection. So you're losing the ordering information. 

Why don't you use the `DataSnapshot` instead? That also frees you from the adapter overhead, but then ensures you stick to the underlying system model. 

I have no doubt you have good reasons for wanting this, but I don't see the advantages in usage. It may be easier to understand if you add some more code snippets on using the `ObservableMap`, comparing it to how you'd use a `DataSnapshot` or `Firebase...Adapter` in the same use-cases.

samtstern:
@jpventura thanks for raising an issue to discuss this rather than going straight to a PR, that's always the right thing to do.

I agree with @puf, I am not familiar with this pattern and would like to see some more examples of how it's used and how that compares with normal Firebase usage patterns to understand the developer benefir.

samtstern:
[...]

Issue #337, opened 2016-10-03
FirebaseArray's Limited Access Modifiers


TheCraftKid:
I'm trying to create an addon to FirebaseUI; this addon includes the FirebaseSpinnerAdapter I proposed in #304 for this library. @puf suggested I put it in a separate library, thus I created the [FirebaseUI-Android-Addons repository](https://github.com/TheCraftKid/FirebaseUI-Android-Addons). 
It (for now) includes the FirebaseSpinnerAdapter, but I ran into a problem during implementation: I had to create a [copy of the FirebaseArray](https://github.com/TheCraftKid/FirebaseUI-Android-Addons/blob/v0.1.0/library/src/main/java/com/craft/libraries/firebaseuiaddon/FirebaseArray.java) for a future feature since the one FirebaseUI provides is package-private.

I was wondering if I could submit a PR to change that since my problem would be fixed with some SOLID programming.

SUPERCILEX:
@puf I don't see the harm in making `FirebaseArray` public. `mSnapshots` is still private so I don't see how using `FirebaseArray` could lead to unexpected behavior.

puf:
You can currently use FirebaseArray in your own adapters by putting them in
the same package, which is how FirebaseUI's adapters do it.

We've so far steered away from making `FirebaseArray` part of the public
contract of the library. The reasoning is that we'd be supporting its API
contract and guaranteeing backwards compatibility. While we may (and
probably should) at some point do that, I'd prefer to push that moment out
a bit further if possible.

samtstern:
#544 has been merged.  You can check out the `version-2.0.0-dev` branch and play around if you're curious.

TheCraftKid:
Thanks!

Issue #409, opened 2016-11-18
displayName returning null   firebase-ui-auth 1.0.0


jetav8r:
I'm trying to add a user using the Firebase auth and it returns the displayName as null, even though it asks for it via the UI on sign up. Any ideas why and how to fix? Here's the code I am using to do this:

startActivityForResult( AuthUI.getInstance()
.createSignInIntentBuilder()
.setProviders(Collections.singletonList(new AuthUI.IdpConfig.Builder(AuthUI.EMAIL_PROVIDER).build()))
.build(), RC_SIGN_IN);

and in the Activity for Result....

protected void onActivityResult(int requestCode, int resultCode, Intent data) {
super.onActivityResult(requestCode, resultCode, data);
if (resultCode == RESULT_OK) {
// user is signed in!

        //save user
        saveUser();

        startMainActivity();
        return;
    }
save user code is as follows:

private void saveUser(){
FirebaseUser fUser = FirebaseAuth.getInstance().getCurrentUser();
final User user = new User();
if (fUser.getDisplayName() != null) {
user.setUserName(fUser.getDisplayName());
} else {
Log.e("FB_info", "User has no display name");
}
user.setUserEmail(fUser.getEmail());

    final DatabaseReference userRef = FirebaseDatabase.getInstance().getReference(FirebaseRefs.getUsersRef()).child(fUser.getUid());
    userRef.addListenerForSingleValueEvent(new ValueEventListener() {
        @Override
        public void onDataChange(DataSnapshot dataSnapshot) {
            if(!dataSnapshot.exists()){
                userRef.updateChildren(user.toMap());
            }
        }

        @Override
        public void onCancelled(DatabaseError databaseError) {

        }
    });

}
Any idea why the display name is returning null?

SUPERCILEX:
Do you ever call `setValue()` to upload the name to your database?

samtstern:
@jetav8r I don't see an obvious bug here, but this is also not likely to be an issue with FirebaseUI.  Please raise this question on StackOverflow as this is more of a troubleshooting request than a library issue.

jetav8r:
I can't upload the value to the database because the user.getDisplayName()
is always null

jetav8r:
I'm gonna disagree with you here, since the only interface is with the
Firebase Auth UI and it doesn't return the displayName when I call
getDisplayName...this comes directly from your methods.  The UI asks for
the users First and Last name, but it always returns null.  thanks for
[...]

PR #544, opened 2017-01-26
The Big Kahuna of firebase-ui-database

SUPERCILEX:
enables #539 (I think)
#337
@TheCraftKid I think this one will make you happy! (and your library obsolete 😄)

@samtstern Sorry about this one, the amount of changes is high up there on my list of biggest PRs, but I believe all of this is necessary for us to be able to commit to maintaining `firebase-ui-database` in the long term.

- [ ] Decide what to do about `equals` and `hashCode`.
- [x] Thoroughly test stuff
- [x] Figure out what to do about the performance issues with `FirebaseArrayOfObjects`
- [x] Explain what's going on in `FirebaseArrayOfObjectsOptimized` (Fixed in fda54177423b57c267b0208ffea280a6a385a564)


TheCraftKid:
I'm looking at the branch, and, if my eyes don't deceive me, I see a public FirebaseArray and a new FirebaseObjectsArray. Is this what you meant, @SUPERCILEX?

SUPERCILEX:
@TheCraftKid Yup! It might be happening! 😄

PS: Don't update your addons quite yet as this PR is targeted for the 1.2.0 release which could take a few weeks.

puf:
This feels like a **major** change, which should not be introduced in a minor version. Why has a completely new adapter (in the `com.firebase.ui.database.adapter` package) been introduced, instead of updating the existing one? Why have the completely functional existing adapters been deprecated? Where is a non-deprecated `FirebaseListAdapter`?

SUPERCILEX:
@puf I submitted this PR prematurely, will explain everything and CC you once I think it's ready.

SUPERCILEX:
@puf I believed we should scrap `ListView` in order to improve performance for devs, but upon further investigation into the uses of `ListView` today, that seems like the wrong move. As such, I've added back a `FirebaseListAdapter`.

@samtstern @puf Wow, this turned out to be a longer post than I had thought, hopefully it isn't all too much for you guys.

# High level overview

## Adapters

### Why deprecate the old adapters?

I believe some bad early decisions were made by not directly implementing `FirebaseArray.OnChangedListener` and instead forwarding calls to protected versions of the interface methods. To implement `ChangeEventListener` today (which is what I've done), we would need to make our access modifiers weaker, thus breaking backwards compatibility.

### What's up with the new `com.firebase.ui.database.adapter` package?

First, expanding the `firebase-ui-database` apis means more classes. I believe these classes fall into two distinct categories: the core FirebaseUI apis such as `FirebaseArray` and then the adapters. Organizing our classes in such a way provides a nice distinction between our core apis and our adapters. Second, I don't believe Java lets you have two classes with the same name in the same package so this is also a necessary move in addition to being the right move in my opinion.

### What are the benefits to directly implementing the interfaces?

1. Documentation! This one is only for us, but if you look at the new adapters, there's no javadoc because everything is an interface now.
2. It fits with the new `FirebaseArray` model of `addChangeEventListener`: you could swap out data sets on an adapter really easily now. This makes for a really nice api where an adapter is nothing more than a receiver of change events.
3. You know the drill, `List<T>`, `Runnable`, etc aren't there for nothing. Directly implementing `ChangeEventListener` is simply a far superior way of doing it than forwarding calls everywhere and it provides that nice abstraction every developer loves.

### One interface to rule them all: `FirebaseAdapter`

Unfortunately, due to the nature of adapters needing to extend stuff, we can't simply call `setAdapter(FirebaseAdapter)`. However, the point of this one is to provide a unified way to create an adapter for today and in the future. Implementing `FirebaseAdapter` provides all of the core methods needed for a Firebase adapter such as `cleanup`, `getRef`, etc. This means documentation is all in one place, and extenders of FirebaseUI will be able to provide consistent apis that people will be familiar with.

### Some new features for adapters

There's now a `startListening()` method which gives devs more control over when their adapter should start doing stuff. I believe not letting devs choose when to start listening was one of the fundamental flaws with `FirebaseArray`, something I don't want to repeat for the adapters. The new method makes doing things like adding a `JoinResolver` in #448 a piece of cake: add the interface and you're good to go. Devs can do stuff like this with no changes on our side:
```java
@Override
public void startListening() {
    ((FirebaseIndexArray) mSnapshots).setJoinResolver(this);
    super.startListening();
}
```

### Documentation

As I mentioned earlier, everything is provided through interfaces meaning we only update stuff once which is really nice. To go even further, I removed the in-depth class documentation leaving only a high-level overview. Instead, we now refer people to our README which would alleviate problems like #550 and ensure that users always see up to date documentation. In addition, this promotes awareness of our sub READMEs which I feel like many people don't know exist for some reason.

## Core apis

### `FirebaseArray`

#### Implementing `List<DataSnapshot>`

This is the biggest change for `FirebaseArray` and I think the reasons are self explanatory: it's in the name "array". Implementing `List` makes `FirebaseArray` far more powerful and flexible making it a true array that can be used as a source of truth for a `Query`. It's always up to date and you can perform all the functions of a list directly on `FirebaseArray`.

#### Following the core Firebase apis as closely as possible

`FirebaseArray` now follows the same add/remove model that `Query` does for listeners which is far superior to the setListener model the previous `FirebaseArray` followed.

### `FirebaseArrayOfObjects`

Because most people care about the objects contained inside the `DataSnapshot`s and not the snapshots themselves, `FirebaseArrayOfObjects` provides a nice wrapper around a `List<DataSnapshot>`.

#### `SnapshotParser` interface

To provide flexibility, `FirebaseArrayOfObjects` uses a `SnapshotParser` to convert `DataSnapshot`s to their model object. (And of course, `FirebaseAdapter extends SnapshotParser`).

# Technical tidbits

## `equals` and `hashCode`

We need to decide how we want to implement `equals` and `hashCode`. Right now, they are not solely based on the contents of the list which feels wrong yet necessary. A `FirebaseArray` with the same contents, but a different query isn't really equal. In the same sense, a `FirebaseArrayOfObjects` with the same contents, but a different model class isn't really equal either.

## `FirebaseArrayOfObjects`

Because `FirebaseArrayOfObjects` tries to provide an up to date list of objects without relying on `FirebaseArray` to be more flexible, it must rebuild the list of objects on every list operation except for `size`, `isEmpty`, and `get` (thankfully). I think we need to either provide a subclass that is more opinionated and uses `FirebaseArray` to keep an up to date copy of the list, or adequately warn developers about the performance implications of calling `indexOf` and the other list methods.

TheCraftKid:
It was a good idea to not scrap the `FirebaseListAdapter`s because Android's RemoteViews (for widgets and whatnot) don't support RecyclerViews (hopefully yet).

samtstern:
@SUPERCILEX by the way this has not been ignored, and thank you for your long explanation.  I plan to spend some time looking at this soon.

Because I want to keep the focus on the main changes here, could you revert any changes to the sample app or non-code files (besides those forced to keep it compiling)?  If this change is approved we can do a follow up PR to fix `README` and sample files.

SUPERCILEX:
@samtstern done, I created #567 to track documentation and such. Do you also want me to move test updates to #567 and let them be broken be broken in this PR?

samtstern:
@SUPERCILEX thanks for moving those changes.  Tests belong here still, so we're good.

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
```diff
@@ -14,52 +14,118 @@
...
+     * @param listener the listener to be called with changes
+     * @return a reference to the listener provided. Save this to remove the listener later
+     * @throws IllegalArgumentException if the listener is null
+     */
+    public ChangeEventListener addChangeEventListener(@NonNull ChangeEventListener listener) {
+        checkNotNull(listener);
+
+        mListeners.add(listener);
+        notifySubscriptionEventListeners(SubscriptionEventListener.EventType.ADDED);
+        if (mListeners.size() <= 1) {
```
samtstern: nit: change to `mListeners.size() == 1` and add a comment explaining this pattern.  As I understand it, we are not listening to Firebase Database until at least one listener has been added.

database/src/main/java/com/firebase/ui/database/FirebaseArrayOfObjects.java
```diff
@@ -0,0 +1,237 @@
...
+public class FirebaseArrayOfObjects<E> extends ImmutableList<E> {
+    protected List<DataSnapshot> mSnapshots;
+    protected Class<E> mEClass;
+    protected SnapshotParser<E> mParser;
+
+    /**
+     * @param snapshots  a list of {@link DataSnapshot}s to be converted to a model type
+     * @param modelClass the model representation of a {@link DataSnapshot}
+     */
+    public FirebaseArrayOfObjects(List<DataSnapshot> snapshots, Class<E> modelClass) {
```
samtstern: Should this constructor be private?  Seems strange to have a public constructor and `newInstance`

database/src/main/java/com/firebase/ui/database/FirebaseArrayOfObjects.java
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return "FirebaseArrayOfObjects{" +
+                "mSnapshots=" + mSnapshots +
+                '}';
+    }
+
+    protected static class FirebaseArrayOfObjectsOptimized<E> extends FirebaseArrayOfObjects<E>
```
samtstern: Need comments all throughout this class to explain what's going on.  Also it should likely be in its own file unless I am missing something that requires it to be a static inner class.

database/src/main/java/com/firebase/ui/database/FirebaseArrayOfObjects.java
```diff
@@ -0,0 +1,237 @@
...
+            implements ChangeEventListener, SubscriptionEventListener {
+        protected List<E> mObjects = new ArrayList<>();
+        protected Pair<Boolean, Boolean> mIsListening$AddedListener = new Pair<>(true, false);
```
samtstern: Why does this variable name contain `$`?  never seen this convention before.

database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
```diff
@@ -14,57 +14,60 @@
...
 
     @Override
-    public void cleanup() {
-        super.cleanup();
-        Set<Query> refs = new HashSet<>(mRefs.keySet());
-        for (Query ref : refs) {
-            ref.removeEventListener(mRefs.remove(ref));
+    public void removeChangeEventListener(@NonNull ChangeEventListener listener) {
+        super.removeChangeEventListener(listener);
+        if (!isListening()) {
```
samtstern: Slightly paranoid: throughout this PR there are operations like this that depend on changes in the `isListening` state where multiple things must happen when the number of listeners goes above or below 1.  Because there is nothing preventing the rapid adding/removal of listeners to any of your classes, these operations would be safer inside of `synchronized` blocks that share a lock.  That way if someone quickly calls `addListener()`, `removeListener()`, `addListener()` we don't get in trouble.

database/src/main/java/com/firebase/ui/database/FirebaseIndexListAdapter.java
```diff
@@ -54,6 +56,6 @@ public FirebaseIndexListAdapter(Activity activity,
                                     @LayoutRes int modelLayout,
                                     Query keyRef,
                                     Query dataRef) {
-        super(activity, modelClass, modelLayout, new FirebaseIndexArray(keyRef, dataRef));
+        super(activity, modelClass, modelLayout, new FirebaseIndexArray(keyRef, dataRef.getRef()));
```
samtstern: Probably missing something, but why accept a `Query` if we are going to call `getRef()` on it immediately? Also it seems strange to introduce a behavior change at the same time as a deprecation.

database/src/main/java/com/firebase/ui/database/FirebaseIndexListAdapter.java
```diff
@@ -31,9 +31,11 @@
  *     listView.setListAdapter(adapter);
  * </pre>
  *
- * @param <T> The class type to use as a model for the data
- *           contained in the children of the given Firebase location
+ * @param <T> The class type to use as a model for the data contained in the children of the given
+ *            Firebase location
+ * @deprecated use {@link com.firebase.ui.database.adapter.FirebaseIndexRecyclerAdapter} instead
  */
+@Deprecated
```
samtstern: Ok this is the biggest thing we need to talk about: the change in package and the mass-deprecation.  This is something we should avoid at nearly all costs.  A change like this would need to be in version `2.0` for sure, as even if SemVer does not have a strong opinion on deprecations this is essentially asking every FBUI developer to change all of their code.

samtstern:
Can you explain this sentence in your explanation:
""Second, I don't believe Java lets you have two classes with the same name in the same package so this is also a necessary move in addition to being the right move in my opinion.""

What are the class names that would conflict?

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
SUPERCILEX: @samtstern Yup, that's correct.

database/src/main/java/com/firebase/ui/database/FirebaseArrayOfObjects.java
SUPERCILEX: @samtstern Thanks for catching that, it was a refactoring mistake.

database/src/main/java/com/firebase/ui/database/FirebaseArrayOfObjects.java
SUPERCILEX: @samtstern Yeah, this shouldn't have been a `Pair` anyway. I was using the `$` as an `&`.

database/src/main/java/com/firebase/ui/database/FirebaseArrayOfObjects.java
SUPERCILEX: @samtstern Totally agree with you about the comments, but I don't even know if it works yet so I'd rather wait so I don't cement possibly false ideas into my head. I've added it to the TODO list.

As for moving it to a separate class, we could do that but my thoughts were that it's supposed to be part of the same class and the same `newInstance` api. I didn't want to move it to a separate class because I don't want it to be public, but I also want people to be able to extend it, thus the `protected` modifier. What do you think?

database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
SUPERCILEX: @samtstern Yup, better be safe than sorry.

Does this locking make sense to you? I'm only locking additions and removals to the list:
```java
public ChangeEventListener addChangeEventListener(@NonNull ChangeEventListener listener) {
    checkNotNull(listener);

    synchronized (this) {
        mListeners.add(listener);
    }
    // ...

    return listener;
}

public void removeChangeEventListener(@NonNull ChangeEventListener listener) {
    synchronized (this) {
        mListeners.remove(listener);
    }
    // ...
}
``

File: database/src/main/java/com/firebase/ui/database/FirebaseIndexListAdapter.java (hunk 6 above)
Comment (SUPERCILEX) [inline database/src/main/java/com/firebase/ui/database/FirebaseIndexListAdapter.java]: @samtstern Yeah, this is the deprecated adapter so we can't break backwards compatibility.

Comment (SUPERCILEX) [inline database/src/main/java/com/firebase/ui/database/FirebaseIndexListAdapter.java]: @samtstern This shouldn't be a behavior change because we ignore ordering anyway. I wanted to make that clearer so `FirebaseIndexArray` only accepts a `DatabaseReference`, hence the need to convert the `Query`.

**Comment by SUPERCILEX** [comment]
@samtstern The conflicting classes would be the adapters: `FirebaseRecyclerAdapter` etc.

File: database/src/main/java/com/firebase/ui/database/FirebaseArrayOfObjects.java (hunk 3 above)
Comment (samtstern) [inline database/src/main/java/com/firebase/ui/database/FirebaseArrayOfObjects.java]: As far as not being able to extend it, making the class `final` is the proper way to do that.

The reason to move it into a separate final (maybe as a `protected final` class) is just because it's strange to see a subclass in the same file as its super class and they share so many methods that it makes it harder to navigate the code.

File: database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java (hunk 5 above)
Comment (samtstern) [inline database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java]: For this method I would lock the entire body (removing the event listeners and clearing data/refs) and use the `mListeners` list as the lock.

It's also worth noting that `isListening()` itself is implemented as a check on mListeners, so that should be inside the lock.

File: database/src/main/java/com/firebase/ui/database/FirebaseIndexListAdapter.java (hunk 6 above)
Comment (samtstern) [inline database/src/main/java/com/firebase/ui/database/FirebaseIndexListAdapter.java]: Ok that makes sense.

File: database/src/main/java/com/firebase/ui/database/FirebaseIndexRecyclerAdapter.java [hunk 8]
```diff
@@ -59,8 +59,11 @@
  * </pre>
  *
  * @param <T>  The Java class that maps to the type of objects stored in the Firebase location.
- * @param <VH> The ViewHolder class that contains the Views in the layout that is shown for each object.
+ * @param <VH> The ViewHolder class that contains the Views in the layout that is shown for each
+ *             object.
+ * @deprecated use {@link com.firebase.ui.database.adapter.FirebaseIndexRecyclerAdapter} instead
  */
+@Deprecated
```
samtstern: Let's take this as a concrete example of the deprecation objection.  Here are the differences I see in the new class vs the old class:

- A change in the order of the constructor arguments (`viewHolderClass` switched with `modelLayout`)
- A change in the package name
- `dataRef` has become a `DatabaseReference` where it was a `Query`

The first change (param ordering) seems to just be random.  The package name change is nice, but certainly the kind of thing we can live with unchanged.  And the third change could be done by simply deprecating the constructor that accepts `Query` and pointing people to a new overloaded constructor that accepts a `DatabaseReference`

So couldn't we avoid deprecating this class?

database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
SUPERCILEX: @samtstern SGTM, I was thinking that locking the whole object when adding or removing listeners to the list would guarantee that calls to `isListening` would be blocked while the list is updated. However, your solution makes it clearer that we're locking the list so I like it better.

database/src/main/java/com/firebase/ui/database/FirebaseArrayOfObjects.java
SUPERCILEX: @samtstern I said I want people to be able to extend it, not the opposite. Here's what I would like:

- People **can** extend `FirebaseArrayOfObjectsOptimized`
- People **cannot** create a new instance from outside `FirebaseArrayOfObjects`

@samtstern as for making a normal class `protected`, you can't do that. It goes `public` -> `default` and that's it. For people to be able to extend `FirebaseArrayOfObjectsOptimized` the class would need to be public, but that makes a confusing api because people will wonder what the difference between `FirebaseArrayOfObjects` and `FirebaseArrayOfObjectsOptimized` is.

@samtstern Given what I said above, do you still think `FirebaseArrayOfObjectsOptimized` should still be in `FirebaseArrayOfObjects`? I know it makes it confusing for us, but I'm trying to think of the api consumer instead of us.

database/src/main/java/com/firebase/ui/database/FirebaseIndexRecyclerAdapter.java
SUPERCILEX: @samtstern Yeah, the permuted arguments was a refactoring mistake.

@samtstern I would never needlessly deprecate something so here the missing key point in that list above: we are now implementing `ChangeEventListener` directly. If you diff the two files, you'll notice that the protected methods that forward calls from `ChangeEventListener` are now public, thus breaking backwards compatibility. I wrote a blurb in my original post about why we should directly implement the interface, but I'll reiterate the main point here: since people can access `FirebaseArray` directly now, it makes no sense to have an anonymous listener forwarding calls to identical methods whose only difference to the interface is being protected. If you look at what I did in the old adapters, we have to store the listener to be able to remove it later and it would be such a pain for devs if they wanted change listeners midway or something. Also, it's just confusing in general if devs try to implement `ChangeEventListener` and they get compile errors because the forwarding methods are protected.

SUPERCILEX:
Just to help keep track of things, these are the two comments that still need to be addressed: https://github.com/firebase/FirebaseUI-Android/pull/544#discussion_r99217115 and https://github.com/firebase/FirebaseUI-Android/pull/544#discussion_r99260360. Once those are resolved, I'll go in a testing phase to make sure everything works. (I know the new adapters and basic `FirebaseArray`/`FirebaseIndexArray` stuff works, but I haven't tested anything else.)

SUPERCILEX:
@samtstern I realized that the only non-stable api was the new adapters so I went ahead and started testing everything. There were quite a few bugs, but I was pleasantly surprised that everything mostly worked the first time.

TheCraftKid:
Hey, just a suggestion, maybe FirebaseArrayOfObjects and FirebaseArrayOfObjectsOptimized should be named to FirebaseObjectsArray and FirebaseOptimzedObjectsArray respectively. The semantics are a bit better for developers that will use the classes.

database/src/main/java/com/firebase/ui/database/FirebaseIndexRecyclerAdapter.java
samtstern: Sorry for the delay here but I have been trying to think of some way out of this situation, been playing with your branch in my IDE.

We need to find *some* way to avoid deprecating `FirebaseRecyclerAdapter` and `FirebaseListAdapter` pretty much at all costs.  99% of our users just want to use the adapters as-is, but we'd be breaking all of them for the 1% who want to extend the adapters in interesting ways.  I see the issue in the changing of `protected onDataChanged` to `public onDataChanged`, that's what we need to get around.

Here's a random idea:
  * Rename the method in `ChangeEventListener` to something else, have that method delegate to `onDataChanged` in the default implementation of each adapter.
  * Deprecate the `onDataChanged` method and point people to use the newly named method.

Would that enable us to keep the adapters in the original package and not deprecated?  I'm sure I'm missing something, hard to reason about these things in a comment box.

samtstern: I experimented with the idea above, you can see what would happen here:
https://github.com/samtstern/FirebaseUI-Android/commit/5d2d34b802c3295505dfd1144cad8f58abd79814

This makes the deprecation blow much softer, still not "ideal".  What do you think of something like that?

SUPERCILEX: @samtstern What do you think of doing a "Screw it all, let's release v2" type of thing? People would just have to make their overridden methods public which doesn't seem so bad. If not, I like your solution, but the method name changes would kill me:

- `onChildChanged` -> `onChangedEvent` I actually kinda like this one.
- `onCancelled` -> `onDatabaseError` I think `onError` would be nicer
- `onDataChanged` - `onNewData` This one is just a blatant lie because it will get called for data changes, removals permutations, etc.
- `parseSnapshot` -> `parseDataSnapshot` This just makes me 😢

If we decided to go the method rename route, `onNewData` absolutely needs to be fixed. Other than that, I guess it's okay. Like you said, it wouldn't be ideal but it's better than being like "Screw y'all, everything you're using is deprecated!" 😄

jpventura:
@SUPERCILEX When I first started #539, I initially made a `FirebaseObservableArray` because I was hoping to use it inside [`ArrayAdapter`](https://developer.android.com/reference/android/widget/ArrayAdapter.html) and no longer relay on [`FirebaseListAdapter`](https://firebaseui.com/docs/android/com/firebase/ui/FirebaseListAdapter.html).

But right at the begining, I felt forced to move from [`List`](https://developer.android.com/reference/java/util/List.html) to [`ArrayMap`](https://developer.android.com/reference/android/support/v4/util/ArrayMap.html) due [`Adapter.add`](https://developer.android.com/reference/android/widget/ArrayAdapter.html#add(T)) and [`Adapter.addAll`](https://developer.android.com/reference/android/widget/ArrayAdapter.html#addAll) methods.

These methods correctly call [`notifyDataSetChanged`](https://developer.android.com/reference/android/widget/ArrayAdapter.html#notifyDataSetChanged()) if the collection changes, however the key generation would have two possible scenarios:

  - A random key is created through [`push().key`](https://firebase.google.com/docs/reference/js/firebase.database.Reference#push)
  - The key would be an [`Adapter.add`](https://developer.android.com/reference/android/widget/ArrayAdapter.html#add(T)) argument

SUPERCILEX:
@jpventura I'm confused, what's your point and how does this relate to my PR? Sorry about the misunderstanding!

SUPERCILEX:
@samtstern @puf Given the discussion in https://github.com/firebase/FirebaseUI-Android/pull/544#discussion_r99927717, FirebaseUI `v2` or sad method names?

puf:
Releasing a new major version (say 2.x) is easy. But that leaves the question what happens to people currently using 1.x versions. If there won't be fixes/maintenance releases on 1.x then they're equally screwed, just in a semver compatible way.

Keeping the existing methods working and adding new alternatives that expose the new functionality will prevent that, at a higher maintenance cost to the current version. Also not pretty to the ones writing code (it's a pretty bad technical debt), but friendlier to the library's users.

I thought we could just add the now-public methods to the interface, but that indeed doesn't work (since existing implementation have those methods marked as `protected` and would not compile). But having to introduce new method names, just so we can make them public is horrible. 

So: I don't have a solution yet, but definitely a better understanding of the problem. It definitely makes me lean towards cutting a major, but that leads back to the maintenance consideration. :-/

SUPERCILEX:
@puf Awesome, thanks for the great explanation! I do feel like a `v2` release makes more sense:
1. If I move the new adapters to their old package, most users won't even notice that anything changed.
2. Since FirebaseUI `1.0.0`, we've mostly just been fixing bug after bug after bug so I don't think the maintenance costs would be that bad especially considering `v1.2.0` which [fixes another slew of bugs](https://github.com/firebase/FirebaseUI-Android/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20label%3Afix-implemented%20label%3Abug%20).

samtstern:
So I have been thinking (sometimes aloud in @puf's direction) about the rename vs major version issue and I am now also leaning towards major version for a few reasons:

- The rename thing is "semver compatible" but we all know it's an abuse of the system :-) also it adds technical debt and makes us look bad to future potential contributors
- I actually don't believe there will be a maintenance cost to moving to 2.0.  This will only be a breaking change in the `database` module and that module hasn't had a serious bug in ages.  So I don't think we'd ever need to release another `1.x` version for database.  As for auth, we won't be making any breaking changes.

So I'd like to hear thoughts from @puf  and @SUPERCILEX but I'd recommend the following path:

  * Let's get out `v1.2.0` without this PR.  It will focus on Auth fixes (as usual).
  * Change this PR to be a breaking change but also smaller by keeping the adapters in the package they're currently in (rather than the move-and-deprecate method).
  * Move this PR to a `version-2.0-dev` branch and make a new "Project" for version 2.0.  We should also think about if there are any other changes we'd like to make while we have the opportunity to bump the major version.
  * Plan for version `2.0` to come as the successor to version `1.2.0` unless major bugs requiring a `1.2.1` are found.

SUPERCILEX:
@samtstern Sounds like a plan to me!

puf:
Yup. I'm also onboard with this. I think a major version is the best way to
go as long as we have a good policy for how we support developers who
remain on 1.x. It could be as simple as "we will release critical bugfixes
for FirebaseUI-Android 1.2 for 6 months after FirebaseUI-Android 2.0 has
been released" (don't worry too much about the exact number of months, I
just needed a number). Is that needed?

SUPERCILEX:
@samtstern @puf If it's decided that `v1.2.0` is going to be the last major release on the `1.x` branch, I think we really need to wait for the next play services version to come out before releasing `v1.2.0` since it fixes some important bugs like #409.

samtstern:
I think @SUPERCILEX raises a very good point, we need to consider that a 'critical bug fix' may come from an underlying dependency.  Play Services makes this especially interesting.

I have this nagging feeling that we could somehow make play services a `provided` dependency but I haven't figured it out and it's far outside the scope of this discussion.

database/src/main/java/com/firebase/ui/database/FirebaseArrayOfObjects.java
```diff
@@ -0,0 +1,231 @@
...
+        @Override
+        public void onSubscriptionRemoved() {
+            FirebaseArray snapshots = (FirebaseArray) mSnapshots;
+            if (!snapshots.isListening()) {
+                snapshots.removeChangeEventListener(this);
+            }
+        }
+
+        @Override
+        public void onSubscriptionAdded() {
```
samtstern: I'm having a hard time wrapping my head around the SubscriptionEventListener concept.  It seems to be a meta-listener, letting you know when a ChangeEventListener has been added or removed from a source.

Here the logic is "if a listener is added to my FirebaseArray and I am not listening to changes on that array, add me as a listener" ... it seems very self-referential and I am trying to figure out the problem it solves.

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
```diff
@@ -14,52 +14,137 @@
...
+        checkNotNull(listener);
+
+        mListeners.add(listener);
+        notifySubscriptionEventListeners(SubscriptionEventListener.EventType.ADDED);
+        if (mListeners.size() <= 1) { // Only start listening when the first listener is added
+            mQuery.addChildEventListener(this);
+            mQuery.addValueEventListener(this);
+        } else {
+            for (int i = 0; i < size(); i++) {
+                listener.onChildChanged(ChangeEventListener.EventType.ADDED, i, -1);
```
samtstern: Should this be `-1` or `i` since the new children are presumably added to the end?

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
```diff
@@ -14,52 +14,137 @@
...
+     *
+     * @param listener the listener to be called with changes
+     * @return a reference to the listener provided. Save this to remove the listener later
+     * @throws IllegalArgumentException if the listener is null
+     */
+    public SubscriptionEventListener addSubscriptionEventListener(@NonNull SubscriptionEventListener listener) {
+        checkNotNull(listener);
+
+        for (SubscriptionEventListener ignored : mSubscribers) {
+            listener.onSubscriptionAdded();
```
samtstern: So what happens here is when we add a listener we tell it about all existing subscribers, but we don't tell existing subscribers about the new listener?  Again I think I am missing the overall concept because I find this block confusing.

database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
```diff
@@ -75,39 +78,39 @@ private int getIndexForKey(String key) {
...
 
     private boolean isKeyAtIndex(String key, int index) {
-        return index >= 0 && index < getCount() && mDataSnapshots.get(index).getKey().equals(key);
+        return index >= 0 && index < size() && mDataSnapshots.get(index).getKey().equals(key);
     }
 
     @Override
     public void onChildAdded(DataSnapshot keySnapshot, String previousChildKey) {
-        super.setOnChangedListener(null);
+        setShouldNotifyListeners(false);
```
samtstern: I don't have the solution, but I always felt that the "remove listener, do operation, add listener" pattern was code smell.  Can we do anything cleaner while we're already in here making all these API changes?  `shouldNotifyListeners` is not much of an improvement (although it's definitely something).

database/src/main/java/com/firebase/ui/database/FirebaseArrayOfObjects.java
SUPERCILEX: @samtstern Yeah, I was trying to be smart here by auto unlistening when the user removed their own listeners, but this code doesn't work. (we only stop listening if the list of listeners is empty, but `FirebaseArrayOfObjectsOptimized` prevents that from ever happening because it adds its own listener) I would need to get the number of listeners and see if they are equal to 1 for `onSubscriptionRemoved()` to work. Anyway, I decided to scrap this and make the standard api be `FirebaseArray#removeAllListeners()` in `onDestoy()`. It makes more sense for safety reasons because it ensure people can guarantee that they've cleaned everything up.

database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
SUPERCILEX: @samtstern Yeah, I would love to improve that too, but I'm not sure how to do so. See the original discussion: https://github.com/firebase/FirebaseUI-Android/pull/276#pullrequestreview-503792. The way I did it originally was to override the `notify...` methods and make them noops, but @MaciejCiemiega pointed out that if devs were to extend `FirebaseIndexArray`, they would expect `notify...` to actually do something. If you can think of another way to solve this problem, I would love to hear it!

database/src/main/java/com/firebase/ui/database/FirebaseArrayOfObjects.java
```diff
@@ -0,0 +1,212 @@
...
+        return mSnapshots.toString();
+    }
+
+    protected static class FirebaseArrayOfObjectsOptimized<E> extends FirebaseArrayOfObjects<E>
+            implements ChangeEventListener {
+        protected List<E> mObjects = new ArrayList<>();
+
+        public FirebaseArrayOfObjectsOptimized(FirebaseArray snapshots, Class<E> modelClass) {
+            super(snapshots, modelClass);
+            snapshots.addChangeEventListener(this);
```
SUPERCILEX: @samtstern this really bothers me because users aren't going to expect it.

I'm thinking of adding a `/* default */ List<ChangeEventListener> mInternalListeners` field that acts like a shadow and doesn't affect listening for db changes. What are your thoughts on this?

SUPERCILEX:
@samtstern I just finished integrating all the new features of this PR into my app and I'm not seeing any noticeable bugs in my testing (apart from 1c31e16 which I fixed). Unless you have more feedback for me to address, there's only https://github.com/firebase/FirebaseUI-Android/pull/544#issue-203268560 and https://github.com/firebase/FirebaseUI-Android/pull/544#discussion_r102144473 to work on.

samtstern:
@SUPERCILEX sorry about the silence on my end.  I have actually been spending a fair amount of time staring at this PR, and yes you are right those are the only outstanding issues from the discussions we've had.  Thanks for continuing to test this!

@puf posed an interesting question to me yesterday: if we had been interested in doing all this refactoring months ago (before `FirebaseIndexArray` or `ChangeEventListener` existed) what would `FirebaseArray` look like?  Would it be the same as this proposal?

Since 2.0 isn't going to come for a little bit (need to let `1.2.0` soak and probably bundle in `Auth` re authentication as well) I am going to keep pondering that question.

samtstern:
@SUPERCILEX don't read too far into this but before I head out for the weekend wanted to at least make a note about what I was working on with respect to this PR.  I almost feel bad sharing something this shaky, but better than silence!

I'm exploring on this branch:
https://github.com/samtstern/FirebaseUI-Android/tree/firebase-array-explore

Some things my changes improve:
  * No more `FirebaseArrayOfObjects` or its `Optimized` cousin
  * No more `shouldNotifyListeners`
  * Unifies `FirebaseArray` / `FirebaseIndexArray` by what they are (observable arrays of DataSnapshots) rather than one being a super class of the other

Still got a long way to go (docs, tests, discussion, etc) before it's as solid as what you have here, but it *feels* like the right path since there are fewer hacks and it moves responsibilities to the right place.

SUPERCILEX:
@samtstern Hmmm, your branch looks really interesting! The gist I got from it was to make the model class a first class citizen instead of `DataSnapshot` which makes way more sense. I don't really like that you have a `getObject` instead of `getSnapshot` method thus moving the model class a little further back. What do you think of rewriting this PR with your changes, but implementing `List<E>` in `ObservableSnapshotArray` instead of `List<DataSnapshot>`? (We could call it `ObservableModelArray<T>` or something better.)

puf:
Agreed with Alex that this looks promising. I only looked at
`FirebaseArray` for now, so still need to see how the extensions (such as
the indexed array or a re-sorted array) would work.

I'd prefer to not expose too many lists though. If it exposes a `List<E>`,
there should also be a `List<String>` with the keys. And when you have both
of those, they're really just syntactic sugar for `List<DataSnapshot>`.
Nothing wrong with syntactic sugar, but three lists feels like a bit much.

samtstern:
Opened https://github.com/SUPERCILEX/FirebaseUI-Android/pull/24

SUPERCILEX:
@samtstern The index adapters were completely broken because we forgot to pass in a parser. Since I feel like this could be a recurring theme, I just removed the possibility of a null parser in 7601ded. If devs really don't want to provide a model class for some reason, they can just pass in `Object.class` and it will return an object:
```java
var1.equals(Object.class)?var0:(var1.isEnum()?zzd(var0, var1):zze(var0, var1));
```

I also had to write a hacky constructor that really bothers me, but I couldn't think of a better way of doing it that would still let devs extend the adapters while passing in their own `FirebaseArray`. What do you think of 7601ded?

samtstern:
@SUPERCILEX 

1. I am fine with removing the `null` parser option, like you did.
2. Can you explain a little more what is enabled by adding the empty constructors for the adapters?  I think I am missing something.

SUPERCILEX:
@samtstern 

1. Awesome!
2. Java forces you to either have a default constructor (the option I opted for) or to call super in the constructor you're extending.

```java
public class First {
    public First(String foo) {}
}

private class Second extends First {
    public Second() {} // Doesn't compile

    public Second(String bar) { super(bar); } // Compiles
}
```

Since we also can't pass `this` into a super call, we're stuck again.

database/src/main/java/com/firebase/ui/database/FirebaseIndexListAdapter.java
```diff
@@ -3,57 +3,23 @@
...
      */
     public FirebaseIndexListAdapter(Activity activity,
                                     Class<T> modelClass,
                                     @LayoutRes int modelLayout,
-                                    Query keyRef,
-                                    Query dataRef) {
-        super(activity, modelClass, modelLayout, new FirebaseIndexArray(keyRef, dataRef));
+                                    Query keyQuery,
+                                    DatabaseReference dataRef) {
+        init(activity, new FirebaseIndexArray<>(keyQuery, dataRef, this), modelClass, modelLayout);
```
samtstern: Hmm now I see what happened here.  Having `init` be protected and called by the super class definitely "smells" and makes me think we got something wrong.

Seems like the problem is the circular reference between the adapter and the array, even though it goes through the `SnapshotParser` interface.

Any thoughts on how we can make this better?

SUPERCILEX: @samtstern Hmmm, I found out that you can't reference non-static fields or methods either in a constructor call so that means no forwarding calls or anything like that. At this point, the only solution I can think of is using a setter for the parser, but I feel like that's way more hacky and makes the intentions less clear. Also, that beats the point of initializing everything for people because they will have to call setParser. @puf Have you encountered this kind of issue before or have any thoughts on this?

samtstern: Here's a possible answer.  Since we're breaking parseSnapshot anyway, what if we removed it from the adapter entirely and instead had the adapters take an optional `SnapshotParser` in their constructor.  They would pass it through to the `FirebaseArray` if provided.

This seems to be a better representation of the hierarchy anyway,

SUPERCILEX: @samtstern As long as you're fine with making more breaking api changes, that solution sounds awesome to me!

SUPERCILEX:
@samtstern I'm getting this really weird crash because `ChildEventListener`s are firing even though we removed the listener. I checked and there is only one pair of add/removing being called so that's not the issue. At this point, I think it's a Firebase problem, not this PR.

Here are some screenshots showing that even though `removeEventListener` was called, `onChildMoved` still gets called:

image

image

samtstern:
@SUPERCILEX looks like maybe it's how we're casting?  We're calling `removeEventListener(this)` twice but maybe we need `removeEventListener((ChildEventListener) this); removeEventListener((ValueEventListener) this);`

SUPERCILEX:
@samtstern I was confused by that too, but I think it's just a decompiling bug (java doesn't let you do what's in the screenshot anyway--I checked). If you look at our real code, we are casting things correctly: https://github.com/firebase/FirebaseUI-Android/pull/544/files#diff-fd5c505ad432166321255bc4541624ddR89 so I'm pretty sure this is an issue with the RTDB.

samtstern:
@SUPERCILEX yep you're right, it's not the casting issue.

I just took a look at this and (I can't believe I'm saying it) but I'm ready to merge this beast!  Could you file a new issue about the possible listener removal race condition thing?  Definitely worth tracking but it does not seem to be a direct result of this PR.

Do you have any oppositions to my doing a 'squash and merge' on this PR?

samtstern:
giphy

SUPERCILEX:
@samtstern AWESOME!!!!

Yeah, I already filed a bug for the slow listener removal.

No objections at all, please don't do anything but a squash. There are way too many commits in this PR!

SUPERCILEX:
Haha, nice gif!

PR #550, opened 2017-01-26
Update docs stating the need to call cleanup

Maragues:
See https://github.com/firebase/FirebaseUI-Android/issues/244

Not invoking cleanup causes a memory leak, and it isn't mentioned.


Maragues:
I signed it!

samtstern:
@Maragues LGTM, thanks!

Don't worry about the Travis build, it's a flaky lint thing and I just pushed a commit to stop it happening in the future.

PR #551, opened 2017-01-27
Add documentation for setAllowNewEmailAccounts and teeny cleanup

SUPERCILEX:
@samtstern Moved the `setAllowNewEmailAccounts` method to a nicer position and added documentation.

Updated butterknife: https://github.com/JakeWharton/butterknife/blob/master/CHANGELOG.md#version-851-2017-01-24.

Removed `otherwise = VisibleForTesting.PRIVATE` because I found out it was the default for `@VisibleForTesting()`.


samtstern:
:+1:

Issue #549, opened 2017-01-26
AuthUI flow starts multiple times when rotated and then crashes when selecting an account for each instance


oviroa:
### Step 2: Describe your environment

  * Android device: Pixel XL
  * Android OS version: 7.1.1
  * Google Play Services version: 10.0.84
  * Firebase/Play Services SDK version: 10.0.1
  * FirebaseUI version: 1.1.1
  
### Step 3: Describe the problem:
Account selector UI that overlays on top of the root UI gets created multiple times, once with every orientation change rotation. Then, I have to tap on the desired account once for each instance, as they disappear one by one, and after the last one, app crashes:

01-26 03:14:10.381 20689-20689/oviroa.bestshot.android E/UncaughtException: java.lang.NullPointerException: Attempt to invoke virtual method 'void android.app.Activity.setResult(int, android.content.Intent)' on a null object reference
                                                                                at com.firebase.ui.auth.ui.BaseHelper.finishActivity(BaseHelper.java:51)
                                                                                at com.firebase.ui.auth.ui.FragmentHelper.finish(FragmentHelper.java:27)
                                                                                at com.firebase.ui.auth.ui.FragmentBase.finish(FragmentBase.java:26)
                                                                                at com.firebase.ui.auth.util.signincontainer.SaveSmartLock.finish(SaveSmartLock.java:178)
                                                                                at com.firebase.ui.auth.util.signincontainer.SaveSmartLock.saveCredentialsOrFinish(SaveSmartLock.java:198)
                                                                                at com.firebase.ui.auth.ui.BaseHelper.saveCredentialsOrFinish(BaseHelper.java:124)
                                                                                at com.firebase.ui.auth.ui.BaseHelper.saveCredentialsOrFinish(BaseHelper.java:112)
                                                                                at com.firebase.ui.auth.ui.idp.CredentialSignInHandler.onComplete(CredentialSignInHandler.java:68)
                                                                                at com.google.android.gms.tasks.zzc$1.run(Unknown Source)
                                                                                at android.os.Handler.handleCallback(Handler.java:751)
                                                                                at android.os.Handler.dispatchMessage(Handler.java:95)
                                                                                at android.os.Looper.loop(Looper.java:154)
                                                                                at android.app.ActivityThread.main(ActivityThread.java:6119)
                                                                                at java.lang.reflect.Method.invoke(Native Method)
[...]

PR #552, opened 2017-01-27
Fix startLogin being called on every configuration change with only one provider enabled

SUPERCILEX:
#549

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/util/signincontainer/IdpSignInContainer.java
+++ b/auth/src/main/java/com/firebase/ui/auth/util/signincontainer/IdpSignInContainer.java
@@ -116,7 +116,16 @@ public class IdpSignInContainer extends FragmentBase implements IdpCallback {
         }
 
         mIdpProvider.setAuthenticationCallback(this);
-        mIdpProvider.startLogin(getActivity());
+
+        if (savedInstanceState == null) {
+            mIdpProvider.startLogin(getActivity());
+        }
+    }
+
+    @Override
+    public void onSaveInstanceState(Bundle outState) {
+        outState.putBoolean(ExtraConstants.HAS_EXISTING_INSTANCE, true);
+        super.onSaveInstanceState(outState);
     }
 
     @Override


samtstern:
LGTM, but can you quickly explain how to reproduce the intial bug?  I didn't really understand the report in #549

SUPERCILEX:
@samtstern Of course!

[Using a single provider with or without smart lock eventually calls `redirectToIdpSignIn`](https://github.com/firebase/FirebaseUI-Android/blob/master/auth/src/main/java/com/firebase/ui/auth/util/signincontainer/SignInDelegate.java#L315) -> [an instance of `IdpSignInContainer` is created with `setRetainInstance(false)` (by default)](https://github.com/firebase/FirebaseUI-Android/blob/master/auth/src/main/java/com/firebase/ui/auth/util/signincontainer/SignInDelegate.java#L329) -> [on every screen rotation, `startLogin` gets called](https://github.com/firebase/FirebaseUI-Android/blob/master/auth/src/main/java/com/firebase/ui/auth/util/signincontainer/IdpSignInContainer.java#L119). The reason we don't see this bug anywhere else in FirebaseUI is because stuff like `AuthMethodPickerActivity` doesn't use `IdpSignInContainer`. `IdpSignInContainer`'s only usage is for `SignInDelegate`.

SUPERCILEX:
@samtstern I just realized you might have been asking for a non-technical explanation for which the answer would be: start login with a single provider -> get to the provider's login section (eg not the smart lock thing) -> rotate the screen.

samtstern:
@SUPERCILEX yep the second one was what I wanted, confirmed the bug.

samtstern:
@SUPERCILEX https://travis-ci.org/firebase/FirebaseUI-Android looks like the build is failing with a lint error here but I don't get any on my machine.  Can you reproduce?

SUPERCILEX:
@samtstern Travis says it's because support lib 25.1.1 came out https://developer.android.com/topic/libraries/support-library/revisions.html#25-1-1

I'll create an update PR later today since it's just bug fixes (and I'll also disable `GradleDependency` to make our builds less flaky like in #542).

PR #559, opened 2017-01-31
Update everything, remove deprecated stuff from updates, make Travis builds less flaky, and update storage sample

SUPERCILEX:
@samtstern Please don't merge yet, I want to do a wee bit of testing with Facebook.


auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -347,9 +346,7 @@ public Void then(@NonNull Task<GoogleApiClient> task) throws Exception {
                 });
 
         // Facebook sign out
-        if (FacebookSdk.isInitialized()) {
```
samtstern: Did something change the Facebook SDK that we no longer need to deal with initialization?

SUPERCILEX: @samtstern Since you noticed, I might as well explain it now instead of in my review. The easiest would be to read Facebook's release notes: https://developers.facebook.com/docs/android/change-log-4x#4_19_0 and the upgrade guide https://developers.facebook.com/docs/android/upgrading-4x#4180to4190.

samtstern: SGTM

SUPERCILEX:
@samtstern If everything looks good to you, both Facebook login and the storage permission stuff work great so this should be good to merge once the Travis build passes.

SUPERCILEX:
@samtstern Urgh. It should be fixed in https://github.com/firebase/FirebaseUI-Android/pull/559/commits/f501d20f45320abb64cc9be3a970f6306666cb26 though!

auth/src/main/AndroidManifest.xml
```diff
@@ -19,6 +19,10 @@
             android:name="com.facebook.sdk.ApplicationId"
             android:value="@string/facebook_application_id"/>
 
+        <meta-data
```
samtstern: @SUPERCILEX Could we be inadvertently disabling something in developer's apps?  What if they have a conflicting value in their `AndroidManifest`?

SUPERCILEX: @samtstern cc4b64b is the best solution I can come up with. Since we are upgrading the Facebook sdk for people, they will expect the Facebook upgrade guide (remove the `initializeApp`) to work. With this latest commit, that will still work, but people will have to override our manifest if they want to disable it:
```xml
<meta-data
    android:name="com.facebook.sdk.AutoLogAppEventsEnabled"
    android:value="false"
    tools:replace="android:value"/>
```

auth/build.gradle
```diff
@@ -15,9 +15,15 @@ android {
...
     buildTypes {
         release {
+            manifestPlaceholders = [enableFbLogging: true]
+
             minifyEnabled false
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
         }
+
+        debug {
+            manifestPlaceholders = [enableFbLogging: false]
```
samtstern: This is fine, just add a comment here saying this is only for testing.  Since in the real world library modules are always compiled in their 'release' variant.  So our end users will see `enableFbLogging=true` always

samtstern:
@SUPERCILEX the manifest placeholders may not be working in the Robolectric tests, the tests are unhappy again.

SUPERCILEX:
@samtstern Yeah, I noticed. I thought tests ran in debug mode, but I guess not. I think I'll just have to copy the manifest for the tests.

SUPERCILEX:
@samtstern I figured it out: `./gradlew check` runs `testReleaseUnitTest`. Now I just have to remove that task from check and we should be good.

SUPERCILEX:
@samtstern It's working locally, will push and we should be good to go hopefully.

samtstern:
Merging.  In the future we should see if we can get release tests running.

SUPERCILEX:
@samtstern Totally agree with you, created https://github.com/facebook/facebook-android-sdk/pull/494 to address this issue.

Issue #522, opened 2017-01-13
Support landscape and tablet layouts on all screens


samtstern:
In particular, the `auth_method_picker_layout.xml` looks bad when in landscape or when on a large tablet screen.

Other screens are functional but could look nicer.

samtstern:
Tried to fix landscape in #515 but introduced bug #521.

aviadmini:
Non-tablet layout in landscape with logo looks pretty ugly too.

Would it be possible to use custom layouts there? (with a requirement to contain button holder layout maybe)

samtstern:
@aviadmini yes agreed the non-tablet layout in landscape looks horrible.  We will fix this in `1.2.0`.

consp1racy:
NestedScrollView with fillViewport I presume.

samtstern:
Assigning this to myself, so if anyone else is working on this or was planning on starting it please contact me

samtstern:
In #562 a fix was put in for the picker Activity.  I believe that's the only one that was totally broken in landscape.

SUPERCILEX:
@samtstern Looks like https://github.com/firebase/FirebaseUI-Android/issues/375 is a duplicate of this issue.

samtstern:
The fix for this issue has been released in version `1.2.0`.

PR #562, opened 2017-01-31
Make auth_method_picker layout work in landscape

samtstern:
Issue #522 

Landscape:
pick_landscape

Portrait:
pick_portrait


samtstern:
@SUPERCILEX take a look if you can

SUPERCILEX:
@samtstern LGTM! Though isn't there a way we can use the include tag to do less copy pasting?

PS: I'm looking into the manifest thing for #559 so the Travis build will be happy after that.

SUPERCILEX:
@samtstern Looks like you need to merge from 1.2 dev to make the Travis builds pass.

Issue #528, opened 2017-01-18
Cannot sign in via email on a device without a Google account


consp1racy:
* Android device: Genymotion Google Nexus 6P
  * Android OS version: 7.0.0
  * Google Play Services version: 10.0.84
  * Firebase/Play Services SDK version: 10.0.1
  * FirebaseUI version: 1.1.1
  
On a device where the user has no Google account yet the user cannot sign in via email address already associated with a Google account.

  1. Use AuthUI to sign in via Google account.
  2. Reset the device so it is not logged in to Google / Get another device that's not signed in.
  3. Use AuthUI to sign in via email address.
  4. Put in the address of previously used Google account and hit next.

#### Observed Results:

  5. Be redirected to Google sign in flow because "you've already used --insert email--".
  6. Fail with "An error has occurred." Toast and the following message in LogCat:
  
    E/GoogleProvider: Error logging in with Google. 5 null

#### Expected Results:

Start the GMS error resolution (Google system login flow) as does directly using the "Sign in with Google" button and actually get into the "Checking credentials" and "Add your account" screens.

aviadmini:
Is the email address a gmail address?

samtstern:
@consp1racy I was able to reproduce this, thanks for pointing it out.  Will try to get a fix in soon.

samtstern:
Ok I figured out where this occurs.  When `AuthMethodPickerActivity` creates `GoogleProvider` it does so with `email = null`.  When `WelcomeBackIdpPrompt` does this it sets `email` equal to the email entered from the email linking flow.

It seems that the Google Sign In API does not launch the add-account-to-device flow when `setAccountName` is called on the `GoogleApiClient.`

I will bring this up with the Google Sign In team to see what the recommended path forward is.

samtstern:
@consp1racy I have brought this up with the Sign In team and, for now, there's nothing FBUI can do.  There is no way for us to launch the "add account" flow with a specific email in mind.  That's why right now we return this error, as there's no way for us to trigger the proper google sign in flow and guarantee it's for the right account.

Hopefully the case of "user signs in with Google on device A and then attempts to sign in with email on device B where the google account is not present" is a very small edge case (I believe it is).

So as far as FBUI is concerned, this is working as intended and I will close the issue.  I will log a feature request with the Android accounts team to see if we can enable this case.

consp1racy:
[...]

PR #565, opened 2017-02-01
Redirect user to add account if user removed Google account

SUPERCILEX:
@samtstern Since it's been awhile since you self assigned #528, I took it on. Hope you don't mind! 😄


auth/src/main/java/com/firebase/ui/auth/provider/GoogleProvider.java
```diff
@@ -138,21 +143,26 @@ public void startLogin(Activity activity) {
...
 
     private void onError(String errorMessage) {
         Log.e(TAG, "Error logging in with Google. " + errorMessage);
         Bundle extra = new Bundle();
         extra.putString(ERROR_KEY, errorMessage);
-        mIDPCallback.onFailure(extra);
-    }
-
-    @Override
-    public void onClick(View view) {
```
SUPERCILEX: This is the second time I've removed an unused click listener!

SUPERCILEX:
@samtstern I just merged and I don't believe this PR has any huge mistakes. 😄

Issue #561, opened 2017-01-31
Memory leak on AuthUI.getInstance()#signOut(Activity)


pedrodanielcsantos:
### Describe your environment

  * Android device: Samsung Galaxy S6 
  * Android OS version: 6.0.1
  * Google Play Services version: 10.0.1
  * Firebase/Play Services SDK version: 10.0.1
  * FirebaseUI version: got it both on 1.0.1 and on 1.1.1
  
### Describe the problem:

After invoking `AuthUI.getInstance.signOut(getActivity())`from a fragment, LeakCanary reported a leak of the parent activity as it can be seen in the leak canary's report screenshot. 

![leak](https://cloud.githubusercontent.com/assets/8170256/22470153/443201ce-e7c6-11e6-802f-43706f86323d.png)

#### Steps to reproduce:

  1. (In our case, we authenticated using the GoogleProvider directly instead of the wrapper. We're using the wrappers directly. [Reference](https://github.com/firebase/FirebaseUI-Android/issues/490) );
  2. Invoke `AuthUI.getInstance.signOut(getActivity())` from a fragment and redirect the user to a new activity (for instance to reauthenticate again);
  3. Wait for `LeakCanary` to run and report the leak;
  
#### Observed Results:

Printscreen of the leak above ( _Describe the problem_ section).

SUPERCILEX:
@samtstern, I've noticed this leak several times, but I don't understand why it's happening since we remove the activity from `INSTANCES`. It's something we either have to rewrite or spend some quality time debugging. :smile:

samtstern:
@pedrodanielcsantos thanks for this report, we'll see what we can do to hunt this down

pedrodanielcsantos:
Thanks a lot @samtstern @SUPERCILEX ! If I can help in some way, just give me a shout 👍

samtstern:
The fix for this issue has been released in version `1.2.0`.

PR #566, opened 2017-02-01
Completely rewrite CredentialsApiHelper.java and hopefully fix memory leak

SUPERCILEX:
@samtstern This should fix the memory leak because we aren't storing the activity anywhere.

#561


auth/src/main/java/com/firebase/ui/auth/util/GoogleApiHelper.java
```diff
@@ -0,0 +1,13 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class GoogleApiHelper {
+    private static final List<Void> IDS = new ArrayList<>();
+
+    public static int getSafeAutoManageId() {
+        IDS.add(null);
+        return IDS.size() - 1;
```
SUPERCILEX: This is sooooo much better than my previous solution. 😄

auth/src/main/java/com/firebase/ui/auth/util/OldCredentialsApiHelper.java
```diff
@@ -0,0 +1,101 @@
...
+import com.google.android.gms.tasks.Continuation;
+import com.google.android.gms.tasks.Task;
+import com.google.android.gms.tasks.TaskCompletionSource;
+
+/**
+ * A {@link com.google.android.gms.tasks.Task Task} based wrapper for the Smart Lock for Passwords
+ * API.
+ */
+@Deprecated
+public class OldCredentialsApiHelper {
```
SUPERCILEX: You don't need to look here, I just copied this from the old `CredentialsApiHelper`.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -313,10 +315,12 @@ public static int getDefaultTheme() {
...
-     * @return a task which, upon completion, signals that the user has been signed out
-     * ({@code result.isSuccess()}, or that the sign-out attempt failed unexpectedly
-     * ({@code !result.isSuccess()}).
+     * @return a task which, upon completion, signals that the user has been signed out ({@code
+     * result.isSuccess()}, or that the sign-out attempt failed unexpectedly ({@code
+     * !result.isSuccess()}).
+     * @deprecated use {@link #signOut(FragmentActivity)} instead
      */
+    @Deprecated
     public Task<Void> signOut(@NonNull Activity activity) {
```
SUPERCILEX: This makes me so happy: because our new `signOut` method uses `FragmentActivity` which extends `Activity`, people will be automatically upgraded. How's that for customer service! 😄

auth/src/main/java/com/firebase/ui/auth/util/GoogleApiHelper.java
```diff
@@ -0,0 +1,13 @@
+package com.firebase.ui.auth.util;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class GoogleApiHelper {
+    private static final List<Void> IDS = new ArrayList<>();
```
samtstern: I think this would do better as an `AtomicInteger` and then your static function would just be wrapping `incrementAndGet()`

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -359,7 +363,9 @@ public Void then(@NonNull Task<GoogleApiClient> task) throws Exception {
      * silently.
      *
      * @param activity the calling {@link Activity}.
+     * @deprecated use {@link #delete(FragmentActivity)} instead
      */
+    @Deprecated
     public Task<Void> delete(@NonNull Activity activity) {
```
samtstern: As far as I can tell the difference between `delete(FragmentActivity)` and `delete(Activity)` is just that the former uses `CredentialApiHelper` and the latter uses `OldCredentialApiHelper`.

Could this all be simplified by having both of the `CredentialApiHelper` classes implement a common interface?  I think it would look like this:

```
interface CredentialApiHelper {

  Task<Status> delete(Credential credential);

  Task<Status> disableAuthSignIn();

}
```

And then these two `delete()` functions would just be a matter of constructing the proper `Helper` and then calling to a private `Task<Void> delete(CredentialApiHelper helper)`

auth/src/main/java/com/firebase/ui/auth/util/OldCredentialsApiHelper.java
samtstern: Let's pick a better name for this, maybe `ActivityCredentialsApiHelper` or something.  "Old" as a prefix does not help to show what was wrong with it.

Also I'd argue that we should keep this class named `CredentialsApiHelper` and the new class is the one that should have a new name since it reduces churn.

auth/src/main/java/com/firebase/ui/auth/util/GoogleApiHelper.java
SUPERCILEX: @samtstern Never knew that existed, sounds awesome!

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
SUPERCILEX: @samtstern To address this comment and [this one](https://github.com/firebase/FirebaseUI-Android/pull/566#discussion_r98960229), I didn't add an interface or a good name because my intention for us to throw this away in FBUI `v2.0`. Do you disagree with that? If so, then we should definitely clean things up and maybe undeprecate thoses classes. What do you think?

samtstern: 2.0 could be a while away, let's try and keep things clean in the meantime.

SUPERCILEX: @samtstern SGTM, I'll leave things deprecated so we know to remove that stuff if we go that route, but I keep things clean in the meantime.

SUPERCILEX:
@samtstern I'm happy with it now, PTAL.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -373,7 +380,84 @@ public Void then(@NonNull Task<GoogleApiClient> task) throws Exception {
         // Initialize SmartLock helper
         GoogleApiClientTaskHelper gacHelper = GoogleApiClientTaskHelper.getInstance(activity);
         gacHelper.getBuilder().addApi(Auth.CREDENTIALS_API);
-        CredentialsApiHelper credentialHelper = CredentialsApiHelper.getInstance(gacHelper);
+        CredentialApiHelper credentialHelper = CredentialsApiHelper.getInstance(gacHelper);
+
+        // Get all SmartLock credentials associated with the user
+        List<Credential> credentials = SmartLockBase.credentialsFromFirebaseUser(firebaseUser);
```
samtstern: From here down to the end of the function seems duplicated between the two `delete()` methods.  Can we abstract into a common private method?

SUPERCILEX: Yup, forgot to refactor that.

auth/src/main/java/com/firebase/ui/auth/util/CredentialsApiHelper.java
```diff
@@ -31,7 +31,8 @@
  * A {@link com.google.android.gms.tasks.Task Task} based wrapper for the Smart Lock for Passwords
  * API.
  */
-public class CredentialsApiHelper {
+@Deprecated
+public class CredentialsApiHelper implements CredentialApiHelper {
```
samtstern: One last (annoying) comment: it's pretty confusing to have these two names only differ by an 's' in the middle.  Let's change the name of the interface?  Maybe `CredentialTaskApi` or something like that.

SUPERCILEX: @samtstern SGTM, I just can't wait for us to throw this stuff away anyway.

auth/src/main/java/com/firebase/ui/auth/util/GoogleSignInHelper.java
```diff
@@ -0,0 +1,64 @@
...
+    public static GoogleSignInHelper getInstance(FragmentActivity activity) {
+        GoogleApiClient.Builder builder = new GoogleApiClient.Builder(activity)
+                .addApi(Auth.CREDENTIALS_API)
+                .addApi(Auth.GOOGLE_SIGN_IN_API, GoogleSignInOptions.DEFAULT_SIGN_IN);
+        return new GoogleSignInHelper(activity, builder);
+    }
+
+    public Task<Status> signOut() {
+        final TaskCompletionSource<Status> statusTask = new TaskCompletionSource<>();
+        getConnectedApiTask().addOnSuccessListener(new OnSuccessListener<Bundle>() {
```
samtstern: Missed this, but shouldn't we also add an `onFailureListener` to `getConnectedApiTask()` that propagates the failure to `statusTask` through `setException`?

auth/src/main/java/com/firebase/ui/auth/util/signincontainer/SignInDelegate.java
```diff
@@ -299,11 +298,11 @@ private void deleteCredentialAndRedirect() {
             return;
         }
 
-        CredentialsApiHelper credentialsApiHelper = CredentialsApiHelper.getInstance(getActivity());
-        credentialsApiHelper.delete(mCredential)
-                .addOnCompleteListener(new OnCompleteListener<Status>() {
+        AuthUI.getInstance()
+                .delete(getActivity())
```
samtstern: This now also deletes the underlying Firebase Auth account in addition to deleting the credentials.  This has two new properties:

- Will fail if there is no currently signed in user
- Will fail if we don't have recent authentication (last 5 minutes)

Did you intend to change this behavior?

SUPERCILEX:
@samtstern Sorry about those mistakes, please let me take another look at this PR to make sure I didn't mess up again. And maybe you could take another look too? (I've been distracted by my other PRs 😄)

SUPERCILEX:
@samtstern Ok, I think it's all good now.

auth/src/main/java/com/firebase/ui/auth/util/GoogleSignInHelper.java
```diff
@@ -26,7 +25,7 @@ public static GoogleSignInHelper getInstance(FragmentActivity activity) {
 
     public Task<Status> signOut() {
         final TaskCompletionSource<Status> statusTask = new TaskCompletionSource<>();
-        getConnectedApiTask().addOnSuccessListener(new OnSuccessListener<Bundle>() {
+        getConnectedApiTask().addOnSuccessListener(new ExceptionForwarder<Bundle>(statusTask) {
```
samtstern: I don't think this will work, it's still setting the `ExceptionForwarder` using `addOnSuccessListener` so even though it implements `onFailure` that will never be called.

You may want to make this an `onCompleteListener` instance instead so you'll get notified of success or failure.

auth/src/main/java/com/firebase/ui/auth/util/GoogleSignInHelper.java
```diff
@@ -0,0 +1,63 @@
...
+    public static GoogleSignInHelper getInstance(FragmentActivity activity) {
+        GoogleApiClient.Builder builder = new GoogleApiClient.Builder(activity)
+                .addApi(Auth.CREDENTIALS_API)
+                .addApi(Auth.GOOGLE_SIGN_IN_API, GoogleSignInOptions.DEFAULT_SIGN_IN);
+        return new GoogleSignInHelper(activity, builder);
+    }
+
+    public Task<Status> signOut() {
+        final TaskCompletionSource<Status> statusTask = new TaskCompletionSource<>();
+        getConnectedApiTask().addOnSuccessListener(new ExceptionForwarder<Bundle>(statusTask) {
```
samtstern: I don't think this will work, it's still setting the ExceptionForwarder using addOnSuccessListener so even though it implements onFailure that will never be called.

You may want to make this an onCompleteListener instance instead so you'll get notified of success or failure.

auth/src/main/java/com/firebase/ui/auth/util/GoogleSignInHelper.java
SUPERCILEX: @samtstern Today's not my day with tasks! 😵 PTAL at 15fbed3, I'm pretty sure that will work.

samtstern:
Woohoo looks good!  I will merge once tests pass.

SUPERCILEX:
@samtstern OMG finally! Thanks for sticking with me on this one! 😄

PR #567, opened 2017-02-01
Update documentation and sample for #544

SUPERCILEX:
This PR is built on top of #544 and updates documentation and the sample with the changes from #544. It should be merged only if #544 is merged.

- [x] Update README links in new adapters


samtstern:
@SUPERCILEX this one's next!

app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
```diff
@@ -131,26 +136,50 @@ public void onAuthStateChanged(@NonNull FirebaseAuth firebaseAuth) {
     }
 
     private void attachRecyclerViewAdapter() {
-        Query lastFifty = mChatRef.limitToLast(50);
-        mAdapter = new FirebaseRecyclerAdapter<Chat, ChatHolder>(
-                Chat.class, R.layout.message, ChatHolder.class, lastFifty) {
+        mAdapter = new FirebaseIndexRecyclerAdapter<Chat, ChatHolder>(
```
SUPERCILEX: I have a really strong preference to keeping this since I was able to find issues with the index stuff by testing with this branch.

app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
```diff
@@ -131,26 +136,50 @@ public void onAuthStateChanged(@NonNull FirebaseAuth firebaseAuth) {
...
+            public void populateViewHolder(final ChatHolder holder, Chat chat, int position) {
+                holder.bind(chat);
+
+                holder.itemView.setOnCreateContextMenuListener(new View.OnCreateContextMenuListener() {
+                    @Override
+                    public void onCreateContextMenu(ContextMenu menu,
+                                                    View v,
+                                                    ContextMenu.ContextMenuInfo menuInfo) {
+                        menu.add("Delete")
+                                .setOnMenuItemClickListener(new MenuItem.OnMenuItemClickListener() {
```
SUPERCILEX: Should we keep this? (I just did it for testing, but it's kinda nice)

SUPERCILEX:
@samtstern Merged!

samtstern:
@SUPERCILEX thanks for merging!  To be honest I have not looked at this one in many moons (was focusing on #544) so I will give it a pass in the morning.  I don't expect this to take more than one round of back-and-forth before merging too, thanks for keeping this one up to date.

app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
samtstern: I like the idea but the ratio of "lines of code" to "how much it teaches you about FirebaseUI" may be too low here.  If you don't have a strong preference I'd say remove it to keep `ChatActivity` as short as possible.

SUPERCILEX: SGTM!

app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
samtstern: While I definitely appreciate your constant testing of the indexed bits, the purpose of `ChatActivity` is to show off the basic FBUI integration, and I think `FirebaseRecyclerAdapter` is the most popular.  People are likely to copy and paste from this class and I want them to get the least surprising thing.

Is there somewhere else we can put the indexed code?  Maybe in another Activity that's not exposed in the main chooser?

SUPERCILEX: @samtstern Hmmm, I didn't think about people copypasta-ing code straight from `ChatActivity`... If that's the case, then I agree with you that the code should be as simple as possible and I like the idea of a separate activity for the index stuff. What do you think of 758b69239048050f163956c3c1e56b1e12edd76d?

samtstern:
LGTM!  Gonna squash and merge.

SUPERCILEX:
Awesome!

PR #569, opened 2017-02-02
Remove in-depth documentation for AuthUI.java and link to README instead

SUPERCILEX:
@samtstern This change is up to debate, but I've noticed that we rarely update the `AuthUI.java` documentation and I feel like it would make it easier for us to directly link users to the README giving them the most up to date documentation and promoting awareness of our sub-READMEs.

@samtstern What do you think? The only con I can think of is that devs on an airplane without wifi might not be happy. 😄

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/AuthUI.java
+++ b/auth/src/main/java/com/firebase/ui/auth/AuthUI.java
@@ -66,7 +66,7 @@ import java.util.Set;
  * {@link AuthUI#getInstance(FirebaseApp)} instead, passing the appropriate app instance.
  * <p>
  * <p>
- * See the <a href="https://github.com/firebase/FirebaseUI-Android/tree/master/auth#using-firebaseui-for-authentication">README</a>
+ * See the <a href="https://github.com/firebase/FirebaseUI-Android/blob/master/auth/README.md#table-of-contents">README</a>
  * for examples on how to get started with FirebaseUI Auth.
  */
 public class AuthUI {
--- a/auth/src/main/java/com/firebase/ui/auth/AuthUI.java
+++ b/auth/src/main/java/com/firebase/ui/auth/AuthUI.java
@@ -60,173 +60,14 @@ import java.util.List;
 import java.util.Set;
 
 /**
- * The entry point to the AuthUI authentication flow, and related utility methods.
- * If your application uses the default {@link FirebaseApp} instance, an AuthUI instance can
- * be retrieved simply by calling {@link AuthUI#getInstance() AuthUI.getInstance()}.
- * If an alternative app instance is in use, call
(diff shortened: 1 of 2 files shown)


auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -60,173 +60,14 @@
...
- * <p>
- * <li>Facebook Sign-in: Please override the string resource
- * {@code facebook_application_id} to provide the
- * <a href="https://developers.facebook.com/docs/apps/register">App ID</a> for your app as
- * registered on the
- * <a href="https://developers.facebook.com/apps">Facebook Developer Dashboard</a>.
- * </li>
- * <p>
- * </ul>
+ * See the <a href="https://github.com/firebase/FirebaseUI-Android/tree/master/auth#using-firebaseui-for-authentication">README</a>
```
samtstern: Let's just link to the README itself rather than a section since those headers could change:
https://github.com/firebase/FirebaseUI-Android/blob/master/auth/README.md

samtstern:
I think this is a good change since you're right, we never update this JavaDoc.  It's unfortunate but it's better to have an up-to-date link than broken docs!

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -304,8 +144,8 @@ public static AuthUI getInstance(FirebaseApp app) {
     }
 
     /**
-     * Default theme used by {@link SignInIntentBuilder#setTheme(int)} if no theme
-     * customization is required.
+     * Default theme used by {@link SignInIntentBuilder#setTheme(int)} if no theme customization is
```
samtstern: Just a note for the future, including formatting changes like this that are unrelated to the functionality of the PR is distracting.  I can normally look around it, but just wanted to point it out.

Is it your IDE settings that do this automatically?

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
SUPERCILEX: @samtstern I changed the link to reference the file directly, and changed the url fragment to reference the Table of Contents (something that shouldn't change anytime soon). If we do change it and forget to update the docs, it'll just go to the top of the page anyway.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
SUPERCILEX: @samtstern Yeah, sorry. Intellij, just does that automatically when formatting stuff. (I should poke around and try to find where the setting is to disable it, but I've been lazy. I'll look into it at some point.)

PR #572, opened 2017-02-05
Remove outdated know issues and cleanup storage README

SUPERCILEX:
Having a huge "KNOWN ISSUES" header is a pretty good way to scare people off and this issue is pretty outdated by now. Also, I'm requesting a merge into `version-1.2.0-dev` so this issue will be even more outdated by the time it disappears from the README. @samtstern What do you think?


samtstern:
Yep I believe this is fixed.

PR #573, opened 2017-02-05
Remove unused proguard files and fix `./gradlew build` not working because of failing tests

SUPERCILEX:
Otherwise, `./gradlew build` fails.


app/proguard-rules.pro
```diff
@@ -1,30 +0,0 @@
...
-# and specify the fully qualified class name to the JavaScript interface
-# class:
-#-keepclassmembers class fqcn.of.javascript.interface.for.webview {
-#   public *;
-#}
-
-# See:
-# https://firebase.google.com/docs/auth/android/start/#proguard
--keepattributes Signature
--keepattributes *Annotation*
```
SUPERCILEX: None of this is necessary because we add `consumerProguardFiles`.

build.gradle
```diff
@@ -25,6 +25,11 @@ allprojects {
     }
 }
 
+// TODO Remove once either Facebook or Robolectric fix the crash: https://github.com/robolectric/robolectric/issues/2579#issuecomment-276526911
+gradle.taskGraph.whenReady { graph ->
+    graph.allTasks.findAll { it.name ==~ /testReleaseUnitTest/ }*.enabled = false
+}
```
SUPERCILEX: This is a more permanent solution with a TODO that disables release unit tests for all gradle tasks. Unfortunately, I couldn't find a way to disable a task from within another task eg disable `testReleaseUnitTest` just for the build task.

samtstern: This LGTM, it's a better solution

PR #574, opened 2017-02-05
Remove unused pom.xml file

SUPERCILEX:
@samtstern Jitpack was saying that it found a `v0.3.1` artifact so I did a Ctrl + Shift + F and found it in this pom file. I don't believe it's being used so we might as well get rid of it.


samtstern:
SGTM, we generate the POM in code now.

PR #575, opened 2017-02-05
Add documentation for `GoogleApiHelper#getSafeAutoManageId()` and do a teeny bit of cleanup in the auth README

SUPERCILEX:
@samtstern I recently added Firebase app indexing to my app and noticed that it was crashing if I tried to sign in or out. Turns out I was duplicating `enableAutoManage` ids with FirebaseUI so I think we should make `GoogleApiHelper#getSafeAutoManageId()` part of the public api so people can use it to get around this problem.

### DIFF
--- a/auth/README.md
+++ b/auth/README.md
@@ -217,7 +217,8 @@ startActivityForResult(
 
 #### Handling the sign-in response
 
-#####Response codes
+##### Response codes
+
 The authentication flow provides several response codes of which the most common are as follows:
 `ResultCodes.OK` if a user is signed in, `ResultCodes.CANCELLED` if the user manually canceled the sign in,
 `ResultCodes.NO_NETWORK` if sign in failed due to a lack of network connectivity,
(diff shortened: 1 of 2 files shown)


auth/README.md
```diff
@@ -352,7 +370,7 @@ represented in the following diagram:
 
 ![FirebaseUI authentication flow on Android](flow.png)
 
-### UI customization
+## UI customization
```
SUPERCILEX: This header is part of the `Table of Contents` so it should be on the same level as `configuration` and `using-firebaseui-for-authentication`.

auth/README.md
```diff
@@ -344,6 +345,23 @@ AuthUI.getInstance()
         });
 ```
 
+### Using the `GoogleApiClient` in your own app
+
+FirebaseUI Auth heavily relies on the `GoogleApiClient` and its `enableAutoManage` method to
+help prevent memory leaks. However, using `enableAutoManage` can be risky because the method will throw
+an `IllegalStateException` if another `GoogleApiClient` is being managed with the same id.
+To help solve this problem, FirebaseUI provides a helper method to get a safe id that is
+not being used: `GoogleApiHelper#getSafeAutoManageId()`
```
SUPERCILEX: @samtstern I think the api is stable enough for this, do you agree?

samtstern:
Hey @SUPERCILEX I don't think this should be part of the public API for a few reasons:

  1. It's tangential to our library's core function, so if this was a pattern we wanted people to use we'd do better to just publish a snippet somewhere (Github Gist, StackOverflow, etc)
  2. It's really just wrapping AtomicInteger, so we're not providing much value
  3. The GoogleApiClient engineers made a decision to not just auto-increment IDs internally so that users would have to track their clients, so while for us this pattern works it's not something we want everyone to do without thinking.

Gonna close this one since the other README fixes can be part of another PR, hope that makes sense!

SUPERCILEX:
@samtstern your arguments totally make sense on not making it part of the public API, but as a counter proposal I think we should at least start the atomic integer with a value of 10 or something. What do you think? This should remove the need for devs to use `getSafeAutoManageId()` because we don't start with the default ID anymore. Also, I didn't think about this but we weren't using auto manage before so this is technically breaking backwards compatibility.

PR #576, opened 2017-02-05
Fix crash because `onConnected` is called multiple times

SUPERCILEX:
Refactor mistake: I didn't notice that the old helper [removed the listeners](https://github.com/firebase/FirebaseUI-Android/blob/version-1.2.0-dev/auth/src/main/java/com/firebase/ui/auth/util/GoogleApiClientTaskHelper.java#L105). I think that method is flawed though because if `onConnectionFailed` is called and later on `onConnected` also gets called, an `IllegalStateException` will still be thrown.

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/util/GoogleApiHelper.java
+++ b/auth/src/main/java/com/firebase/ui/auth/util/GoogleApiHelper.java
@@ -42,6 +42,9 @@ public abstract class GoogleApiHelper implements GoogleApiClient.ConnectionCallb
 
     @Override
     public void onConnected(@Nullable Bundle bundle) {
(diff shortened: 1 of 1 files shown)


samtstern:
This does seem functionally the same as unregistering the listeners as we're basically just ignoring follow-up calls, but can you add some comments above the `trySet{Foo}` calls explaining why we need to use that style?

SUPERCILEX:
@samtstern Will do! The other reason I wanted to use a `trySet` is because extenders of `GoogleApiHelper` might depend on `onConnected` being called each time the client connects, not just the first time.

SUPERCILEX:
@samtstern Done.

PR #578, opened 2017-02-06
Start safe ID at non-default value for getSafeAutoManageId

SUPERCILEX:
@samtstern See https://github.com/firebase/FirebaseUI-Android/pull/575#issuecomment-277764262. I think this is the best course of action, any objections?


auth/src/main/java/com/firebase/ui/auth/util/GoogleApiHelper.java
```diff
@@ -32,6 +32,10 @@ protected GoogleApiHelper(FragmentActivity activity, GoogleApiClient.Builder bui
         mClient = builder.build();
     }
 
+    /**
+     * @return a safe id for {@link GoogleApiClient.Builder#enableAutoManage(FragmentActivity, int,
+     * GoogleApiClient.OnConnectionFailedListener)}
+     */
```
SUPERCILEX: Any objections to leaving the documentation?

samtstern: Nope documentation is always good

PR #579, opened 2017-02-07
Modernize database tests and cleanup

SUPERCILEX:
@samtstern Since it looks like #481 isn't going to be merged anytime soon, this PR gives @CodingDoug credit for his work while allowing us to move forward with modernized tests.


database/src/androidTest/java/com/firebase/ui/database/FirebaseArrayOfObjectsTest.java
```diff
@@ -35,54 +33,46 @@
...
-    @Test
-    public void testSize() throws Exception {
-        assertEquals(3, mArray.getCount());
+        mArray.cleanup();
+        mRef.getRoot().removeValue();
     }
 
     @Test
     public void testPushIncreasesSize() throws Exception {
-        assertEquals(3, mArray.getCount());
```
samtstern: Was dropping this line intentional?

SUPERCILEX: @CodingDoug removed it in his PR. If you'd like I'll add it back (I'm not sure why he removed it).

samtstern:
@CodingDoug any issues with merging in your test changes now since we all agree they are good?  You can then rebase your PR and it will only contain the `Map` performance improvements.

CodingDoug:
@samtstern Yeah, that's fine.  I promise I'll get back to the rest of this PR some time!

database/src/androidTest/java/com/firebase/ui/database/FirebaseArrayOfObjectsTest.java
samtstern: It does seem a little strange to assert that the count after is "4" but not assert that it was "3" before.  4 is only correct if it's one more than the initial count.

CodingDoug: This check is redundant.  The '@Before' annotated function gets run before every single '@Test' function, and it verifies the size of the array ahead of time.

SUPERCILEX: @CodingDoug thanks for the explanation!

PR #582, opened 2017-02-08
Add firebase repository to README

marceloslacerda:
Added an instruction to include the fabric repository to app/build.gradle in README.md.

This have been discussed in #392 and was added to FirebaseUI-Android/auth/README.md but not to FirebaseUI-Android/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ dependency on the appropriate Firebase SDK so there is no need to include
 those separately in your app.
 
 In your `app/build.gradle` file add a dependency on one of the FirebaseUI
-libraries.
+libraries:
 
 ```groovy
 dependencies {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,6 +41,18 @@ dependencies {
 }
 ```
 
+and add the Fabric repository
```
SUPERCILEX: @marceloslacerda I agree that this needs to be in the main README because it's part of the installation process, but I think it would be better to link to the sub README like so:
```
If you're including the `firebase-ui-auth` dependency, there's a little [more setup](https://github.com/firebase/FirebaseUI-Android/tree/master/auth#configuration) required.
```

samtstern: I think your approach is better.  Also I am wondering if we should drop the prominent reference to the combined `firebase-ui` target which causes more problems than it solves.

samtstern:
Pulled this into #604

Issue #563, opened 2017-01-31
Add an AuthUI.reauthenticate() function


samtstern:
There are some Firebase Auth actions that require a recent (~5mins) sign-in to succeed:

  * Delete account (#478)
  * Change email
  * Etc

We should offer a simple method to launch into a re-authentication flow that allows the user to sign in choosing from any of their linked providers.

Considerations:

  * This would likely have to be a method that returns an `Intent` similar to the sign-in intent
  * The developer will probably want to provide a reason for the re-authentication, maybe passed in
  * This would be similar to the normal sign-in flow with a few alterations:
    * Account creation disabled
    * Account picker screen lists only methods the user has already linked to the `FirebaseUser`

amandle:
Unless there are objections I'm going to start taking a crack at this.

samtstern:
Go for it, thanks!

nikahmadz:
May I know , what is the current status of this implementation?

samtstern:
@nikahmadz this is indefinitely on hold ... @amandle left the project and we are focusing on some other priorities for now.

maillouxc:
FWIW, I'd really love to see this implemented. If I'm reading the comments here correctly, it seems like the work was already done for Android, but never released because it wasn't available on the other platforms. Is it possible to just push the release out for android?  I would implement the other devices myself if I had the time, but it seems like a shame to have this stuff ready to go but just waiting on other implementations.

TeunVR:
This is really a necessary feature. 
I am currently using the unreleased iOS FUIAccountSettingsViewController, and it seems to be working for me.
But it should be officially released and also available for Android.
I am a bit sad that i chose for FirebaseUI-Auth now that i am running into this issue :(

jackz314:
It's been more than 2 years from its initial progress and more than a year from the last update, did you guys abandon this key feature?

Heldenkrieger01:
When can we expect the release of this necessary feature? @samtstern

katowulf:
Closing this conversation to prevent more "ETA requested" responses (which actually take time from feature work). 

Latest status update: #issuecomment-367736441

Firebase team will update here and potentially unlock thread if/when there is an update.

demolaf:
This should be resolved in this PR https://github.com/firebase/FirebaseUI-Android/pull/2332, feel free to open a new issue if you come across any bugs.

PR #583, opened 2017-02-09
Add reauthentication

amandle:
Adds reauthentication as discussed in #563 

- There is now `AuthUI.createReauthIntentBuilder()` which returns an intent builder very similar to `AuthUI.createSignInIntent()` but adds the methods `setReauthReason` and `setReauthEmail` and removes `setAllowNewEmailAccounts`.  Both builders extend a common builder which contains the common methods.
- Account creation is skipped during the reauth flow
- Account picker screen only lists the methods the user has linked.
- The reauth reason gets displayed directly above the auth method buttons. This is totally open for discussion but it's where made the most sense to me.

### DIFF
--- a/app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
+++ b/app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
@@ -126,15 +126,23 @@ public class AuthUiActivity extends AppCompatActivity {
     @Override
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+        setContentView(R.layout.auth_ui_layout);
+        ButterKnife.bind(this);
 
         FirebaseAuth auth = FirebaseAuth.getInstance();
         if (auth.getCurrentUser() != null) {
-            startActivity(SignedInActivity.createIntent(this, null));
+            startActivity(
+                    SignedInActivity.createIntent(
+                            this,
+                            null,
+                            new SignedInActivity.SignedInConfig(
+                                    getSelectedLogo(),
+                                    getSelectedTheme(),
+                                    getSelectedProviders(),
+                                    getSelectedTosUrl(),
+                                    mEnableSmartLock.isEnabled())));
             finish();
         }
-
-        setContentView(R.layout.auth_ui_layout);
-        ButterKnife.bind(this);
 
         if (!isGoogleConfigured()) {
             mUseGoogleProvider.setChecked(false);
@@ -208,7 +216,16 @@ public class AuthUiActivity extends AppCompatActivity {
 
         // Successfully signed in
         if (resultCode == ResultCodes.OK) {
-            startActivity(SignedInActivity.createIntent(this, response));
+            startActivity(
+                    SignedInActivity.createIntent(
+                            this,
+                            response,
+                            new SignedInActivity.SignedInConfig(
+                                    getSelectedLogo(),
+                                    getSelectedTheme(),
+                                    getSelectedProviders(),
+                                    getSelectedTosUrl(),
+                                    mEnableSmartLock.isEnabled())));
             finish();
             return;
         } else {
--- a/app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
+++ b/app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
@@ -242,7 +242,7 @@ public class AuthUiActivity extends AppCompatActivity {
                                 getSelectedTheme(),
                                 getSelectedProviders(),
                                 getSelectedTosUrl(),
-                                mEnableSmartLock.isEnabled())));
+                                mEnableSmartLock.isChecked())));
     }
 
     @MainThread
--- a/app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
+++ b/app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
@@ -131,16 +131,7 @@ public class AuthUiActivity extends AppCompatActivity {
 
         FirebaseAuth auth = FirebaseAuth.getInstance();
         if (auth.getCurrentUser() != null) {
-            startActivity(
-                    SignedInActivity.createIntent(
-                            this,
-                            null,
-                            new SignedInActivity.SignedInConfig(
-                                    getSelectedLogo(),
-                                    getSelectedTheme(),
-                                    getSelectedProviders(),
-                                    getSelectedTosUrl(),
-                                    mEnableSmartLock.isEnabled())));
+            startSignedInActivity(null);
             finish();
         }
 
@@ -216,16 +207,7 @@ public class AuthUiActivity extends AppCompatActivity {
 
         // Successfully signed in
         if (resultCode == ResultCodes.OK) {
-            startActivity(
-                    SignedInActivity.createIntent(
-                            this,
-                            response,
-                            new SignedInActivity.SignedInConfig(
-                                    getSelectedLogo(),
-                                    getSelectedTheme(),
-                                    getSelectedProviders(),
-                                    getSelectedTosUrl(),
-                                    mEnableSmartLock.isEnabled())));
+            startSignedInActivity(response);
             finish();
             return;
         } else {
@@ -248,6 +230,19 @@ public class AuthUiActivity extends AppCompatActivity {
         }
 
         showSnackbar(R.string.unknown_sign_in_response);
+    }
+
+    private void startSignedInActivity(IdpResponse response) {
+        startActivity(
+                SignedInActivity.createIntent(
+                        this,
+                        response,
+                        new SignedInActivity.SignedInConfig(
+                                getSelectedLogo(),
+                                getSelectedTheme(),
+                                getSelectedProviders(),
+                                getSelectedTosUrl(),
+                                mEnableSmartLock.isEnabled())));
     }
 
     @MainThread
--- a/auth/src/main/java/com/firebase/ui/auth/AuthUI.java
+++ b/auth/src/main/java/com/firebase/ui/auth/AuthUI.java
@@ -431,6 +431,7 @@ public class AuthUI {
     /**
      * Base builder for both {@link SignInIntentBuilder} and {@link ReauthIntentBuilder}
      */
+    @SuppressWarnings(value="unchecked")
     private abstract class AuthIntentBuilder<T extends AuthIntentBuilder> {
         int mLogo = NO_LOGO;
         int mTheme = getDefaultTheme();
--- a/auth/src/main/java/com/firebase/ui/auth/AuthUI.java
+++ b/auth/src/main/java/com/firebase/ui/auth/AuthUI.java
@@ -305,6 +305,13 @@ public class AuthUI {
     }
 
     /**
+     * Starts the reauthentication flow.
+     */
+    public ReauthIntentBuilder createReauthIntentBuilder() {
+        return new ReauthIntentBuilder();
+    }
+
+    /**
      * Configuration for an identity provider.
      * <p>
      * In the simplest case, you can supply the provider ID and build the config like this:
@@ -422,47 +429,48 @@ public class AuthUI {
     }
 
     /**
-     * Builder for the intent to start the user authentication flow.
+     * Base builder for both {@link SignInIntentBuilder} and {@link ReauthIntentBuilder}
      */
-    public final class SignInIntentBuilder {
-        private int mLogo = NO_LOGO;
-        private int mTheme = getDefaultTheme();
-        private List<IdpConfig> mProviders = new ArrayList<>();
-        private String mTosUrl;
-        private boolean mIsSmartLockEnabled = true;
-        private boolean mAllowNewEmailAccounts = true;
+    private abstract class AuthIntentBuilder<T extends AuthIntentBuilder> {
+        int mLogo = NO_LOGO;
+        int mTheme = getDefaultTheme();
+        List<IdpConfig> mProviders = new ArrayList<>();
+        String mTosUrl;
+        boolean mIsSmartLockEnabled = true;
 
-        private SignInIntentBuilder() {
+        private AuthIntentBuilder() {
+            mProviders.add(new IdpConfig.Builder(EMAIL_PROVIDER).build());
         }
+
 
         /**
          * Specifies the theme to use for the application flow. If no theme is specified,
          * a default theme will be used.
          */
-        public SignInIntentBuilder setTheme(@StyleRes int theme) {
+        public T setTheme(@StyleRes int theme) {
             Preconditions.checkValidStyle(
-                    mApp.getApplicationContext(),
-                    theme,
-                    "theme identifier is unknown or not a style definition");
+                mApp.getApplicationContext(),
+                theme,
+                "theme identifier is unknown or not a style definition");
             mTheme = theme;
-            return this;
+            return (T) this;
         }
 
         /**
          * Specifies the logo to use for the {@link AuthMethodPickerActivity}. If no logo
          * is specified, none will be used.
          */
-        public SignInIntentBuilder setLogo(@DrawableRes int logo) {
+        public T setLogo(@DrawableRes int logo) {
             mLogo = logo;
-            return this;
+            return (T) this;
         }
 
         /**
          * Specifies the terms-of-service URL for the application.
          */
-        public SignInIntentBuilder setTosUrl(@Nullable String tosUrl) {
+        public T setTosUrl(@Nullable String tosUrl) {
             mTosUrl = tosUrl;
-            return this;
+            return (T) this;
         }
 
         /**
@@ -476,7 +484,7 @@ public class AuthUI {
          *                   configuration parameters for the IDP.
          * @see IdpConfig
          */
-        public SignInIntentBuilder setProviders(@NonNull List<IdpConfig> idpConfigs) {
+        public T setProviders(@NonNull List<IdpConfig> idpConfigs) {
             mProviders.clear();
             for (IdpConfig config : idpConfigs) {
                 if (mProviders.contains(config)) {
@@ -487,7 +495,7 @@ public class AuthUI {
                     mProviders.add(config);
                 }
             }
-            return this;
+            return (T) this;
         }
 
         /**
@@ -503,7 +511,7 @@ public class AuthUI {
          * @see #GOOGLE_PROVIDER
          */
         @Deprecated
-        public SignInIntentBuilder setProviders(@NonNull String... providers) {
+        public T setProviders(@NonNull String... providers) {
             mProviders.clear(); // clear the default email provider
             for (String provider : providers) {
                 if (isIdpAlreadyConfigured(provider)) {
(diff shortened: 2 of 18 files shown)


auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -401,48 +408,49 @@ public IdpConfig build() {
...
-    public final class SignInIntentBuilder {
-        private int mLogo = NO_LOGO;
-        private int mTheme = getDefaultTheme();
-        private LinkedHashSet<IdpConfig> mProviders = new LinkedHashSet<>();
-        private String mTosUrl;
-        private boolean mIsSmartLockEnabled = true;
-        private boolean mAllowNewEmailAccounts = true;
-
-        private SignInIntentBuilder() {
+    @SuppressWarnings(value="unchecked")
```
SUPERCILEX: @amandle the build should pass if you put space around the `=`.

auth/src/main/java/com/firebase/ui/auth/ui/FlowParameters.java
```diff
@@ -51,21 +51,36 @@
 
     public final boolean allowNewEmailAccounts;
 
+    public final boolean isReauth;
```
SUPERCILEX: @amandle Couldn't this be a method instead? It seems weird to have a parameter just for something that can be inferred programmatically.

Something like:
```java
public boolean isReauth() {
    return !TextUtils.isEmpty(reauthEmail);
}
```

auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
```diff
@@ -77,7 +83,46 @@ protected void onCreate(Bundle savedInstanceState) {
...
+                    }
+                }
+            }
+        } else {
+            visibleProviders = allConfiguredProviders;
+        }
+        if (visibleProviders.size() == 0) {
+            Log.e(TAG, "There are no configured providers associated with the user's account");
+        }
+        populateIdpList(visibleProviders);
```
SUPERCILEX: Before this happens I think we should check to see if `visibleProviders.size() == 1` and just launch the sign in right away if that is the case.

auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
```diff
@@ -185,9 +234,25 @@ public void onFailure(Bundle extra) {
     @Override
     public void onClick(View view) {
         if (view.getId() == R.id.email_provider) {
-            startActivityForResult(
-                    RegisterEmailActivity.createIntent(this, mActivityHelper.getFlowParams()),
-                    RC_EMAIL_FLOW);
+            if (mActivityHelper.getFlowParams().isReauth) {
+                String reauthEmail = mActivityHelper.getFlowParams().reauthEmail;
+                if (reauthEmail == null) {
+                    Log.e(TAG, "Attempting to reauth with email but no reauthEmail provided");
```
SUPERCILEX: I think the `build()` method should prevent this and throw an `IllegalArgumentException` if the dev tries to do this.

auth/src/main/res/layout/auth_method_picker_layout.xml
```diff
@@ -11,16 +11,28 @@
...
-            android:id="@+id/email_provider"
-            style="@style/FirebaseUI.Button.AccountChooser.EmailButton"
-            android:text="@string/sign_in_with_email"
-            android:visibility="gone"
-            tools:visibility="visible"/>
+      android:layout_height="wrap_content"
+      android:layout_width="wrap_content"
+      android:orientation="vertical">
+
+        <TextView
```
SUPERCILEX: It looks like this is just copy past from the landscape layout so we could use a `<merge />` tag for this.

auth/src/test/java/com/firebase/ui/auth/testhelpers/TestHelper.java
```diff
@@ -64,9 +64,12 @@ public static FlowParameters getFlowParameters(List<String> providerIds) {
...
                 AuthUI.NO_LOGO,
-                null,
-                true,
-                true);
+                null  /* tosUrl */,
+                true  /* smartLockEnabled */,
+                true  /* allowNewEmailAccounts */,
+                false /* isReauth */,
+                null  /* reauthEmail */,
+                null  /* reauthReason */);
```
SUPERCILEX: 👍

auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
```diff
@@ -185,9 +234,25 @@ public void onFailure(Bundle extra) {
...
+                        WelcomeBackPasswordPrompt.createIntent(
+                                this,
+                                mActivityHelper.getFlowParams(),
+                                new IdpResponse(
+                                        AuthUI.EMAIL_PROVIDER,
+                                        reauthEmail)),
+                        RC_EMAIL_DIRECT_SIGN_IN);
+            } else {
+                startActivityForResult(
+                        RegisterEmailActivity.createIntent(this, mActivityHelper.getFlowParams()),
```
SUPERCILEX: @amandle I'm just being nit picky here (Sorry! 😄), but I feel like `mActivityHelper.getFlowParams()` could be `Ctrl + Alt + V`ed into something like `params`.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -401,48 +408,49 @@ public IdpConfig build() {
...
             Preconditions.checkValidStyle(
-                    mApp.getApplicationContext(),
-                    theme,
-                    "theme identifier is unknown or not a style definition");
+                mApp.getApplicationContext(),
+                theme,
+                "theme identifier is unknown or not a style definition");
             mTheme = theme;
-            return this;
+            return (T) this;
```
samtstern: Sigh.  I looked at this and thought "there's no way that Java really requires this generic + cast situation" but it looks like we do!

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -529,14 +527,94 @@ public Intent build() {
...
+        }
+
+        /**
+         * Sets the email address for which reauthentication will be tried.  If email accounts are
+         * enabled this MUST be set or email reauthentication will fail.
+         *
+         * @param reauthEmail The email address of the account for which reauthentication is being
+         * requested
+         */
+        public ReauthIntentBuilder setReauthEmail(String reauthEmail) {
```
samtstern: Is it possible to do re-auth and not use the currently signed in user?  In which case this email would be `FirebaseAuth.getInstance().getCurrentUser().getEmail()`  I feel like we could remove this setter and therefore reduce the chance of error/misunderstanding.

auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
```diff
@@ -77,7 +83,46 @@ protected void onCreate(Bundle savedInstanceState) {
...
 
-        populateIdpList(mActivityHelper.getFlowParams().providerInfo);
+        Map<String, IdpConfig> providerIdToConfig = new HashMap<>();
+        List<IdpConfig> allConfiguredProviders = mActivityHelper.getFlowParams().providerInfo;
+        for (IdpConfig providerConfig : allConfiguredProviders) {
+            providerIdToConfig.put(providerConfig.getProviderId(), providerConfig);
+        }
+        List<IdpConfig> visibleProviders = new ArrayList<>();
+        if (mActivityHelper.getFlowParams().isReauth) {
+            // display the reauthReason if available
```
samtstern: We should ask UX about this, but instinctively a dialog feels like the "normal" reauth flow.

  1. User clicks a button requiring reauth ("Delete Account" for example)
  2. A dialog shows up that says "This action requires you to sign in again, would you like to continue"
  3. Yes launches the flow, No returns `RESULT_CANCELED`

auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
```diff
@@ -77,7 +83,46 @@ protected void onCreate(Bundle savedInstanceState) {
...
+                TextView reauthReasonTextView = (TextView) findViewById(R.id.reauth_reason);
+                reauthReasonTextView.setText(reauthReason);
+                reauthReasonTextView.setVisibility(View.VISIBLE);
+            }
+
+            // For reauth flow we only want to show the IDPs which the user has associated with
+            // their account.
+            List<String> providerIds = mActivityHelper.getCurrentUser().getProviders();
+            if (providerIds.size() == 0) {
+                // zero providers indicates that it is an email account
```
samtstern: Doesn't the providers array contain just `"password"` in this case or is that only when you call `fetchProvidersForEmail()`?

auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
samtstern: Agreed and I think we already have this logic somewhere to steal.

auth/src/main/java/com/firebase/ui/auth/ui/FlowParameters.java
SUPERCILEX: Nevermind, I thought the `reauthenticate` method required an email but it doesn't so this won't work.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
amandle: Done

app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
```diff
@@ -208,7 +216,16 @@ private void handleSignInResponse(int resultCode, Intent data) {
...
+            startActivity(
+                    SignedInActivity.createIntent(
+                            this,
+                            response,
+                            new SignedInActivity.SignedInConfig(
+                                    getSelectedLogo(),
+                                    getSelectedTheme(),
+                                    getSelectedProviders(),
+                                    getSelectedTosUrl(),
+                                    mEnableSmartLock.isEnabled())));
```
amandle: good point, done

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -529,14 +527,94 @@ public Intent build() {
...
+        private ReauthIntentBuilder() {
+            super();
+        }
+
+        /**
+         * Set an explanation for why reauth was requested e.g. "To delete your account you must
+         * reauthenticate."
+         *
+         * @param reauthReason A string explaining why reauthentication was requested.
+         * @return
```
amandle: I think you're referring to the empty return? I have now removed it.

auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
amandle: Done.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
amandle: Nice, that simplifies things considerably.

auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
amandle: As per Sam's comment reauthEmail is now removed, and the user is required to be logged in to start, so this whole block can go away

SUPERCILEX: Awesome!

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -529,14 +527,94 @@ public Intent build() {
...
+
+        @Override
+        public FlowParameters getFlowParams() {
+            return new FlowParameters(
+                    mApp.getName(),
+                    new ArrayList<>(mProviders),
+                    mTheme,
+                    mLogo,
+                    mTosUrl,
+                    mIsSmartLockEnabled,
```
SUPERCILEX: @amandle Have you tried reauthenticating with smart lock enabled? Given @samtstern's comment about showing a dialog before we show auth stuff, I feel like things will get complicated with smart lock enabled. @samtstern @amandle Am I missing something or could this be a problem?

auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
SUPERCILEX: @samtstern Yup, it's right here: https://github.com/firebase/FirebaseUI-Android/blob/master/auth/src/main/java/com/firebase/ui/auth/util/signincontainer/SignInDelegate.java#L233.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
amandle: Yeah, it's a little complicated/confusing, but I think we have to have it as an option for the case where the person doesn't know their password but has it stored in smartlock

samtstern:
@amandle FYI there are merge conflicts between your branch and the base.

samtstern:
I checked out the branch and tried a few things.  I am noticing some strange behavior when using Google Sign In:

- If I have SmartLock enabled I get the "choose an account saved with Google Smart Lock" dialog, but if I click "None of the above" then the whole reauth flow ends.  This does not happen with password login.
- If I don't have SmartLock enabled I still get the dialog above when I try to reauth.  It does not seem that I can actually disable SmartLock (maybe this is just a sample issue?)

Also the re authentication dialog does not seem to respect my theme colors, the buttons are always straight black.  Ideally this dialog would inherit the theme passed to the sign in flow.

auth/src/main/java/com/firebase/ui/auth/ui/FlowParameters.java
```diff
@@ -94,9 +106,12 @@ public FlowParameters createFromParcel(Parcel in) {
             String termsOfServiceUrl = in.readString();
             int smartLockEnabledInt = in.readInt();
             int allowNewEmailAccountsInt = in.readInt();
+            int isReauthInt = in.readInt();
+            String reauthReason = in.readString();
 
             boolean smartLockEnabled = smartLockEnabledInt != 0;
             boolean allowNewEmailAccounts = allowNewEmailAccountsInt != 0;
+            boolean isReauth = isReauthInt != 0;
```
SUPERCILEX: @amandle The merge conflicts are my fault, sorry about that! I removed this unnecessary abstraction so this should work once you merge:
```java
boolean smartLockEnabled = in.readInt() != 0;
boolean allowNewEmailAccounts = in.readInt() != 0;
boolean isReauth = in.readInt() != 0;
String reauthReason = in.readString();
```

amandle:
> If I have SmartLock enabled I get the "choose an account saved with Google Smart Lock" dialog, but if I click "None of the above" then the whole reauth flow ends. This does not happen with password login.

The reauth does happen though. Since the user is already signed in to Google and we know the email address it happens automatically. You can verify this by putting a breakpoint at `IdpSignInContainer.java:134`

> If I don't have SmartLock enabled I still get the dialog above when I try to reauth. It does not seem that I can actually disable SmartLock (maybe this is just a sample issue?)

Ah, yep this is a bug in the sample app. Should be fixed now. 

> Also the re authentication dialog does not seem to respect my theme colors, the buttons are always straight black. Ideally this dialog would inherit the theme passed to the sign in flow.

Hmm, I'm not sure how to do this without going back to using standard activities instead of AlertDialogs. The RecoveryEmailSentDialog uses the same theme: https://github.com/firebase/FirebaseUI-Android/blob/e14170d2006bdffc767f69b740e1dd7cbc3c79b5/auth/src/main/java/com/firebase/ui/auth/ui/email/RecoveryEmailSentDialog.java#L32

auth/src/main/java/com/firebase/ui/auth/ui/FlowParameters.java
amandle: :+1:

auth/src/main/java/com/firebase/ui/auth/util/signincontainer/SignInDelegate.java
```diff
@@ -232,17 +237,44 @@ private void handleCredential(Credential credential) {
...
+            providerIdToConfig.put(providerConfig.getProviderId(), providerConfig);
+        }
 
+        List<IdpConfig> visibleProviders = new ArrayList<>();
+        if (mHelper.getFlowParams().isReauth) {
+            // For reauth flow we only want to show the IDPs which the user has associated with
+            // their account.
+            List<String> providerIds = mHelper.getCurrentUser().getProviders();
+            if (providerIds.size() == 0) {
+                // zero providers indicates that it is an email account
```
amandle: It does seem to be different, asking some of the Firebase folks for more information

samtstern:
@amandle I moved this to target the `2.0.0` release since this is a notable enough feature that we should pursue iOS parity.  Hopefully @yramanchuk can implement a mirror of this.

The only remaining concern I still have is the bad UX with SmartLock and "None of the Above" where GSI succeeds even though the user never really did anything.  Can we clarify this somehow?  Options include:
  * For reauth change SmartLock settings to only include password accounts
  * Display a Toast after reauth success
  * Other (?)

amandle:
@samtstern I think this is ready to review now. I added a toast when the automatic Google Sign In happens and skip Smart Lock if it's a reauth with a non-password account.

samtstern:
@amandle approved!  Going to squash and merge.

@yramanchuk @morganchen12 @bojeil-google FYI we need to coordinate the same abilities across platforms.  This is merging into our next release branch.

develuxe:
Would love to have this feature back but can't find it in the repo. Was it removed?

samtstern:
This feature was never released actually since amandle left the project and
we were focused on stability work.

develuxe:
Thanks for your quick response. Is there any best practice on how to reauthenticate the user?

ViniciusR:
So sad they abandoned this feature. I'm actually facing this situation, it's so much complex to retrieve credentials from Google and other providers, Firebase's reauthenticate function is pretty easy, but it requires that credentials. 

I think the easiest way is forcing sign out and ask user to sign in again - it would needed to login back to reauthenticate anyway.

bojeil-google:
Hey folks, we have not abandoned our plans to support re-authentication in FirebaseUI (it is indeed a missing key functionality that we should support). The priority has been lowered in favor of some other popular/critical feature requests such as support for email link sign-in and account management UI.

jackz314:
Any updates on this feature? And would there be an ETA for this?

Heldenkrieger01:
Will this be released any time soon?

PR #584, opened 2017-02-11
Improve Gradle build times

SUPERCILEX:
I got 4X on my machine, but on Travis the speed improvements are less significant but still pretty good (~20s): `4 mins 11.941 secs -> 3 mins 49.533 secs`.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,19 +1,3 @@
-# Project-wide Gradle settings.
-
-# IDE (e.g. Android Studio) users:
-# Gradle settings configured through the IDE *will override*
-# any settings specified in this file.
-
-# For more details on how to configure your build environment visit
-# http://www.gradle.org/docs/current/userguide/build_environment.html
-
-# Specifies the JVM arguments used for the daemon process.
-# The setting is particularly useful for tweaking memory settings.
-# Default value: -Xmx10248m -XX:MaxPermSize=256m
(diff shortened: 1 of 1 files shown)


gradle.properties
```diff
@@ -1,19 +1,3 @@
...
-# Default value: -Xmx10248m -XX:MaxPermSize=256m
-# org.gradle.jvmargs=-Xmx2048m -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
-org.gradle.jvmargs=-Xmx1536m
-
-# When configured, Gradle will run in incubating parallel mode.
-# This option should only be used with decoupled projects. More details, visit
-# http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
-# org.gradle.parallel=true
+org.gradle.jvmargs=-Xmx2048m -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
+org.gradle.parallel=true
```
SUPERCILEX: We can use this because our modules aren't circular dependencies, eg `app` depends on `auth` but `auth` doesn't depend on `app`.

gradle.properties
```diff
@@ -1,19 +1,3 @@
...
+org.gradle.configureondemand=true
```
SUPERCILEX: > The standard project dependencies are supported and makes relevant projects configured. If project A has a compile dependency on project B then building A causes configuration of both projects.

This means that is we touch `auth`, `app` will properly be configured, but `database` won't be because we didn't touch it, thus saving build time.

samtstern:
Cool with me, thanks!

Issue #588, opened 2017-02-13
ProgressDialog of SignInDelegate in KickOffActivity uses the wrong progress bar color


Alex100:
KickOffActivity does not set the customized theme. That's the reason why the SignInDelegate fragment uses the default progress bar color. KickOffActivity should extend from AppCompatBase to fix this bug.
Tested in FB UI 1.1.1

SUPERCILEX:
@Alex100 this was a calculated decision. If we use the custom theme, `KickOffActivity` will not be transparent and it will look gross. If you can think of a way to use the custom theme while keeping `KickOffActivity` transparent, that would be awesome!

Alex100:
This line in FragmentHelper inside of showLoadingDialog() would do the trick:
Context context = getFlowParams().themeId !=  0 ? new ContextThemeWrapper(mFragment.getContext(), getFlowParams().themeId) : mFragment.getContext();

Here the whole method:

    @Override
    public void showLoadingDialog(String message) {
        dismissDialog();

        Context context = getFlowParams().themeId !=  0 ? new ContextThemeWrapper(mFragment.getContext(), getFlowParams().themeId) : mFragment.getContext();
        mProgressDialog = ProgressDialog.show(context, "", message, true);
    }

samtstern:
@Alex100 huh I had no idea about `ContextThemeWrapper`, thanks for the tip!  I think you're right, we can fix this.

SUPERCILEX:
@samtstern It worked!!!!!! That's so amazing! @Alex100 You're a god, thanks! 😆

![screenshot_20170213-130010](https://cloud.githubusercontent.com/assets/9490724/22903184/a4bfd39c-f1ec-11e6-968a-9c54e43d086e.png)

SUPERCILEX:
@Alex100 @samtstern Uh oh, the whole class is marked as a restricted api.

Alex100:
@SUPERCILEX There are two implementations of `ContextThemeWrapper`:

1. `android.support.v7.view.ContextThemeWrapper` (restricted)
2. `android.view.ContextThemeWrapper` (not restricted)

(https://code.google.com/p/android/issues/detail?id=227580)

SUPERCILEX:
@Alex100 yeah, it was my bad. #589 was already merged so we're good. Thanks for the explanation though!

samtstern:
The fix for this issue has been released in version `1.2.0`.

PR #589, opened 2017-02-13
Fix custom style not being applied to `KickOffActivity` progress dialogs

SUPERCILEX:
#588

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/ui/FragmentHelper.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/FragmentHelper.java
@@ -6,7 +6,7 @@ import android.content.IntentSender;
 import android.os.Bundle;
 import android.support.annotation.RestrictTo;
 import android.support.v4.app.Fragment;
-import android.support.v7.view.ContextThemeWrapper;
+import android.view.ContextThemeWrapper;
 
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 public class FragmentHelper extends BaseHelper {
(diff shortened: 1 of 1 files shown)


auth/src/main/java/com/firebase/ui/auth/ui/FragmentHelper.java
```diff
@@ -30,7 +31,9 @@ public void finish(int resultCode, Intent intent) {
     @Override
     public void showLoadingDialog(String message) {
         dismissDialog();
-        mProgressDialog = ProgressDialog.show(mFragment.getContext(), "", message, true);
+        ContextThemeWrapper context =
+                new ContextThemeWrapper(mFragment.getContext(), getFlowParams().themeId);
```
SUPERCILEX: We shouldn't need to check for `0` because we do this: `private int mTheme = getDefaultTheme();`.

auth/src/main/java/com/firebase/ui/auth/ui/FragmentHelper.java
```diff
@@ -6,6 +6,7 @@
 import android.os.Bundle;
 import android.support.annotation.RestrictTo;
 import android.support.v4.app.Fragment;
+import android.support.v7.view.ContextThemeWrapper;
```
samtstern: I think we need to use android.view.ContextThemeWrapper (https://developer.android.com/reference/android/view/ContextThemeWrapper.html) instead of the support library version.  The framework version should not be restricted, and was added in API Level 1.

SUPERCILEX: Oops, I'm so used to defaulting to support lib classes that I don't even think twice about it. 😄

PR #591, opened 2017-02-15
Bug fixes and perf improvements for setting providers in AuthUI

SUPERCILEX:
@samtstern We currently throw an exception, but I think we should be smarter about this. Now, everything gets merged together into one unique set of providers and scopes.

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/AuthUI.java
+++ b/auth/src/main/java/com/firebase/ui/auth/AuthUI.java
@@ -55,7 +55,6 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.IdentityHashMap;
-import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 
@@ -315,9 +314,9 @@ public class AuthUI {
         private final String mProviderId;
         private final List<String> mScopes;
 
-        private IdpConfig(@NonNull String providerId, List<String> scopes) {
-            mScopes = scopes;
+        private IdpConfig(String providerId, List<String> scopes) {
             mProviderId = providerId;
+            mScopes = scopes;
         }
 
         private IdpConfig(Parcel in) {
@@ -356,10 +355,32 @@ public class AuthUI {
             parcel.writeStringList(mScopes);
         }
 
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+
+            IdpConfig config = (IdpConfig) o;
+
+            return mProviderId.equals(config.mProviderId);
+        }
+
+        @Override
+        public int hashCode() {
+            return mProviderId.hashCode();
+        }
+
+        @Override
+        public String toString() {
+            return "IdpConfig{" +
+                    "mProviderId='" + mProviderId + '\'' +
+                    ", mScopes=" + mScopes +
+                    '}';
+        }
+
         public static class Builder {
             private String mProviderId;
             private List<String> mScopes = new ArrayList<>();
-
 
             /**
              * Builds the configuration parameters for an identity provider.
@@ -406,7 +427,7 @@ public class AuthUI {
     public final class SignInIntentBuilder {
         private int mLogo = NO_LOGO;
         private int mTheme = getDefaultTheme();
-        private LinkedHashSet<IdpConfig> mProviders = new LinkedHashSet<>();
+        private List<IdpConfig> mProviders = new ArrayList<>();
         private String mTosUrl;
         private boolean mIsSmartLockEnabled = true;
         private boolean mAllowNewEmailAccounts = true;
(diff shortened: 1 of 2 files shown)


auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -356,11 +355,33 @@ public void writeToParcel(Parcel parcel, int i) {
...
         }
 
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+
+            IdpConfig config = (IdpConfig) o;
+
+            return mProviderId.equals(config.mProviderId);
```
SUPERCILEX: @samtstern Identifying providers by provider id only and completely ignoring scopes makes sense to me, do you agree?

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -406,7 +427,7 @@ public IdpConfig build() {
     public final class SignInIntentBuilder {
         private int mLogo = NO_LOGO;
         private int mTheme = getDefaultTheme();
-        private LinkedHashSet<IdpConfig> mProviders = new LinkedHashSet<>();
+        private List<IdpConfig> mProviders = new ArrayList<>();
```
SUPERCILEX: This is much better, we were converting this to a list anyway which seems kinda pointless.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -458,19 +479,30 @@ public SignInIntentBuilder setTosUrl(@Nullable String tosUrl) {
          */
         public SignInIntentBuilder setProviders(@NonNull List<IdpConfig> idpConfigs) {
             mProviders.clear();
-            Set<String> configuredProviders = new HashSet<>();
```
SUPERCILEX: Less memory usage if you don't overlap providers! 😄

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -495,6 +527,15 @@ public SignInIntentBuilder setProviders(@NonNull String... providers) {
             return this;
         }
 
+        private boolean isIdpAlreadyConfigured(@NonNull String providerId) {
```
SUPERCILEX: I moved it to where it's being used (in the deprecated `setProviders`).

auth/src/test/java/com/firebase/ui/auth/AuthUITest.java
```diff
@@ -63,13 +66,30 @@ public void testCreateStartIntent_shouldHaveEmailAsDefaultProvider() {
...
+        List<IdpConfig> providers = builder.getFlowParams().providerInfo;
+        assertTrue(providers.size() == 2);
+
+        IdpConfig email = providers.get(0);
+        List<String> emailScopes = email.getScopes();
+        assertTrue(email.getProviderId().equals(AuthUI.EMAIL_PROVIDER));
+        assertTrue(emailScopes.size() == 2);
+        assertTrue(emailScopes.contains("A") && emailScopes.contains("B"));
+
+        assertTrue(providers.get(1).getProviderId().equals(AuthUI.GOOGLE_PROVIDER));
```
SUPERCILEX: @samtstern Did I get all the possible cases?

samtstern:
@SUPERCILEX what is the use case for specifying the same provider multiple times in the builder?  I think throwing an Exception is good since I doubt the developer meant to specify duplicate providers and we are probably helping them catch a mistake.

SUPERCILEX:
@samtstern If a user enables a setting in the setup of the app, they get logged in with different permissions than the standard setup. This means I have to take my default configuration, remove the `IdpConfig` for Google, and add it back with the new permissions by doing an `addAll` to my existing permissions. Whew! It's kind of a pain, but also not that big of a deal. It would just be easier to add another `IdpConfig` with the extra permission I need and have FirebaseUI merge it for me.

samtstern:
@SUPERCILEX my personal opinion is that there will be more people saved from doing the wrong thing by the error-throwing behavior than there will be people who appreciate the merging of IdpConfig.

SUPERCILEX:
@samtstern Okay, that makes sense. I made some improvements to increase readability and reduce memory thrashing along the way so I'll leave this PR open and focus its changes on that instead.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -515,23 +543,18 @@ public SignInIntentBuilder setAllowNewEmailAccounts(boolean enabled) {
...
-
         public Intent build() {
             return KickoffActivity.createIntent(mApp.getApplicationContext(), getFlowParams());
         }
 
         @VisibleForTesting()
         public FlowParameters getFlowParams() {
+            if (mProviders.isEmpty()) {
+                mProviders.add(new IdpConfig.Builder(EMAIL_PROVIDER).build());
+            }
```
SUPERCILEX: This is why we should always do this kind of thing in the `build` method: if people call `setProviders(Collections.emptyList())`, we will start the auth flow with no providers. Now, we can guarantee that at least the email provider is provided.

samtstern:
@SUPERCILEX SGTM, we can move forward with the general improvements.

samtstern:
Forgot that you had updated this, but LGTM.  Squashing and merging.

SUPERCILEX:
@samtstern Oops, I forgot to ask you to wait for @amandle's PR to be merged since this one caused a bunch of merge conflicts. I created https://github.com/amandle/FirebaseUI-Android/pull/1 to fix the conflicts. Sorry about that!

Issue #535, opened 2017-01-21
How to change from setReadPermissions in Facebook provider to setPublishPermissions


judeebene:
What is the best method to change from  read permission to publish permission in facebook Provider Using Firebase UI.

By default, Firebase UI uses  "facebook read permission for login"  as seen in FacebookProvider.java in  line 114   loginManager.logInWithReadPermissions(activity, permissionsList);

 which is quite OK ,but when the user want to share on facebook, it require that the read permission should be changed to  publish permission.
and trying to change directly without firebase UI, seems not to work 
 LoginManager.getInstance().logInWithPublishPermissions(...)

so What is the best method to do it with Firebase UI?

samtstern:
@judeebene thanks pointing this out.  I don't think we had considered the two types of permissions, I will try to address this in the next version

samtstern:
@judeebene unfortunately after some investigation we can't support this feature please see discussion in #592 for the reasons.

Facebook makes it intentionally hard to request read and publish permissions at the same time so it won't fit within the flow of FirebaseUI.  The best thing to do is provide only your read permissions to FirebaseUI and then after the user has signed in ask again for the publish permissions (which is what Facebook's guidelines suggest)

PR #592, opened 2017-02-15
[DON'T MERGE] Support publish permissions for Facebook

SUPERCILEX:
#535


auth/src/main/java/com/firebase/ui/auth/provider/FacebookProvider.java
```diff
@@ -43,7 +43,10 @@
 import java.util.List;
 
 public class FacebookProvider implements IdpProvider, FacebookCallback<LoginResult> {
+    public static final String PUBLISH_PERMISSION = "publish";
```
SUPERCILEX: I don't think this should be here because it suddenly makes `FacebookProvider` public api, but I'm not sure where to put it that makes sense. Would `AuthUI.FACEBOOK_PUBLISH_PERMISSION` work? Or should we not even make it a constant?

samtstern: Ok so Facebook's permissions docs leave something to be desired but I have a few questions:

  * Looks like the `publish_actions` permissions already acts as the signaling value here.  You need that permission to do any publishing (https://developers.facebook.com/docs/facebook-login/permissions#reference-publish_actions)
  * Facebook recommends against asking for publish permissions at login time, so not even sure we should offer this (but I guess it can't hurt)
  * What happens if we get the split wrong?  Can you try requesting a bunch of read-only permissions using the `loginWithPublish` method and see what happens?

auth/src/main/java/com/firebase/ui/auth/provider/FacebookProvider.java
```diff
@@ -182,9 +201,32 @@ private IdpResponse createIdpResponse(String email, LoginResult loginResult) {
...
+                public void onCancel() {
+                    FacebookProvider.this.onCancel();
+                }
+
+                @Override
+                public void onError(FacebookException error) {
+                    FacebookProvider.this.onError(error);
+                }
+            });
+            loginManager.logInWithPublishPermissions((Activity) mCallbackObject,
```
SUPERCILEX: @samtstern Here's where I'm stuck. The code works ([see video](https://goo.gl/photos/t37UGVG4FiuQRTdx9)) for `AuthMethodPickerActivity` of course, but we can't do this anywhere else. This means we would have to make Facebook special and save the activity in some other way. (We can't save it from `startLogin` because the activity might die between then and when we get the callback.)

#### Some context

You were right, Facebook makes this harder than I originally thought. You can only request read permissions in the read method and write permissions in the write method. This raises the problem that we have to request a login twice which I feel is bad UX. In addition, keeping track of all write permissions is pain and I doubt anyone will be paying attention if Facebook updates their permission list.

#### My thoughts

The way it turned out is kinda gross and I feel like it's only going to get worse if we do it right. In addition, I think this feature extends beyond the scope of **Firebase**UI. The main point of this library is to connect various auth providers with a Firebase account. I feel like messing around with complex permissions that require approval from Facebook shouldn't really be our problem, especially since asking for those permissions happens after we've already accomplished our goal of creating a Firebase account. Given the above, I think we **shouldn't** support requesting publish permissions. @samtstern What are your thoughts on this?

samtstern: I agree with you.  Facebook's guidelines say not to request publish permissions at the time of the first login and that's all we really handle.  If this had turned out super easy and clean I would have said we can put it in, but not worth the trouble!

Thanks for investigating.

PR #593, opened 2017-02-15
Magically apply codestyle for all FirebaseUI developers

SUPERCILEX:
@samtstern I'm really excited about this one!!! 😄 I learned about this while working on another open-source project that doesn't use hungarian notation: when working on projects like those, I'm used to having to switch to a different codestyle that doesn't make Intellij automatically add m and s prefixes to fields, but in this case my code style had already been updated which blew my mind. Anyway, the trick is to have the `codeStyleSettings.xml` file pre-loaded with the following option:
```xml
<option name="USE_PER_PROJECT_SETTINGS" value="true" />
```
This makes Intellij prioritize the project codestyle vs your own style. Mind blown! 💥 🚀 😄 This also means that any updates we make to the style will be used by all contributors. Pretty awesome, right?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.iml
 /local.properties
 .idea
+!.idea/codeStyleSettings.xml
 .DS_Store
 build
 google-services.json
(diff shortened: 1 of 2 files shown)


samtstern:
Cool!  Let's give this a try.  If it causes pain, we can roll this back easily.

PR #594, opened 2017-02-15
Bump SDK version to 10.2.0

samtstern:


samtstern:
This should contain the fix for #409

SUPERCILEX:
OMG finally!!!!! Can't wait to check the new method count!!!

On a more serious note, do you think this concerns us?
> Improved the behavior of ID tokens with Smart Lock. Apps must now explicitly request a token by calling setIdTokenRequested(true). Also, apps can specify the audience and nonce values for a token using the setServerClientId() and setIdTokenNonce() methods.

samtstern:
@SUPERCILEX nope we don't get ID tokens directly from SmartLock, although there's a strong argument that we should!

SUPERCILEX:
@samtstern SGTM!

PR #595, opened 2017-02-16
Lock IdpConfigs and scopes

SUPERCILEX:
While working on #592, I accidentally modified the `IdpConfig` scopes so this PR prevents that.


.idea/codeStyleSettings.xml
```diff
@@ -493,4 +493,4 @@
     </option>
     <option name="USE_PER_PROJECT_SETTINGS" value="true" />
   </component>
-</project>
+</project>
```
SUPERCILEX: Intellij obsessively removes the extra line at the end of this file so whatever.

auth/src/main/java/com/firebase/ui/auth/ui/FlowParameters.java
```diff
@@ -92,11 +94,8 @@ public FlowParameters createFromParcel(Parcel in) {
             int themeId = in.readInt();
             int logoId = in.readInt();
             String termsOfServiceUrl = in.readString();
-            int smartLockEnabledInt = in.readInt();
-            int allowNewEmailAccountsInt = in.readInt();
-
-            boolean smartLockEnabled = smartLockEnabledInt != 0;
-            boolean allowNewEmailAccounts = allowNewEmailAccountsInt != 0;
```
SUPERCILEX: This seems like unnecessary abstraction.

samtstern:
No objections here!

PR #596, opened 2017-02-16
Fix broken `minSdkVersion`

SUPERCILEX:
@samtstern I started working on #544 again, but nothing would compile. Eventually I realized that we forgot to bump the minSdkVersion which really confuses me because the Travis build should have failed. Not sure why they passed... 😐

### DIFF
--- a/constants.gradle
+++ b/constants.gradle
@@ -1,10 +1,15 @@
-project.ext.firebase_version = '10.2.0'
-project.ext.support_library_version = '25.1.1'
+project.ext {
+    submodules = ['database', 'auth', 'storage']
+    group = 'com.firebaseui'
+    version = '1.1.1'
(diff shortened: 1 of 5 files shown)


constants.gradle
```diff
@@ -1,10 +1,15 @@
...
-project.ext.compileSdk = 25
-project.ext.targetSdk = 25
+    compileSdk = 25
+    targetSdk = 25
+    minSdk = 14
+
+    buildTools = '25.0.2'
+
+    firebaseVersion = '10.2.0'
+    supportLibraryVersion = '25.1.1'
```
SUPERCILEX: I cleaned this up a little and made these names more consistent with `compileSdk`, `targetSdk`, etc.

samtstern:
No idea why the build didn't fail either ... strange indeed.  Thanks!

PR #609, opened 2017-02-23
Remove all deprecated stuff

SUPERCILEX:
@samtstern What do you think, should we do it? On the one hand, this removes 400+ lines of code and cleans things up nicely, but on the other hand it could make it harder for people to upgrade if they were still using the deprecated stuff.


SUPERCILEX:
Something else I just thought of: making any decision sets a precedent so if we don't remove stuff, people might expect deprecations to never be removed in the future.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -155,87 +149,23 @@ public static int getDefaultTheme() {
...
-     * Signs the current user out, if one is signed in.
-     *
      * @param activity the activity requesting the user be signed out
      * @return A task which, upon completion, signals that the user has been signed out ({@link
      * Task#isSuccessful()}, or that the sign-out attempt failed unexpectedly !{@link
      * Task#isSuccessful()}).
      */
     public Task<Void> signOut(@NonNull FragmentActivity activity) {
         // Get Credentials Helper
-        GoogleSignInHelper credentialsHelper = GoogleSignInHelper.getInstance(activity);
```
samtstern: Note: in the future let's try to avoid changes like this where we do a cosmetic refactor within another change.  It makes the diff larger and distracts from your main purpose (which is killing deprecated code).

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -254,12 +184,8 @@ public Void then(@NonNull Task<GoogleApiClient> task) throws Exception {
      */
     public Task<Void> delete(@NonNull FragmentActivity activity) {
         // Initialize SmartLock helper
-        CredentialTaskApi credentialHelper = GoogleSignInHelper.getInstance(activity);
-
-        return getDeleteTask(credentialHelper);
```
samtstern: Same comment here, I have no issue with inlining this method but it's not relevant to the title of the PR.

auth/src/main/java/com/firebase/ui/auth/ui/ResultCodes.java
```diff
@@ -1,20 +0,0 @@
...
-@Deprecated
-public class ResultCodes {
-
-    /**
-     * Sign in failed due to lack of network connection
-     *
-     * @deprecated Please use {@link com.firebase.ui.auth.ErrorCodes#NO_NETWORK}
-     **/
-    @Deprecated
-    public static final int RESULT_NO_NETWORK = com.firebase.ui.auth.ErrorCodes.NO_NETWORK;
```
samtstern: We should probably keep a running list of 1.x --> 2.x upgrade tips since something like this will be a little confusing.  Not sure what the best format for this is, maybe an MD file?

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
SUPERCILEX: Yeah, sorry about that. These changes were introduced when deprecating the old `delete(Activity)` method which is why I undid them here, but I can see your point of how that's confusing.

auth/src/main/java/com/firebase/ui/auth/ui/ResultCodes.java
SUPERCILEX: @samtstern Totally agree with you, 1113cc0 should take care of this.

samtstern:
@SUPERCILEX this CL has been here forever ... I've been thinking about it and 2.0 is the time to kill anything that was `@Deprecated`.  Let's do this. 

Do you have any reservations about merging this change?

SUPERCILEX:
@samtstern Nope, LGTM! 😄 

We're still keeping the deprecated [`setProviders` method](https://github.com/firebase/FirebaseUI-Android/blob/version-2.0.0-dev/auth/src/main/java/com/firebase/ui/auth/AuthUI.java#L534-L539), but I think that's the right desision because we don't want to confuse or frustrate people too much.

samtstern:
Yeah we have to keep that method since it was never actually released as `@Deprecated`.  But for things that have been deprecated for a release or two it's a good time to nuke em.

SUPERCILEX:
Cool beans! 😄

PR #610, opened 2017-02-23
Make ChooserActivity.java way prettier

SUPERCILEX:
The new one is so refreshing! 😄 

# Old

image

# New

image

### DIFF
--- a/app/src/main/java/com/firebase/uidemo/ChooserActivity.java
+++ b/app/src/main/java/com/firebase/uidemo/ChooserActivity.java
@@ -14,15 +14,18 @@
 
 package com.firebase.uidemo;
 
-import android.content.Context;
 import android.content.Intent;
+import android.graphics.Typeface;
 import android.os.Bundle;
+import android.support.annotation.StringRes;
 import android.support.v7.app.AppCompatActivity;
+import android.support.v7.widget.LinearLayoutManager;
+import android.support.v7.widget.RecyclerView;
+import android.util.TypedValue;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
-import android.widget.ArrayAdapter;
-import android.widget.ListView;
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/firebase/uidemo/ChooserActivity.java
```diff
@@ -31,73 +34,99 @@
...
-    public static class MyArrayAdapter extends ArrayAdapter<Class> {
-
-        private Context mContext;
+        @Override
+        public ActivityStarterHolder onCreateViewHolder(ViewGroup parent, int viewType) {
+            View itemView = LayoutInflater.from(parent.getContext())
+                    .inflate(R.layout.activity_chooser_item, parent, false);
+
+            // Set custom margins
+            int marginsInPx = (int) TypedValue.applyDimension(
```
samtstern: What are we doing in this custom margin code that we can't do in XML?

SUPERCILEX: I'm [including `android.R.layout.simple_list_item_2`](https://github.com/firebase/FirebaseUI-Android/pull/610/files#diff-a6396a78e0409ffd093003e115647a15R11) which doesn't let me set margins in XML. If you'd like, we can use our own custom layout instead of the native one. I just felt that it was easier to do this in code rather than copying the XML, but I'm fine with it either way.

samtstern: We don't really need that layout, all we need are two TextViews in a vertical layout where one is bold.  We can steal the ids (text1 and text2) if we want.

SUPERCILEX: @samtstern SGTM, I copied over the layout in 872f3724fe97b2fef0d1ad800d1695c818086d24.

samtstern:
Nice and simple now, will merge.

PR #618, opened 2017-03-03
Permanently fix NPEs when attempting to finish Smart Lock fragments

SUPERCILEX:
See https://github.com/firebase/FirebaseUI-Android/issues/549#issuecomment-283827188.


samtstern:
I don't love this hack but it's a few lines of code and seems to work, so I will merge.

This is why I hate Fragments :smile:

SUPERCILEX:
@samtstern 😆. Yeah, we make it especially hard on ourselves because of `setRetainInstance(true)`. Maybe in the future we could find a better solution... Anyway, glad to see this merged!

Issue #623, opened 2017-03-06
Support Facebook login without email


puilp0502:
The gist is same as #352.  

Currently, if the GraphRequest response object does not have an email field for whatever reason (e.g. User's denial, [Facebook's error](http://stackoverflow.com/questions/16630972/facebook-graph-api-wont-return-email-address), etc), it just logs JSONException and fail([src](https://github.com/firebase/FirebaseUI-Android/blob/master/auth/src/main/java/com/firebase/ui/auth/provider/FacebookProvider.java#L148-L151)). 
 However, [TwitterProvider.java](https://github.com/firebase/FirebaseUI-Android/blob/master/auth/src/main/java/com/firebase/ui/auth/provider/TwitterProvider.java#L99-L102) just creates user with no email, and I think FacebookProvider.java should do the same. 
 The current implementation, if not provided with email, just returns to the AuthMethodPicker without providing user with additional information, which is very bad in terms of UX.

The source doesn't seem to be complicated, so I would be able to write a PR for this if desired.

SUPERCILEX:
@puilp0502 Smart lock won't work if we do that, but I can see how it's better to just get the user logged in anyway. Unless @samtstern has any objects I think we should let the user deny us the email permission. 

@puilp0502 do you want to submit a PR for this? If not, I'll go ahead and do it.

puilp0502:
@SUPERCILEX Yeah, I'll write a PR for that.

SUPERCILEX:
SGTM!

samtstern:
@puilp0502 thanks for the PR.  I have a question about this in general: do you have the 'multiple accounts per email' switch turned on in the Firebase console?

puilp0502:
No, it's set to 'One account per email address'.

cpauwels:
Hi, is there an estimated release date for these fixes?

This is quite the issue in developing countries where many users are signed in with their phone number instead of an email, leading to the errors.

We launched our beta yesterday and did 8 user testing interviews, and 2 of our users could not sign in. This is a huge issue if 25% of our user base can't create an account.

samtstern:
@cpauwels this will be in version `2.0.0`.  No release date chosen yet.

You can always do this:
  * Clone the repo
  * Checkout `version-2.0.0-dev`
  * Edit `constants.gradle` to make the version number 2.0.0
  * Run `./gradlew :libraryPublishAllToMavenLocal`
  * Use the local version in your app

PR #624, opened 2017-03-06
Support Facebook Login without email

puilp0502:
Addresses #623.

I didn't know if I should bump up the version, so I left it as it is now.

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/provider/FacebookProvider.java
+++ b/auth/src/main/java/com/firebase/ui/auth/provider/FacebookProvider.java
@@ -146,8 +146,8 @@ public class FacebookProvider implements IdpProvider, FacebookCallback<LoginResu
                                 String email = object.getString("email");
                                 onSuccess(email, loginResult);
                             } catch (JSONException e) {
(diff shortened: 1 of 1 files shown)


samtstern:
@puilp0502 just wanted to say I know this is lingering, I have been caught up on some other stuff and will test this soon.  Thank you for doing it!

samtstern:
@puilp0502 FYI I changed the target branch to `version-2.0.0-dev` as that's where the next version will come from.  You may have to change your local setup a bit, let me know if that's confusing.

samtstern:
Ok I just went and got a new phone number (thanks Google Voice) and then made a Facebook account that has no email.  Then I signed in and got the following (see below).

I am surprised that signing in like that even works!  I guess email is not a hard requirement for Facebook accounts (learn something new about your own products every day haha).

I am fine to merge this change, but @amandle can you chime in and let me know if this could cause any potential problems?

screenshot from 2017-03-09 12 02 03

no_email

amandle:
LGTM, Twitter also doesn't require email be shared, so I don't foresee this change causing problems

PR #629, opened 2017-03-13
Finish the KickOffActivity on dialog tap outside

amandle:
Fixes an issue where tapping outside the reauth dialog would result in an invisible activity hanging around blocking clicks.


samtstern:
@amandle good catch!

PR #632, opened 2017-03-14
Cleanup index array

SUPERCILEX:


database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
```diff
@@ -162,20 +162,18 @@ protected void onKeyMoved(DataSnapshot data, int index, int oldIndex) {
...
-                                       oldIndex);
+            notifyChangeEventListeners(EventType.MOVED, snapshot, index, oldIndex);
         }
     }
 
     protected void onKeyRemoved(DataSnapshot data, int index) {
         String key = data.getKey();
-        mDataRef.child(key).removeEventListener(mRefs.remove(mDataRef.getRef().child(key)));
+        ValueEventListener listener = mRefs.remove(mDataRef.getRef().child(key));
+        if (listener != null) mDataRef.child(key).removeEventListener(listener);
```
SUPERCILEX: My app keeps crashing randomly with NPEs so I'm starting to think this could be getting called twice in really weird cases. Unfortunately, I've never been able to consistently reproduce it, but it can't hurt to add a null check.

samtstern:
LGTM

Issue #631, opened 2017-03-14
[Feature request] Callbacks for data population 


TheCraftKid:
I'm probably wrong, but I don't think the `FirebaseRecyclerAdapter` and its children have callbacks for when the data is about to be initially populated or when it's finished populating items. I checked docs and code for the current and 2.0.0 branches, and I didn't see listeners or callbacks that would allow someone to display a progress bar if the database is taking a lot of time to fetch some items on a slow network and hide it when the items are populated. I know it's possible to do this for simple references like so:
```java
dataRef.addValueEventListener(new ValueEventListener() {
    @Override
    public void onDataChange(DataSnapshot dataSnapshot) {
        Data data = dataSnapshot.getValue(Data.class);
        textView.setText(data.getText());
        progressBar.setVisibility(View.GONE);
    }

    @Override
    public void onCancelled(DatabaseError databaseError) {
        // ...
    }
});
```
However, I'm wondering if something like this should be implemented in the FirebaseRecyclerAdapters:
```java
dataAdapter.setOnPopulatedListener(new OnPopulatedListener() {
    @Override
    public void onPopulation() {
        progressBar.setVisibility(View.GONE);
    }
});

// Or perhaps, probably the right choice, by subclassing

public void onInitialPopulation() {
    progressBar.setVisibility(View.GONE);
}
```

SUPERCILEX:
The callback is [`onDataChanged`](https://github.com/firebase/FirebaseUI-Android/blob/master/database/src/main/java/com/firebase/ui/database/FirebaseRecyclerAdapter.java#L217) which is available on the master branch.

A note of caution: there is little support for this method. It is completely broken for the index adapters and will never get called in certain situations when offline. I've just been using `onChildChanged` to know if I should show an empty list hint.

TheCraftKid:
No wonder I didn't catch that small line of code! :open_mouth: Does the 2.0.0 branch include this method?

SUPERCILEX:
@TheCraftKid Yup! (Though it still isn't supported for the index stuff and probably won't be.)

TheCraftKid:
Why won't it be supported for the indexing? Is it an implementation issue? Also, in what situations is onDataChanged not called?

SUPERCILEX:
@TheCraftKid So many questions! 😄 Well, now #633 should take care of it.

See https://github.com/firebase/FirebaseUI-Android/pull/477#issuecomment-270283875 for when `onDataChanged` isn't called.

Issue #546, opened 2017-01-26
onDataChanged is completely broken for FirebaseIndexArray


SUPERCILEX:
@samtstern my brain is about to melt with everything that's going on right now, but I upgraded our sample to use the `FirebaseIndexRecyclerView` in #544 and immediately noticed that the "start chatting at the bottom" message wasn't going away when the chats were populated. I didn't think about this when reviewing @puf's PR, but `FirebaseIndexArray` isn't done updating when `FirebaseArray` updates because it just uses `FirebaseArray`'s keys to get the actual data. So basically what happens is that `FirebaseArray` calls `onDataChanged`, but the new data isn't actually there yet in `FirebaseIndexArray`.

PR #633, opened 2017-03-14
Add support for batch events in the index stuff

SUPERCILEX:
#631 and #546

### DIFF
--- a/database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
+++ b/database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
@@ -46,7 +46,8 @@ public class FirebaseIndexArray<T> extends CachingObservableSnapshotArray<T> imp
     private List<String> mKeysWithPendingData = new ArrayList<>();
     /**
      * Moves or deletions don't need to fetch new data so they can be performed instantly once the
-     * backing {@link FirebaseArray} is done updating.
+     * backing {@link FirebaseArray} is done updating. This will be true if the backing {@link
+     * FirebaseArray} is in the middle of an update, false otherwise.
      */
     private boolean mHasPendingMoveOrDelete;
 
--- a/database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
+++ b/database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
@@ -38,6 +38,9 @@ public class FirebaseIndexArray<T> extends CachingObservableSnapshotArray<T> imp
     private FirebaseArray<String> mKeySnapshots;
     private List<DataSnapshot> mDataSnapshots = new ArrayList<>();
 
+    private List<String> mQueuedAdditions = new ArrayList<>();
+    private boolean mHasQueuedInstantOps;
+
     /**
      * Create a new FirebaseIndexArray that parses snapshots as members of a given class.
      *
@@ -97,7 +100,10 @@ public class FirebaseIndexArray<T> extends CachingObservableSnapshotArray<T> imp
 
     @Override
     public void onDataChanged() {
-        // No-op, we don't listen to batch events for the key ref
+        if (mHasQueuedInstantOps) {
+            notifyListenersOnDataChanged();
+            mHasQueuedInstantOps = false;
+        }
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
```diff
@@ -221,16 +231,21 @@ public void onDataChange(DataSnapshot snapshot) {
...
                     mDataSnapshots.add(index, snapshot);
                     notifyChangeEventListeners(ChangeEventListener.EventType.ADDED, snapshot, index);
+
+                    mQueuedAdditions.remove(key);
+                    if (mQueuedAdditions.isEmpty()) notifyListenersOnDataChanged();
                 } else {
                     // We already know about this data, just update it
                     updateData(index, snapshot);
                     notifyChangeEventListeners(ChangeEventListener.EventType.CHANGED, snapshot, index);
+                    notifyListenersOnDataChanged();
```
SUPERCILEX: I can't think of a better way to do this without having some sort of looper and/or timer. @samtstern Do you have any ideas?

database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
```diff
@@ -38,6 +38,9 @@
     private FirebaseArray<String> mKeySnapshots;
     private List<DataSnapshot> mDataSnapshots = new ArrayList<>();
 
+    private List<String> mQueuedAdditions = new ArrayList<>();
```
samtstern: Took me a little while to figure out what these variables do.  I'd suggest a rename and some comments.  How about this (and these are just my first stab so feel free to think up better ones):

```java
// Keys for which we have received an 'ADDED' event but have not yet downloaded the indexed data
private List<String> mKeysWithPendingData = new ArrayList();

// True if a key has been modified, moved, or deleted but we have not propagated the change yet. False otherwise.
private boolean mHasPendingModifyOrDelete;
```

database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
```diff
@@ -97,7 +100,10 @@ public void onChildChanged(EventType type, DataSnapshot snapshot, int index, int
 
     @Override
     public void onDataChanged() {
-        // No-op, we don't listen to batch events for the key ref
+        if (mHasQueuedInstantOps) {
+            notifyListenersOnDataChanged();
```
samtstern: For an "instant operation" will there be two batch events filed?  I see one here, and one in the DataRefListener.

SUPERCILEX: @samtstern No, there shouldn't be. For deletions, there will be two notifications if the dev deletes the content and then the key. However, if they do it the suggested way of deleting the key first and then the data, they'll be fine. As for moves, there definitely will only be one notification.

samtstern:
@SUPERCILEX can you explain your very first comment "This adds support for batch additions/removals, but it doesn't really work for changes to a ref."

database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
SUPERCILEX: @samtstern Dang, those are much better names. Thanks! 😄

I changed the documentation a bit and I think you're confused about `mHasPendingModifyOrDelete`. It only support moves and deletions. This is the big issue with this PR, it doesn't work for changes. See my comment here: https://github.com/firebase/FirebaseUI-Android/pull/633#discussion_r105979080. We would need some sort of gross timer or something to wait for a group of writes. 😢

SUPERCILEX:
@samtstern Oh, didn't noticed you posted your other comment. See https://github.com/firebase/FirebaseUI-Android/pull/633#discussion_r113342093. Basically, changes require listening to all the different nodes so we would have to somehow group together changes from all the different nodes to notify the dev properly. Right now, we just notify them for every change.

samtstern:
@SUPERCILEX I see, I think that's reasonable and it's not a regression.

SUPERCILEX:
@samtstern Awesome!

PR #635, opened 2017-03-15
Add getSnapshots method to adapters

SUPERCILEX:
@samtstern I think it would be really useful to have this available for when you are trying to get the `DataSnapshot`s outside of the adapter. For example, I have a method to add an item at the end of the list by priority. Since I don't update all refs when an item is deleted, I just get the highest priority + 1:
```java
public static int getHighestIntPriority(ObservableSnapshotArray snapshots) {
    int highest = 0;
    for (DataSnapshot snapshot : snapshots) {
        int priority = (int) snapshot.getPriority();
        if (priority > highest) highest = priority;
    }
    return highest;
}
```

This would be harder to do without an `Adapter#getSnapshots()` method easily available to retrieve the `FirebaseArray`.

### DIFF
--- a/database/src/main/java/com/firebase/ui/database/FirebaseAdapter.java
+++ b/database/src/main/java/com/firebase/ui/database/FirebaseAdapter.java
@@ -20,4 +20,6 @@ interface FirebaseAdapter<T> extends ChangeEventListener {
     T getItem(int position);
 
     DatabaseReference getRef(int position);
+
+    ObservableSnapshotArray<T> getSnapshots();
 }
--- a/database/src/main/java/com/firebase/ui/database/FirebaseListAdapter.java
+++ b/database/src/main/java/com/firebase/ui/database/FirebaseListAdapter.java
@@ -113,6 +113,11 @@ public abstract class FirebaseListAdapter<T> extends BaseAdapter implements Fire
     }
 
     @Override
+    public ObservableSnapshotArray<T> getSnapshots() {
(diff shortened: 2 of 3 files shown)


puf:
If you expose the underlying array of snapshots from an adapter, you remove
all options to change anything about the implementation in the future.

If you have a use-case that requires you to have access to all snapshots,
can't you construct the adapter out of the `ObservableSnapshotArray`
yourself? That way make it explicit that you own the underlying array for
this implementation, instead of putting it in the adapter's "contract" by
definition.

SUPERCILEX:
@puf Hmmm, I guess that sounds reasonable. However, we can't change the underlying data structure anyway because the `mSnapshots` field is `protected`. @samtstern @puf As a side note, shouldn't all our protected fields be final? (I just noticed that)

puf:
Why do you think they should be?

SUPERCILEX:
@puf I don't think we want people to be able to do `mSnapshots = new FirebaseArray(...)` in the middle of stuff. (That should be done with `RecyclerView#swapAdapter(...)`.) Either way, I don't see any downsides to reducing mutability... do you?

puf:
By "In the middle of stuff" do you mean in their derived classes? Please be
explicit or provide an example of something that you would want to disallow.


If the field should not be access by subclasses we should mark it private
and provide a protected getter.

SUPERCILEX:
@puf See #639.

PR #639, opened 2017-03-17
Make protected fields final in adapters

SUPERCILEX:
@puf @samtstern Without this PR, devs could do stuff like this which would be nonsensical:
```java
new FirebaseRecyclerAdapter(...) {
    @Override
    protected void populateViewHolder(RecyclerView.ViewHolder holder, Object model, int position) {
        mSnapshots = new FirebaseArray<Team>(...);
        mViewHolderClass = null;
        mModelLayout = R.layout.new_layout;
    }
};

### DIFF
--- a/database/src/main/java/com/firebase/ui/database/FirebaseRecyclerAdapter.java
+++ b/database/src/main/java/com/firebase/ui/database/FirebaseRecyclerAdapter.java
@@ -31,9 +31,9 @@ public abstract class FirebaseRecyclerAdapter<T, VH extends RecyclerView.ViewHol
         extends RecyclerView.Adapter<VH> implements FirebaseAdapter<T> {
     private static final String TAG = "FirebaseRecyclerAdapter";
 
-    protected ObservableSnapshotArray<T> mSnapshots;
-    protected Class<VH> mViewHolderClass;
-    protected int mModelLayout;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: database/src/main/java/com/firebase/ui/database/FirebaseRecyclerAdapter.java [hunk 1]
```diff
@@ -31,9 +31,9 @@
         extends RecyclerView.Adapter<VH> implements FirebaseAdapter<T> {
     private static final String TAG = "FirebaseRecyclerAdapter";
 
-    protected ObservableSnapshotArray<T> mSnapshots;
-    protected Class<VH> mViewHolderClass;
-    protected int mModelLayout;
+    protected final ObservableSnapshotArray<T> mSnapshots;
```
SUPERCILEX: @puf Were you saying we should make a `protected ObservableSnapshotArray<T> getSnapshots()` method and make this field private?

SUPERCILEX: @samtstern What do you think? (Gentle bump 😁)

samtstern:
@SUPERCILEX sorry this sat for so long.  I am happy with this change, there's no *good* reason to change class, layout, or the underlying array on the fly.

PR #643, opened 2017-03-21
Update README.md

koceeng:
I've seen (and answered some) a lot of question in StackOverflow that complains about how their RecyclerView not get updated. And some of them got it fixed after they only change the recyclerView.setHasFixedSize to false. And I wonder why they make it true in the first place if they don't know what it does, so I come here and check, and there it is. I don't know why it is there, so please decline my change if it does any kind of harm to this great repository.

### DIFF
--- a/database/README.md
+++ b/database/README.md
@@ -268,7 +268,7 @@ Just like we did for `FirebaseListAdapter`, we'll create an anonymous subclass f
 
 ```java
 RecyclerView recycler = (RecyclerView) findViewById(R.id.messages_recycler);
-recycler.setHasFixedSize(true);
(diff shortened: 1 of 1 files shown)


samtstern:
@koceeng thanks for such a clear PR description.

I am happy to merge this once you sign the CLA (see instructions by @googlebot).  If you don't want to do that for some reason, let me know and I will make this change myself.

koceeng:
I've signed it, sorry for the late answer

samtstern:
@koceeng thanks again!

koceeng:
Happy to help

Issue #644, opened 2017-03-21
Twitter Session Even After Logout are Persistent


chiragbhatia94:
My environment
  * Android device: Moto G3
  * Android OS version: Nougat 7.1.1
  * Google Play Services version: 10.2.98
  * Firebase/Play Services SDK version: 1.2.0
  * FirebaseUI version: 1.2.0
  
Describe the problem:
Once a user logs in using twitter as a provider. Its twitter session is persistent even after he logs out of the firebase account.

Changes required in 
com.firebase.ui.auth.provider.TwitterProvider

Proposed solution:
It can be removed by flushing the cookies everytime we call TwitterProvider.startLogin function

This reduces the effort for logging out every time you want to login using different user account
This is similar to what it is there in GoogleProvider  
#### Relevant Code:

Problem starts here
@Override
    public void startLogin(Activity activity) {
        mTwitterAuthClient.authorize(activity, this);
    }

Solution
@SuppressWarnings("deprecation")
    private static void clearCookies(Context context) {

        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP_MR1) {
            Log.d(TAG, "Using clearCookies code for API >=" + String.valueOf(Build.VERSION_CODES
                    .LOLLIPOP_MR1));
            CookieManager.getInstance().removeAllCookies(null);
            CookieManager.getInstance().flush();
        } else {
            Log.d(TAG, "Using clearCookies code for API <" + String.valueOf(Build.VERSION_CODES
                    .LOLLIPOP_MR1));
            CookieSyncManager cookieSyncMngr = CookieSyncManager.createInstance(context);
            cookieSyncMngr.startSync();
            CookieManager cookieManager = CookieManager.getInstance();
            cookieManager.removeAllCookie();
            cookieManager.removeSessionCookie();
            cookieSyncMngr.stopSync();
            cookieSyncMngr.sync();
        }
    }

@Override
    public void startLogin(Activity activity) {
        clearCookies(activity);
        mTwitterAuthClient.authorize(activity, this);
    }

samtstern:
@chiragbhatia94 yikes, this is a pretty obvious miss by us.  Can't believe we went so long with this behavior!

In the signOut function, looks like we're only calling Facebook, Google, and Firebase sign out:
https://github.com/firebase/FirebaseUI-Android/blob/master/auth/src/main/java/com/firebase/ui/auth/AuthUI.java#L165

[...]

PR #645, opened 2017-03-22
Logout of Twitter

SUPERCILEX:
#644

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/AuthUI.java
+++ b/auth/src/main/java/com/firebase/ui/auth/AuthUI.java
@@ -26,6 +26,7 @@ import android.support.annotation.StyleRes;
 import android.support.v4.app.FragmentActivity;
 
 import com.facebook.login.LoginManager;
+import com.firebase.ui.auth.provider.TwitterProvider;
 import com.firebase.ui.auth.ui.FlowParameters;
 import com.firebase.ui.auth.ui.idp.AuthMethodPickerActivity;
 import com.firebase.ui.auth.util.CredentialTaskApi;
@@ -49,6 +50,7 @@ import com.google.firebase.auth.FirebaseAuth;
 import com.google.firebase.auth.FirebaseUser;
 import com.google.firebase.auth.GoogleAuthProvider;
 import com.google.firebase.auth.TwitterAuthProvider;
+import com.twitter.sdk.android.Twitter;
 
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -239,6 +247,10 @@ public Void then(@NonNull Task<GoogleApiClient> task) throws Exception {
         // Facebook sign out
         LoginManager.getInstance().logOut();
 
+        // Twitter sign out
```
SUPERCILEX: @samtstern Signing out of Twitter and Facebook should probably a method (something like `signOutProviders`), but that will cause more thrashing in #609. Would you like me to do the refactor anyway and then undo it in #609 or are we not going through with #609?

samtstern: Let's keep sign out all as one method for now.

SUPERCILEX: SGTM!

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -192,6 +196,10 @@ public Void then(@NonNull Task<GoogleApiClient> task) throws Exception {
...
+        if (!Fabric.isInitialized()) TwitterProvider.initialize(activity);
+        Twitter.logOut();
```
samtstern: I think we also want to call:
`Twitter.getSessionManager().clearActiveSession();`

SUPERCILEX: @samtstern I thought so too, but `Twitter.logOut()` already does that for us:
```java
// In Twitter.java
public static void logOut() {
    checkInitialized();
    getInstance().core.logOut(); // core == TwitterCore
}

// In TwitterCore.java
public void logOut() {
    checkInitialized();
    final SessionManager<TwitterSession> sessionManager = getSessionManager();
    if (sessionManager != null) {
        sessionManager.clearActiveSession();
    }
}

public static SessionManager<TwitterSession> getSessionManager() {
    checkInitialized();
    return getInstance().core.getSessionManager(); // Same reference as logOut in TwitterCore.java
}
```

Actually, now that I'm paying more attention, it looks like that's all `Twitter#logOut()` does.

The docs for `TwitterCore#logOut()` say it logs the user out, so I think that's all we have to do:
```java
/**
 * Logs out the user, clearing user session. This will not make a network request to invalidate
 * the session.
 *
 * @throws java.lang.IllegalStateException if {@link io.fabric.sdk.android.Fabric}
 *          or {@link TwitterCore} has not been initialized.
 */
```

samtstern: Thanks for digging, works for me!

The Twitter docs don't have anything at all on signOut ... oversight I guess.

PR #670, opened 2017-04-23
Improve email UX

SUPERCILEX:
- Add done action to IME
- Capitalize words in the name field

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/ui/ImeHelper.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/ImeHelper.java
@@ -1,0 +1,23 @@
+package com.firebase.ui.auth.ui;
+
+import android.view.KeyEvent;
+import android.view.inputmethod.EditorInfo;
+import android.widget.EditText;
+import android.widget.TextView;
+
+public class ImeHelper {
+    public static void addImeOnDoneListener(EditText doneEditText, final Runnable doneAction) {
+        doneEditText.setOnEditorActionListener(new TextView.OnEditorActionListener() {
+            @Override
+            public boolean onEditorAction(TextView view, int actionId, KeyEvent event) {
+                if (event != null
+                        && event.getKeyCode() == KeyEvent.KEYCODE_ENTER
+                        || actionId == EditorInfo.IME_ACTION_DONE) {
+                    doneAction.run();
+                    return true;
+                }
+                return false;
+            }
+        });
+    }
+}
--- a/auth/src/main/java/com/firebase/ui/auth/ui/ImeHelper.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/ImeHelper.java
@@ -1,19 +1,26 @@
 package com.firebase.ui.auth.ui;
 
+import android.support.annotation.RestrictTo;
 import android.view.KeyEvent;
 import android.view.inputmethod.EditorInfo;
 import android.widget.EditText;
 import android.widget.TextView;
 
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 public class ImeHelper {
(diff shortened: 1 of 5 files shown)


auth/src/main/res/layout/register_email_layout.xml
```diff
@@ -46,7 +46,7 @@
                 style="@style/FirebaseUI.EditText"
                 android:layout_width="match_parent"
                 android:layout_height="wrap_content"
-                android:inputType="textPersonName"
+                android:inputType="textCapWords"
```
SUPERCILEX: `textPersonName` wasn't working for me for some reason and it doesn't work in my app either so I'm not sure what's going on. `textCapWords` replicates the behavior we actually want so it should be fine.

auth/src/main/java/com/firebase/ui/auth/ui/ImeHelper.java
```diff
@@ -0,0 +1,23 @@
+package com.firebase.ui.auth.ui;
+
+import android.view.KeyEvent;
+import android.view.inputmethod.EditorInfo;
+import android.widget.EditText;
+import android.widget.TextView;
+
+public class ImeHelper {
```
samtstern: nit: for public classes let's add JavaDoc and also `@RestrictTo(LIBRARY)`

auth/src/main/java/com/firebase/ui/auth/ui/ImeHelper.java
```diff
@@ -0,0 +1,23 @@
...
+    public static void addImeOnDoneListener(EditText doneEditText, final Runnable doneAction) {
```
samtstern: I'd prefer to make up an interface here (`OnDonePressedListener`) or similar rather than using the generic `Runnable` since it makes usage clearer.  Even though Runnable is a compatible interface for what we want to do, if we used it everywhere we needed a no-args function we'd have a hard time managing.

auth/src/main/java/com/firebase/ui/auth/ui/accountlink/WelcomeBackPasswordPrompt.java
```diff
@@ -117,7 +120,7 @@ protected void onCreate(Bundle savedInstanceState) {
     public void onClick(View view) {
         final int id = view.getId();
         if (id == R.id.button_done) {
-            next(mEmail, mPasswordField.getText().toString());
+            run();
```
samtstern: Rather than calling `run` here, make `run()` call out to a no-args method like this:

```java
    // ...
    if (id == R.id.button_done) {
        nextWithEmailAndPassword();
    }
    // ...
}

// ...


@Override
public void run() {
    nextWithEmailAndPassword();
}

private voidNextWithEmailAndPassword() {
   next(mEmail, mPasswordField.getText().toString());
}
```

auth/src/main/java/com/firebase/ui/auth/ui/accountlink/WelcomeBackPasswordPrompt.java
```diff
@@ -55,7 +56,7 @@
  * the password before initiating a link.
  */
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-public class WelcomeBackPasswordPrompt extends AppCompatBase implements View.OnClickListener {
+public class WelcomeBackPasswordPrompt extends AppCompatBase implements View.OnClickListener, Runnable {
```
samtstern: My comments on this class extend to all the other classes that are using the new listener.

SUPERCILEX:
@samtstern Yup, all of your comments make total sense! 😄

samtstern:
Looks good to me now!  Gonna merge.

Issue #673, opened 2017-04-25
println needs a message - SaveSmartLock


plugie:
Android Device:
Manufacturer: MITO
Model: MITO_A10
Android API: 23
Android OS: 6.0.1
Brand: Google
RAM: 965.54MB
Google Play Service : 10.2.98
Firebase/Play Services SDK version : 10.2.0
FirebaseUI version : 1.2.0

Describe the problem:
Crash when try to login. But login is success. It means auth.getCurrentUser() is not null after crash is happen. 
The problem **doesn't happen** in other two of my mobile phone, i.e.: Samsung Galaxy S5 and honor 6

Steps to reproduce:
1. Login using firebase-ui. There are two methods : Google and Facebook
2. Touch Google sign in button
3. Choose my Google account
4. Then crash with error below : 

> Exception java.lang.NullPointerException: println needs a message
> android.util.Log.println_native (Log.java)
> android.util.Log.w (Log.java:181)
> com.firebase.ui.auth.util.signincontainer.SaveSmartLock.getInstance (SaveSmartLock.java)
> <OR>.onConnected (SaveSmartLock.java)
> <OR>.onConnectionFailed (SaveSmartLock.java)
> <OR>.onResult (SaveSmartLock.java)
> <OR>.onActivityResult (SaveSmartLock.java)
> <OR>.saveCredentialsOrFinish (SaveSmartLock.java)
> <OR>.onResult (SaveSmartLock.java)
> com.firebase.ui.auth.util.signincontainer.SaveSmartLock.getInstance (SaveSmartLock.java)
> <OR>.onConnected (SaveSmartLock.java)
> <OR>.onConnectionFailed (SaveSmartLock.java)
> <OR>.onResult (SaveSmartLock.java)
> <OR>.onActivityResult (SaveSmartLock.java)
> <OR>.saveCredentialsOrFinish (SaveSmartLock.java)
> <OR>.onResult (SaveSmartLock.java)
> com.google.android.gms.internal.zzaaf$zza.zzb (zzaaf.java)
> com.google.android.gms.internal.zzaaf$zza.handleMessage (zzaaf.java)
> android.os.Handler.dispatchMessage (Handler.java:102)
> android.os.Looper.loop (Looper.java:148)
> android.app.ActivityThread.main (ActivityThread.java:5417)
> java.lang.reflect.Method.invoke (Method.java)
[...]

PR #674, opened 2017-04-25
Fix all possible null Log.x calls

SUPERCILEX:
#673

@samtstern I looked at every `TaskFailureLogger` and `Log.x` call and I think I've got them all. Phew, glad we caught that! 😄 

PS: I wonder why the Google Play team decided to break backwards compatibility? Do you think that's a bug we should report?

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/ui/TaskFailureLogger.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/TaskFailureLogger.java
@@ -23,7 +23,7 @@ public class TaskFailureLogger implements OnFailureListener {
     private String mTag;
     private String mMessage;
 
-    public TaskFailureLogger(String tag, String message) {
+    public TaskFailureLogger(@NonNull String tag, @NonNull String message) {
         mTag = tag;
         mMessage = message;
     }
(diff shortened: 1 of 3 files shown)


SUPERCILEX:
@samtstern The Google Play Services version was bumped from `10.3.x` to `10.5.x` recently. I'm guessing they changed stuff in that version, but I'm just speculating at this point.

screenshot_20170425-110139

samtstern:
The behavior of `Log.x` shouldn't change, unless you're saying that Play Services is now returning different error messages?  That could be.  I am happy to merge this PR but want to know the root cause so I can bother someone internally about it.

SUPERCILEX:
@samtstern Yup, that's exactly what I'm saying. The two instances where we saw crashes was when we were passing in a `com.google.android.gms.common.api.Status#getStatusMessage()` for the `Log.x(..., message)`. AFIK, this never returned null before, but suddenly it's now returning `null` and `Log` throws a NPE if the message is null.

samtstern:
@SUPERCILEX ok makes sense, I have reported it.

SUPERCILEX:
@samtstern Awesome!

Issue #526, opened 2017-01-18
Improve IdpResponse error handling for auth sign-in


aviadmini:
Would be great to get original error codes (like [these](https://developers.google.com/android/reference/com/google/android/gms/auth/api/signin/GoogleSignInStatusCodes)) to get more flexibility in error resolution.

Also possibly move constants from [ErrorCodes](FirebaseUI-Android/auth/src/main/java/com/firebase/ui/auth/ErrorCodes.java) to IdpResponse itself (and annotate with `@IntDef `?)

samtstern:
@aviadmini one of the goals of this library is to hide the underlying errors from all of the different auth APIs that we use under the hood.  So exposing *all* error messages would be somewhat counterproductive.

Is there a specific error case you had in mind that we're not handling properly?

aviadmini:
@samtstern totally makes sense, it definitely shouldn't be primary way of error resolution

In my case I'd like to both have all the login logic solved in FUI auth activity and then in case of specific error codes like [INTERNAL_ERROR](https://developers.google.com/android/reference/com/google/android/gms/common/api/CommonStatusCodes#INTERNAL_ERROR) I'd like to retry automatically/show specific error message. I actually get this error more often than expected.

Actually it might be a good idea to catch this particular one in FUI and make a retry. For me it always succeeded login after INTERNAL_ERROR was returned (well, not exactly returned, but shown in logcat)

samtstern:
@aviadmini thanks for the clarification, that's very interesting.  Do you know under what conditions you get `INTERNAL_ERROR`?

aviadmini:
Unfortunately it seems to be random. Although I'm doing a lot of login-logout cycles with different constraints. Potentially it could mess up things

SUPERCILEX:
@aviadmini How would you use the different error codes? I have a solution in mind, but I'm having trouble seeing the end value to the dev once they receive a more detailed error...

Here's my proposed solution:
Error codes are kinda messy and they are limited in the amount of information they can provide. Instead, I was thinking of returning a custom `Exception` with a detailed error message. I can already think of a few main categories:
[...]

PR #675, opened 2017-04-26
Add solution to fetchProvidersForEmail failure

SUPERCILEX:
I moved this change from #309 to a separate PR where it makes more sense.

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/ui/idp/CredentialSignInHandler.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/idp/CredentialSignInHandler.java
@@ -20,7 +20,9 @@ import android.support.annotation.Nullable;
 import android.support.annotation.RestrictTo;
 import android.util.Log;
 
+import com.firebase.ui.auth.ErrorCodes;
 import com.firebase.ui.auth.IdpResponse;
+import com.firebase.ui.auth.ResultCodes;
 import com.firebase.ui.auth.ui.BaseHelper;
 import com.firebase.ui.auth.ui.TaskFailureLogger;
 import com.firebase.ui.auth.ui.User;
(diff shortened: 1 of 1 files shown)


auth/src/main/java/com/firebase/ui/auth/ui/idp/CredentialSignInHandler.java
```diff
@@ -82,9 +84,10 @@ public void onComplete(@NonNull Task<AuthResult> task) {
                             .addOnFailureListener(new OnFailureListener() {
                                 @Override
                                 public void onFailure(@NonNull Exception e) {
-                                    // TODO: What to do when signing in with Credential fails
-                                    // and we can't continue to Welcome back flow without
-                                    // knowing providers?
+                                    mHelper.finishActivity(
+                                            mActivity,
+                                            ResultCodes.CANCELED,
+                                            IdpResponse.getErrorCodeIntent(ErrorCodes.UNKNOWN_ERROR));
```
SUPERCILEX: I think any solution is better than none at this point since we don't even notify the user that anything has happened (failing silently is almost always a bad idea IMO). Another possibility would be to show a toast to the user, but this solution could in theory give more control to the dev if we refactor the `ErrorCodes`. @samtstern What are your thoughts?

samtstern: Agree, having some solution is better here.  When you say "refactor the ErrorCodes" what do you have in mind?  I'd say 'UNKNOWN_ERROR' is actually somewhat fitting here since this is a corner case situation.

SUPERCILEX: @samtstern See my reply here: https://github.com/firebase/FirebaseUI-Android/issues/526#issuecomment-297532702. This would be a case where we could return the exception directly. However, with our current system in place, I agree with you that returning `UNKNOWN_ERROR` is the right thing to do in this case.

samtstern: Alright well let's go with this for now.

PR #678, opened 2017-04-27
[CLEANUP] Refactor various classes

SUPERCILEX:
@samtstern In my quest to move non-100% related changes from #309 to separate PRs, I've kinda lopped all the changes together into one rather large PR. I hope that's ok with you! 😄 Otherwise, I can try to pull stuff out into a bunch of separate PRs.

Here are the changes contained in this PR:
- Refactor `IdpResponse` to use a builder. I think our current non-builder version is arguably already a bit messy and adding yet another set of constructor overloads to add a field in #309 made things very sad very fast. This PR brings down the number of constructors to just one (excluding the error code constructor). In addition, we perform sanity checks in the builder to prevent future errors.
- Refactor `IdpResponse.getIntent(IdpResponse)` to make more sense and enable extensibility: `IdpResponse#toIntent()`
- Fix a typo in `CredentialSignInHandler`: `mAccountLink**Result**Code` should be `mAccountLink**Request**Code`
- Fixup tests to enable returning custom uid such as [here](https://github.com/firebase/FirebaseUI-Android/pull/309/files#diff-08294632c68ed353405bc96cccdedf2bR120) in #309


auth/src/main/java/com/firebase/ui/auth/util/Builder.java
```diff
@@ -0,0 +1,5 @@
+package com.firebase.ui.auth.util;
+
+public interface Builder<T> {
```
samtstern: This interface is not necessary, builders having a `build()` method is a known Java convention and I don't see any methods that are dependent on the generic `Builder` interface in this PR ( correct me if I am wrong)

auth/src/test/java/com/firebase/ui/auth/testhelpers/FakeAuthResult.java
```diff
@@ -17,15 +17,11 @@
...
 import com.google.firebase.auth.FirebaseUser;
 
-public class FakeAuthResult implements AuthResult {
-    private FirebaseUser mFirebaseUser;
-
-    public FakeAuthResult(FirebaseUser firebaseUser) {
-        mFirebaseUser = firebaseUser;
-    }
+public enum FakeAuthResult implements AuthResult {
+    INSTANCE;
```
samtstern: What is this line?

auth/src/main/java/com/firebase/ui/auth/util/Builder.java
SUPERCILEX: Yup, that makes sense.

auth/src/test/java/com/firebase/ui/auth/testhelpers/FakeAuthResult.java
SUPERCILEX: That's just me being lazy since this is in a test. 😄 I can remove it if you want, but basically it's the easiest way of ensuring the singleton property since I'm using an Enum and I have a single instance of it.

samtstern:
@SUPERCILEX fine to merge this once the funny `INSTANCE` thing is gone!  Thanks for separating out these changes.

SUPERCILEX:
@samtstern Lol, yeah I removed it. 😄

SUPERCILEX:
@samtstern Fixed the build! 😄

SUPERCILEX:
@samtstern Awesome, time to merge stuff! 😛

Issue #642, opened 2017-03-20
Add back the Adapter constructors that take a DatabaseReference


puf:
In FirebaseUI 0.4 the database adapters had an [overloaded constructor that would take a `DatabaseReference`](https://github.com/firebase/FirebaseUI-Android/blob/0.4.0/database/src/main/java/com/firebase/ui/database/FirebaseRecyclerAdapter.java#L122) (just called `Firebase` back then) instead of a `Query`. 

While this overload was not strictly needed (as `DatabaseReference` extends `Query`), we found in developer feedback sessions that not all developers are aware that they can pass their database reference. 

We had one case where the developer spent quite some time researching Firebase queries. While that was a great learning experience for the developer, it would've been better if the API was harder to mis-interpret, which is the reason we added an overload that takes a database location.

SUPERCILEX:
@samtstern Do you think #679 is enough? Also, this should become less relevant in FirebaseUI `v3.0`. 😉

samtstern:
I have not seen any confusion about this issue, although puf does offer evidence from the original developer feedback sessions on FirebaseUI.

My theory is that over time `Query` has become a more essential piece of the puzzle for RTDB developers and most of them have heard of it before they get here.  So I am inclined to close this issue (with your doc comments) but I'll give @puf a final chance to tell me that I am very wrong :-)

puf:
I chatted to Sam about this offline. We haven't seen as much confusion in practice as in the initial API feedback tests. I still have a preference to be able to pass in either a `Query` or a `DatabaseReference`, but I'm also fine leaving it out for now.

PR #679, opened 2017-04-27
Add hint that a DatabaseReference is also a Query

SUPERCILEX:
@samtstern I would like to compromise on @puf's proposal (#642). Currently, we already have 3 constructors with 2 of them using queries. This would mean we'll have to add two more "useless" constructors which is 😢. Instead, I propose we add a bolded note in the javadoc for each constructor that has a `Query` explaining that you can also pass in a `DatabaseReference`. Here's what it looks like:

image

The `README` has also be revamped and clearly shows that you can pass in a ref [here](https://github.com/firebase/FirebaseUI-Android/blob/version-2.0.0-dev/database/README.md#create-custom-firebaserecycleradapter-subclass):

image


samtstern:
Well this certainly cant hurt so happy to add these comments.  Still open to @puf making an argument for adding all the extra constructors.  We pretty clearly show the use of a `DatabaseReference` in our docs and I've never heard the feedback that using the `Query` superclass is confusing but puf did get pretty strong feedback about this once.

SUPERCILEX:
@samtstern Ok, SGTM! 😄 I'll leave it up to you and @puf to make the decision to add the extra constructors then.

PR #680, opened 2017-04-27
Fixup db README

SUPERCILEX:
@samtstern While working on #679 I noticed that there was some broken code in the readme so this PR fixes that and a few other issues:
- Add link to `RecyclerView` since it's an advanced topic that devs may want to research further
- Make README consistent with sample by changing all refs of `text` to `message`
- Reorder setting up db and creating `ViewHolder` to make more sense


samtstern:
:+1: thanks for this!  Nice work with the image too.

SUPERCILEX:
You're welcome! 😁

Issue #586, opened 2017-02-12
Changing order of auth providers


davidvavra:
When I setup the library, I always get this provider order on the first screen:
1. Facebook
2. Google
3. Email

I would like to have Google first. There doesn't need to be a new API for this, just following the order in the provider list.

SUPERCILEX:
@davidvavra Just change the order in which you provide the `IdpConfig` builders:
```java
Collections.unmodifiableList(
        Arrays.asList(new AuthUI.IdpConfig.Builder(AuthUI.TWITTER_PROVIDER).build(), // It's backwards for some reason, the top ones are last visually.
                      new AuthUI.IdpConfig.Builder(AuthUI.FACEBOOK_PROVIDER).build(),
                      new AuthUI.IdpConfig.Builder(AuthUI.GOOGLE_PROVIDER).build(),
                      new AuthUI.IdpConfig.Builder(AuthUI.EMAIL_PROVIDER).build())); // Email is always on the bottom.
```

davidvavra:
OK, thanks this workaround works. But it's still a bug - the library should follow the specified order.

SUPERCILEX:
@davidvavra Yeah, but at this point changing it would break the order for everyone already using FirebaseUI so we're kind of stuck in a tough situation.

@samtstern Do you think we should fix the order bug or leave it given what I said above?

samtstern:
Yeah I'd rather not break current people, even though @davidvavra is right that it would have been better to follow the specified order in the first place!

davidvavra:
Well, I think at least this should be documented somewhere.

Issue #554, opened 2017-01-29
Email authentication typo results in poor UX


togic8:
#### Versions

  * Google Play Services version: 10.0.1
  * FirebaseUI version: 1.1.1

#### Steps to reproduce:

  1. Authenticate using email.
  2. Sign out.
  3. Try to authenticate again, using the same email, but "accidentally" introduce a typo (for example, replace an "n" with an "m" in the email address).
  
#### Observed Results:

  * Firebase correctly assumes this is a new email address, rather than a returning user; however, after the user realizes they have entered their email address incorrectly and proceeds to fix their mistake on the "new user" page, they must re-enter their name and password. This is not the problem. The problem is when they try to log in and then they are met with the error "an account already exists with that email address" rather than being authenticated using the correct email & password combination.
  
#### Expected Results:

  * After correcting their email address and entering their name and password, if their email address & password hash match a record in the user base, they should be authenticated as that user. If you really want to be paranoid, make it so that the name also has to match (ignoring case, probably) the name already on record for that account.

SUPERCILEX:
@samtstern @togic8  This was a conscious decision https://github.com/firebase/FirebaseUI-Android/pull/447#discussion_r91544515 to make the UX feel less janky, but I'm wondering if we could compromise and forward the user along to the sign in page while showing a toast or something.

https://github.com/firebase/FirebaseUI-Android/blob/master/auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailFragment.java#L288

PR #681, opened 2017-04-28
Fix displayed provider order being the inverse of the order in which they were supplied

SUPERCILEX:
@samtstern In both #309 and a PR I'm working on to fix \#554, I've wanted to get providers in an orderly fashion with the most important provider being on top and the least at the bottom. In #309, I simply made the assumption that the order should be `Google` -> `Facebook` -> `Twitter` -> `Email`. However, I think it would be better to use the dev's supplied provider order, but unfortunately, the order in which providers are displayed and passed in is wonky so this PR fixes that.

Addresses https://github.com/firebase/FirebaseUI-Android/issues/586#issuecomment-279246587

PS: I made this change in a backwards compatible way because I think people might be surprised if they just update FirebaseUI and suddenly their provider order is all messed up.

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/provider/Provider.java
+++ b/auth/src/main/java/com/firebase/ui/auth/provider/Provider.java
@@ -5,9 +5,7 @@ import android.content.Context;
 import android.content.Intent;
 
 public interface Provider {
-    /**
-     * Retrieves the name of the IDP, for display on-screen.
-     */
+    /** Retrieves the name of the IDP, for display on-screen. */
     String getName(Context context);
 
     String getProviderId();
--- a/auth/src/main/java/com/firebase/ui/auth/provider/Provider.java
+++ b/auth/src/main/java/com/firebase/ui/auth/provider/Provider.java
@@ -1,0 +1,18 @@
+package com.firebase.ui.auth.provider;
+
+import android.app.Activity;
+import android.content.Context;
+import android.content.Intent;
+
+public interface Provider {
+    /**
+     * Retrieves the name of the IDP, for display on-screen.
+     */
+    String getName(Context context);
+
+    String getProviderId();
+
+    void startLogin(Activity activity);
+
+    void onActivityResult(int requestCode, int resultCode, Intent data);
+}
--- a/auth/src/main/java/com/firebase/ui/auth/provider/Provider.java
+++ b/auth/src/main/java/com/firebase/ui/auth/provider/Provider.java
@@ -3,6 +3,7 @@ package com.firebase.ui.auth.provider;
 import android.app.Activity;
 import android.content.Context;
 import android.content.Intent;
+import android.support.annotation.LayoutRes;
 
 import com.google.firebase.auth.GoogleAuthProvider;
 
@@ -12,6 +13,10 @@ public interface Provider {
 
     /** Retrieves the id of the IDP, e.g. {@link GoogleAuthProvider#PROVIDER_ID}. */
     String getProviderId();
+
+    /** Retrieves the layout id of the button to inflate and/or set a click listener. */
+    @LayoutRes
+    int getButtonLayout();
 
     /** Start the login process for the IDP, e.g. show the Google sign-in activity. */
     void startLogin(Activity activity);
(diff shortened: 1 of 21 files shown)


puf:
Good find. I always wondered how the order they're displayed in relates to the code. :-)

samtstern:
LGTM, @amandle any thoughts on this one?

auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
```diff
@@ -140,7 +140,8 @@ public void onClick(View view) {
                     }
                 });
                 provider.setAuthenticationCallback(this);
-                btnHolder.addView(loginButton, 0);
+                // Account for the email provider which should always be at the bottom
+                btnHolder.addView(loginButton, btnHolder.getChildCount() - 1);
```
SUPERCILEX: Now that I think about this, do we really want the email provider to always be at the bottom? Why aren't we using the supplied order from the dev?

samtstern: If we're going to have a deprecated version of the old behavior anyway then I agree, let's just fully respect the order they provide.

SUPERCILEX: Awesome, I'm on it! 😄

auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
```diff
@@ -127,6 +126,10 @@ private void populateIdpList(List<IdpConfig> providers) {
                     loginButton = getLayoutInflater()
                             .inflate(R.layout.idp_button_twitter, btnHolder, false);
                     break;
+                case EmailAuthProvider.PROVIDER_ID:
+                    loginButton = getLayoutInflater()
+                            .inflate(R.layout.provider_button_email, btnHolder, false);
```
SUPERCILEX: This is where things would have gotten messy without the `Provider` refactor: I would have had to keep track of the email provider index and somehow insert it somewhere inside this for loop without it actually being part of the `mProviders` list. It would have been possible, but very messy.

auth/src/main/res/layout/provider_button_email.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Button
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:id="@+id/email_button"
+    style="@style/FirebaseUI.Button.AccountChooser.EmailButton"
+    android:text="@string/sign_in_with_email"
+    tools:ignore="UnusedIds"/> <!-- Used in tests -->
```
SUPERCILEX: It took me a while to figure out that my build was failing because lint thinks these ids are unused so I moved the suppression from the `lint-baseline` to each button.

auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
samtstern: Could we take this refactor a bit further?  `Provider` could define `@LayoutRes int getButtonLayout()` and then this switch could disappear and just become `loginButton = getLayoutInflater().inflate(provider.getButtonLayout(), btnHolder, false)`

auth/src/main/java/com/firebase/ui/auth/provider/Provider.java
```diff
@@ -0,0 +1,16 @@
+package com.firebase.ui.auth.provider;
+
+import android.app.Activity;
+import android.content.Context;
+import android.content.Intent;
+
+public interface Provider {
```
samtstern: Let's add some comments to each method here

SUPERCILEX:
@samtstern Done! 😄

samtstern:
@SUPERCILEX I'm gonna merge this when the tests pass unless you have lingering concerns?

SUPERCILEX:
@samtstern nope, I think it's good!

Issue #671, opened 2017-04-25
Distinguish between register/login for backend server auth validation


AndroidRio:
### Step 3: Describe the problem:
My request is quite simple. A feature to distinguish between login/register in FirebaseUI so I can go to different endpoints in my backend depending on which situation happened.

SUPERCILEX:
@AndroidRio This feature has been requested several times, but it hasn't been done because it's relatively difficult to implement on the FirebaseUI side. The simple solution usually mentioned is store user info in your backend and then check if there is any data for the given uid. If not, then it's a new user. Otherwise, it's an existing one. Hope this helps! 😄

shanukag:
@SUPERCILEX I have the same issue as @AndroidRio, but how can i save the UID if the user object is null when starting the signinintent?

SUPERCILEX:
@shanukag Take a look at how to [handle the sign-in response](https://github.com/firebase/FirebaseUI-Android/tree/master/auth#handling-the-sign-in-response). You can also register an [AuthStateListener](https://firebase.google.com/docs/reference/android/com/google/firebase/auth/FirebaseAuth.html#addAuthStateListener(com.google.firebase.auth.FirebaseAuth.AuthStateListener)). In both cases, that's where you'll want to [get the currently signed-in user](https://firebase.google.com/docs/auth/android/manage-users#get_the_currently_signed-in_user). Hope this helps! 😄

AndroidRio:
Thanks for the support!

samtstern:
FYI the new `FirebaseUserMetadata` class should help with this:
https://firebase.google.com/docs/reference/android/com/google/firebase/auth/FirebaseUserMetadata

PR #687, opened 2017-05-02
Fix incorrect link in README to auth state changes

SUPERCILEX:
@samtstern While working on https://github.com/firebase/FirebaseUI-Android/issues/671#issuecomment-298699037, I noticed that the links to the Firebase documentation weren't very clear so this PR fixes that.


morganchen12:
Thanks!

PR #696, opened 2017-05-09
Massively simplify styling

SUPERCILEX:
@samtstern I was going through the sign-in process for my app because I updated some Facebook/Twitter settings and it looked pretty ugly: 😱

Long story short, I forgot to style FirebaseUI correctly. Oops! 😄 Anyway, the point of this PR is to make styling easier, or more accurately, unnecessary. At the moment, styling requires a bunch of manual work so this PR basically styles FirebaseUI with the app's default look n' feel for free. Yay! 🚀 🎉 😄 

Here's a really large and gnarly table with comparison screenshots of the default no config theme before and after this PR:

Old | New
--- | ---
screenshot_20170508-201458 | screenshot_20170508-211507
screenshot_20170508-201559 | screenshot_20170508-211525
screenshot_20170508-201634 | screenshot_20170508-212635
screenshot_20170508-201707 | screenshot_20170508-212429
screenshot_20170508-201713 | screenshot_20170508-212757
screenshot_20170508-201718 | screenshot_20170508-212913
screenshot_20170508-212933 | screenshot_20170508-213425
screenshot_20170508-201428 | screenshot_20170508-221604
screenshot_20170508-201530 | screenshot_20170508-221626

@samtstern Lol, it took me so long to make that table. 😆 This PR is a WIP because I still haven't looked through my code to clean stuff up and the existing themes look pretty dang ugly right now (their `colorAccent` is yucky which makes all the buttons pretty bad).

@samtstern What are your thoughts on this before I move forward and update our sample app themes? I was almost thinking of removing theme customization completely and just using the app's default theme, but I feel like that might be a little extreme... thoughts, comments, concerns, random outbursts? 😆


auth/src/main/res/values/colors.xml
```diff
@@ -1,11 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-    <color name="authui_colorPrimary">#3E50B4</color>
-    <color name="authui_colorPrimaryDark">#2F3E9E</color>
-    <color name="authui_colorAccent">#3E50B4</color>
-    <color name="authui_inputTextColor">#FFFFFF</color>
-    <color name="authui_inputTextColorDark">#000000</color>
+    <color name="colorPrimary">#3F51B5</color>
+    <color name="colorPrimaryDark">#303F9F</color>
+    <color name="colorAccent">#FF4081</color>
```
SUPERCILEX: This is where all the magic happens: the app's theme overrides this so we get theme customization for free. However, we need to keep these values to be able to build the `auth` module.

auth/src/main/res/values/styles.xml
```diff
@@ -2,17 +2,10 @@
 <resources>
 
     <style name="FirebaseUI" parent="Theme.AppCompat.Light.DarkActionBar">
-        <item name="colorPrimary">@color/authui_colorPrimary</item>
```
samtstern: So are all of the deleted values OK because they default to relying on `colorPrimary`, `colorPrimaryDark`, and `colorAccent`?

auth/src/main/res/values/colors.xml
samtstern: I do think your new colors look better, however I think the original color schemes were set by the person who made the UX mocks so I am hesitant to change them based on our personal preference.

I'll see if someone can weigh in on that.  But warning I may have to ask you to revert the colors, which will make your beautiful table obsolete :-/

auth/src/main/res/values/styles.xml
```diff
@@ -133,13 +108,9 @@
     </style>
 
     <style name="FirebaseUI.EditText">
-        <item name="android:textColor">?android:textColor</item>
         <item name="android:textSize">16sp</item>
-        <item name="android:lineSpacingExtra">8sp</item>
```
samtstern: Are these all the same as defaults too?

auth/src/main/res/values/styles.xml
```diff
@@ -207,6 +177,7 @@
         <item name="android:layout_gravity">left</item>
         <item name="android:gravity">left|center_vertical</item>
         <item name="android:typeface">normal</item>
+        <item name="android:foreground">?android:attr/selectableItemBackground</item>
```
samtstern: Another day, another new trick I learned!

SUPERCILEX: Yeah, I was really happy when I saw this in a gnarly SO post.

auth/src/main/res/values/colors.xml
samtstern: I see.  I wasn't thinking it through all the way.  **Before** this change the colors in our theme mattered since we were not leaning on AppCompat, so if the developer did not pass a theme those colors would show up.

Now because we are leaning on AppCompat our theme will almost never show up (unless the dev is not using AppCompat, right?)

SUPERCILEX: @samtstern Yup and yup! I'm happy to change the colors back to what they were since they shouldn't matter; what I really want is a way to error if the dev doesn't override our colors, but I don't think that's possible.

PR #698, opened 2017-05-10
Improve email UX when a user makes a typo in their email

SUPERCILEX:
#554


auth/src/main/java/com/firebase/ui/auth/ui/email/CheckEmailFragment.java
```diff
@@ -94,6 +96,52 @@ public static CheckEmailFragment getInstance(@NonNull FlowParameters flowParamet
         return fragment;
     }
 
+    public static void checkAccountExists(Activity activity,
+                                          final BaseHelper helper,
+                                          final CheckEmailListener listener,
+                                          @NonNull final String email,
+                                          @Nullable final String name,
+                                          @Nullable final Uri photoUri) {
```
SUPERCILEX: The refactor came out a little messier than I would have liked, but at least we're sharing the same logic.

auth/src/main/java/com/firebase/ui/auth/ui/email/CheckEmailFragment.java
```diff
@@ -94,6 +96,52 @@ public static CheckEmailFragment getInstance(@NonNull FlowParameters flowParamet
...
+        helper.showLoadingDialog(R.string.progress_dialog_checking_accounts);
+
+        if (!TextUtils.isEmpty(email)) {
```
samtstern: nit: generally if the whole method is wrapped in an `if()` it's better to just short-circuit to reduce nesting.  So

```java
if (TextUtils.isEmpty(email)) {
   return;
}

helper.getFirebaseAuth()
  // ...

}

File: auth/src/main/java/com/firebase/ui/auth/ui/email/CheckEmailFragment.java (hunk 1 above)
Comment (samtstern) [inline auth/src/main/java/com/firebase/ui/auth/ui/email/CheckEmailFragment.java]: There are a few things I don't really like here:

  * Taking an instance of  `BaseHelper` so that we can start/stop the progress dialog on behalf of the caller does not seem right.  This method should not be concerned with UI.
   * We take in `name` and `photoUri` only to give it right back
   * `CheckEmailListener` really doesn't seem like it saves much code.  For example:

```java
if (providers == null || providers.isEmpty()) {
    // New user
} else if (EmailAuthProvider.PROVIDER_ID.equalsIgnoreCase(providers.get(0))) {
   // Email user
} else {
   // IDP user
}
```
becomes

```java
@Override
onNewUser(User user) {

}

@Override
onExistingEmailUser(User user) {

}

@Override
onExistingIdpUser(User user) {

}
```

So what if we defined this method to be like this (coding in a comment field so excuse typos):

```java
public static Task<String> getTopProvider(FirebaseAuth auth, String email) {
   TaskCompletionSource<String> source = new TaskCompletionSource<>();

   auth.fetchProvidersForEmail(email)
       .addOnCompleteListener(Task<ProviderQueryResult> task) {
          if (!task.isSuccessful) {
              source.setException(task.getException());
              return;
          }

           List<String> providers = task.getResult().getProviders();
           if (providers == null || providers.isEmpty()) {
               return null;
           } else {
               return providers.get(0);
           }
       });

    return source.getTask();
}
```

and then consumers can just call this `Task` and branch on if the result is `EmailAuthProvider.PROVIDER_ID`, `null` or some IDP.

SUPERCILEX: @samtstern Yeah, that's a much better way of doing it! 😄

SUPERCILEX:
@samtstern Still need to do some refactoring e.g. remove `getTopProvider` from `CheckEmailFragment`.

auth/src/main/java/com/firebase/ui/auth/ui/email/CheckEmailFragment.java
```diff
@@ -94,6 +97,26 @@ public static CheckEmailFragment getInstance(@NonNull FlowParameters flowParamet
...
+        return auth.fetchProvidersForEmail(email)
+                .addOnFailureListener(
+                        new TaskFailureLogger(TAG, "Error fetching providers for email"))
+                .continueWith(new Continuation<ProviderQueryResult, String>() {
+                    @Override
+                    public String then(@NonNull Task<ProviderQueryResult> task) throws Exception {
+                        if (!task.isSuccessful()) return null;
+
+                        List<String> providers = task.getResult().getProviders();
+                        if (providers == null || providers.isEmpty()) return null;
```
samtstern: Maybe I am dumb but why switch from `.get(0)` to `.get(size() - 1)` ?
 
Also style nit but for blocks with `if` and `else` can we use `{}` even when each clause is one line?

SUPERCILEX: @samtstern Lol, of course not. 😆 I started working on the refactor and I'll post an explanatory comment once the method is in stable place where my comment won't be folded away.

And yeah, I'll add braces. 😄

auth/src/main/java/com/firebase/ui/auth/provider/AuthCredentialHelper.java
```diff
@@ -1,39 +0,0 @@
...
-
-import android.support.annotation.Nullable;
-
-import com.firebase.ui.auth.IdpResponse;
-import com.google.firebase.auth.AuthCredential;
-import com.google.firebase.auth.FacebookAuthProvider;
-import com.google.firebase.auth.GoogleAuthProvider;
-import com.google.firebase.auth.TwitterAuthProvider;
-
-public final class AuthCredentialHelper {
```
SUPERCILEX: Whoops, it looks like git didn't catch the rename. 😢

auth/src/main/java/com/firebase/ui/auth/provider/ProviderUtils.java
```diff
@@ -0,0 +1,75 @@
...
+                .addOnFailureListener(
+                        new TaskFailureLogger(TAG, "Error fetching providers for email"))
+                .continueWith(new Continuation<ProviderQueryResult, String>() {
+                    @Override
+                    public String then(@NonNull Task<ProviderQueryResult> task) throws Exception {
+                        if (!task.isSuccessful()) return null;
+
+                        List<String> providers = task.getResult().getProviders();
+                        return providers == null || providers.isEmpty()
+                                ? null : providers.get(providers.size() - 1);
```
SUPERCILEX: @samtstern To address your earlier comment, I did some debugging and discovered that the order of the providers is the order in which the user adds/links a provider to their account. I think it's better UX for the user's last added provider to be the one they are asked to sign in with since it's the one that's become more relevant to them. I'm open to going back to using the first provider the user signed up with, but again, I feel like that would be worse UX. @samtstern What are your thoughts on this?

SUPERCILEX: PS: are you okay with a ternary or should I go back to an if statement with braces?

samtstern: Let's stick with providers(0) since if we're going to make a sneaky (but important) change like that I'd like to think about it more.

Fine with ternary!  My opposition to no-braces if statements is that it's easy to try to modify them and forgot to add braces

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailFragment.java
```diff
@@ -297,12 +300,63 @@ public void onFailure(@NonNull Exception e) {
...
+                                                                        email).build()),
+                                                        RegisterEmailActivity.RC_WELCOME_BACK_IDP);
+                                            } else {
+                                                onExistingIdpUser(new User.Builder(email)
+                                                                          .setProvider(provider)
+                                                                          .build());
+                                            }
+                                        }
+
+                                        private void onExistingIdpUser(User user) {
```
samtstern: Let's not add private methods to listeners.  It's hard to follow and makes them non-lambda-compatible in the future.

auth/src/main/java/com/firebase/ui/auth/provider/AuthCredentialHelper.java
samtstern: Just FYI: doing renames like this (or `getInstance()` --> `newInstance()`) increases the PR size a lot and makes it harder to pay attention to the important non-cosmetic changes.  I know there are a lot of poorly named things in this lib and it's tempting to fix them while you're in there but I think we should try to keep cosmetic changes in separate CLs that can be rubber-stamp approved.

auth/src/main/java/com/firebase/ui/auth/provider/ProviderUtils.java
SUPERCILEX: @samtstern Ok, I'll make another PR for that so we can think about it some more. 😄

Oh, haha that makes sense! I just hate all the boilerplate braces! 😆

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailFragment.java
SUPERCILEX: I like it, thinking about our glorious lambda future! 😆

auth/src/main/java/com/firebase/ui/auth/provider/AuthCredentialHelper.java
SUPERCILEX: @samtstern Yeah, sorry about that; I have a habit of always trying to lop in unrelated changes into my PRs. 😊 I'll try to move those changes to other PRs in the future.

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailFragment.java
```diff
@@ -297,12 +300,59 @@ public void onFailure(@NonNull Exception e) {
...
+                                                           Toast.LENGTH_LONG)
+                                                    .show();
+
+                                            if (provider == null) {
+                                                throw new IllegalStateException(
+                                                        "User has no providers even though we got a FirebaseAuthUserCollisionException");
+                                            } else if (EmailAuthProvider.PROVIDER_ID.equalsIgnoreCase(
+                                                    provider)) {
+                                                getActivity().startActivityForResult(
+                                                        WelcomeBackPasswordPrompt.createIntent(
```
SUPERCILEX: @samtstern One last note before you merge: do we want to pass the user's password to the `WelcomeBackPasswordPrompt`? It could make the sign-in faster and less annoying to the user since I doubt they want to type their password in yet again. 😄

samtstern: Nope let's leave it as is.  I can think of a few strange situations with that

SUPERCILEX: @samtstern Alrighty, SGTM!

PR #701, opened 2017-05-11
When linking two accounts, use the provider the user last added to their account instead of the first one

SUPERCILEX:
Since the last provider the user adds to their account is likely to be the one that is most practical for them to use, I think this change would provider better UX.

@samtstern @amandle What are your thoughts on this?


SUPERCILEX:
@samtstern What do you think about this one?

samtstern:
@SUPERCILEX I like the idea, but I have to draw the line on `2.0.0` somewhere and I am drawing it at "things we have already done + phone auth" :-)

SUPERCILEX:
@samtstern cool beans, that works for me! 😀

SUPERCILEX:
@samtstern What are your thoughts on this?

I'm wondering if using our own custom ranking system would simply be better. We know that Google auth is way easier and faster, then I would say Facebook, then Twitter, then email. What do you think?

samtstern:
@SUPERCILEX I think custom ranking gets a little political, I don't want to favor one IDP over another because of implementation details.

I am fine to merge this into `2.1.0` now if you're happy with it as-is.

SUPERCILEX:
> I think custom ranking gets a little political

Haha, yeah good catch! 😂

@samtstern yup, I still think using the user's newer provider is better then their origin one so go for it! 😀

PR #704, opened 2017-05-11
Fix future lint errors

SUPERCILEX:
There's some code lint will complain about in the Gradle plugin `v2.4.x` so this PR fixes that in advance.


auth/src/main/res/values/styles.xml
```diff
@@ -31,33 +31,6 @@
         <item name="android:fontFamily">sans-serif</item>
     </style>
 
-    <style name="FirebaseUI.Text.T01">
```
samtstern: Can we just tell lint to ignore the unused styles?  It's sort of useful to have these here so that when we make new screens we don't have to pick new magic numbers.

SUPERCILEX: @samtstern Sure! I learned something new today: you can add the `tools` namespace to your styles.

SUPERCILEX: I also had to rebuild the `baseline` for some reason. It was probably something on my machine, but it cleaned stuff up so why not.

PR #705, opened 2017-05-11
Fix broken database tests

SUPERCILEX:
@samtstern Ouch, we forgot to update the database tests after making some braking changes in the `FirebaseArray`s. This PR incorporates the updated `FirebaseArray`s and also ensures the tests pass.


samtstern:
@SUPERCILEX I just noticed this the other day and thought "I must be crazy, my git is out of date" and shook it off.

Here's the real question (before I merge this): Why isn't Travis complaining?

SUPERCILEX:
@samtstern lol, I had almost the same reaction! 😄 This was a decision by @puf to make the build faster: running the database tests requires booting up an emulator since they aren't unit tests... We could make them run on every Travis build, but it would slow things down significantly.

samtstern:
@SUPERCILEX can we at least make the tests compile on each build?

SUPERCILEX:
@samtstern Yup, I found the task.

PR #706, opened 2017-05-11
Make layouts better on small screens

### LINKED ISSUE #626: can not login using email on small devices (e.g. Nexus One) - buttons are offscreen
* Android device: Nexus One
  * Android OS version: any
  * Google Play Services version: 25.2.0
  * Firebase/Play Services SDK version: 10.0.1
  * FirebaseUI version: 1.2.0
  
It is impossible to re-login using email on small devices, like the Nexus One.
The "Welcome back!" message takes up the entire screen and buttons are not visible.

#### Steps to reproduce:

  1. create Nexus One emulator
  2. create a test program using Firebase UI
  3. login
  4. logout
  5. login again (it is impossible now)
  
#### Observed Results:

  when I press enter or tab on the keyboard, focus should move to the SIGN IN button.
  it does not do this.  instead focus moves to the "eye" then back to the login.
  Please add KEYCODE_ENTER to input processing and move focus to the SIGN-IN button.
  or put everything into a scroll view.   There needs to be a way to get the button in view.
login_problem

  
#### Expected Results:

  I expected to be able to press the login button.
  
#### Relevant Code:

  ```
  // TODO(you): code here to reproduce the problem
  ```

samtstern:
Fixes #626

A few improvements:

  * Puts tall layouts in `ScrollViews`
  * Makes horiztonal padding smaller on small screens
  * Makes field separation smaller on small screens

### DIFF
--- a/auth/src/main/res/layout/register_email_layout.xml
+++ b/auth/src/main/res/layout/register_email_layout.xml
@@ -36,7 +36,7 @@
             style="@style/Widget.Design.TextInputLayout"
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
-            android:paddingTop="112dp"
+            android:paddingTop="@dimen/register_field_padding_vert"
             android:hint="@string/name_hint">
 
             <android.support.design.widget.TextInputEditText
@@ -52,7 +52,7 @@
             style="@style/FirebaseUI.TextInputLayout.PasswordField"
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
-            android:paddingTop="224dp"
+            android:paddingTop="@dimen/register_field_padding_vert_2x"
             app:passwordToggleEnabled="true">
 
             <android.support.design.widget.TextInputEditText
@@ -66,7 +66,7 @@
             style="@style/FirebaseUI.Text.BodyText"
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
-            android:layout_marginTop="336dp"
+            android:layout_marginTop="@dimen/register_field_padding_vert_3x"
             android:textIsSelectable="false"/>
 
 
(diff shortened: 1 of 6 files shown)


samtstern:
@SUPERCILEX or @amandle  wanna take a look at this?

auth/src/main/res/layout/register_email_layout.xml
```diff
@@ -66,7 +66,7 @@
             style="@style/FirebaseUI.Text.BodyText"
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
-            android:layout_marginTop="336dp"
+            android:layout_marginTop="@dimen/register_field_padding_vert_3x"
```
samtstern: The comment above in the source says:

```
 Since the TextInputLayouts change size based on whether or not an error
 is showing we use a relative layout to position them absolutely
```

So for that reason I am hesitant to change it.  Clearly Aaron found something better about the LinearLayout when trying to match the mocks.

I may be able to replace with a FrameLayout though since we're not using any of the RelativeLayout's relational bits?  I'll try.

SUPERCILEX: @samtstern Oh crud, I completely forgot about that. ☹️ There's got to be a better way. We're using margins right now so when the hint slides up and expands, it pushes itself down. What if we used raw `paddingTop`? I think that won't push the view down since it should just expand inside the padding. If you don't have the time, I'm happy to play around with it! 😄

samtstern: And I found what we need:
https://developer.android.com/reference/android/support/design/widget/TextInputLayout.html#attr_android.support.design:errorEnabled

Woot!

SUPERCILEX:
@samtstern Awesome, LGTM! 😁

PR #711, opened 2017-05-13
Fix broken picker layout on small screens

SUPERCILEX:
@samtstern With the addition of the GitHub provider, the main auth picker layout looked really bad. This PR fixes that with `ConstraintLayout`. Yay!!! 😄 🚀 🎉 

I also wrapped the auth buttons in a `ScrollView` so they are guaranteed to be usable.

Old | New
--- | ---
screenshot_20170513-110820 | screenshot_20170513-111209
screenshot_20170513-110935 | screenshot_20170513-111223
screenshot_20170513-110939 | screenshot_20170513-111235
screenshot_20170513-110945 | screenshot_20170513-111242

## New layout edge cases

No logo | Minimum providers
--- | ---
screenshot_20170513-112353 | screenshot_20170513-112422


samtstern:
@SUPERCILEX thank you for this!  I want to hold this until after 2.0, this is clearly an improvement but I am wondering if we can do even better (esp for the scrolling buttons part).  Once 2.0 goes out I will have some UX people take a look and we can put this into 2.1 (or whatever the next version is)

SUPERCILEX:
@samtstern yeah, it would be nice to have designers look at this (I don't know about you, but UX isn't my strong suit 😂).

SUPERCILEX:
@samtstern I know you said you wanted designers to take a look at this, but in the meantime I feel like any improvement is better than none... **Especially since you can now have 5 providers with phone auth which looks like 💩 and is completely unusable in landscape mode.**

SUPERCILEX:
@samtstern lol now that I'm trying to break the picker, I've found more 💩😂. If we enable accessibility settings to boost the text size, the following happens both with and without this PR:
screenshot_20170615-151139

I'll try to submit a PR fixing that at some point soon.

SUPERCILEX:
@samtstern Merged! 😄 Dunno if you read my previous comment, but this PR should really be merged to make the picker layout look nice when using all the providers. The screenshot with GitHub oauth is how it looks with phone auth since we now have 5 providers.

SUPERCILEX:
@samtstern gentle bump

samtstern:
@SUPERCILEX thanks for bumping this, will take a look soon.  Good candidate for `2.1.1`.

SUPERCILEX:
@samtstern Merged! 😄

I also fixed the word wrapping while I was at it:

Old | New
--- | ---
image | image

samtstern:
Thanks for merging everything into this.  We will still need some follow up improvements (`ScrollView` is not ideal) but it's certainly better than what we have.

This is the worst I could do to it.  Nexus 4, split screen, landscape:
image

SUPERCILEX:
@samtstern Yeah, we need designers cuz I can't think of anything better than `ScrollView` 😢😆.

For your case, I think we can set a `minWidth` and wrap the buttons in a `ScrollView` (again 😢).

PR #727, opened 2017-05-26
Add support for privacy policy URLs in sign-up

### LINKED ISSUE #690: [Feature request] Privacy policy link in FirebaseUI Auth sign-in flow
I understand that one can add a Terms of Service URL to the sign-in flow like so:

    AuthUI.getInstance()
        .createSignInIntentBuilder()
        .setProviders(...)
        .setTosUrl("https://superapp.example.com/terms-of-service.html")
        .setTheme(...)
        .build()
However, since Google Play [policies](https://play.google.com/about/privacy-security/personal-sensitive/) require the disclosure of how personal data is used in apps, it would be nice to increase the visibility of this information during sign-up/sign-in. I'd be happy to add a method `setPrivacyPolicyUrl(String url)` to the sign-in builder if there aren't any major problems with the idea.

TheCraftKid:
This addresses and resolves #690.
(Sorry it took a bit, I had to find a way to add URL to the existing text spans.)

### DIFF
--- a/app/src/main/res/layout/auth_ui_layout.xml
+++ b/app/src/main/res/layout/auth_ui_layout.xml
@@ -160,6 +160,12 @@
                     android:checked="true"
                     android:text="@string/google_privacy_label"/>
 
+                <RadioButton
+                    android:id="@+id/firebase_privacy"
+                    android:layout_width="wrap_content"
+                    android:layout_height="wrap_content"
+                    android:text="@string/firebase_privacy_label"/>
+
             </RadioGroup>
 
             <TextView
--- a/app/src/main/res/layout/auth_ui_layout.xml
+++ b/app/src/main/res/layout/auth_ui_layout.xml
@@ -146,6 +146,28 @@
                 android:layout_height="wrap_content"
                 android:layout_marginBottom="8dp"
                 android:layout_marginTop="16dp"
+                android:text="@string/privacy_policy_header"/>
+
+            <RadioGroup
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:orientation="vertical">
+
+                <RadioButton
+                    android:id="@+id/google_privacy"
+                    android:layout_width="wrap_content"
+                    android:layout_height="wrap_content"
+                    android:checked="true"
+                    android:text="@string/google_privacy_label"/>
+
+            </RadioGroup>
+
+            <TextView
+                style="@style/Base.TextAppearance.AppCompat.Subhead"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:layout_marginBottom="8dp"
+                android:layout_marginTop="16dp"
                 android:text="@string/logo_header"/>
 
             <RadioGroup
--- a/auth/src/main/java/com/firebase/ui/auth/ui/email/PreambleHandler.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/email/PreambleHandler.java
@@ -1,0 +1,122 @@
+package com.firebase.ui.auth.ui.email;
+
+import android.content.Context;
+import android.net.Uri;
+import android.support.annotation.ColorInt;
+import android.support.annotation.StringRes;
+import android.support.customtabs.CustomTabsIntent;
+import android.support.v4.content.ContextCompat;
+import android.text.SpannableStringBuilder;
+import android.text.TextUtils;
+import android.text.method.LinkMovementMethod;
+import android.text.style.ClickableSpan;
+import android.text.style.ForegroundColorSpan;
+import android.util.TypedValue;
+import android.view.View;
+import android.widget.TextView;
+
+import com.firebase.ui.auth.R;
+import com.firebase.ui.auth.ui.FlowParameters;
+
+public class PreambleHandler {
+    private static final String TOS_TARGET = "%TOS%";
+    private static final String PP_TARGET = "%PP%";
+
+    private final Context mContext;
+    private final FlowParameters mFlowParameters;
+    private final ForegroundColorSpan mLinkSpan;
+
+    private SpannableStringBuilder mBuilder;
+
+    public PreambleHandler(Context context, FlowParameters parameters) {
+        mContext = context;
+        mFlowParameters = parameters;
+        mLinkSpan = new ForegroundColorSpan(ContextCompat.getColor(mContext, R.color.linkColor));
+
+        setupCreateAccountPreamble();
+    }
+
+    public void setPreamble(TextView textView) {
+        textView.setMovementMethod(LinkMovementMethod.getInstance());
+        textView.setText(mBuilder);
+    }
+
+    private void setupCreateAccountPreamble() {
+        int preambleType = getPreambleType();
+        if (preambleType == -1) {
+            return;
+        }
+
+        String[] preambles =
+                mContext.getResources().getStringArray(R.array.create_account_preamble);
+        mBuilder = new SpannableStringBuilder(preambles[preambleType]);
+
+        replaceTarget(TOS_TARGET, R.string.terms_of_service, mFlowParameters.termsOfServiceUrl);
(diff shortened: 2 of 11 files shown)


samtstern:
@TheCraftKid this looks great!  Can you provide a quick screenshot for me?

samtstern:
The code looks great but tests are broken, I think there is somewhere in tests where we are using the `new FlowParameters()` constructor and you're not passing a privacy URL, so it's breaking.

TheCraftKid:
Hey, I fixed the tests, but I'm trying to build the app to get a screenshot, and keep getting this error:
     Error:org.gradle.api.GradleException: No matching client found for package name 'com.firebase.uidemo'.

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailFragment.java
```diff
@@ -210,6 +213,39 @@ public void onClick(View view) {
...
+
+        ForegroundColorSpan foregroundColorSpan =
+                new ForegroundColorSpan(ContextCompat.getColor(getContext(), R.color.linkColor));
+
+        String link = getString(R.string.privacy_policy);
+        SpannableStringBuilder spannableStringBuilder = new SpannableStringBuilder(link);
+        int start = mAgreementText.length();
+        spannableStringBuilder.setSpan(foregroundColorSpan, start, start + link.length(), 0);
+
+        mAgreementText.append(" and the ");
```
SUPERCILEX: Instead of doing this which isn't translatable, could we instead combine the `setUpTermsOfService` and `setUpPrivacyPolicy` methods to make the two clickable bits of text in one go? This should simplify things and fix the bug below.

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailFragment.java
```diff
@@ -210,6 +213,39 @@ public void onClick(View view) {
...
+        mAgreementText.append(spannableStringBuilder);
+        mAgreementText.setOnClickListener(new View.OnClickListener() {
```
SUPERCILEX: I don't think this will work. Try clicking the Terms of Service link and it will open the privacy policy instead. Oops! 😄 You'll need to make only the bits of text clickable instead of the entire view like so: https://stackoverflow.com/a/10697453/4548500.

SUPERCILEX:
@TheCraftKid For your build error, make sure the Firebase test project you setup has a matching package: https://stackoverflow.com/a/39301178/4548500.

Here's mine:
image

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailFragment.java
SUPERCILEX: Hmmm, there's another bug I just thought about: if the dev provides a PP but not a TOS, the agreement text will be something like `and the privacy policy`. I'm thinking about how we can rewrite this in a sustainable way.

app/src/main/res/layout/auth_ui_layout.xml
```diff
@@ -146,6 +146,28 @@
...
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:orientation="vertical">
+
+                <RadioButton
+                    android:id="@+id/google_privacy"
+                    android:layout_width="wrap_content"
+                    android:layout_height="wrap_content"
+                    android:checked="true"
+                    android:text="@string/google_privacy_label"/>
```
TheCraftKid: I was looking for some kind of Fireabse privacy policy, but I could only find terms of service. That's a good idea.

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailFragment.java
TheCraftKid: I was debating whether to combine setting the terms of service and privacy policy text in a single method, but I was a bit afraid to touch anything else. We can use two template strings in strings.xml for the possibility that either link doesn't exist.

SUPERCILEX: Yeah, that's what I did. Since you mentioned having issues with the Spans, I went ahead and created a patch for you: [Fix_various_bugs.zip](https://github.com/firebase/FirebaseUI-Android/files/1030730/Fix_various_bugs.zip).

If you want to give it a go yourself, hack at it! 😄 Otherwise, just extract the zip and then apply the patch in Android Studio by calling up the shortcut: `Ctrl/Cmd + Shift + A` -> "Apply patch".

TheCraftKid: Wow, thanks! With the string templates, I'm curious why you didn't use the normal Java ones. Is there any technical problem with doing so?

SUPERCILEX: I needed a way to manually inject the strings so that I could get their start and end indices:
```java
    private void replaceTarget(String target, @StringRes int replacementRes, String url) {
        int targetIndex = mBuilder.toString().indexOf(target);
        if (targetIndex != -1) {
            String replacement = mContext.getString(replacementRes);
            mBuilder.replace(targetIndex, targetIndex + target.length(), replacement);

            int end = targetIndex + replacement.length();
            mBuilder.setSpan(mLinkSpan, targetIndex, end, 0);
            mBuilder.setSpan(new CustomTabsSpan(url), targetIndex, end, 0);
        }
    }
``` 

If you have any idea how to improve this mess, go for it! 😄

auth/src/main/java/com/firebase/ui/auth/ui/email/PreambleHandler.java
```diff
@@ -0,0 +1,122 @@
...
+        replaceTarget(TOS_TARGET, R.string.terms_of_service, mFlowParameters.termsOfServiceUrl);
+        replaceTarget(PP_TARGET, R.string.privacy_policy, mFlowParameters.privacyPolicyUrl);
+    }
+
+    private void replaceTarget(String target, @StringRes int replacementRes, String url) {
+        char[] currentPreambleChars = new char[mBuilder.length()];
+        mBuilder.getChars(0, mBuilder.length(), currentPreambleChars, 0);
+        String currentPreamble = String.valueOf(currentPreambleChars);
+
+        int targetIndex = currentPreamble.indexOf(target);
```
SUPERCILEX: @TheCraftKid whoops, forgot to mention that I updated my patch. Turns out the `SpannableStringBuilder#toString()` method does exactly what I just did so all of that can be replaced with `mBuilder.toString().indexOf(target);`

auth/src/main/java/com/firebase/ui/auth/ui/email/PreambleHandler.java
```diff
@@ -0,0 +1,122 @@
...
+import android.text.style.ClickableSpan;
+import android.text.style.ForegroundColorSpan;
+import android.util.TypedValue;
+import android.view.View;
+import android.widget.TextView;
+
+import com.firebase.ui.auth.R;
+import com.firebase.ui.auth.ui.FlowParameters;
+
+public class PreambleHandler {
```
SUPERCILEX: @samtstern LGTM now, but can you think of ways to improve this?

samtstern:
@TheCraftKid  and @SUPERCILEX thank you both for getting this all ready.  I won't actually merge this until after the long weekend, so just warning you in advance of my silence.  But we're gonna get this in.

SUPERCILEX:
@samtstern Awesome! 😄

@TheCraftKid Could you push the code simplification with `toString` I mentioned?

This
```java
char[] currentPreambleChars = new char[mBuilder.length()];
mBuilder.getChars(0, mBuilder.length(), currentPreambleChars, 0);
String currentPreamble = String.valueOf(currentPreambleChars);

int targetIndex = currentPreamble.indexOf(target);
```

should turn into this

```java
int targetIndex = mBuilder.toString().indexOf(target);
```

PR #734, opened 2017-06-01
Revert reauthentication

### LINKED ISSUE #715: Revert reauthentication
Reauth is not going to make it into the next version for iOS or Web so we are going to drop it from version 2.0 for Android :-(

samtstern:
This fixes #715, we have to remove re-authentication for version 2.0.0 because there has not been enough bandwidth to implement on other platforms and we want to conserve a consistent UI surface.

### DIFF
--- a/app/src/main/java/com/firebase/uidemo/auth/SignedInActivity.java
+++ b/app/src/main/java/com/firebase/uidemo/auth/SignedInActivity.java
@@ -53,9 +53,8 @@ import butterknife.ButterKnife;
 import butterknife.OnClick;
 
 public class SignedInActivity extends AppCompatActivity {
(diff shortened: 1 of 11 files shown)


SUPERCILEX:
@samtstern Aw dang, that's so sad. 😢 I guess it had to be done.

samtstern:
@SUPERCILEX the good news is I deleted only what I had to.  The intent builder is still generic, etc.  So I think when it comes time to re-implement we will have no trouble.

SUPERCILEX:
@samtstern Oh, that's true! I looked at the changes and it seems like there shouldn't be too many merge conflicts with existing PRs. Awesome! 😄

PR #735, opened 2017-06-03
Update Auth documentation for sign-in intent builder

TheCraftKid:
This documents the new behavior implemented in #727.


samtstern:
@TheCraftKid thanks!

PR #736, opened 2017-06-05
Add call to onDataChanged when adding first listener

SUPERCILEX:
@samtstern This one should really go in before 2.0 is released.

Basically, I got a weird bug because of this and we should be doing it anyway for consistency's sake.

### DIFF
--- a/database/src/main/java/com/firebase/ui/database/ObservableSnapshotArray.java
+++ b/database/src/main/java/com/firebase/ui/database/ObservableSnapshotArray.java
@@ -23,6 +23,8 @@ public abstract class ObservableSnapshotArray<E> extends ImmutableList<DataSnaps
     protected final List<ChangeEventListener> mListeners = new CopyOnWriteArrayList<>();
     protected final SnapshotParser<E> mParser;
 
+    private boolean mHasDataChanged = false;
+
     /**
      * Create an ObservableSnapshotArray where snapshots are parsed as objects of a particular
      * class.
@@ -56,7 +58,9 @@ public abstract class ObservableSnapshotArray<E> extends ImmutableList<DataSnaps
         for (int i = 0; i < size(); i++) {
             listener.onChildChanged(ChangeEventListener.EventType.ADDED, get(i), i, -1);
         }
-        listener.onDataChanged();
+        if (mHasDataChanged) {
+            listener.onDataChanged();
+        }
 
         return listener;
     }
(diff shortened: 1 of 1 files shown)


samtstern:
@SUPERCILEX I agree with this one but just curious, what was the bug you ran into?

SUPERCILEX:
@samtstern Lol, as I was writing about my bug, I realized this PR introduced another bug.

For my bug, it was just something in my app. Basically, I'm trying to cleanup my "no content hint" logic and I wasn't getting the first `onDataChanged` callback because I was adding my listener to a `FirebaseArray` that was already live. (So my no content hint wouldn't show up even though there were no items.)

As for this PR, I realized that `onDataChanged` would always be called right away when adding the first listener regardless of whether or not any data had changed. I then considered checking if we had been listening for changes and only send the callback if it wasn't the first listener, but that doesn't scale to the second, third, or `n + 1` listener. So instead, the solution is slightly more messy: we only send an `onDataChanged` event if one has already occurred.

database/src/main/java/com/firebase/ui/database/ObservableSnapshotArray.java
```diff
@@ -66,6 +71,11 @@ public ChangeEventListener addChangeEventListener(@NonNull ChangeEventListener l
     @CallSuper
     public void removeChangeEventListener(@NonNull ChangeEventListener listener) {
         mListeners.remove(listener);
+
+        // Reset mHasDataChanged if there are no more listeners
+        if (!isListening()) {
+            mHasDataChanged = false;
```
samtstern: This block is a little surprising to me.  Wouldn't you want the first listener attached to still fire immediately if the FirebaseArray contains data?

SUPERCILEX: @samtstern Yeah, that still happens. This just ensures that adding the first listener is the same whether or not this class was first created or is being reused. The contract for **`remove`**`ChangeEventListener` is that the list must be cleared when there are no more listeners, as if the class was first created. Here's the `FirebaseArray` implementation:
```java
// Clear data when all listeners are removed
if (!isListening()) {
    mQuery.removeEventListener((ValueEventListener) this);
    mQuery.removeEventListener((ChildEventListener) this);

    clearData();
}
```

SUPERCILEX:
@samtstern Would you like some more clarification on what's going on? Here are the cases I'm trying to solve:
1. First listener is added, shouldn't get any callbacks
1. `n + 1` listener is added, should get the current items to catch up with the other listeners, but not `onDataChanged` unless the other listeners already received that call
1. `n + 1` listener is added after a full cycle of adding items and calling `onDataChanged`. In this case, we play catch up and call the items added method and `onDataChanged`.
1. All listeners are reset: everything should go back to being in the same state as if the class was first loaded.

samtstern:
Thanks for explaining!  SGTM.

PR #737, opened 2017-06-05
[2.0] Restrict unnecessary immutable stuff to library so we can change it later

SUPERCILEX:
@samtstern Sorry for being late about all this, I thought I had submitted a PR and it turns out I didn't. 😄

Basically, all I want to do is not have to support our custom iterator implementations because they aren't needed so I'm making them private for now. **_This has to be done before 2.0 is released to ensure backwards compatibility!_** Another PR is coming up for which 2.0 or not doesn't matter.


samtstern:
Thanks! This is a good idea.

PR #738, opened 2017-06-05
Remove unnecessary immutable stuff

SUPERCILEX:
Built on top of #737 

@samtstern When I first wrote the refactor, I didn't know [`AbstractList`](https://docs.oracle.com/javase/8/docs/api/java/util/AbstractList.html) was a thing. Now I do! 😄 It's immutable by default and handles a ton of stuff for us so we don't have to deal with it which is really nice.

Also, as long as #737 is merged this can happen after 2.0 is released.


samtstern:
@SUPERCILEX this is great, thank you for finding this improvement.  I am happy to include this in 2.0 once the merge conflict is fixed.

SUPERCILEX:
@samtstern Cool beans, merged! 😄

Issue #558, opened 2017-01-30
Consider allowing email hints even when smartlock disabled


samtstern:
We should consider allowing email autofill in the email flow even when smartlock is disabled for actual retrieval/storage of credentials.

Pros:
  * We know this increases conversion rates for email flows significantly
  * Does not really hurt development

Cons:
  * The email hint selector looks like the Google account picker and may be confusing

SUPERCILEX:
@samtstern I'm in strong favor of doing this, but there's another con you mentioned before: still using smart lock when the developer explicitly said we shouldn't breaks the API contact. However, this use of smart lock isn't related to saving passwords so I think it should be fine.

samtstern:
Yeah @SUPERCILEX we'd likely have to change the name of the flag

SUPERCILEX:
@samtstern Hmmmm, this would cause problems for stuff like #200. 😢

PR #740, opened 2017-06-07
Merge phone auth into version-2.0.0

samtstern:
As mentioned we've been working on phone auth internally.  Now that the Play services `11.0.0` SDK is out we can move into the open.

This merges all of the phone auth changes into `version-2.0.0-dev`


ashwinraghav:
LGTM

samtstern:
Here is an enumeration of the changes at a high level (in no particular order):

- Phone auth (obviously)
- Smart Lock enable flag split into two pieces: one for the hint selector and one for actual credentials
- Facebook and Twitter made `provided` dependencies.  This means apps not using these providers don't need the dep.
- Adding `translation_description` attributes to all library strings (preparing for internationalization)

SUPERCILEX:
@samtstern Ouch! That's a lot of changes! 😕 Glad to finally see phone auth come out though!!! 😄 

I have a few questions:
1. The facebook and twitter things aren't particularly related to phone auth, right? 
Also, I disagree with making people add them to their own deps so I think we should definitely try to figure out testing compatibility to upgrade to the 3.0 plugin and use `implementation`. That should be what we use for basically all our deps except the Firebase ones. Dunno if you've looked into it yet, but `implementation` basically means "my module needs this dep so let people compile it even if they aren't including the dep, but don't let them use it e.g. they have to include the dep if they want to use it". That seems like a much better solution than breaking everyone's dependencies when they try to upgrade.
1. I haven't looked at the PR, but does it also fix #558?

SUPERCILEX:
PS: The failing test is because you guys aren't using `Integer.parseInt()` which does unnecessary boxing:
```html
<a href="#DM_BOXED_PRIMITIVE_FOR_PARSING">Bug type DM_BOXED_PRIMITIVE_FOR_PARSING (click for details)</a>
<br/>In class com.firebase.ui.auth.ui.phone.CountryListSpinner<br/>In method com.firebase.ui.auth.ui.phone.CountryListSpinner.setSelectedForCountry(Locale, String)<br/>Called method Integer.intValue()<br/>Should call Integer.parseInt(String) instead<br/>At CountryListSpinner.java:[line 83]</p>
```

samtstern:
@SUPERCILEX there's a few huge benefits of doing twitter/facebook this way:

- Shaves 2MB+ off of your app if you were not using those providers before by avoiding the depdendency
- No more adding that pesky twitter maven repo

I agree we should use implementation down the road, but I think in the meantime this is a big improvement.

SUPERCILEX:
@samtstern Ohhhh, wait so you don't have to include them if you aren't using Facebook/Twitter? Won't there be a `ClassDefNotFound` exception somewhere down the line?

samtstern:
@SUPERCILEX we're using a trick to check for Facebook/Twitter SDK presence if you enable those providers (and at other entry points to the flow like `signOut`) and then throwing a RuntimeException before you can proceed saying "add the dependency"

samtstern:
Two remaining test failures:

1. AuthMethodPickerActivityTest. testFacebookLoginFlow

```
java.lang.NoClassDefFoundError: com/facebook/R$style
	at com.facebook.FacebookSdk.<clinit>(FacebookSdk.java:87)
	at com.facebook.internal.Validate.sdkInitialized(Validate.java:144)
	at com.facebook.login.LoginManager.__constructor__(LoginManager.java:72)
	at com.facebook.login.LoginManager.<init>(LoginManager.java)
	at com.facebook.login.LoginManager.getInstance(LoginManager.java:83)
	at com.firebase.ui.auth.provider.FacebookProvider.startLogin(FacebookProvider.java:97)
	at com.firebase.ui.auth.ui.idp.AuthMethodPickerActivity$1.onClick(AuthMethodPickerActivity.java:127)
	at android.view.View.performClick(View.java:5639)
```

2. AuthMethodPickerActivityTest. testGoogleLoginFlow

```
junit.framework.ComparisonFailure: expected:<test@example.com> but was:<null>
	at junit.framework.Assert.assertEquals(Assert.java:100)
	at junit.framework.Assert.assertEquals(Assert.java:107)
	at com.firebase.ui.auth.testhelpers.TestHelper.verifySmartLockSave(TestHelper.java:82)
```

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -179,12 +184,18 @@ public static int getDefaultTheme() {
         Task<Status> signOutTask = signInHelper.signOut();
 
         // Facebook sign out
-        LoginManager.getInstance().logOut();
+        try {
+            LoginManager.getInstance().logOut();
+        } catch (NoClassDefFoundError e) {
```
SUPERCILEX: > @SUPERCILEX we're using a trick to check for Facebook/Twitter SDK presence if you enable those providers (and at other entry points to the flow like signOut) and then throwing a RuntimeException before you can proceed saying "add the dependency"

Gotcha, just saw this. Now that I think about it, I'm definitely starting to really like the idea. For Instant apps, you could limit the providers to just Google and reap massive benefits. Snazzy! 😄

samtstern:
cc: @ashwinraghav two remaining test failures in the phone auth branch.  I pushed commits to fix all the lint/checkstyle/findbugs errors.

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailFragment.java
```diff
@@ -157,7 +157,7 @@ public void onActivityCreated(@Nullable Bundle savedInstanceState) {
         getActivity().setTitle(R.string.title_register_email);
 
         mSaveSmartLock = mHelper.getSaveSmartLockInstance(getActivity());
-        new PreambleHandler(getContext(), mHelper.getFlowParams()).setPreamble(mAgreementText);
+        mAgreementText.showTerms(mHelper.getFlowParams(), R.string.button_text_save);
```
SUPERCILEX: Yeah, this seems kinda pointless, why not?:
```java
PreambleHandler.showTerms(mAgreementText, mHelper.getFlowParams(), R.string.button_text_save);

public static void showTerms(TextView tv, FlowParameters params, @StringRes int buttonText) {
    new PreambleHandler(tv.getContext(), params, buttonText).setPreamble(tv);
}
```

auth/src/main/java/com/firebase/ui/auth/ui/phone/CompletableProgressDialog.java
```diff
@@ -0,0 +1,74 @@
...
+import android.content.Context;
+import android.os.Bundle;
+import android.view.View;
+import android.widget.ImageView;
+import android.widget.ProgressBar;
+import android.widget.TextView;
+
+import com.firebase.ui.auth.R;
+
+public final class CompletableProgressDialog extends ProgressDialog {
```
SUPERCILEX: Ouch, we're going to have to throw all this away eventually. 😢

auth/src/main/java/com/firebase/ui/auth/ui/phone/CountryInfo.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        final CountryInfo that = (CountryInfo) o;
+
+        if (countryCode != that.countryCode) return false;
+        return locale != null ? locale.equals(that.locale) : that.locale == null;
```
SUPERCILEX: This `equals` and `hashCode` look hand generated, but I would trust Intellij's implementation more. Don't see anything wrong with it so looks fine I guess. I still think it would just be safer to have Intellij generate it...

PS: `Alt + Insert` -> `Hashcode and equals`

auth/src/main/java/com/firebase/ui/auth/ui/phone/CountryInfo.java
```diff
@@ -0,0 +1,78 @@
...
+        result = 31 * result + countryCode;
+        return result;
+    }
+
+    @Override
+    public String toString() {
+        return localeToEmoji(locale) + " " + this.locale.getDisplayCountry() + " +" + countryCode;
+    }
+
+    public static String localeToEmoji(Locale locale) {
```
SUPERCILEX: I'm confused, where do emojis come in and how are they related to country codes?

auth/src/main/java/com/firebase/ui/auth/ui/phone/CountryListLoadTask.java
```diff
@@ -0,0 +1,302 @@
...
+        countryInfoList.add(new CountryInfo(new Locale("", "VU"), 678));
+        countryInfoList.add(new CountryInfo(new Locale("", "VA"), 379));
+        countryInfoList.add(new CountryInfo(new Locale("", "VE"), 58));
+        countryInfoList.add(new CountryInfo(new Locale("", "VN"), 84));
+        countryInfoList.add(new CountryInfo(new Locale("", "WF"), 681));
+        countryInfoList.add(new CountryInfo(new Locale("", "EH"), 212));
+        countryInfoList.add(new CountryInfo(new Locale("", "YE"), 967));
+        countryInfoList.add(new CountryInfo(new Locale("", "ZM"), 260));
+        countryInfoList.add(new CountryInfo(new Locale("", "ZW"), 263));
+        Collections.sort(countryInfoList);
```
SUPERCILEX: O-M-F-G, that list is so long with so much memory allocation!!! This is so 😢!!! Yeah, I definitely think we should consider using RecyclerView for 2.0. (I almost had a heart attack there 😆)

auth/src/main/java/com/firebase/ui/auth/ui/phone/CountryListSpinner.java
```diff
@@ -0,0 +1,185 @@
...
+
+    public CountryListSpinner(Context context, AttributeSet attrs, int defStyle) {
+        super(context, attrs, defStyle);
+
+        init();
+    }
+
+    /**
+     * Only for Testing
+     */
```
SUPERCILEX: That's what `@VisibleForTesting` is for 😄.

auth/src/main/java/com/firebase/ui/auth/ui/phone/CountryListSpinner.java
```diff
@@ -0,0 +1,185 @@
...
+    void setDialogPopup(DialogPopup dialog) {
+        this.dialogPopup = dialog;
+    }
+
+    private void init() {
+        super.setOnClickListener(this);
+
+        countryListAdapter = new CountryListAdapter(getContext());
+        dialogPopup = new DialogPopup(countryListAdapter);
+        textFormat = "%1$s  +%2$d";
```
SUPERCILEX: Is this a bug or is the `+` supposed to be there?

auth/src/main/java/com/firebase/ui/auth/ui/phone/SpacedEditText.java
```diff
@@ -0,0 +1,108 @@
...
+import android.util.AttributeSet;
+
+import com.firebase.ui.auth.R;
+
+/**
+ * This element inserts spaces between characters in the edit text and expands the width of the
+ * spaces using spannables.
+ * This is required since Android's letter spacing is not available until API 21.
+ */
+public final class SpacedEditText extends android.support.v7.widget.AppCompatEditText {
```
SUPERCILEX: nit: `AppCompatEditText` instead of `android.support.v7.widget.AppCompatEditText`

auth/src/main/java/com/firebase/ui/auth/ui/phone/SubmitConfirmationCodeFragment.java
```diff
@@ -0,0 +1,277 @@
...
+
+    TextView mEditPhoneTextView;
+    TextView mResendCodeTextView;
+    TextView mCountDownTextView;
+    SpacedEditText mConfirmationCodeEditText;
+    Button mSubmitConfirmationButton;
+    CustomCountDownTimer mCountdownTimer;
+    PhoneVerificationActivity mVerifier;
+    TermsTextView mAgreementText;
+    long mMillisUntilFinished;
```
SUPERCILEX: Any reason these aren't private?

auth/src/main/java/com/firebase/ui/auth/ui/phone/VerifyPhoneNumberFragment.java
```diff
@@ -0,0 +1,241 @@
...
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+public class VerifyPhoneNumberFragment extends FragmentBase implements View.OnClickListener {
+    public static final String TAG = "VerifyPhoneFragment";
+
+    CountryListSpinner countryListSpinner;
+    EditText mPhoneEditText;
+    TextView errorEditText;
+    Button sendCodeButton;
+    PhoneVerificationActivity mVerifier;
+    TextView mSmsTermsText;
```
SUPERCILEX: Same, any reason to not make them `private`?

auth/src/main/java/com/firebase/ui/auth/ui/phone/VerifyPhoneNumberFragment.java
```diff
@@ -0,0 +1,241 @@
...
+
+    private static final int RC_PHONE_HINT = 22;
+
+    public static VerifyPhoneNumberFragment getInstance(FlowParameters flowParameters, String
```
SUPERCILEX: Same, `newInstance`

auth/src/main/java/com/firebase/ui/auth/ui/phone/SubmitConfirmationCodeFragment.java
```diff
@@ -0,0 +1,277 @@
...
+
+
+
+    public static SubmitConfirmationCodeFragment getInstance(FlowParameters flowParameters,
```
SUPERCILEX: Should be `newInstance` to reduce confusion.

auth/src/main/java/com/firebase/ui/auth/util/signincontainer/SaveSmartLock.java
```diff
@@ -93,7 +93,6 @@ public void onConnected(Bundle bundle) {
             finish();
             return;
         }
-
```
SUPERCILEX: ? nit: this space reduces cram so we should keep it.

samtstern:
@SUPERCILEX got all the tests to pass now but `findbugs` is erroring out.  It only happens when I do a `clean` in the same command (`./gradlew findbugs` works `./gradlew clean assembleDebug check` fails)

auth/src/main/java/com/firebase/ui/auth/util/signincontainer/SmartLockBase.java
```diff
@@ -21,6 +21,7 @@
 import com.google.firebase.auth.FacebookAuthProvider;
 import com.google.firebase.auth.FirebaseUser;
 import com.google.firebase.auth.GoogleAuthProvider;
+import com.google.firebase.auth.PhoneAuthProvider;
```
SUPERCILEX: This doesn't look necessary since there are no other changes in this class. Should there be some though? Wouldn't we want to update and use `providerIdToAccountType` somewhere?

auth/src/test/java/com/firebase/ui/auth/testhelpers/ActivityHelperShadow.java
```diff
@@ -16,6 +16,7 @@
 
 import com.firebase.ui.auth.ui.ActivityHelper;
 import com.firebase.ui.auth.util.signincontainer.SaveSmartLock;
+import com.google.firebase.auth.PhoneAuthProvider;
```
SUPERCILEX: Probs a refactor mistake, we don't need it.

auth/src/test/java/com/firebase/ui/auth/testhelpers/TestConstants.java
```diff
@@ -18,6 +18,7 @@
 
 public class TestConstants {
     public static final String EMAIL = "test@example.com";
+    public static final String PHONE = "+14347565353";
```
SUPERCILEX: Lol, is this a real phone number? We should probably go with `+15555555555` or something similar.

auth/src/test/java/com/firebase/ui/auth/testhelpers/TestHelper.java
```diff
@@ -81,4 +83,26 @@ public static void verifySmartLockSave(String providerId, String email, String p
         assertEquals(password, passwordCaptor.getValue());
         assertEquals(providerId, idpResponseCaptor.getValue().getProviderType());
     }
+
+    public static void verifySmartLockSave(String providerId, String email, String phoneNumber,
```
SUPERCILEX: Could we fold this into the method above? I know this is a test, but still the method below is almost unreadable. We could just pass in null for the `phoneNumber` or have a simple overload that does that for us.

SUPERCILEX:
>@SUPERCILEX got all the tests to pass now but findbugs is erroring out. It only happens when I do a clean in the same command (`./gradlew findbugs` works but `./gradlew clean assembleDebug check` fails)

@samtstern Looks like an issue with your machine 😄 (The Travis tests are passing)

auth/src/main/java/com/firebase/ui/auth/IdpResponse.java
```diff
@@ -34,22 +34,25 @@
 public class IdpResponse implements Parcelable {
     private final String mProviderId;
     private final String mEmail;
+    private final String mPhoneNumber;
```
samtstern: We are setting it in `PhoneVerificationActivity` so I'll add a getter.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -438,6 +450,26 @@ public T setAvailableProviders(@NonNull List<IdpConfig> idpConfigs) {
                 } else {
                     mProviders.add(config);
                 }
+
+                if (config.getProviderId().equals(FACEBOOK_PROVIDER)) {
+                    try {
+                        Class c = com.facebook.FacebookCallback.class;
```
samtstern: Yep, good call.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -438,6 +450,26 @@ public T setAvailableProviders(@NonNull List<IdpConfig> idpConfigs) {
...
+                    } catch (NoClassDefFoundError e) {
+                        throw new RuntimeException("Facebook provider cannot be configured " +
+                               "without dependency. Did you forget to add " +
+                               "'com.facebook.android:facebook-android-sdk:VERSION' dependency?");
+                    }
+                }
+
+                if (config.getProviderId().equals(TWITTER_PROVIDER)) {
+                    try {
+                        Class c = com.twitter.sdk.android.core.TwitterCore.class;
```
samtstern: We're actually only depending on Twitter core now (which means we don't pull in Fabric).  This is new with the upgrade to Twitter SDK 3.0

auth/src/main/java/com/firebase/ui/auth/provider/PhoneProvider.java
```diff
@@ -0,0 +1,53 @@
...
+        mHelper = helper;
+    }
+
+    @Override
+    public String getName(Context context) {
+        return context.getString(R.string.provider_name_phone);
+    }
+
+    @Override
+    public String getProviderId() {
```
samtstern: Done, added on email too.

auth/src/main/java/com/firebase/ui/auth/provider/PhoneProvider.java
```diff
@@ -0,0 +1,53 @@
...
+        return PhoneAuthProvider.PROVIDER_ID;
+    }
+
+    @Override
+    public int getButtonLayout() {
```
samtstern: Done.

auth/src/main/java/com/firebase/ui/auth/provider/TwitterProvider.java
```diff
@@ -46,7 +47,19 @@ public static void initialize(Context context) {
...
+        TwitterConfig config = new TwitterConfig.Builder(context)
+                .twitterAuthConfig(authConfig)
+                .build();
+        Twitter.initialize(config);
+    }
+
+    public static void signout(Context context) {
+        try {
+            signOut();
+        } catch (IllegalStateException e) {
```
samtstern: Done

auth/src/main/java/com/firebase/ui/auth/ui/FragmentHelper.java
```diff
@@ -13,7 +13,7 @@
     private Fragment mFragment;
 
     public FragmentHelper(Fragment fragment) {
-        super(fragment.getContext().getApplicationContext(),
+        super(fragment.getContext(),
```
samtstern: Great catch, this was a merge mistake

auth/src/main/java/com/firebase/ui/auth/ui/TermsTextView.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    /**
+     * @param params     FlowParameters containing terms URLs.
+     * @param buttonText for the button that represents the "action" described in the terms.
+     */
+    public void showTerms(FlowParameters params, @StringRes int buttonText) {
+        PreambleHandler handler = new PreambleHandler(getContext(), params, buttonText);
+        handler.setPreamble(this);
+    }
```
samtstern: @ashwinraghav made the change to introduce the custom view at the exact same time that the PR to include privacy policy came in, so this was the simplest merge resolution.  I don't think it's harmful in anyway so I'll leave the refactoring for later.

auth/src/main/java/com/firebase/ui/auth/ui/phone/CompletableProgressDialog.java
samtstern: Maybe!  This specific style of dialog comes from UX testing on digits, so we don't have to take the word of material design as law unless they actually delete the ProgressDialog class.

auth/src/main/java/com/firebase/ui/auth/ui/phone/CountryInfo.java
samtstern: Tested the generated version, it was identical (assuming we don't want to include `collator`, which we don't).

auth/src/main/java/com/firebase/ui/auth/ui/phone/CountryInfo.java
samtstern: I believe this is to generate the flag for each country.

auth/src/main/java/com/firebase/ui/auth/ui/phone/CountryListLoadTask.java
samtstern: Copied from digits code, I agree that we should refactor this significantly.  However right now the primary goal of this phone auth implementation is to enable digits customers to migrate so leaning towards "if it ain't broke don't fix it" for the initial release.

auth/src/main/java/com/firebase/ui/auth/ui/phone/SubmitConfirmationCodeFragment.java
samtstern: Done

auth/src/main/java/com/firebase/ui/auth/ui/phone/SubmitConfirmationCodeFragment.java
samtstern: Done

auth/src/main/java/com/firebase/ui/auth/ui/phone/VerifyPhoneNumberFragment.java
samtstern: Done

auth/src/main/java/com/firebase/ui/auth/ui/phone/VerifyPhoneNumberFragment.java
samtstern: Done

auth/src/test/java/com/firebase/ui/auth/testhelpers/TestHelper.java
samtstern: Good idea, will make this DRY

auth/src/main/java/com/firebase/ui/auth/ui/phone/CountryListSpinner.java
samtstern: This is intentional.  It's the `+` before a country code.

samtstern:
@SUPERCILEX  thank you for the thorough review!  I believe I have addressed all feedback with the following TODO items (which can be handled later):

- Use `RecyclerView` for the country list
- Consider removing `TermsTextView` and just making a static method in `PreambleHandler`

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
SUPERCILEX: Oh, awesome! Gotta love that smaller APK size! 😄

auth/src/main/java/com/firebase/ui/auth/IdpResponse.java
SUPERCILEX: Oh, K so we do want the dev to get the phone number from `IdpResponse`. Cool beans! 😄

auth/src/main/java/com/firebase/ui/auth/provider/PhoneProvider.java
SUPERCILEX: Awesome!

auth/src/main/java/com/firebase/ui/auth/ui/TermsTextView.java
SUPERCILEX: Yup, makes sense.

auth/src/main/java/com/firebase/ui/auth/ui/phone/CompletableProgressDialog.java
SUPERCILEX: Oh, interesting! 😄 I haven't actually tried phone auth for myself, so dunno what the UX feels like.

auth/src/main/java/com/firebase/ui/auth/ui/phone/CountryInfo.java
SUPERCILEX: Awesome!

auth/src/main/java/com/firebase/ui/auth/ui/phone/CountryInfo.java
SUPERCILEX: Ohhhhhhh! Wow, that's a really interesting usage of emojis! 😄

auth/src/main/java/com/firebase/ui/auth/ui/phone/CountryListLoadTask.java
SUPERCILEX: Gotcha, totally makes sense.

auth/src/main/java/com/firebase/ui/auth/ui/phone/CountryListSpinner.java
SUPERCILEX: Ah, cool beans.

SUPERCILEX:
@samtstern Awesome, LGTM!

PR #742, opened 2017-06-08
Fix `onDataChanged` not being called in FirebaseIndexArray if empty

SUPERCILEX:
@samtstern I think I've finally eradicated all the `onDataChanged` bugs! 😄

This PR fixes the edge case where if the key array calls `onDataChanged` but is empty, the index array won't notify its client because there were no `onChildChanged` events to switch the `mHasPendingMoveOrDelete` flag.


samtstern:
LGTM!

Issue #743, opened 2017-06-08
Change documentation to reflect phone auth


samtstern:
We need to check all READMEs and such to make sure we document Phone Auth as a provider.

jeandavid:
When do you expect to release 2.0.0 with phone auth? Thanks!

samtstern:
I don't want to make any promises but the goal is to release it in the next few days.

Issue #676, opened 2017-04-26
Use API compatibility report to check Version 2.0.0 against Version 1.2.0


samtstern:
Should use [japi-compliance-checker](https://github.com/lvc/japi-compliance-checker) to generate a full report and inform the release notes.

Furthermore, we should add instructions for how to do this in the `README` so we can easily do it for all future releases.

PR #744, opened 2017-06-08
Upgrade guide and compat reports

samtstern:
Issues #676 and #743 are addressed.

### DIFF
--- a/docs/upgrade-to-2.0.md
+++ b/docs/upgrade-to-2.0.md
@@ -14,15 +14,19 @@ For a full, machine-generated compatibility report see the `docs/compat_reports`
     Google/Email/Phone authentication.
   * **Sign Out and Delete** - The `AuthUI#delete(Activity)` and `AuthUI#signOut(Activity)` 
     methods have been removed, please use the versions that accept `FragmentActivity`.
-  * **Error Codes** - The `ResultCodes` class has been removed, instead in `onActivityResult` you 
-    should check for `RESULT_CANCELED` and then use `IdpResponse#getErrorCode()` to determine the \
-    source of the error.
+  * **Error Codes** - `ResultCodes.RESULT_NO_NETWORK` has been removed. Instead you should use
+    `IdpResponse.fromResultIntent(data)` to get the response from the `AuthUI` intent and then
+    check if `IdpResponse#getErrorCode()` is `ErrorCodes.NO_NETWORK`.
   * **Choosing Providers** - `setProviders` has been deprecated, please use `setAvailableProviders`
-    which now respects the order in which you provide them.
+    which now respects the order in which you provide them.  The varargs version of `setProviders`
+    has been removed completely.
   * **Facebook and Google Scopes** - setting scopes via `string` resources is no longer supported,
     please set scopes in code using `IdpConfig`.
-  * **Themes** - AppCompat theme properties such as `colorPrimary` or `colorAccent` are now
-    respected, please double-check to make sure your UI still looks as expected.
+  * **Themes** - AppCompat theme properties such as `colorPrimary` and `colorAccent` are now used 
+    to style FirebaseUI automatically without any need for customization. Unless your auth UI 
+    needs a different theme than the rest of your app, please remove 
+    `AuthUI.SignInIntentBuilder#setTheme(int)` and its related xml theme from your auth intent
+    builder and check to make sure that the auth UI has been themed correctly.
   * **SmartLock for Passwords** - `setIsSmartLockEnabled` has added a two-argument overload. 
     There are now separate flags for enabling the hint select and enabling
     the saving/retrieving of full credentials from the API. Setting the same value for each flag
--- a/docs/upgrade-to-2.0.md
+++ b/docs/upgrade-to-2.0.md
@@ -1,0 +1,42 @@
+# Upgrading to FirebaseUI 2.0
+
+This document outlines the _breaking_ changes made between versions 1.2.0 and 2.0.0 as well as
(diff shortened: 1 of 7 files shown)


samtstern:
@SUPERCILEX did I miss anything?  I feel like I did for database but I am not sure.

docs/upgrade-to-2.0.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Auth
+
+  * **Facebook and Twitter Dependencies** - FirebaseUI no longer directly depends on the Facebook
+    or Twitter SDKs. If you plan to enable either of these providers you should include the
+    appropriate SDK in your app to avoid a crash. This saves space for developers only using
+    Google/Email/Phone authentication.
+  * **Sign Out and Delete** - The `AuthUI#delete(Activity)` and `AuthUI#signOut(Activity)` 
+    methods have been removed, please use the versions that accept `FragmentActivity`.
+  * **Error Codes** - The `ResultCodes` class has been removed, instead in `onActivityResult` you
```
SUPERCILEX: @samtstern We actually haven't done that yet, though I'd be happy to submit a PR. (The class is [here](https://github.com/firebase/FirebaseUI-Android/blob/version-2.0.0-dev/auth/src/main/java/com/firebase/ui/auth/ResultCodes.java) and just references the activity results).

docs/upgrade-to-2.0.md
```diff
@@ -0,0 +1,42 @@
...
+## Auth
+
+  * **Facebook and Twitter Dependencies** - FirebaseUI no longer directly depends on the Facebook
+    or Twitter SDKs. If you plan to enable either of these providers you should include the
+    appropriate SDK in your app to avoid a crash. This saves space for developers only using
+    Google/Email/Phone authentication.
+  * **Sign Out and Delete** - The `AuthUI#delete(Activity)` and `AuthUI#signOut(Activity)` 
+    methods have been removed, please use the versions that accept `FragmentActivity`.
+  * **Error Codes** - The `ResultCodes` class has been removed, instead in `onActivityResult` you 
+    should check for `RESULT_CANCELED` and then use `IdpResponse#getErrorCode()` to determine the \
```
SUPERCILEX: The `\` looks like a typo.

docs/upgrade-to-2.0.md
```diff
@@ -0,0 +1,42 @@
...
+    source of the error.
+  * **Choosing Providers** - `setProviders` has been deprecated, please use `setAvailableProviders`
```
SUPERCILEX: Also, we removed the vargs `setProviders`

docs/upgrade-to-2.0.md
```diff
@@ -0,0 +1,42 @@
...
+    which now respects the order in which you provide them.
+  * **Facebook and Google Scopes** - setting scopes via `string` resources is no longer supported,
+    please set scopes in code using `IdpConfig`.
+  * **Themes** - AppCompat theme properties such as `colorPrimary` or `colorAccent` are now
+    respected, please double-check to make sure your UI still looks as expected.
```
SUPERCILEX: I think this should be rewritten to something like so:
> AppCompat theme properties such as `colorPrimary` and `colorAccent` are now used to style FirebaseUI automatically without any need for customization. Unless your auth UI needs a different theme than the rest of your app, please remove `AuthUI.SignInIntentBuilder#setTheme(int)` and its related xml theme from your auth intent builder and check to make sure that the auth UI has been themed correctly.

docs/upgrade-to-2.0.md
```diff
@@ -0,0 +1,42 @@
...
+  * **SmartLock for Passwords** - `setIsSmartLockEnabled` has added a two-argument overload. 
+    There are now separate flags for enabling the hint select and enabling
+    the saving/retrieving of full credentials from the API. Setting the same value for each flag
+    will emulate the previous single-flag behavior.
+
+## Database
```
SUPERCILEX: @samtstern I think you got them all. 😄

docs/upgrade-to-2.0.md
SUPERCILEX: Created #745

samtstern: Oh interesting.  We did do something about the network error code though, do you remember what that change was?

SUPERCILEX: Yeah, we used to have two error code classes, one in the UI package and the new one in the root package.

I'm not sure, but I think a long time ago we make it so that everything returned result code ok or cancelled and you would check for the idp response error code to see if it was a network error. I believe we used to return the error code directly in the result code before that.

samtstern: Found it and fixed.

docs/upgrade-to-2.0.md
samtstern: Done

docs/upgrade-to-2.0.md
samtstern: done.

docs/upgrade-to-2.0.md
samtstern: Done.

SUPERCILEX:
@samtstern Awesome! Do we _not_ want to go through with #745 then? I still think it's a nice improvement and v2.0 is going to be our only chance for a while.

samtstern:
@SUPERCILEX just commented on that PR but yeah let's not go through with #745.

Also while I have your attention, see my comment on your "Key doesn't exist in a ref" PR since that's the last outstanding thing.

PR #745, opened 2017-06-09
Remove ResultCodes in favor of Activity#RESULT_...

SUPERCILEX:
See https://github.com/firebase/FirebaseUI-Android/pull/744#discussion_r121025943


samtstern:
@SUPERCILEX in the spirit of `2.0` being closed I'd say we drop this one.  Since the values are actually the same under the hood there's no reason we can't document that people should just look for `Activity.RESULT_OK` and `Activity.RESULT_CANCELED`.  So this is really internal cleanup.

SUPERCILEX:
@samtstern SGTM! I'll keep this branch around for after `v2.0`.

Issue #750, opened 2017-06-13
onActivityResult is not called for phone auth.


owcs:
* Android device: OnePlus One
  * Android OS version: 7.1.2
  * Google Play Services version: 11.0.55
  * Firebase/Play Services SDK version: 11.0.0
  * FirebaseUI version: 2.0.0
  
### Describe the problem:

onActivityResult is not called for cancelled/successful phone auth

#### Steps to reproduce:

1. Start phone number verification.
2. Cancel phone number verification.
  
#### Observed Results:

onActivityResult is not called.
  
#### Expected Results:

onActivityResult is called when cancelled or successful.
  
#### Relevant Code:

  ```
startActivityForResult(
    AuthUI.getInstance()
        .createSignInIntentBuilder()
        .setAvailableProviders(
            Arrays.asList(
                new AuthUI.IdpConfig.Builder(AuthUI.PHONE_VERIFICATION_PROVIDER).build())
            )
        .build(),
    RC_SIGN_IN);

  ```
Maybe due to `RC_PHONE_FLOW` not being handle in [SignInDelegate.java](https://github.com/firebase/FirebaseUI-Android/blob/master/auth/src/main/java/com/firebase/ui/auth/util/signincontainer/SignInDelegate.java#L178)?

samtstern:
@owcs thanks for reporting this and also for pointing at a possible solution in `SignInDelegate`.  I will take a look.

geolyth:
The same issue happens to me...in code of SignInDelegate.java I read:

```
@Override
    public void onActivityResult(int requestCode, int resultCode, Intent data) {
        super.onActivityResult(requestCode, resultCode, data);
        switch (requestCode) {
            case RC_CREDENTIALS_READ:
                if (resultCode == ResultCodes.OK) {
                    // credential selected from SmartLock, log in with that credential
                    Credential credential = data.getParcelableExtra(Credential.EXTRA_KEY);
                    handleCredential(credential);
                } else {
                    // Smart lock selector cancelled, go to the AuthMethodPicker screen
                    startAuthMethodChoice();
                }
                break;
            case RC_IDP_SIGNIN:
            case RC_AUTH_METHOD_PICKER:
            case RC_EMAIL_FLOW:
                finish(resultCode, data);
                break;
            default:
                IdpSignInContainer signInContainer = IdpSignInContainer.getInstance(getActivity());
                if (signInContainer != null) {
[...]

PR #751, opened 2017-06-13
Make onActivityResult called from phone flow (#750)

samtstern:


SUPERCILEX:
@samtstern LGTM! Do we want to merge this into master and release a patch (`v2.0.1`) instead of waiting for `2.1`?

samtstern:
@SUPERCILEX yeah I think we will want to do that.  I can make a `2.0.1` branch.

PR #754, opened 2017-06-14
translate to hebrew

baruchiro:
Hey there! So you want to contribute to FirebaseUI? Before you file this pull request, follow these steps:

  * Read [the contribution guidelines](CONTRIBUTING.md).

### DIFF
--- a/app/src/main/res/values-iw/strings.xml
+++ b/app/src/main/res/values-iw/strings.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="app_name">Firebase UI</string>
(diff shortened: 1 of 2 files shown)


baruchiro:
What should I do? can you help me?

samtstern:
@baruchiro thanks for the contribution!  Unfortunately I can't accept this Pull Request for a few reasons:

  * For simplicity's sake, we do not intend to translate the contents of the `app` module, which is only a demonstration of API usage.
  * In the `auth` module we do intend to translate everything (working on it now) but we're going to use Google's internal translation service for the first round so we can't take community translation contributions right now
  * Our RTL language support is not yet finished, so Hebrew could cause some issues.

I hope that all makes sense, we do appreciate your taking the time to send this over.

PR #756, opened 2017-06-15
Update and fix deps

SUPERCILEX:
@samtstern See https://github.com/firebase/FirebaseUI-Android/pull/752#issuecomment-308628100


library/google-services.json
```diff
@@ -1,33 +1,25 @@
 {
   "project_info": {
-    "project_number": "",
-    "firebase_url": "",
-    "project_id": "",
-    "storage_bucket": ""
+    "project_number": "316052285104",
```
SUPERCILEX: `com.google.gms:google-services:3.1.0` complained with these empty values so I just provided my dummy project ones.

samtstern: Can we just put some actual dummy values then?  I know there's nothing bad about leaking yours but it feels a little odd...

samtstern:
@SUPERCILEX thank you for solving this!  Much better, just left one comment.

SUPERCILEX:
@samtstern Yup, rebased.

library/google-services.json
SUPERCILEX: Yup, rebased.

samtstern:
@SUPERCILEX d'oh looks like the plugin can detect obviously fake values.

Here's a fake JSON file I've made that works:
https://github.com/firebase/quickstart-android/blob/master/mock-google-services.json

You can copy the values you need out of that.

SUPERCILEX:
@samtstern Dang, that's a hardcore `google-services.json` you have there!😆 Thanks!

PR #758, opened 2017-06-15
Changed order of parameters in constructor of "Chat" class

pizza:
All examples of the `new Chat(...)` constructor being used in the code assume the message is the 3rd parameter passed to the constructor and not the 2nd.

### DIFF
--- a/database/README.md
+++ b/database/README.md
@@ -27,7 +27,7 @@ public static class Chat {
         // Needed for Firebase
     }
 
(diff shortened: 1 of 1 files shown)


samtstern:
@pizza thanks for catching that!  Looks like we do the opposite outside of the README, but this at least keeps the README consistent with itself.

Also your username is fantastic :pizza: :+1:

pizza:
Thanks Sam! 😍

PR #774, opened 2017-06-22
Update FirebaseListAdapter to use Context instead of Activity

### LINKED ISSUE #773: Use Context instead of Activity for layout inflation
I'm updating an addon library to this library, and it appears the `FirebaseListAdapter` is using an `Activity` parameter in the constructor where only a `Context` is needed. This would help out in situations where a home screen widget uses the adapter but doesn't have access to an `Activity`.
I say replacing it with a `Context` is a good idea because we only need it to inflate the layout, and the system already provides a more generalized way to access the layout inflater.

Instead of `mActivity.getLayoutInflater().inflate(mLayout, viewGroup, false);` 
we should do this: `LayoutInflater.from(mContext).inflate(mLayout, viewGroup, false);` in the `FirebaseListAdapter`.

TheCraftKid:
Fixes #773.

### DIFF
--- a/database/src/main/java/com/firebase/ui/database/FirebaseListAdapter.java
+++ b/database/src/main/java/com/firebase/ui/database/FirebaseListAdapter.java
@@ -1,8 +1,10 @@
 package com.firebase.ui.database;
 
 import android.app.Activity;
+import android.content.Context;
 import android.support.annotation.LayoutRes;
 import android.util.Log;
+import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.BaseAdapter;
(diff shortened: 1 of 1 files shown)


TheCraftKid:
I know this would be a breaking change, but is there any reason we can't just use `LayoutInflater.from(view.getContext()).inflate(mLayout, viewGroup, false);` instead of having a dedicated constructor parameter? This would of course be changed in the next major release. (@samtstern)

samtstern:
@TheCraftKid you're probably right, but given that it's easy to get a `Context` anywhere you'd want to use ` ListAdapter` I don't know if that's worth the break.

Issue #776, opened 2017-06-26
Kill 'helper' objects


samtstern:
Been thinking about this for a little while, but I hate the `BaseHelper` object and its children.  It's just a place to build up technical debt and pass around `FlowParameters`.

I am going to work on killing it.

samtstern:
The fix for this issue has been released in `v2.1.0`.

PR #777, opened 2017-06-26
Kill BaseHelper and friends

samtstern:
See #776 

Steps:
  1. Kill `ActivityHelper` and `FragmentHelper`, make everything a `BaseHelper`
  2. Help Fragments and Activities manage their own `FlowParameters` and `ProgressDialog`
  3. Separate out all factory-type responsibilities of the helpers (`getFirebaseAuth()`, etc) into a dedicated class for that.


auth/src/main/java/com/firebase/ui/auth/ui/DialogBase.java
```diff
@@ -5,24 +5,38 @@
 import android.support.annotation.Nullable;
 import android.support.annotation.RestrictTo;
 import android.support.v4.app.DialogFragment;
+import android.view.ContextThemeWrapper;
 
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 public class DialogBase extends DialogFragment {
```
SUPERCILEX: @samtstern Could we just get rid of this class? The only dialog using it is the `RecoveryEmailSentDialog` and it doesn't use any of this. I also noticed that there's dead code in the `RecoveryEmailSentDialog`:
```java
public static void show(String email, FlowParameters parameters, FragmentManager manager) {
    // We never use the flow params
    bundle.putParcelable(ExtraConstants.EXTRA_FLOW_PARAMS, parameters);
}
```

auth/src/main/java/com/firebase/ui/auth/ui/FragmentBase.java
```diff
@@ -1,28 +1,49 @@
...
 import android.support.annotation.Nullable;
 import android.support.annotation.RestrictTo;
 import android.support.v4.app.Fragment;
+import android.view.ContextThemeWrapper;
 
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 public class FragmentBase extends Fragment {
-    protected FragmentHelper mHelper;
+
+    protected FlowParameters mFlowParameters;
```
SUPERCILEX: Shouldn't the flow params be private since we have a getter to handle initialization?

auth/src/main/java/com/firebase/ui/auth/ui/HelperActivityBase.java
```diff
@@ -1,30 +1,87 @@
...
+import com.google.firebase.auth.FirebaseUser;
+
+import static com.firebase.ui.auth.util.Preconditions.checkNotNull;
+
 @SuppressWarnings("Registered")
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 public class HelperActivityBase extends AppCompatActivity {
 
-    protected ActivityHelper mActivityHelper;
+    protected FlowParameters mFlowParameters;
```
SUPERCILEX: Same.

auth/src/main/java/com/firebase/ui/auth/ui/HelperActivityBase.java
```diff
@@ -1,30 +1,87 @@
...
+    protected ProgressDialogHolder mProgressDialogHolder;
```
SUPERCILEX: It looks like the dialog can also be private since you have a getter. Or we could remove the getter, seems easier and would be more consistent with the other code you've written.

auth/src/main/java/com/firebase/ui/auth/ui/email/RecoveryEmailSentDialog.java
```diff
@@ -13,14 +13,16 @@
...
 
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 public class RecoveryEmailSentDialog extends DialogBase {
     private static final String TAG = "RecoveryEmailSentDialog";
 
-    public static void show(String email, FragmentManager manager) {
+    public static void show(String email, FlowParameters parameters, FragmentManager manager) {
         RecoveryEmailSentDialog result = new RecoveryEmailSentDialog();
         Bundle bundle = new Bundle();
+        bundle.putParcelable(ExtraConstants.EXTRA_FLOW_PARAMS, parameters);
```
SUPERCILEX: Oh, just noticed you added the flow params. Yeah, lol. We should just kill `DialogBase` since we aren't using any of it and it's forcing us to add dead code to the `RecoveryEmailSentDialog`.

auth/src/main/java/com/firebase/ui/auth/util/AuthInstances.java
```diff
@@ -0,0 +1,42 @@
...
+
+    public static FirebaseAuth getFirebaseAuth(FlowParameters parameters) {
+        return FirebaseAuth.getInstance(FirebaseApp.getInstance(parameters.appName));
+    }
+
+    public static CredentialsApi getCredentialsApi() {
+        return Auth.CredentialsApi;
+    }
+
+    public static FirebaseUser getCurrentUser(FlowParameters parameters) {
```
SUPERCILEX: @samtstern We should probably mark this as `@Nullable`.

auth/src/main/java/com/firebase/ui/auth/ui/DialogBase.java
samtstern: Good catch!

samtstern:
@SUPERCILEX thanks for the comments!  Think I fixed them all.

SUPERCILEX:
@samtstern LGTM! 😄

Issue #780, opened 2017-06-28
NullPointerException in PhoneNumberUtils


krawa:
Android 4.4
Samsung Galaxy Ace 4 (vivaltods5m)

```
java.lang.RuntimeException: 
  at android.app.ActivityThread.deliverResults(ActivityThread.java:3710)
  at android.app.ActivityThread.handleSendResult(ActivityThread.java:3753)
  at android.app.ActivityThread.access$1400(ActivityThread.java:179)
  at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1372)
  at android.os.Handler.dispatchMessage(Handler.java:102)
  at android.os.Looper.loop(Looper.java:146)
  at android.app.ActivityThread.main(ActivityThread.java:5641)
  at java.lang.reflect.Method.invokeNative(Method.java:0)
  at java.lang.reflect.Method.invoke(Method.java:515)
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1288)
  at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1104)
  at dalvik.system.NativeStart.main(NativeStart.java:0)
Caused by: java.lang.NullPointerException: 
  at com.firebase.ui.auth.ui.phone.PhoneNumberUtils.getSimBasedLocale(PhoneNumberUtils.java:1342)
  at com.firebase.ui.auth.ui.phone.PhoneNumberUtils.getCurrentCountryInfo(PhoneNumberUtils.java:105)
  at com.firebase.ui.auth.ui.phone.PhoneNumberUtils.formatPhoneNumberUsingCurrentCountry(PhoneNumberUtils.java:99)
  at com.firebase.ui.auth.ui.phone.VerifyPhoneNumberFragment.onActivityResult(VerifyPhoneNumberFragment.java:144)
  at android.support.v4.app.FragmentActivity.onActivityResult(FragmentActivity.java:164)
  at android.app.Activity.dispatchActivityResult(Activity.java:5676)
  at android.app.ActivityThread.deliverResults(ActivityThread.java:3706)
```

samtstern:
The method in question is:

```java
    private static Locale getSimBasedLocale(@NonNull Context context) {
        final TelephonyManager tm =
                (TelephonyManager) context.getSystemService(Context.TELEPHONY_SERVICE);
        final String countryIso = tm != null ? tm.getSimCountryIso() : null;
[...]

PR #781, opened 2017-06-28
Fix Fragment getContext NPE

SUPERCILEX:
#780

@samtstern Sigh, classic Fragment `getContext()` NPE. We've [done this before](https://github.com/firebase/FirebaseUI-Android/blob/c449be1c316c321fcbb45e1f4d7d99bec1ea1ea5/auth/src/main/java/com/firebase/ui/auth/util/signincontainer/SaveSmartLock.java#L83-L87) so it should work just fine.


samtstern:
Nice!  LGTM.

PR #784, opened 2017-07-02
Style refactor

SUPERCILEX:
@samtstern I started working on making my anonymous auth PR support phone auth, but I kept shuddering some really gross phone auth code so this PR fixes that messiness. 😄

The PR
- Auto styles all documentation so it's nice and aligned and stuff
- Restyles the manifests with the trailing space
- Cleans up phone auth code
- Deletes a whole bunch of zombie code we weren't using (Yay! 😄)

PR #785, opened 2017-07-02
Code inspect cleanup

SUPERCILEX:
This PR does an Intellij code inspect and changes some semantics which is why I separated it from the style refactor PR.


SUPERCILEX:
Oh, and I forgot to mention that it's built on top of #784.

app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
```diff
@@ -156,7 +156,7 @@ public void onCreate(Bundle savedInstanceState) {
             return;
         }
 
-        if (!isGoogleConfigured()) {
+        if (isGoogleMisconfigured()) {
```
SUPERCILEX: Intellij said we were always inverting these.

auth/src/main/java/com/firebase/ui/auth/provider/IdpProvider.java
```diff
@@ -14,15 +14,13 @@
...
-import android.os.Bundle;
-
 import com.firebase.ui.auth.IdpResponse;
 
 public interface IdpProvider extends Provider {
     interface IdpCallback {
         void onSuccess(IdpResponse idpResponse);
 
-        void onFailure(Bundle extra);
+        void onFailure();
```
SUPERCILEX: We never use this bundle anywhere.

auth/src/main/java/com/firebase/ui/auth/provider/Provider.java
```diff
@@ -5,15 +5,10 @@
...
 import android.support.annotation.LayoutRes;
 
-import com.google.firebase.auth.GoogleAuthProvider;
-
 public interface Provider {
     /** Retrieves the name of the IDP, for display on-screen. */
     String getName(Context context);
 
-    /** Retrieves the id of the IDP, e.g. {@link GoogleAuthProvider#PROVIDER_ID}. */
-    String getProviderId();
```
SUPERCILEX: We never use this anywhere which makes sense since most of the time we need the provider id to figure out which class to instantiate. Thus, we would already have the provider id by the time we get an instance of a `Provider`.

auth/src/main/java/com/firebase/ui/auth/ui/phone/SubmitConfirmationCodeFragment.java
```diff
@@ -108,10 +107,7 @@ public void onActivityCreated(@Nullable Bundle savedInstanceState) {
         super.onActivityCreated(savedInstanceState);
 
         if (savedInstanceState != null) {
-            Long t = savedInstanceState.getLong(EXTRA_MILLIS_UNTIL_FINISHED);
-            if (t != null) {
-                mCountdownTimer.update(t);
-            }
+            mCountdownTimer.update(savedInstanceState.getLong(EXTRA_MILLIS_UNTIL_FINISHED));
```
SUPERCILEX: `getLong` returns a `long`, not a `Long` so it's always going to be non-null.

samtstern:
LGTM!

PR #786, opened 2017-07-02
Auth instances cleanup

SUPERCILEX:
@samtstern While working on the anonymous auth PR, it was frustrating to have to pass in flow params each time I wanted to get an auth or user instance. This PR fixes that issue with a slightly hacky solution of storing the flow params in a static variable that's reset each time the helper activity is created.

This PR is built on top of #785 which is built on top of #784.


samtstern:
This change strikes me as a little strange.  We've avoided, until now, keeping `FlowParams` as static.  Instead each `Activity` and `Fragment` has kept the flow params as a member variable and passed it along through the flow.

If we do this change, then we'd have a mixed approach.  A more consistent thing to do would be to make this change and then add a `static FlowParameters getFlowParams()` method to `AuthInstances` but I don't really like that idea since `static` variables are generally code smell.

SUPERCILEX:
@samtstern Yeah, upon looking back at this PR it does seem kinda gross. I've updated it to convert the `AuthInstances` into an `AuthHelper` object. While I know you were trying to get rid of the helpers, I feel like keeping a simple and contained helper just for the auth stuff makes sense and reduces the bloat of having to pass in those `FlowParams` each time. What do you think?

SUPERCILEX:
@samtstern Well that was rough. Finally got the build to pass though! 😄

samtstern:
Sorry I forgot about this, I am fine with this change!

SUPERCILEX:
@samtstern awesome possum! 😀

PR #788, opened 2017-07-03
Removed static from class definition.

gazialankus:
The actual code does not have static. With static it won't compile.


samtstern:
@gazialankus this looks like an improvement, thanks!

PR #796, opened 2017-07-10
Use resource prefix for library resources

### LINKED ISSUE #794: name of resources
Please change all names of resources.

For example:
    `<string name="done" translation_description="Progress dialog success transition once sms is sent">Done</string>`

has name "done". In my projects I using the same name with another content.
Please change all resource names according library pattern. 
For example: "firebase_ui_done"
    `<string name="firebase_ui_done" translation_description="Progress dialog success transition once sms is sent">Done</string>`

TheCraftKid:
This fixes #794.


auth/src/test/java/com/firebase/ui/auth/ui/phone/SpacedEditTextTest.java
```diff
@@ -52,9 +52,9 @@ public void setUp() throws Exception {
         context = mock(Context.class);
         array = mock(TypedArray.class);
 
-        when(array.getFloat(R.styleable.SpacedEditText_spacingProportion, 1)).thenReturn
+        when(array.getFloat(R.styleable.fui_SpacedEditText_fui_spacingProportion, 1)).thenReturn
```
samtstern: Any idea why this one go a double `fui`?

database/src/main/java/com/firebase/ui/database/FirebaseListAdapter.java
```diff
@@ -27,21 +29,21 @@
 public abstract class FirebaseListAdapter<T> extends BaseAdapter implements FirebaseAdapter<T> {
     private static final String TAG = "FirebaseListAdapter";
 
-    protected final Activity mActivity;
+    protected final Context mContext;
```
SUPERCILEX: Can we remove the DB changes since they aren't related to this PR? Otherwise LGTM! 😀

auth/src/main/java/com/firebase/ui/auth/ui/phone/CountryListAdapter.java
```diff
@@ -38,7 +38,7 @@
     private String[] sections;
 
     public CountryListAdapter(Context context) {
-        super(context, R.layout.dgts__country_row, android.R.id.text1);
+        super(context, R.layout.fui_dgts__country_row, android.R.id.text1);
```
SUPERCILEX: While we're renaming, might as well remove that extra `_`.

database/src/main/java/com/firebase/ui/database/FirebaseListAdapter.java
samtstern: I think those will be fixed after the merge conflicts are fixed, those changes have already been accepted in another PR.

SUPERCILEX:
@TheCraftKid might I also suggest you try branching? 😂 Nah, I'm just kidding with you but you should always try starting a PR by checking out a fresh version of `upstream/dev`, or in this case the 2.1 dev branch. Otherwise you end up with a whole bunch of conflicts and unrelated change like that DB stuff. Best of luck resolving those conflicts! 😀

database/src/main/java/com/firebase/ui/database/FirebaseListAdapter.java
SUPERCILEX: @samtstern yeah, once I saw the conflicts I realized this PR was built using personal commits instead of dev. Hopefully it won't be too tough to resolve them! 😀

auth/src/main/java/com/firebase/ui/auth/ui/phone/CountryListAdapter.java
TheCraftKid: Will do.

database/src/main/java/com/firebase/ui/database/FirebaseListAdapter.java
TheCraftKid: Yeah, I'm still learning how to use git. Will do.

auth/src/test/java/com/firebase/ui/auth/ui/phone/SpacedEditTextTest.java
TheCraftKid: I think that's an effect of the build tools appending the two resource names together. I'll make sure it's reduced to `SpacedEditText_fui_spacingProportion`.

samtstern:
@TheCraftKid we're getting close to `2.1.0` and I am happy to include this in the release if you have time to finish it in the next day or two.  What do you think?  If you have any questions or need some help just let me know.

TheCraftKid:
Yep, sorry about the delay; a Windows update messed up my desktop environment.

On a side note, I learned how to use branches and remove the irrelevant code to this PR. Is there a way for me to add my commits to this PR, or will I have to make a new one?

samtstern:
@TheCraftKid just commit again and push to the same branch, the PR will update like magic.

TheCraftKid:
**Wow**, that took way too long. That updated like magic, all right. I'm really sorry if there are some screwy commits or anything else in this PR.

Final note: just to get the linter to stop complaining, I added `tools:ignore="ResourceName"` to the root of the `strings.xml` file. Just making sure this is okay.

SUPERCILEX:
@TheCraftKid glad you pulled through!!! 😀

I hate to be the bearer of bad news again, but the issue you're trying to resolve—#794—is about the string res names, not the layouts. We definitely should keep the layout name changes (thanks!), but the `styles.xml` file should have a root res name prefix ignore tag instead of the string res which should each have the prefix. More work for you, sorry! 😊

auth/src/main/res/values/config.xml
```diff
@@ -1,4 +1,4 @@
-<resources>
+<resources xmlns:tools="http://schemas.android.com/tools">
```
SUPERCILEX: We should also ignore at root here.

TheCraftKid:
Okay, that should solve the issues @SUPERCILEX brought up.

samtstern:
@TheCraftKid we still have a ton of merge issues though, you need to merge the firebase `version-2.1.0-dev` branch into your branch and then resolve all conflicts.

samtstern:
@TheCraftKid most of these conflicts are my fault I will send a PR to your branch that will fix it.

samtstern:
Opened https://github.com/TheCraftKid/FirebaseUI-Android/pull/1 which will fix it.

Here's an issue though: because this adds the `fui_` prefix to all of the strings it messes up all of the translations.  Obviously I could just add the `fui_` prefix to all of our translated strings as well but the problem is in the translation portal we use the string `name` is used as a unique ID so I'm actually not sure how it would react to this refactoring.

Since the major goal of `2.1.0` was to get translations out I am concerned about how this PR affects it.  I'd like to punt this to `2.1.1` so I can think about the implications of adding this prefix.  

@TheCraftKid @SUPERCILEX what do you think?

TheCraftKid:
I'd be willing to keep v2.1.0 translation-based and move the resource name changes to v2.1.1 if that means we can push the essential v2.1.0 changes.

samtstern:
@TheCraftKid thanks for your flexibility, I think that's the safest move.  I definitely want to get your resource prefix change in ASAP after `2.1.0` though, it's the right thing to do for developers.

samtstern:
Woot no more conflicts!  The build is failing now but that's working as intended since we have not answered the translation question yet.

samtstern:
@TheCraftKid I opened https://github.com/TheCraftKid/FirebaseUI-Android/pull/2 to fix the final issues.

samtstern:
@TheCraftKid nice!  This one is all set to merge, thanks for the hard work.

PR #807, opened 2017-07-18
Translations prefix scripts

samtstern:
This solves the main issue in #796 

@TheCraftKid


samtstern:
Note to self: need to check on plurals handling.

samtstern:
Whoops I should be opening this against TheCraftKid

PR #808, opened 2017-07-19
Merge improvements from anonymous auth PR (#309)

SUPERCILEX:
@samtstern While working on automatically merging user profiles for #309, I made several improvements worth merging back into master:
- We no longer show that annoying Twitter request email screen!!! 😄🚀 I always knew we were doing it wrong somehow since the Firebase apis could get the Twitter email without showing that screen and that's correct--you can get profile info from the current Twitter session.
- The `IdpResponse` user data is unified and uses our `User` object under the hood now
- We collect all possible user info from all providers (we aren't using this yet, but it's used in #309 and could be useful for other things in the future.)

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/IdpResponse.java
+++ b/auth/src/main/java/com/firebase/ui/auth/IdpResponse.java
@@ -24,7 +24,6 @@ import android.text.TextUtils;
 
 import com.firebase.ui.auth.ui.ExtraConstants;
 import com.google.firebase.auth.FacebookAuthProvider;
-import com.google.firebase.auth.GithubAuthProvider;
 import com.google.firebase.auth.GoogleAuthProvider;
 import com.google.firebase.auth.TwitterAuthProvider;
 
@@ -33,23 +32,20 @@ import com.google.firebase.auth.TwitterAuthProvider;
  */
 public class IdpResponse implements Parcelable {
     private final User mUser;
-    private final String mPhoneNumber;
     private final String mToken;
     private final String mSecret;
     private final int mErrorCode;
 
     private IdpResponse(int errorCode) {
-        this(null, null, null, null, errorCode);
+        this(null, null, null, errorCode);
     }
 
     private IdpResponse(
             User user,
-            String phoneNumber,
             String token,
             String secret,
             int errorCode) {
         mUser = user;
-        mPhoneNumber = phoneNumber;
         mToken = token;
         mSecret = secret;
         mErrorCode = errorCode;
@@ -102,7 +98,7 @@ public class IdpResponse implements Parcelable {
      */
     @Nullable
     public String getPhoneNumber() {
-        return mPhoneNumber;
+        return mUser.getPhoneNumber();
     }
 
     /**
@@ -136,7 +132,6 @@ public class IdpResponse implements Parcelable {
     @Override
     public void writeToParcel(Parcel dest, int flags) {
         dest.writeParcelable(mUser, flags);
-        dest.writeString(mPhoneNumber);
         dest.writeString(mToken);
         dest.writeString(mSecret);
         dest.writeInt(mErrorCode);
@@ -147,7 +142,6 @@ public class IdpResponse implements Parcelable {
         public IdpResponse createFromParcel(Parcel in) {
             return new IdpResponse(
                     in.<User>readParcelable(User.class.getClassLoader()),
-                    in.readString(),
                     in.readString(),
                     in.readString(),
                     in.readInt()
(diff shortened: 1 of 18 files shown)


auth/src/main/java/com/firebase/ui/auth/IdpResponse.java
```diff
@@ -195,22 +187,22 @@ public Builder setSecret(String secret) {
...
         public IdpResponse build() {
-            if ((mProviderId.equalsIgnoreCase(GoogleAuthProvider.PROVIDER_ID)
-                    || mProviderId.equalsIgnoreCase(FacebookAuthProvider.PROVIDER_ID)
-                    || mProviderId.equalsIgnoreCase(TwitterAuthProvider.PROVIDER_ID)
-                    || mProviderId.equalsIgnoreCase(GithubAuthProvider.PROVIDER_ID))
+            String providerId = mUser.getProviderId();
+            if ((providerId.equalsIgnoreCase(GoogleAuthProvider.PROVIDER_ID)
+                    || providerId.equalsIgnoreCase(FacebookAuthProvider.PROVIDER_ID)
+                    || providerId.equalsIgnoreCase(TwitterAuthProvider.PROVIDER_ID)
+                    || providerId.equalsIgnoreCase(GithubAuthProvider.PROVIDER_ID))
```
samtstern: What's github doing here?

auth/src/main/java/com/firebase/ui/auth/provider/FacebookProvider.java
```diff
@@ -133,19 +135,29 @@ public void onCompleted(JSONObject object, GraphResponse response) {
                             Log.w(TAG, "Received null response from Facebook GraphRequest");
                             onFailure();
                         } else {
+                            String email = null;
+                            String name = null;
+                            Uri photoUri = null;
+
+                            try {
+                                email = object.getString("email");
+                            } catch (JSONException ignored) { }
```
samtstern: Shouldn't we at least log a warning if we can't get email?

auth/src/main/java/com/firebase/ui/auth/provider/TwitterProvider.java
```diff
@@ -92,9 +92,27 @@ public void startLogin(Activity activity) {
...
 
     @Override
-    public void success(Result<TwitterSession> result) {
-        mTwitterAuthClient.requestEmail(result.data,
-                new EmailCallback(result.data, mCallbackObject));
+    public void success(final Result<TwitterSession> sessionResult) {
+        TwitterCore.getInstance()
+                .getApiClient()
+                .getAccountService()
+                .verifyCredentials(false, false, true)
```
samtstern: Can you add a comment documenting the `false, false, true` magic sequence?

auth/src/main/java/com/firebase/ui/auth/IdpResponse.java
```diff
@@ -32,26 +32,23 @@
...
     private IdpResponse(int errorCode) {
-        this(null, null, null, null, null, errorCode);
+        this(null, null, null, null, errorCode);
     }
 
     private IdpResponse(
-            String providerId,
-            String email,
+            User user,
             String phoneNumber,
```
samtstern: any reason why email was merged into `User` but not phoneNumber?

SUPERCILEX: Good catch!

auth/src/main/java/com/firebase/ui/auth/IdpResponse.java
SUPERCILEX: Lol, must have been a merge mistake from a long time ago.

auth/src/main/java/com/firebase/ui/auth/provider/FacebookProvider.java
SUPERCILEX: Yup.

auth/src/main/java/com/firebase/ui/auth/provider/TwitterProvider.java
SUPERCILEX: TBH, I just copied that from StackOverflow 😆. I did a bit of debugging and figured some of it out (in order): `includeEntities` will populate the `User#entities` field but we don't need it, `skipStatus` I have no idea what that does, `includeEmail` includes the email (duh! 😄).

Do you want me to document the parameter names? Intellij shows them, but it could be useful for GitHub:
image

samtstern: That's alright, thanks for clarifying!

samtstern:
@SUPERCILEX last thing left is to target the `2.1.1-dev` branch.

SUPERCILEX:
@samtstern Oops, I've updated the base now.

PR #813, opened 2017-07-21
Fix AuthMethodPickerActivity title being "ui_flow" instead of app name

SUPERCILEX:
We want it to be overridden.

### DIFF
--- a/auth/src/main/AndroidManifest.xml
+++ b/auth/src/main/AndroidManifest.xml
@@ -63,7 +63,7 @@
 
         <activity
             android:name="com.facebook.FacebookActivity"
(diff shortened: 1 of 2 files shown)


TheCraftKid:
Hey, we should probably document this since we're requiring an `app_name` string to override the library's. Is the best idea for usability, though?

samtstern:
@SUPERCILEX there was no behavior change in recent versions regarding `app_name` right?

SUPERCILEX:
@samtstern Nope, we've always used `app_name` because it's the default thing for AppCompat. We should probably put it on our mental list of stuff to fail the build in `v3.0` if the dev doesn't supply it.

PR #814, opened 2017-07-21
Add true support for vector drawables

SUPERCILEX:
@samtstern I didn't know this at the time, but we were never actually getting the true benefits of vector drawables because the build tools would just convert them to pngs at the end of the day. I went through the pains of figuring out how to enable true vector drawable support for my app and I'm bringing this knowledge over to FUI. Yay! 😄 

Turns out we actually save quite a few kbs considering how few drawables we have:

Old | New | Diff | % of original
--- | --- | --- | ---
496KB | 468KB | 28KB | 94%


auth/src/main/res/values/styles.xml
```diff
@@ -199,13 +199,13 @@
     </style>
 
     <style name="FirebaseUI.Button.AccountChooser.EmailButton">
-        <item name="android:drawableLeft">@drawable/fui_ic_mail_white_24dp</item>
+        <item name="drawableStartCompat">@drawable/fui_ic_mail_white_24dp</item>
```
SUPERCILEX: Pre-L this will crash with a res not found exception if we don't manually get the drawable with `AppCompatResources.getDrawable`.

samtstern:
Oh nice!  I'll take another look at this early next week (and your other PRs, I have been slacking on this repo).  I remember when we went with vectors for Santa Tracker there were a bunch of gotchas and I just want to make sure they don't apply here (I think we're OK).

SUPERCILEX:
@samtstern SGTM! (I also have another PR coming up 😊)

puf:
Very cool to see this change! #vectorallthethings

SUPERCILEX:
@samtstern Was this closed by accident or did you decide against it?

samtstern:
@SUPERCILEX accident!  Sorry haha.

SUPERCILEX:
@samtstern 😄👍

app/src/main/res/values/config.xml
```diff
@@ -25,5 +25,5 @@
     operations as your app. Embedding the Consumer Key and Consumer Secret in your app, no matter
     how obfuscated, makes it vulnerable to being stolen.
     -->
-    <string name="twitter_consumer_secret" translatable="false">CHANGE-ME</string>
+    <string name="twitter_consumer_secret" translatable="false">di6UQRwDaxsRQ7RT5M5QcUgq19Dn4NY04QLMjaDhyxqmUy92Pe</string>
```
samtstern: This seems like a mistake haha

auth/src/main/java/com/firebase/ui/auth/ui/idp/SupportVectorDrawablesButton.java
```diff
@@ -0,0 +1,83 @@
...
+import android.content.res.TypedArray;
+import android.graphics.drawable.Drawable;
+import android.os.Build;
+import android.support.v7.content.res.AppCompatResources;
+import android.support.v7.widget.AppCompatButton;
+import android.util.AttributeSet;
+
+import com.firebase.ui.auth.R;
+
+public class SupportVectorDrawablesButton extends AppCompatButton {
```
samtstern: This class seems to have some cool but non-obvious compatibility tricks.  Can you add comments (class level and in the one big method) explaining what's going on?

auth/src/main/res/values/styles.xml
samtstern: So is this the reason for the button subclass?

app/src/main/res/values/config.xml
SUPERCILEX: Oh lol, yep.

auth/src/main/res/values/styles.xml
SUPERCILEX: Yup.

samtstern:
LGTM!

PR #815, opened 2017-07-21
Fix broken auto styling

SUPERCILEX:
@samtstern Found some more broken res refactors.

### DIFF
--- a/auth/src/main/res/values/colors.xml
+++ b/auth/src/main/res/values/colors.xml
@@ -1,9 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-    <color name="fui_colorPrimary">#3F51B5</color>
-    <color name="fui_colorPrimaryDark">#303F9F</color>
-    <color name="fui_colorAccent">#FF4081</color>
-
     <color name="fui_errorColor">#DD2C00</color>
     <color name="fui_linkColor">#4285F4</color>
     <color name="fui_transparent">#00000000</color>
--- a/auth/src/main/res/values/colors.xml
+++ b/auth/src/main/res/values/colors.xml
@@ -1,5 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
-<resources>
+<resources xmlns:tools="http://schemas.android.com/tools">
+    <!-- Eventually, we should remove these so the developer knows they have to style FirebaseUI -->
+    <color name="colorPrimary" tools:ignore="ResourceName">#3F51B5</color>
+    <color name="colorPrimaryDark" tools:ignore="ResourceName">#303F9F</color>
+    <color name="colorAccent" tools:ignore="ResourceName">#FF4081</color>
+
     <color name="fui_errorColor">#DD2C00</color>
     <color name="fui_linkColor">#4285F4</color>
     <color name="fui_transparent">#00000000</color>
--- a/auth/src/main/res/values/styles.xml
+++ b/auth/src/main/res/values/styles.xml
@@ -3,9 +3,9 @@
            tools:ignore="ResourceName">
 
     <style name="FirebaseUI" parent="Theme.AppCompat.DayNight.DarkActionBar">
-        <item name="colorPrimary">@color/fui_colorPrimary</item>
-        <item name="colorPrimaryDark">@color/fui_colorPrimaryDark</item>
-        <item name="colorAccent">@color/fui_colorAccent</item>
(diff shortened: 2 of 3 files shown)


auth/src/main/res/values/colors.xml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
-<resources>
-    <color name="fui_colorPrimary">#3F51B5</color>
-    <color name="fui_colorPrimaryDark">#303F9F</color>
-    <color name="fui_colorAccent">#FF4081</color>
+<resources xmlns:tools="http://schemas.android.com/tools">
+    <color name="colorPrimary" tools:ignore="ResourceName">#3F51B5</color>
```
samtstern: Oops.  @TheCraftKid tried to stop me from doing this and I told him it would be fine.

Does this come into play if the app including FirebaseUI does not use AppCompat at all so `colorPrimary` is undefined?

SUPERCILEX: @samtstern oh sweet!!! You gave me an idea, I'll update this PR in just a few.

auth/src/main/res/values/styles.xml
```diff
@@ -3,9 +3,9 @@
            tools:ignore="ResourceName">
 
     <style name="FirebaseUI" parent="Theme.AppCompat.DayNight.DarkActionBar">
-        <item name="colorPrimary">@color/fui_colorPrimary</item>
-        <item name="colorPrimaryDark">@color/fui_colorPrimaryDark</item>
-        <item name="colorAccent">@color/fui_colorAccent</item>
+        <item name="colorPrimary">@color/colorPrimary</item>
+        <item name="colorPrimaryDark">@color/colorPrimaryDark</item>
+        <item name="colorAccent">@color/colorAccent</item>
```
SUPERCILEX: @samtstern The reason we defined those colors was so that we could use them here. However, we are expecting them to be overridden so we don't actually need them to be fully defined colors, but my knowledge of the Android resource system wasn't that great back then. Anyway, now we just define them as ids of type `color` so the build will properly fail if devs using FirebaseUI don't define their colors and they'll be properly overridden if they are defined.

TheCraftKid: Is the intended effect to require developers to override `colorPrimary` and etc.?

SUPERCILEX: @TheCraftKid Yeah, since FirebaseUI `v2.0.0` they already do, but this got broken in #796

SUPERCILEX: See https://github.com/firebase/FirebaseUI-Android/pull/696#discussion_r115584842 for an in-depth explanation.

auth/src/main/res/values/colors.xml
```diff
@@ -1,9 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-    <color name="fui_colorPrimary">#3F51B5</color>
```
samtstern: Why are we now deleting these entirely rather than just removing the `fui_`?

samtstern:
@SUPERCILEX this is the last thing I want to fix and then we can get out `2.1.1`.

auth/src/main/res/values/colors.xml
SUPERCILEX: See my [earlier comment](https://github.com/firebase/FirebaseUI-Android/pull/815#discussion_r128875877). Those colors are basically a lie and we aren't expecting them to ever be used. The dev should have to override them so that FirebaseUI is styled correctly. In theory, 99% of people won't have to do anything because they'll already have those colors defined since AppComat uses them.

If you want, we can keep the defined colors but it makes it less clear to the developer that they need to override them because they won't get a build error. Your call! 😀

samtstern: Since it's technically a "breaking change" to remove them and cause a build error for that 1%, let's keep the default colors in place.

In a future version we can really pull the rug out and make sure developers override them.

SUPERCILEX: Ah, true. SGTM!

SUPERCILEX:
@samtstern Green! 😄

samtstern:
@SUPERCILEX thanks for being so helpful/responsive on this one.

SUPERCILEX:
@samtstern 😄

samtstern:
@SUPERCILEX thanks for being so helpful/responsive on this one.

PR #816, opened 2017-07-21
Fix bad ellipsis spacing

SUPERCILEX:
@samtstern For some reason, some strings have a space before the ellipsis so I added a fix to the scripts.


samtstern:
Nice! Will merge once green.

Issue #819, opened 2017-07-23
NPE in SaveSmartLock


RedCider:
* Android device: Emulator 
  * Android OS version: Android 7.1.1 (Google APIs)
  * Google Play Services version: 11.0.2
  * Firebase/Play Services SDK version: 11.0.2
  * FirebaseUI version: 2.1.0
  
### Description:
NPE occurs when Smart Lock is used for re-login of user previously signed in via google. It works without Problems with Version 2.0.0 of Firebase UI

#### Steps to reproduce:

  1. Sign in with google
  2. Log out => Logout returns to Login Activity
  3. Pick account
  4. Progress Dialog appears and app crashes with NPE (Activity is null, so getSupportFragmentManager() cannot be called)
  
#### Observed Results:

 Stack Trace:
`07-23 12:37:33.345 6396-6396/com.sonebase.apps.whoop E/UncaughtException: java.lang.NullPointerException: Attempt to invoke virtual method 'android.support.v4.app.FragmentManager com.firebase.ui.auth.ui.HelperActivityBase.getSupportFragmentManager()' on a null object reference
                                                                              at com.firebase.ui.auth.util.signincontainer.SaveSmartLock.getInstance(SaveSmartLock.java:63)
                                                                              at com.firebase.ui.auth.util.AuthHelper.getSaveSmartLockInstance(AuthHelper.java:40)
                                                                              at com.firebase.ui.auth.util.signincontainer.IdpSignInContainer.onCreate(IdpSignInContainer.java:88)
                                                                              at android.support.v4.app.Fragment.performCreate(Fragment.java:2324)
                                                                              at android.support.v4.app.FragmentManagerImpl.moveToState(FragmentManager.java:1376)
                                                                              at android.support.v4.app.FragmentManagerImpl.moveFragmentToExpectedState(FragmentManager.java:1739)
                                                                              at android.support.v4.app.FragmentManagerImpl.moveToState(FragmentManager.java:1808)
                                                                              at android.support.v4.app.BackStackRecord.executeOps(BackStackRecord.java:799)
                                                                              at android.support.v4.app.FragmentManagerImpl.executeOps(FragmentManager.java:2579)
                                                                              at android.support.v4.app.FragmentManagerImpl.executeOpsTogether(FragmentManager.java:2366)
                                                                              at android.support.v4.app.FragmentManagerImpl.removeRedundantOperationsAndExecute(FragmentManager.java:2321)
[...]

PR #821, opened 2017-07-23
Fix NPE getting fragment manager

SUPERCILEX:
#819


samtstern:
Thanks for the quick fix!

Issue #718, opened 2017-05-21
Use Android Architecture Components


TheCraftKid:
Google I/O unveiled a new set of [Android Architecture Components](https://developer.android.com/topic/libraries/architecture/index.html), and it seems some components, like LiveData and ViewModel, fit almost perfectly to the functions performed by the library. Are there any plans to use these components when they' are stable and are compatible with the support libraries?

SUPERCILEX:
@TheCraftKid Yeah, I definitely think that would be awesome to support!!! At the moment I've been reeling from all the announcement and updates at I/O so I've been overwhelmed with trying to learn Kotin and updating my app with all the new stuff. Once I have the time, I'd love to take a look at that and how we could implement it. 😄 For now though I think I have enough open PRs 😆

TheCraftKid:
@SUPERCILEX I'd be happy to submit a PR while the rest of FirebaseUI is being sorted out.

SUPERCILEX:
@TheCraftKid Go for it! 😄 (Just make sure to base your PR off the `version-2.0.0-dev` branch)

TheCraftKid:
Okay, I got a bit excited, and while the components will change before release, I think I can begin the process of implementing `LiveData`. 
The first decision to make when implementing this is composition vs inheritance. In the 2.0.0 branch, `FirebaseArray` extends `CachingObservableSnapshotArray` which extends `ObservableSnapshotArray` which extends `ImmutableList`. Making `ImmutableList` extend `LiveData` doesn't make sense, so I'm thinking we make `ObservableSnapshotArray` implement `List` and have it extend `LiveData` instead.

What are your thoughts, @SUPERCILEX?

SUPERCILEX:
@TheCraftKid Nope. 😆 You're getting there though! 😀 You're right about the composition part, but extending `ImmutableList` is a hard requirement so we can't change that. Instead of updating the existing classes, let's make a new class that extends `LiveData` and takes in an `ObservableSnapshotArray` in the constructor. Then, you can add a listener to the snapshot array and notify the livedata of changes whenever they come in.

b95505017:
I'm curious about how `LiveData` notify the partial data updated (onChildChanged).
[...]

PR #827, opened 2017-07-26
Add support for Android Architecture Components

SUPERCILEX:
#718

### DIFF
--- a/app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
+++ b/app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
@@ -43,6 +43,9 @@ public class ChatActivity extends AppCompatActivity
         implements FirebaseAuth.AuthStateListener, View.OnClickListener, LifecycleRegistryOwner {
     private static final String TAG = "RecyclerViewDemo";
 
+    // TODO remove once arch components are merged into support lib
+    private final LifecycleRegistry mRegistry = new LifecycleRegistry(this);
+
     private FirebaseAuth mAuth;
     protected DatabaseReference mChatRef;
     private Button mSendButton;
@@ -176,9 +179,6 @@ public class ChatActivity extends AppCompatActivity
         mSendButton.setEnabled(isSignedIn());
         mMessageEdit.setEnabled(isSignedIn());
     }
-
-    // TODO remove once arch components are merged into support lib
-    private final LifecycleRegistry mRegistry = new LifecycleRegistry(this);
 
     @Override
     public LifecycleRegistry getLifecycle() {
--- a/app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
+++ b/app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
@@ -14,6 +14,8 @@
 
 package com.firebase.uidemo.database;
 
+import android.arch.lifecycle.LifecycleRegistry;
+import android.arch.lifecycle.LifecycleRegistryOwner;
 import android.os.Bundle;
 import android.support.annotation.NonNull;
 import android.support.v7.app.AppCompatActivity;
@@ -26,6 +28,8 @@ import android.widget.EditText;
 import android.widget.TextView;
 import android.widget.Toast;
 
+import com.firebase.ui.database.ClassSnapshotParser;
+import com.firebase.ui.database.FirebaseArray;
 import com.firebase.ui.database.FirebaseRecyclerAdapter;
 import com.firebase.uidemo.R;
 import com.firebase.uidemo.util.SignInResultNotifier;
@@ -37,7 +41,8 @@ import com.google.firebase.database.DatabaseReference;
 import com.google.firebase.database.FirebaseDatabase;
 import com.google.firebase.database.Query;
 
-public class ChatActivity extends AppCompatActivity implements FirebaseAuth.AuthStateListener, View.OnClickListener {
+public class ChatActivity extends AppCompatActivity
+        implements FirebaseAuth.AuthStateListener, View.OnClickListener, LifecycleRegistryOwner {
     private static final String TAG = "RecyclerViewDemo";
 
     private FirebaseAuth mAuth;
@@ -47,7 +52,7 @@ public class ChatActivity extends AppCompatActivity implements FirebaseAuth.Auth
 
     private RecyclerView mMessages;
     private LinearLayoutManager mManager;
-    protected FirebaseRecyclerAdapter<Chat, ChatHolder> mAdapter;
+    private FirebaseRecyclerAdapter<Chat, ChatHolder> mAdapter;
     protected TextView mEmptyListMessage;
 
     @Override
@@ -70,8 +75,10 @@ public class ChatActivity extends AppCompatActivity implements FirebaseAuth.Auth
         mManager.setReverseLayout(false);
 
         mMessages = (RecyclerView) findViewById(R.id.messagesList);
-        mMessages.setHasFixedSize(false);
+        mMessages.setHasFixedSize(true);
         mMessages.setLayoutManager(mManager);
+
+        if (isSignedIn()) { attachRecyclerViewAdapter(); }
     }
 
     @Override
@@ -81,19 +88,7 @@ public class ChatActivity extends AppCompatActivity implements FirebaseAuth.Auth
         // Default Database rules do not allow unauthenticated reads, so we need to
         // sign in before attaching the RecyclerView adapter otherwise the Adapter will
         // not be able to read any data from the Database.
-        if (isSignedIn()) {
-            attachRecyclerViewAdapter();
-        } else {
-            signInAnonymously();
-        }
-    }
-
-    @Override
-    public void onStop() {
-        super.onStop();
-        if (mAdapter != null) {
-            mAdapter.cleanup();
-        }
+        if (!isSignedIn()) { signInAnonymously(); }
     }
 
     @Override
@@ -144,10 +139,10 @@ public class ChatActivity extends AppCompatActivity implements FirebaseAuth.Auth
     protected FirebaseRecyclerAdapter<Chat, ChatHolder> getAdapter() {
         Query lastFifty = mChatRef.limitToLast(50);
         return new FirebaseRecyclerAdapter<Chat, ChatHolder>(
-                Chat.class,
+                new FirebaseArray<>(lastFifty, new ClassSnapshotParser<>(Chat.class)),
                 R.layout.message,
                 ChatHolder.class,
-                lastFifty) {
+                this) {
             @Override
             public void populateViewHolder(ChatHolder holder, Chat chat, int position) {
                 holder.bind(chat);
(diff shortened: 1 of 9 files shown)


SUPERCILEX:
@samtstern see https://github.com/firebase/FirebaseUI-Android/issues/718#issuecomment-317928588.

database/build.gradle
```diff
@@ -27,6 +27,10 @@ dependencies {
     // Needed to override play services
     compile "com.android.support:support-v4:$supportLibraryVersion"
 
+    compile 'android.arch.lifecycle:runtime:1.0.0-alpha5'
+    compile 'android.arch.lifecycle:extensions:1.0.0-alpha5'
+    annotationProcessor 'android.arch.lifecycle:compiler:1.0.0-alpha5'
```
SUPERCILEX: I wanted to make these `provided` so devs didn't have to pull them in, but I kept getting compile errors for the `database` module so I gave up. [Their](http://www.methodscount.com/?lib=android.arch.lifecycle%3Aruntime%3A1.0.0-alpha5) [size](http://www.methodscount.com/?lib=android.arch.lifecycle%3Aextensions%3A1.0.0-alpha5) isn't too big though.

database/src/main/java/com/firebase/ui/database/FirebaseRecyclerAdapter.java
```diff
@@ -90,6 +110,17 @@ public void cleanup() {
         mSnapshots.removeChangeEventListener(this);
     }
 
+    @SuppressWarnings("unused")
+    @OnLifecycleEvent(Lifecycle.Event.ON_ANY)
+    void cleanup(LifecycleOwner source, Lifecycle.Event event) {
+        if (event == Lifecycle.Event.ON_STOP) {
+            cleanup();
+            notifyDataSetChanged();
```
SUPERCILEX: Without this I was getting `Inconsistency detected. Invalid view holder adapter positionViewHolder` exceptions. I also got worried about the RecyclerView querying the `FirebaseArray` while it still had another listener and still getting the items so I added [the check](https://github.com/firebase/FirebaseUI-Android/pull/827/files#diff-724a30131c3284a3a9d723d6bb38561dR168) in `getItemCount`.

samtstern:
@SUPERCILEX thanks for experimenting with this.  I am not as familiar with these components as you are, so I'm going to ask all the dumb questions that come to mind.

app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
```diff
@@ -70,8 +75,10 @@ protected void onCreate(Bundle savedInstanceState) {
         mManager.setReverseLayout(false);
 
         mMessages = (RecyclerView) findViewById(R.id.messagesList);
-        mMessages.setHasFixedSize(false);
+        mMessages.setHasFixedSize(true);
```
samtstern: Kinda unrelated to the point of this PR but I feel like we keep flipping this boolean back and forth, what did this change do?

app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
```diff
@@ -182,4 +177,12 @@ private void updateUI() {
         mSendButton.setEnabled(isSignedIn());
         mMessageEdit.setEnabled(isSignedIn());
     }
+
+    // TODO remove once arch components are merged into support lib
```
samtstern: Is the idea that eventually an `AppCompatActivity` will automatically contain the registry?

database/src/main/java/com/firebase/ui/database/FirebaseRecyclerAdapter.java
```diff
@@ -88,6 +108,17 @@ public void startListening() {
...
         mSnapshots.removeChangeEventListener(this);
+        notifyDataSetChanged();
+    }
+
+    @SuppressWarnings("unused")
+    @OnLifecycleEvent(Lifecycle.Event.ON_ANY)
+    void cleanup(LifecycleOwner source, Lifecycle.Event event) {
+        if (event == Lifecycle.Event.ON_STOP) {
+            cleanup();
+        } else if (event == Lifecycle.Event.ON_DESTROY) {
```
samtstern: What are the guarantees that we ever get the destroy event?  And if it's not guaranteed, does it matter?

app/src/main/java/com/firebase/uidemo/database/ChatIndexActivity.java
```diff
@@ -34,12 +35,13 @@ public void onClick(View v) {
 
     @Override
     protected FirebaseRecyclerAdapter<Chat, ChatHolder> getAdapter() {
-        return new FirebaseIndexRecyclerAdapter<Chat, ChatHolder>(
-                Chat.class,
+        return new FirebaseRecyclerAdapter<Chat, ChatHolder>(
+                new FirebaseIndexArray<>(
```
samtstern: Why do we need to create the `IndexArray` manually now?

app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
SUPERCILEX: @samtstern since adding architecture components was all about improving performance for my app, I finally decided to figure out what this flag does. Basically, if you can guarantee that the recyclerview's size isn't affected by its children, then it doesn't have to invalidate itself whenever the data changes. This is especially important for a chat app since messages are coming in all the time.

To get this to work, I changed the recyclerview height to `match_parent` instead of `wrap_content`.

app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
SUPERCILEX: Yeah, you can see the [documentation for `LifecycleActivity`](https://developer.android.com/reference/android/arch/lifecycle/LifecycleActivity.html). Basically, Fragments and activities will be lifecycle owners once AC hits stable.

PS: I didn't extend `LifecycleActivity` because it extends `FragmentActivity` instead of `AppCompatActivity`.

database/src/main/java/com/firebase/ui/database/FirebaseRecyclerAdapter.java
SUPERCILEX: Hmmm, I'm not sure what you mean. AFAIK, we'll get an `onDestroy` if one occurred—always.

app/src/main/java/com/firebase/uidemo/database/ChatIndexActivity.java
SUPERCILEX: Yeah, I wasn't sure about that. I only added the `LifecycleOwner` parameter to the constructor that takes an `ObservableSnapshotArray` because it looked insane to have 7 constructor overloads.

If you think we should provide an overload with a `LifecycleOwner` for each constructor then I'll be happy to add them, I just wasn't sure if we should go that route.

database/src/main/java/com/firebase/ui/database/FirebaseRecyclerAdapter.java
SUPERCILEX: Yeah, I'm 99.9% sure it will work because I copied this from the Android source code for removing `LiveData` observers:
```java
@SuppressWarnings("unused")
@OnLifecycleEvent(Lifecycle.Event.ON_ANY)
void onStateChange() {
    if (owner.getLifecycle().getCurrentState() == DESTROYED) {
        removeObserver(observer);
        return;
    }
    // immediately set active state, so we'd never dispatch anything to inactive
    // owner
    activeStateChanged(isActiveState(owner.getLifecycle().getCurrentState()));
}
```

app/src/main/java/com/firebase/uidemo/database/ChatIndexActivity.java
samtstern: I think if we merge this PR we will recommend that everyone use the `LifecycleOwner` version of the adapters, but I wouldn't want those people to also have to understand `FirebaseArray`.  So we either need to do all the overloads or find some other way to get the owner in there.

database/src/main/java/com/firebase/ui/database/FirebaseRecyclerAdapter.java
samtstern: SGTM, thanks for checking.

app/src/main/java/com/firebase/uidemo/database/ChatIndexActivity.java
SUPERCILEX: @samtstern Ok, I decided to go with outside developer happiness instead of ours and added a whole bunch of overloads.

app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
```diff
@@ -182,4 +176,12 @@ private void updateUI() {
...
+    private final LifecycleRegistry mRegistry = new LifecycleRegistry(this);
```
samtstern: nit: let's move this up to be with the other instance variables.

TheCraftKid:
It's a shame `LiveData` wasn't a fit for FirebaseUI Database, but I'm not complaining. This LGTM.

app/src/main/java/com/firebase/uidemo/database/ChatActivity.java
SUPERCILEX: SGTM!

SUPERCILEX:
@TheCraftKid not really. Dunno if you've used AC much, but the point of `LiveData` as I understand it is to be able to add a listener that's lifecycle aware and won't leak contexts. This PR basically retrofits the Firebase listeners to be lifecycle aware which is what `LiveData` is for.

As I see it, `LiveData`'s main use for retrofitting libraries that don't yet support AC and for internal app listeners.

Feel free to correct me if I'm wrong! 😄

keyboardr:
Kind of a late response here, but LiveData is about more than just having a lifecycle-aware listener (although that's its main benefit). MediatorLiveData (esp. via Transformations) can create a whole system of chained LiveDatas where updates in one can trigger updates in others, leading to a much more functional style of programming.

I'm curious what would preclude a small shim class to translate between a ValueEventListener and a LiveData? I've written an example at https://gist.github.com/keyboardr/8259533ec74f703f9e999f103a6b1135

samtstern:
@keyboardr you are right and I think simple Live Data classes like that could be useful!  Do you mind filing a separate issue for that in particular?

Also I had this same discussion with someone at the Firebase summit in Amsterdam, was that you?

keyboardr:
Not me, but glad to see I'm not the only one with this idea. FWIW the shim linked above has been working decently well for me, but hasn't yet been rigorously validated. I'll file an issue to get something canonical made.

SUPERCILEX:
@keyboardr Oh no, you found my old comments! 😆 Yeah, I've learned way more over the past few months.

Anyway, that sounds like a really good idea. I've written some classes to manage a `FirestoreArray`'s lifecycle, but not a specific document. I think it would be really nice to add a whole bunch of lifecycle aware classes like that. For example, I'd like to have the adapter save its `FirestoreArray` across screen rotations, and lifecyclize the other database classes. Would you like to open an issue or should I?

keyboardr:
You're probably more familiar with FirebaseUI in particular (I've used Firebase in a few projects, but haven't used it for directly driving UI), so you'd probably have more context about what kinds of wins can be had here.

SUPERCILEX:
SGTM, will do! 😄

PR #828, opened 2017-07-26
Ouch! Fix terrible db listener leak

SUPERCILEX:


database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
```diff
@@ -85,8 +85,32 @@ public String parseSnapshot(DataSnapshot snapshot) {
...
 
-        mKeySnapshots.addChangeEventListener(this);
+        return listener;
+    }
+
+    @Override
+    public void removeChangeEventListener(@NonNull ChangeEventListener listener) {
+        super.removeChangeEventListener(listener);
+        if (!isListening()) {
+            mKeySnapshots.removeChangeEventListener(this);
```
SUPERCILEX: I can't believe I never noticed this, but we never actually removed the `FirebaseArray` listener! 😱

database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
```diff
@@ -85,8 +85,32 @@ public String parseSnapshot(DataSnapshot snapshot) {
...
+    public ChangeEventListener addChangeEventListener(@NonNull ChangeEventListener listener) {
+        boolean wasListening = isListening();
+        super.addChangeEventListener(listener);
+
+        // Only start listening when the first listener is added
+        if (!wasListening) {
+            mKeySnapshots.addChangeEventListener(this);
+        }
 
-        mKeySnapshots.addChangeEventListener(this);
```
SUPERCILEX: And on top of that, we would start listening immediately when the class was initialized instead of waiting for the first listener like we properly do in `FirebaseArray`. Double ouch! 😱

database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
```diff
@@ -85,8 +85,32 @@ public String parseSnapshot(DataSnapshot snapshot) {
...
         });
+    }
+
+    @Override
+    public ChangeEventListener addChangeEventListener(@NonNull ChangeEventListener listener) {
+        boolean wasListening = isListening();
+        super.addChangeEventListener(listener);
+
+        // Only start listening when the first listener is added
+        if (!wasListening) {
```
samtstern: This is obviously a good fix.  It looks pretty similar to the code we have in `FirebaseArray` though, can we add a TODO to investigate if we should push some of this logic up into a base class?

SUPERCILEX: Yeah, we should just have `onCreate` and `onDestroy` callbacks. I'm on it! 😄

database/src/main/java/com/firebase/ui/database/CachingObservableSnapshotArray.java
```diff
@@ -43,11 +43,19 @@ public T getObject(int index) {
...
+        super.onDestroy();
         getSnapshots().clear();
         mObjectCache.clear();
+
+        clearData();
     }
 
+    /** @deprecated use {@link ObservableSnapshotArray#onDestroy()} instead */
+    @Deprecated
+    protected void clearData() {}
```
samtstern: In order to not make this a breaking change this method should still actually clear the data.

database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
```diff
@@ -85,11 +84,26 @@ public String parseSnapshot(DataSnapshot snapshot) {
...
 
     @Override
+    protected void onDestroy() {
+        super.onDestroy();
+        mKeySnapshots.removeChangeEventListener(this);
+
+        for (DatabaseReference ref : mRefs.keySet()) {
+            ref.removeEventListener(mRefs.get(ref));
+        }
+        mRefs.clear();
```
samtstern: In the previous version we called `clearData()` but now we only call `mRefs.clear()`, is this intentional?

database/src/main/java/com/firebase/ui/database/ObservableSnapshotArray.java
```diff
@@ -59,20 +61,27 @@ public ChangeEventListener addChangeEventListener(@NonNull ChangeEventListener l
             listener.onDataChanged();
         }
 
+        if (!wasListening) { onCreate(); }
+
         return listener;
     }
 
+    @CallSuper
```
samtstern: We should add JavaDoc on `onCreate` and `onDestroy` to describe what they do and when a subclass should override them.

database/src/main/java/com/firebase/ui/database/CachingObservableSnapshotArray.java
SUPERCILEX: Good catch!

database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
SUPERCILEX: Yeah because `clearData()` is called in `onDestroy()` now so this is the same behavior.

database/src/main/java/com/firebase/ui/database/ObservableSnapshotArray.java
SUPERCILEX: Yup!

samtstern:
@SUPERCILEX nice!  This change came out really clean, thanks for doing the refactor.

SUPERCILEX:
@samtstern Awesome, you're welcome! 😄

Issue #832, opened 2017-07-30
Email login fail without error after email authenticated user has been replaced by gmail account 


flavio20002:
Good morning every one.
I've noticed a regression with respect of version 1.2.0 of Firebase-ui Auth library for Android.
This is the behaviour you can see in the previous version:

1.  Login with Email authentication provider using a gmail address
2. Logout and login with google provider. This is going to replace your existing email provider
3.  If you logout and try to login with email provider you will get an error, inviting you to login via Google.

This is the bad behaviour I can see in last version:

1.  Login with Email authentication provider using a gmail address
2. Logout and login with google provider. This is going to replace your existing email provider
3.  If you logout and try to login with email provider you will get no error and you will be moved again to main login activity

It is a bug, or a configuration error?

Thanks.

Flavio

SUPERCILEX:
@flavio20002 Thanks for catching that! Yeah, it's a bug and I'm investigating.

I've also noticed a smart lock bug while debugging so yay to finding more bugs. 😝

SUPERCILEX:
@flavio20002 Yup, this was my bad in the #808 refactor. I accidentally [changed the provider to a hardcoded `EmailAuthProvider` instead of the `provider` variable](https://github.com/firebase/FirebaseUI-Android/commit/d995eb7a71afbbbc71ea9ee96aef6a0d80e4ead1#diff-66ba372876674b36898f873b8d2f3526R210).

flavio20002:
Thank you very much for your effort.

Flavio

samtstern:
The fix for this issue was released in version `2.2.0`.

samtstern:
The fix for this issue was released in version `2.2.0`.

PR #836, opened 2017-07-30
Fix email -> Google account linking regression

SUPERCILEX:
@samtstern See https://github.com/firebase/FirebaseUI-Android/issues/832#issuecomment-318920158.


auth/src/main/java/com/firebase/ui/auth/ui/accountlink/WelcomeBackIdpPrompt.java
```diff
@@ -64,7 +65,7 @@ public static Intent createIntent(
             Context context,
             FlowParameters flowParams,
             User existingUser,
-            IdpResponse newUserResponse) {
+            @Nullable IdpResponse newUserResponse) {
```
SUPERCILEX: This was only used to get the login credential of another idp so we had another crash where the `IdpResponse.Builder#build()` would throw an `IllegalStateException` because the email clients didn't have a `token`. So double ouch! 😱😆

PR #839, opened 2017-08-01
Fix sample app and GSI error handling

### LINKED ISSUE #833: idpResponse.toIntent() in sample app
The following line

https://github.com/firebase/FirebaseUI-Android/blob/master/app/src/main/java/com/firebase/uidemo/auth/SignedInActivity.java#L86

only works because the sample app has the the same package stem com.firebase. 
The average Joe trying to use the sample app as a template for getting FirebaseUI working in his own app will have a different package name, and it will fail for him.

samtstern:
Fixes #833

Also makes it clearer when your Google Sign In is not configured correctly, which was failing silently for me.


SUPERCILEX:
@samtstern Rerun the Travis build? If that doesn't help, I would delete all the Travis [cache](https://travis-ci.org/firebase/FirebaseUI-Android/caches).

samtstern:
@SUPERCILEX hmm Travis says this build ran in their new 'trusty environment'.  Interestingly two of the failures are tests that always fail on my machine but never failed in Travis before.  This may be a moment of reckoning.

SUPERCILEX:
@samtstern Haha, ok. 😄 Those tests aren't failing on my machine (but it's Windows 😕😆). Sooooo... maybe update Robolectric? Dunno where to start, sorry I can't be of much help. 😊

samtstern:
I can't possibly imagine how it depends on environment, but I think I know how to try and fix it anyway,

SUPERCILEX:
@samtstern SGTM as long as you know what you're doing! 😆

Issue #812, opened 2017-07-20
Remove libphonenumber validation from client-side


samtstern:
Some valid numbers (esp Indian it seems) are being rejected on the client side due to phone number validation in Android platform being out of date.

The resolution is to remove all validation from the client side and allow the server side to reject badly formatted numbers.

samtstern:
cc @ashwinraghav

ashwinraghav:
The phone number validation is used in two flows:
1) When a user types in their phone and submits.
2) When the hint selector returns the number.

2) is fixed in the ongoing gms release. Once complete, it might be safe to simply remove this validation. A good test to ensure the change does not regress would be to test if the hint selector response results in acceptable behavior on TMobile AND Verizon sim cards.

HammadNasir:
This issue has not been resolved yet and I'm facing similar issue while trying it with by 'Airtel' phone number here in India.

samtstern:
The fix for the first part of this issue was released in version `2.2.0`.  A complete fix will be available in a week or two after the next release of Google Play services.

100rabhkr:
Hello, is this issue resolved ? I am facing similar problems with my number

samtstern:
@100rabhkr this issue should be resolved in almost all cases.  The only remaining case is when the phone number is provided to FirebaseUI by SmartLock.

100rabhkr:
so should I change it to latest versions will it work?

samtstern:
I believe #1023 removes the last of this.

samtstern:
This is fixed in version `3.1.1`.

Ankur008:
but this problem is still there( only happens with Reliance Jio Phone no), If I am sending phone no with country code than it is working. but when I am trying without country code then it is not working. 
using firebase auth lib version 11.8.0

PR #840, opened 2017-08-01
Remove phone number validation on the client side

samtstern:
See #812 

@ashwinraghav is this the intended behavior? The libphonenumber call is only made for SmartLock hints but not for manually entered phone numbers.


samtstern:
cc: @isaisachen

ashwinraghav:
;tldr
It was intentional, but the wrong decision in hindsight.

Here was how we thought of this:
1) Smartlock hints needed it (at the time) since the phone number formats returned were not in the E164 format. We started using telephony's normalization (assuming it would somehow be updated regularly enough since libPhoneNumber that backs it changes every week).
2) We piggybacked on the phone number normalization for manually entered phone numbers to save on service calls for fat-fingered phone numbers.

To state the obvious, both 1 and 2 are yielding false negatives since the lib is not updated as often as we would like. I would be ok getting rid of this validation for both 1 and 2 or just 2.

auth/src/main/java/com/firebase/ui/auth/ui/phone/PhoneNumberUtils.java
```diff
@@ -93,6 +88,14 @@ static String formatPhoneNumber(@NonNull String phoneNumber, @NonNull CountryInf
     static String formatPhoneNumberUsingCurrentCountry(
             @NonNull String phoneNumber, Context context) {
         final CountryInfo currentCountry = PhoneNumberUtils.getCurrentCountryInfo(context);
+
+        // TODO(samstern): Remove this call once the next version of Play servics is released
+        //                 estimated timelime August 10th
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
```
ashwinraghav: Might be worth adding a (known bug) to the commend that hint selector is still susceptible to the false negatives of this method.

ashwinraghav:
LGTM

Issue #842, opened 2017-08-01
No drop shadow on auth picker buttons on low API devices


samtstern:
Here are some screenshots showing API 25 next to API 19.

|  **API 19 - No Shadow** |  
**API 25 - Shadow** |
|---|---|
| <img width="300" src="https://user-images.githubusercontent.com/8466666/28849043-13b602e6-76ca-11e7-8b8e-89adf03d629f.png" />  | <img  width="300" src="https://user-images.githubusercontent.com/8466666/28849044-13b69030-76ca-11e7-84a0-2d515b1544ab.png" /> |

SUPERCILEX:
@samtstern Pretty sure shadows just aren't a thing pre-L. I remember reading a thing about `Toolbar` adding an extra view that faked the shadow, are you suggesting we should do that?

samtstern:
@SUPERCILEX I am hacking on something right now involving the XML drawable.  I think I can make it work without additional views.

SUPERCILEX:
@samtstern Oh sweet! 😄

samtstern:
See #843.  Afaict it has no effect on post-L devices but on pre-L devices adds a shadow-looking thing.  It doesn't have the scatter / blur that a shadow would have but it at least adds contrast.

samtstern:
The fix for this issue was released in version `2.2.0`.

samtstern:
The fix for this issue was released in version `2.2.0`.

PR #843, opened 2017-08-01
Make shadows have buttons on pre-L devices.

samtstern:
See #842

**Before**

**After**

### DIFF
--- a/auth/src/main/res/drawable-v21/fui_idp_button_background_email.xml
+++ b/auth/src/main/res/drawable-v21/fui_idp_button_background_email.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layer-list xmlns:android="http://schemas.android.com/apk/res/android">
+    <item>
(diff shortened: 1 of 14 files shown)


SUPERCILEX:
Oh, just saw https://github.com/firebase/FirebaseUI-Android/issues/842#issuecomment-319514930. So yeah, awesome! 😄

SUPERCILEX:
@samtstern ooooooooooh, actually I think we can add some gradients! This [StackOverflow post](https://stackoverflow.com/a/8722864/4548500) is for rounded corners but I think the same technique could apply. What do you think? 😀

samtstern:
@SUPERCILEX nice, I will try something like that.

samtstern:
@SUPERCILEX played with it for a little bit, none of the gradient types offered in XML drawables are suited for this.  Linear gradients band out from the center, and the radial and sweep gradients just don't offer enough configuration to look good on a big rectangle.

SUPERCILEX:
@samtstern oh well, we tried. 😀

PR #845, opened 2017-08-02
Fix consumer proguard rules

samtstern:
https://issuetracker.google.com/64177053

cc @SUPERCILEX

### DIFF
--- a/auth/build.gradle
+++ b/auth/build.gradle
@@ -19,7 +19,7 @@ android {
             manifestPlaceholders = [enableFbLogging: true]
 
             minifyEnabled false
-            consumerProguardFiles getDefaultProguardFile('proguard-android.txt'), 'auth-proguard.pro'
+            consumerProguardFiles 'auth-proguard.pro'
         }
 
         debug {


bendowski:
LGTM. Can you think of any other build files in some related libraries that could be making the same mistake?

samtstern:
Nope. 

The bug was introduced here:
https://github.com/firebase/FirebaseUI-Android/commit/7e13be7c6df3d0d472a154a08b7f80e052a5b779

The only other major library I manage is EasyPermissions, which does not have the bug:
https://github.com/googlesamples/easypermissions/blob/master/easypermissions/build.gradle

SUPERCILEX:
@bendowski I just checked my libs and only FJD looks a [little sketchy](https://github.com/firebase/firebase-jobdispatcher-android/blob/master/jobdispatcher/build.gradle#L28). They have the same exact setup FirebaseUI used to have and might make the mistake. With `minifyEnabled false` proguard doesn't do anything anyway, right? If so, I'll send in a PR to simply remove that line since they don't have a custom config anyway.

bendowski:
The linked snippet is fine: if you wish to minify your library before it gets packaged, you can do so and you would use the default file for that. It's just the "consumer" files which don't get applied but are packaged in the AAR and applied by the application build later.

PR #846, opened 2017-08-02
Improve phone auth UX

SUPERCILEX:
@samtstern While making sure the vectors still worked for phone auth, I noticed some UX that could be improved:
- The keyboard "Next" buttons should be "Done" buttons
- They should actually do something 😄
- When a user selects a phone number from the picker, like with the email one we just move on instead of forcing the user to go through that extra click


samtstern:
Can you explain this sentence:

> When a user selects a phone number from the picker, like with the email one we just move on instead of forcing the user to go through that extra click

I don't see that behavior change in this PR, but maybe I am missing something.

auth/src/main/java/com/firebase/ui/auth/ui/phone/VerifyPhoneNumberFragment.java
```diff
@@ -159,24 +167,29 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
                             PhoneNumberUtils.getPhoneNumber(formattedPhone);
                     setPhoneNumber(phoneNumberObj);
                     setCountryCode(phoneNumberObj);
+                    onNext();
```
SUPERCILEX: @samtstern Yup, sorry for not making that clearer. In the email sign-in flow, we move on to the next screen once the user selects an email from the picker. This line inside `onActivityResult` does the same thing as in the email sign-in flow where if a user selects an item from the picker we also "click" the Send Code button for them.

PR #847, opened 2017-08-02
Fix bug where onDataChanged isn't called if no data is found at index key

SUPERCILEX:
I noticed this in my app when I incorrectly deleted some stuff in my database: if the key isn't found, `onDataChanged` won't fire for addition events in `FirebaseIndexArray`.

### DIFF
--- a/database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
+++ b/database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
@@ -253,6 +253,11 @@ public class FirebaseIndexArray<T> extends CachingObservableSnapshotArray<T> imp
                 }
             }
 
+            // In theory, we would only want to pop the queue if this listener was just added
+            // i.e. `snapshot.value != null && isKeyAtIndex(...)`. However, if the developer makes a
+            // mistake and `snapshot.value == null`, we will never pop the queue and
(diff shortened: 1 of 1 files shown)


database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
```diff
@@ -255,6 +252,9 @@ public void onDataChange(DataSnapshot snapshot) {
                     Log.w(TAG, "Key not found at ref: " + snapshot.getRef());
                 }
             }
+
+            mKeysWithPendingUpdate.remove(key);
```
samtstern: Previously we only executed these two lines if:

- Snapshot had a non-null value AND
- `isKeyAtIndex() matched`

Now we are executing it regardless of snapshot data.  Can you add a comment explaining what we are doing here and when it matters?

SUPERCILEX: I'm glad you slowed me down on this one because I didn't notice I was introducing a bug where `onDataChanged` would be called twice in quick succession. Everything LGTM now.

PS: I didn't want to make the code explanation too verbose, but basically if this line is reached:
```java
Log.w(TAG, "Key not found at ref: " + snapshot.getRef());
```
then the bug occurs because we won't pop the queue.

PR #849, opened 2017-08-03
Update buildToolsVersion to 26.0.1, supportLibrary and SdkVersion to 26

### LINKED ISSUE #848: Update buildTool, sdk and libraries to 26

annalabellarte:
Closes #848

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ android:
     - tools
 before_script:
   - mv library/google-services.json app/google-services.json
-  - ${ANDROID_HOME}tools/bin/sdkmanager --channel=3 "tools" "platform-tools" "build-tools;25.0.3" "platforms;android-25" "extras;google;m2repository"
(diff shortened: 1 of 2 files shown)


samtstern:
@annalabellarte thanks for sending this!  Will merge once the build passes.

SUPERCILEX:
@samtstern Ahhhhh, I'm too late. I was going to say there's far more we need to do before we can just update the target sdk version, but I guess it'll force me to submit a PR. 😄 I was holding out because we'll need to update all our `findViewById`s, Robolectric, add autofill hints, update the sample with adaptive icons, etc. There's a lot! 😆

samtstern:
All of those are *enabled* by this, but none of them are required right?  The only required API 26 changes I know there are big changes to background work execution, but that should not apply to us.

Here's the big list:
https://developer.android.com/preview/behavior-changes.html

SUPERCILEX:
@samtstern True, though I have seen autofill make mistakes and show incorrect suggestions at times. That's the main thing we have to test for compatibility, but I'll just get the other stuff done as well since it was on my TODO list anyway. 😄

PR #850, opened 2017-08-03
Add full support for Android O and update deps

SUPERCILEX:
@samtstern I've got a string of PRs lined up. This one updates deps, adds explicit autofill support, and does Android O cleanup.


auth/src/main/res/values-v26/styles.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources
+    xmlns:tools="http://schemas.android.com/tools"
+    tools:ignore="ResourceName">
+
+    <style name="FirebaseUI.TextInputEditText.EmailField">
+        <item name="android:autofillHints">emailAddress</item>
```
samtstern: Hmm I wonder how this will play with smartlock.  It could be annoying if they just turned down the smartlock hint and then autofill offers them a new hint.

SUPERCILEX: Oh, thanks for reminding me! I was going to disable password autofill if smart lock is enabled (because getting a save password prompt twice is really annoying).

As for the email, we could also disable autofill anytime smart lock is enabled because the only way the user could have reached the email field is if they didn't take the hint with smart lock enabled.

samtstern: I think that's a good idea, when smartlock is enabled disable other autofills.  That's the path of least annoyance.

samtstern:
Woot!

PR #858, opened 2017-08-08
Oops, remove unnecessary lint suppression

SUPERCILEX:


samtstern:
Thanks for following up on that.

Issue #856, opened 2017-08-08
NullPointerException on saveCredentialsOrFinish (2.0.1)


victoraldir:
Even after 2.0 I am still seeing such error and I don't know exactly why. It happens after creating a new account using email. Here's a copy of crash report (firebase)

```
com.google.android.gms.common.internal.af.a ()
com.google.android.gms.internal.aio.<init> ()
com.google.android.gms.common.api.e$a.a ()
com.firebase.ui.auth.util.signincontainer.SaveSmartLock.saveCredentialsOrFinish (SaveSmartLock.java:221)
com.firebase.ui.auth.ui.BaseHelper.saveCredentialsOrFinish (BaseHelper.java:113)
com.firebase.ui.auth.ui.email.RegisterEmailFragment$3$1.onComplete (RegisterEmailFragment.java:237)
com.google.android.gms.c.o.run ()
android.os.Handler.handleCallback (Handler.java:739)
android.os.Handler.dispatchMessage (Handler.java:95)
android.os.Looper.loop (Looper.java:145)
android.app.ActivityThread.main (ActivityThread.java:7007)
java.lang.reflect.Method.invoke (Method.java)
java.lang.reflect.Method.invoke (Method.java:372)
com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:1404)
com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1199)
```

Manufacturer: Samsung
Model: SM-J320M
Board: Sc9830i
Android API: 22
Android OS: 5.1.1
Brand: Samsung
RAM: 1.33GB
Orientation: Portrait
Proximity to user: —

Dependencies:
compile 'com.firebaseui:firebase-ui-auth:2.0.1'

Smells like a bug, doesn't it?

TheCraftKid:
Hey, could you try reproducing this error on the latest version of FirebaseUI, 2.2.0?

SUPERCILEX:
@victoraldir Thanks for the bug report, I've got all the info I need! 😄 

The right line of code is [here](https://github.com/firebase/FirebaseUI-Android/blob/f34f5ab1e7f0fa2b728cbb313337fb2e912a5899/auth/src/main/java/com/firebase/ui/auth/util/signincontainer/SaveSmartLock.java#L221).

[...]

PR #859, opened 2017-08-08
Merge more improvements from anonymous-auth PR

SUPERCILEX:
Also addresses #856

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailFragment.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailFragment.java
@@ -25,6 +25,7 @@ import com.firebase.ui.auth.ui.FragmentBase;
 import com.firebase.ui.auth.ui.HelperActivityBase;
 import com.firebase.ui.auth.ui.ImeHelper;
 import com.firebase.ui.auth.ui.TaskFailureLogger;
+import com.firebase.ui.auth.util.accountlink.ProfileMerger;
 import com.firebase.ui.auth.ui.accountlink.WelcomeBackIdpPrompt;
 import com.firebase.ui.auth.ui.accountlink.WelcomeBackPasswordPrompt;
 import com.firebase.ui.auth.ui.email.fieldvalidators.EmailFieldValidator;
@@ -41,8 +42,6 @@ import com.google.firebase.auth.FirebaseAuth;
 import com.google.firebase.auth.FirebaseAuthInvalidCredentialsException;
 import com.google.firebase.auth.FirebaseAuthUserCollisionException;
 import com.google.firebase.auth.FirebaseAuthWeakPasswordException;
-import com.google.firebase.auth.FirebaseUser;
-import com.google.firebase.auth.UserProfileChangeRequest;
 
 /**
  * Fragment to display an email/name/password sign up form for new users.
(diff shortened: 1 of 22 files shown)


auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailFragment.java
```diff
@@ -228,40 +227,25 @@ private void validateAndRegisterUser() {
...
+                        .setPhotoUri(mUser.getPhotoUri())
+                        .build())
+                .build();
+
         getAuthHelper().getFirebaseAuth()
                 .createUserWithEmailAndPassword(email, password)
+                .continueWithTask(new ProfileMerger(response))
                 .addOnFailureListener(new TaskFailureLogger(TAG, "Error creating user"))
-                .addOnSuccessListener(new OnSuccessListener<AuthResult>() {
+                .addOnSuccessListener(getActivity(), new OnSuccessListener<AuthResult>() {
```
SUPERCILEX: Here's the magic for #856. We can do this easily thanks to `ProfileMerger`.

samtstern:
@SUPERCILEX thanks for being thoughtful and picking apart the useful improvements from the anon auth PR.  I'll take a look at this one soon.

SUPERCILEX:
@samtstern SGTM!

auth/src/main/java/com/firebase/ui/auth/util/accountlink/ProfileMerger.java
```diff
@@ -0,0 +1,66 @@
...
+
+        firebaseUser.updateProfile(
+                new UserProfileChangeRequest.Builder()
+                        .setDisplayName(name)
+                        .setPhotoUri(photoUri)
+                        .build())
+                .addOnCompleteListener(this)
+                .addOnFailureListener(new TaskFailureLogger(TAG, "Error updating profile"));
+
+        return mProfileUpdateCompleteTask.getTask();
```
samtstern: This pattern is strange.  How about removing the `TaskCompletionSource` and doing:

```java
Task<AuthResult> updateTask = firebaseUser.updateProfile(....);
updateTask.addOnFailureListener(new TaskFailureLogger(...));

return updateTask;
```

Then this class can also not implement `OnCompleteListener`.

auth/src/test/java/com/firebase/ui/auth/testhelpers/AuthHelperShadow.java
```diff
@@ -17,37 +17,30 @@
 @Implements(AuthHelper.class)
 public class AuthHelperShadow {
 
-    public static final FirebaseAuth sFirebaseAuth;
-    public static final CredentialsApi sCredentialsApi;
-    public static final SaveSmartLock sSaveSmartLock;
-    public static final PhoneAuthProvider sPhoneAuthProvider;
-
+    private static FirebaseAuth sFirebaseAuth;
```
samtstern: :+1: can't imagine why I was too lazy to make this change before haha, thanks!

auth/src/main/java/com/firebase/ui/auth/util/accountlink/ProfileMerger.java
SUPERCILEX: @samtstern I had two reasons to do this:
1. I want the `ProfileMerger` to always return a successful task regardless of whether or not the profile update succeeded and AFAIK, the only way to do that is with a custom `TaskCompletionSource` because `then` only moves forward if the task before it was successful.
1. `updateProfile(...)` returns a `Task<Void>`, not an `AuthResult` because we aren't signing-in to anything.

samtstern: Ah I see.  

How about `updateTask.continueWithTask(.... something that returns Tasks.forResult(mAuthResult) ....)`

SUPERCILEX: **I'VE BEEN ENLIGHTENED!!!** 😄😂 I thought for some reason that the `then(...)` calls would only propagate if the tasks before them succeeded, but I've dug into the docs and I now understand that `getResult()` is used to blow up the continuation and propagate the error. Anyway, thanks for feedback! 😄

samtstern:
@SUPERCILEX thanks!

Issue #857, opened 2017-08-08
 Can not perform this action after onSaveInstanceState


thelong1EU:
* Android device: Nexus 9
  * Android OS version: 25
  * Firebase/Play Services SDK version: 11.0.4
  * FirebaseUI version: 2.2.0

#### Steps to reproduce:

  1. go to home screen when the verification dialog is on screen

  * What happened?  This could be a description, `logcat` output, etc.
 ```
 Fatal Exception: java.lang.IllegalStateException: Can not perform this action after onSaveInstanceState
       at android.support.v4.app.FragmentManagerImpl.checkStateLoss(FragmentManager.java:1884)
       at android.support.v4.app.FragmentManagerImpl.enqueueAction(FragmentManager.java:1902)
       at android.support.v4.app.BackStackRecord.commitInternal(BackStackRecord.java:650)
       at android.support.v4.app.BackStackRecord.commit(BackStackRecord.java:609)
       at android.support.v4.app.DialogFragment.dismissInternal(DialogFragment.java:205)
       at android.support.v4.app.DialogFragment.dismiss(DialogFragment.java:171)
       at com.firebase.ui.auth.ui.phone.PhoneVerificationActivity.dismissLoadingDialog(PhoneVerificationActivity.java:390)
       at com.firebase.ui.auth.ui.phone.PhoneVerificationActivity.access$000(PhoneVerificationActivity.java:56)
       at com.firebase.ui.auth.ui.phone.PhoneVerificationActivity$1.run(PhoneVerificationActivity.java:189)
       at android.os.Handler.handleCallback(Handler.java:751)
       at android.os.Handler.dispatchMessage(Handler.java:95)
       at android.os.Looper.loop(Looper.java:154)
       at android.app.ActivityThread.main(ActivityThread.java:6119)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:776)
```

samtstern:
@SUPERCILEX this is probably related to #817, where the dialog was changed to a `DialogFragment`.  I've seen this before, it has to do with the fact that dialog fragments don't have `commitAllowingStateLoss()` and also the fact that Fragments are truly a creation of demons meant to torment Android developers for all time.

thelong1EU:
:rofl:

[...]

PR #861, opened 2017-08-08
Fix dismiss dialog bugs

SUPERCILEX:
#857

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/ui/phone/PhoneVerificationActivity.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/phone/PhoneVerificationActivity.java
@@ -192,36 +192,33 @@ public class PhoneVerificationActivity extends AppCompatBase {
     }
 
     private void onVerificationFailed(@NonNull FirebaseException ex) {
-        VerifyPhoneNumberFragment verifyPhoneNumberFragment = (VerifyPhoneNumberFragment)
-                getSupportFragmentManager().findFragmentByTag(VerifyPhoneNumberFragment.TAG);
+        dismissLoadingDialog();
 
-        if (verifyPhoneNumberFragment == null) {
-            return;
-        }
         if (ex instanceof FirebaseAuthException) {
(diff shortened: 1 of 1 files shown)


auth/src/main/java/com/firebase/ui/auth/ui/phone/PhoneVerificationActivity.java
```diff
@@ -386,7 +383,7 @@ private void showLoadingDialog(String message) {
 
     private void dismissLoadingDialog() {
         if (mProgressDialog != null) {
-            mProgressDialog.dismiss();
+            mProgressDialog.dismissAllowingStateLoss();
```
samtstern: Whaaaattt?  I never knew this existed!

SUPERCILEX: @samtstern Lol, it was added [way back in `v22.0.0`](https://developer.android.com/reference/android/support/v4/app/DialogFragment.html#dismissAllowingStateLoss()). 😆

samtstern: This would have prevented literally thousands of crashes in Santa Tracker :-( :santa: :mrs_claus:

SUPERCILEX: You work on Santa Tracker too??? Dang, you're all over the place! 😆

At least you know now 😄

samtstern: Santa Tracker was started as a holiday side project by someone on Maps Developer Relations, so it's stayed in the DevRel family.  It's a good way for us to remember what it's like to make a web/android app in the real world, and at this point it's a Google holiday tradition.

SUPERCILEX: Nice, thanks for the history—sounds like fun! 😄

PR #862, opened 2017-08-09
Update README.md to include app_name and color requirements

TheCraftKid:
This adds some information that has been discussed in #696, #815 and others.

### DIFF
--- a/auth/README.md
+++ b/auth/README.md
@@ -204,7 +204,7 @@ startActivityForResult(
     RC_SIGN_IN);
 ```
 
-If a terms of service URLb and privacy policy URL are required:
+If a terms of service URL and privacy policy URL are required:
 
 ```java
 startActivityForResult(
--- a/auth/README.md
+++ b/auth/README.md
@@ -418,16 +418,6 @@ Second, ensure the three standard AppCompat color resources are defined with you
     <item name="colorAccent">@color/colorAccent</item>
 </style>
 
-```
-You don't have to call `setTheme()` on the sign in intent builder as FirebaseUI
-will automatically style the activities. This is sufficient:
-
-```java
-startActivityForResult(
-    AuthUI.getInstance(this).createSignInIntentBuilder()
-        .build());
-```
-
 If you would like more control over FirebaseUI's styling, you can define your own custom style
 to override certain or all styling attributes. For example, a green sign-in theme:
 
@@ -478,6 +468,7 @@ and simply redefine a string to change it:
     <string name="fui_progress_dialog_signing_up">Creating your shiny new account...</string>
 </resources>
 ```
+
 **Note:** String resource names aren't considered part of the public API and might
 therefore change and break your app between library updates. We recommend looking
 at a diff of the `strings.xml` file before updating FirebaseUI.
--- a/auth/README.md
+++ b/auth/README.md
@@ -204,7 +204,7 @@ startActivityForResult(
     RC_SIGN_IN);
 ```
 
-If a terms of service URL, privacy policy URL, and a custom theme are required:
+If a terms of service URLb and privacy policy URL are required:
 
 ```java
 startActivityForResult(
@@ -213,7 +213,6 @@ startActivityForResult(
         .setAvailableProviders(...)
         .setTosUrl("https://superapp.example.com/terms-of-service.html")
         .setPrivacyPolicyUrl("https://superapp.example.com/privacy-policy.html")
-        .setTheme(R.style.SuperAppTheme)
         .build(),
     RC_SIGN_IN);
 ```
@@ -400,25 +399,28 @@ represented in the following diagram:
 To use FirebaseUI Auth's sign-in flows, you must provide an `app_name` string and use the
 AppCompat color attributes in your app.
 
-For example, something like this goes in your `strings.xml`:
+First, ensure an `app_name` resource is defined your `strings.xml` file like so:
+
 ```xml
 <resources>
-    <!-- ... -->
     <string name="app_name">My App</string>
+    <!-- ... -->
 </resources>
 ```
 
-This goes in your `styles.xml`:
+Second, ensure the three standard AppCompat color resources are defined with your own values:
+
 ```xml
-<style name="AppTheme" parent="FirebaseUI">
+<style name="AppTheme" parent="Theme.AppCompat.Light.DarkActionBar">
     <!-- Required for sign-in flow styling -->
-    <item name="colorPrimary">@color/material_green_500</item>
-    <item name="colorPrimaryDark">@color/material_green_700</item>
-    <item name="colorAccent">@color/material_purple_a700</item>
+    <item name="colorPrimary">@color/colorPrimary</item>
+    <item name="colorPrimaryDark">@color/colorPrimaryDark</item>
+    <item name="colorAccent">@color/colorAccent</item>
 </style>
+
 ```
-You don't have to call `setStyle()` on the sign in intent builder as FirebaseUI
-will automatically style the activities. For example:
+You don't have to call `setTheme()` on the sign in intent builder as FirebaseUI
+will automatically style the activities. This is sufficient:
 
 ```java
 startActivityForResult(
@@ -426,10 +428,8 @@ startActivityForResult(
         .build());
 ```
 
-Note: If your app is already using an AppCompat theme, you don't have to create 
-a new theme, but you can still optionally style other parts of the sign-in flow. 
-Standard material design colorand typography properties will take effect as expected. 
-For example, to define a green theme:
+If you would like more control over FirebaseUI's styling, you can define your own custom style
+to override certain or all styling attributes. For example, a green sign-in theme:
 
 ```xml
 <style name="GreenTheme" parent="FirebaseUI">
@@ -446,6 +446,7 @@ For example, to define a green theme:
 ```
 
 With associated colors:
+
 ```xml
 <color name="material_green_50">#E8F5E9</color>
 <color name="material_green_500">#4CAF50</color>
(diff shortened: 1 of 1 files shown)


auth/README.md
```diff
@@ -397,16 +397,47 @@ represented in the following diagram:
...
-the flow, a new theme can be declared. Standard material design color
-and typography properties will take effect as expected. For example, to define
-a green theme:
+To use FirebaseUI Auth's sign-in flows, you must provide an `app_name` string and use the
+AppCompat color attributes in your app.
+
+For example, something like this goes in your `strings.xml`:
+```xml
+<resources>
+    <!-- ... -->
```
SUPERCILEX: nit: can we put the `...` after `app_name` since the latter should in theory be first?

auth/README.md
```diff
@@ -397,16 +397,47 @@ represented in the following diagram:
...
+    <string name="app_name">My App</string>
+</resources>
+```
+
+This goes in your `styles.xml`:
+```xml
+<style name="AppTheme" parent="FirebaseUI">
```
SUPERCILEX: The parent should be something like `Theme.AppCompat.DayNight.DarkActionBar` or `Theme.AppCompat.Light.DarkActionBar`, not `FirebaseUI`.

auth/README.md
```diff
@@ -397,16 +397,47 @@ represented in the following diagram:
...
+    <!-- Required for sign-in flow styling -->
+    <item name="colorPrimary">@color/material_green_500</item>
+    <item name="colorPrimaryDark">@color/material_green_700</item>
+    <item name="colorAccent">@color/material_purple_a700</item>
```
SUPERCILEX: The standard code Android Studio generates has the app style reference each color name like so instead of referencing the color directly:
```xml
<item name="colorPrimary">@color/colorPrimary</item>
<item name="colorPrimaryDark">@color/colorPrimaryDark</item>
<item name="colorAccent">@color/colorAccent</item>
```

auth/README.md
```diff
@@ -397,16 +397,47 @@ represented in the following diagram:
...
 
-To provide customization of the visual style of the activities that implement
-the flow, a new theme can be declared. Standard material design color
-and typography properties will take effect as expected. For example, to define
-a green theme:
+To use FirebaseUI Auth's sign-in flows, you must provide an `app_name` string and use the
+AppCompat color attributes in your app.
+
+For example, something like this goes in your `strings.xml`:
+```xml
```
SUPERCILEX: nit: to follow the existing style in this `README`, we should have a space between all the ` ``` ` and their text.

auth/README.md
```diff
@@ -397,16 +397,47 @@ represented in the following diagram:
...
 ## UI customization
 
-To provide customization of the visual style of the activities that implement
-the flow, a new theme can be declared. Standard material design color
-and typography properties will take effect as expected. For example, to define
-a green theme:
+To use FirebaseUI Auth's sign-in flows, you must provide an `app_name` string and use the
+AppCompat color attributes in your app.
+
+For example, something like this goes in your `strings.xml`:
```
SUPERCILEX: Could we improve the clarity of this statement to something like this?
```
First, ensure an `app_name` resource is defined your `strings.xml` file like so:
```

auth/README.md
```diff
@@ -397,16 +397,47 @@ represented in the following diagram:
...
+```xml
+<resources>
+    <!-- ... -->
+    <string name="app_name">My App</string>
+</resources>
+```
+
+This goes in your `styles.xml`:
+```xml
```
SUPERCILEX: Could we improve the clarity of this statement to something like this?
```
Second, ensure the three standard AppCompat color resources are defined with your own values:
```

nit: add line break

auth/README.md
```diff
@@ -435,20 +464,20 @@ startActivityForResult(
...
-one.
+Your application theme could also simply be used, rather than defining a new one.
 
-If you wish to change the string messages, the existing strings can be
-easily overridden by name in your application. See
-[the built-in strings.xml](src/main/res/values/strings.xml) and simply
-redefine a string to change it, for example:
+If you wish to change the string messages, the existing strings can be overridden
+by name in your application. See the module's [strings.xml](src/main/res/values/strings.xml) file
+and simply redefine a string to change it, for example:
```
SUPERCILEX: We could simplify this to `and simply redefine a string to change it:`

auth/README.md
```diff
@@ -435,20 +464,20 @@ startActivityForResult(
...
 
 ```xml
 <resources>
     <!-- was "Signing up..." -->
-    <string name="progress_dialog_signing_up">Creating your shiny new account...</string>
+    <string name="fui_progress_dialog_signing_up">Creating your shiny new account...</string>
 </resources>
 ```
+Note: String resource names are internal to the library and there are currently no
+guarantees that they won't change between updates.
```
SUPERCILEX: This should be a little scarier: 👻😆
```
**Note:** string resource names aren't considered part of the public API and might
therefore change and break your app between library updates. We recommend looking
at a diff of the `strings.xml` file before updating FirebaseUI.
```

auth/README.md
```diff
@@ -435,20 +464,20 @@ startActivityForResult(
...
-redefine a string to change it, for example:
+If you wish to change the string messages, the existing strings can be overridden
+by name in your application. See the module's [strings.xml](src/main/res/values/strings.xml) file
+and simply redefine a string to change it, for example:
 
 ```xml
 <resources>
     <!-- was "Signing up..." -->
-    <string name="progress_dialog_signing_up">Creating your shiny new account...</string>
+    <string name="fui_progress_dialog_signing_up">Creating your shiny new account...</string>
```
SUPERCILEX: 👍

auth/README.md
```diff
@@ -397,16 +397,47 @@ represented in the following diagram:
...
+```java
+startActivityForResult(
+    AuthUI.getInstance(this).createSignInIntentBuilder()
+        .build());
+```
+
+Note: If your app is already using an AppCompat theme, you don't have to create 
+a new theme, but you can still optionally style other parts of the sign-in flow. 
+Standard material design colorand typography properties will take effect as expected. 
+For example, to define a green theme:
```
SUPERCILEX: I think this paragraph up to `You don't have to call...` should be replaced with something simpler since we don't talk about `setTheme` anywhere else:
```
If you would like more control over FirebaseUI's styling, you can define your own custom style to
override certain or all styling attributes. For example, a green sign-in theme:
```

PS: we don't talk about `setTheme` anywhere except [here](https://github.com/firebase/FirebaseUI-Android/pull/862/files/8b8707c244580002d809342f715ac89d129e05e4#diff-46b10a690795758a8db6b0f86c9afddcR216) which we should remove.

auth/README.md
TheCraftKid: Whoops, I knew I forgot to change something while copy-pasting.

auth/README.md
TheCraftKid: Yeah, phrasing this was a problem for me. That is way better.

auth/README.md
TheCraftKid: In the Android Studio template, `app_name` is first... Will do.

auth/README.md
TheCraftKid: And I thought mine was too abrasive. :laugh:

samtstern:
I just want to say I love seeing a PR where I am neither the author nor the reviewer!  I'll merge this once you two have worked through the comments.

TheCraftKid:
Hey, do we still require the [Fabric Maven repo](https://github.com/firebase/FirebaseUI-Android/pull/862/commits/8b8707c244580002d809342f715ac89d129e05e4#diff-46b10a690795758a8db6b0f86c9afddcR121) in the `build.gradle` for FirebaseUI Auth?

auth/README.md
TheCraftKid: Instead of removing the earlier mention completely, why don't we just provide a link in a to the lower section?

samtstern:
@TheCraftKid we only require it if you're using Twitter auth, since you need it to resolve the twitter dependency.

auth/README.md
SUPERCILEX: @TheCraftKid haha, yeah. We've broken the strings without remorse several times now, so we want to make sure devs understand that we aren't kidding. 😄

auth/README.md
SUPERCILEX: @TheCraftKid Hmmm, the problem I see with this section is that it feels kinda like an "upgrade guide" of sorts. I'm trying to think of it from a newcomer's perspective, and for all they care, we've always auto-styled FirebaseUI so why would they have to call `setTheme()`? If you believe a newcomer would find this information useful, then feel free to keep it around.

auth/README.md
```diff
@@ -435,20 +466,21 @@ startActivityForResult(
...
+and simply redefine a string to change it:
 
 ```xml
 <resources>
     <!-- was "Signing up..." -->
-    <string name="progress_dialog_signing_up">Creating your shiny new account...</string>
+    <string name="fui_progress_dialog_signing_up">Creating your shiny new account...</string>
 </resources>
 ```
+**Note:** String resource names aren't considered part of the public API and might
```
SUPERCILEX: nit: add space between text and ` ``` `.

auth/README.md
```diff
@@ -397,16 +396,48 @@ represented in the following diagram:
...
+
+```
+You don't have to call `setTheme()` on the sign in intent builder as FirebaseUI
+will automatically style the activities. This is sufficient:
+
+```java
+startActivityForResult(
+    AuthUI.getInstance(this).createSignInIntentBuilder()
+        .build());
+```
```
SUPERCILEX: Are you sure you want to keep this section?

Issue #870, opened 2017-08-17
FirebaseIndexArray :java.lang.IndexOutOfBoundsException when key data is changed


nivarp:
### Step 2: Describe your environment

  * Android device: All
  * Android OS version: All
  * Firebase/Play Services SDK version: 11.0.4
  * FirebaseUI version: 2.2.0
  
### Step 3: Describe the problem:
Using FirebaseIndexRecyclerAdapter in the code with key ref and data ref nodes. This issue typically comes when the key ref is changed.

Some points which can be of help:
a. KeyRef is changed using a server side operation and performs set instead of update on the node. This means there can be changes in size and will include addition, deletion and movement of key nodes.
b. DataRef keeps on getting updated irrespective of the above operation and some times we do observe duplicate entries as well, not sure if they are related though.

Let me know if more information is needed.

Crash Log:
java.lang.IndexOutOfBoundsException: Index: 11, Size: 9
                                         at java.util.ArrayList.add(ArrayList.java:457)
                                         at com.firebase.ui.database.FirebaseIndexArray.onKeyMoved(FirebaseIndexArray.java:179)
                                         at com.firebase.ui.database.FirebaseIndexArray.onChildChanged(FirebaseIndexArray.java:113)
                                         at com.firebase.ui.database.ObservableSnapshotArray.notifyChangeEventListeners(ObservableSnapshotArray.java:122)
                                         at com.firebase.ui.database.FirebaseArray.onChildMoved(FirebaseArray.java:116)
                                         at com.google.android.gms.internal.px.zza(Unknown Source)
                                         at com.google.android.gms.internal.vj.zzHX(Unknown Source)
                                         at com.google.android.gms.internal.vp.run(Unknown Source)
                                         at android.os.Handler.handleCallback(Handler.java:751)
                                         at android.os.Handler.dispatchMessage(Handler.java:95)
                                         at android.os.Looper.loop(Looper.java:154)
                                         at android.app.ActivityThread.main(ActivityThread.java:6121)
                                         at java.lang.reflect.Method.invoke(Native Method)
                                         at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:889)
                                         at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:779)

SUPERCILEX:
[...]

PR #871, opened 2017-08-18
Fix tons of bugs relating to null `snapshot.value`s and improve perf for good citizens

SUPERCILEX:
https://github.com/firebase/FirebaseUI-Android/issues/870#issuecomment-323249333

### DIFF
--- a/app/src/main/java/com/firebase/uidemo/database/ChatIndexActivity.java
+++ b/app/src/main/java/com/firebase/uidemo/database/ChatIndexActivity.java
@@ -1,6 +1,5 @@
 package com.firebase.uidemo.database;
 
-import android.os.Bundle;
 import android.view.View;
 
 import com.firebase.ui.database.FirebaseIndexRecyclerAdapter;
@@ -12,12 +11,6 @@ import com.google.firebase.database.FirebaseDatabase;
 
 public class ChatIndexActivity extends ChatActivity {
     private DatabaseReference mChatIndicesRef;
-
-    @Override
-    protected void onCreate(Bundle savedInstanceState) {
-        super.onCreate(savedInstanceState);
-        mChatIndicesRef = FirebaseDatabase.getInstance().getReference().child("chatIndices");
-    }
 
     @Override
     public void onClick(View v) {
@@ -34,6 +27,11 @@ public class ChatIndexActivity extends ChatActivity {
 
     @Override
     protected FirebaseRecyclerAdapter<Chat, ChatHolder> getAdapter() {
+        mChatIndicesRef = FirebaseDatabase.getInstance()
+                .getReference()
+                .child("chatIndices")
+                .child(FirebaseAuth.getInstance().getCurrentUser().getUid());
+
         return new FirebaseIndexRecyclerAdapter<Chat, ChatHolder>(
                 Chat.class,
                 R.layout.message,
--- a/database/src/main/java/com/firebase/ui/database/FirebaseArray.java
+++ b/database/src/main/java/com/firebase/ui/database/FirebaseArray.java
@@ -14,6 +14,8 @@
 
 package com.firebase.ui.database;
 
+import android.support.annotation.Nullable;
+
 import com.google.firebase.database.ChildEventListener;
 import com.google.firebase.database.DataSnapshot;
 import com.google.firebase.database.DatabaseError;
@@ -27,8 +29,19 @@ import java.util.List;
  * This class implements a collection on top of a Firebase location.
  */
 public class FirebaseArray<T> extends CachingObservableSnapshotArray<T> implements ChildEventListener, ValueEventListener {
+    /**
+     * Hack for {@link FirebaseIndexArray}
+     *
+     * @see FirebaseIndexArray#mPendingMoveIndex
+     */
+    interface PreChangeEventListener {
+        void onPreMove(DataSnapshot data);
+    }
+
     private Query mQuery;
     private List<DataSnapshot> mSnapshots = new ArrayList<>();
+
+    @Nullable private PreChangeEventListener mPreChangeEventListener;
 
     /**
      * Create a new FirebaseArray that parses snapshots as members of a given class.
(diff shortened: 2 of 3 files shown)


database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
```diff
@@ -161,23 +185,33 @@ private boolean isKeyAtIndex(String key, int index) {
...
 
-    protected void onKeyAdded(DataSnapshot data) {
+    /**
+     * @deprecated
+     */
+    @Deprecated
+    protected void onKeyAdded(DataSnapshot data) {}
+
+    protected void onKeyAdded(DataSnapshot data, int index) {
+        onKeyAdded(data);
```
SUPERCILEX: Unfortunately, this isn't backwards compatible if people call super on the old method. 😢 I totally understand if you don't want to proceed with this change though we would be missing out on some significant perf gains: O(n) with the old method, O(1) with the new method on simple additions.

database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
```diff
@@ -140,18 +156,26 @@ public void onCancelled(DatabaseError error) {
...
-                index++;
+    private int returnOrFindIndexForKey(int index, String key) {
+        int realIndex;
+        if (isKeyAtIndex(key, index)) {
+            realIndex = index;
+        } else {
+            int dataCount = size();
+            int dataIndex = 0;
+            for (int keyIndex = 0; dataIndex < dataCount; keyIndex++) {
+                if (keyIndex == mKeySnapshots.size()) { break; }
```
SUPERCILEX: I added this line to the old method to handle being called by a deletion at the last index.

app/src/main/java/com/firebase/uidemo/database/ChatIndexActivity.java
```diff
@@ -34,6 +27,11 @@ public void onClick(View v) {
 
     @Override
     protected FirebaseRecyclerAdapter<Chat, ChatHolder> getAdapter() {
+        mChatIndicesRef = FirebaseDatabase.getInstance()
+                .getReference()
+                .child("chatIndices")
+                .child(FirebaseAuth.getInstance().getCurrentUser().getUid());
```
SUPERCILEX: Added a `uid` to the indices to make them more realistic.

database/src/main/java/com/firebase/ui/database/FirebaseArray.java
```diff
@@ -58,6 +71,10 @@ private void init(Query query) {
         mQuery = query;
     }
 
+    void setPreChangeEventListener(PreChangeEventListener listener) {
```
SUPERCILEX: Not sure what to do about this api... seems like other devs might want to use it, but it's kinda gross.

samtstern:
@SUPERCILEX sorry I never responded here.  I think we're both on the same page about this one.  It's definitely an improvement in some ways, but `PreChangeListener` seems like a hacky API change and losing backwards compatibility isn't really worth it here.

If there are any small improvements from this you'd like to keep we can move forward with those, otherwise I'd just say we document the limitations of the indexed adapters.

SUPERCILEX:
Oops, didn't mean to close this PR.

@samtstern yep, I'll keep as many changes as I can as long as they're backwards compatible.

SUPERCILEX:
@samtstern Could you please restart the build? My Travis is [passing](https://travis-ci.org/SUPERCILEX/FirebaseUI-Android/builds/269859182). Thanks!

database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
```diff
@@ -140,18 +140,26 @@ public void onCancelled(DatabaseError error) {
...
-            } else if (mDataSnapshots.get(index).getKey().equals(superKey)) {
-                index++;
+    private int returnOrFindIndexForKey(int index, String key) {
+        int realIndex;
+        if (isKeyAtIndex(key, index)) {
+            realIndex = index;
+        } else {
+            int dataCount = size();
+            int dataIndex = 0;
+            for (int keyIndex = 0; dataIndex < dataCount; keyIndex++) {
```
samtstern: This is a very strange `for` loop.  It initializes `keyIndex` but then waits for a condition on `dataIndex`, all while incrementing `keyIndex` on each pass.

Took me a very long time to decipher.  I'd suggest converting it to something like this:

```java
int dataIndex = 0;
int keyIndex = 0;

while (dataIndex < dataCount) {
   // ... stuff

   keyIndex++;
}
```

Since the first line is `if (keyIndex == mKeySnapshots.size()) { break; }` you may even want to make the main condition `while(dataIndex < dataCount && keyIndex < mKeySnapshots.size())`

Also add comments throughout the method please!

SUPERCILEX: @samtstern Same! Another contributor who worked on my original PR wrote that for loop and it always takes me 5+ mins to get my head around it again. I thought that was just me! 😆 Anyway, your suggestions sound like great improvements!

database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
```diff
@@ -140,18 +140,32 @@ public void onCancelled(DatabaseError error) {
...
-        int dataCount = size();
-        int index = 0;
-        for (int keyIndex = 0; index < dataCount; keyIndex++) {
-            String superKey = mKeySnapshots.getObject(keyIndex);
-            if (key.equals(superKey)) {
-                break;
-            } else if (mDataSnapshots.get(index).getKey().equals(superKey)) {
-                index++;
+    private int returnOrFindIndexForKey(int index, String key) {
+        int realIndex;
```
samtstern: Should we initialize realIndex to 0 for clarity?  If  `key.equals(superKey)` is `true` on the first loop then we will break and return `realIndex` without ever having initialized it.

SUPERCILEX: Hmmm, we could but intellij complains because we do [initialize it to `dataIndex`](https://github.com/firebase/FirebaseUI-Android/pull/871/files/9dbd38b76cf88a795db270bc5f44b0dc11aab1b2#diff-848b2a3fc375b8ab8b5f22f22157614aR166) while out of the loop. Do you still want to proceed?

samtstern: Oh oops I missed that, thanks!

PR #883, opened 2017-08-31
Update to Gradle Plugin 3.0-rc1

samtstern:
This is still a long way from working.


database/build.gradle
```diff
@@ -23,17 +23,16 @@ android {
...
 
-    compile 'android.arch.lifecycle:runtime:1.0.0-alpha8'
-    compile 'android.arch.lifecycle:extensions:1.0.0-alpha8'
-    annotationProcessor 'android.arch.lifecycle:compiler:1.0.0-alpha8'
+    api 'android.arch.lifecycle:runtime:1.0.0-alpha9'
+    api 'android.arch.lifecycle:extensions:1.0.0-alpha9'
+    annotationProcessor 'android.arch.lifecycle:compiler:1.0.0-alpha9'
 
-    compile "com.google.firebase:firebase-database:$firebaseVersion"
+    api "com.google.firebase:firebase-database:$firebaseVersion"
```
TheCraftKid: Do we really want to do this? This would require library users to match the version that we use. The auth module doesn't do this for its respective Firebase dependencies.

samtstern: I don't consider myself an expert on the new `3.0` stuff, but as far as I know you have to use `api` if you expose your dependencies in your public API.  Because we have classes that accept a `Query` or a `DatabaseReference`, this is our only option.

With auth, we completely hide all of the Firebase classes from our public API.  So we can get away with `implementation`.

TheCraftKid: Darn. So much for non-transient dependencies. I wish this wasn't the case as we'd still have to upload the update the library for updates to the Firebase Database SDK to avoid a version mismatch. Or is there a way to still avoid this scenario?

samtstern: There may be, but I am not aware.

app/build.gradle
```diff
@@ -33,38 +36,33 @@ android {
...
+    implementation project(path: ':firestore')
+    implementation project(path: ':database')
+    implementation project(path: ':storage')
 
-    compile "android.arch.lifecycle:runtime:$architectureVersion"
-    compile "android.arch.lifecycle:extensions:$architectureVersion"
+    implementation 'com.facebook.android:facebook-login:4.27.0'
+    // Needed to override Facebook
+    implementation "com.android.support:cardview-v7:$supportLibraryVersion"
+    implementation "com.android.support:customtabs:$supportLibraryVersion"
```
SUPERCILEX: I couldn't figure how to override these on our side so devs will have to do this themselves.

auth/src/main/AndroidManifest.xml
```diff
@@ -11,15 +11,10 @@
         <meta-data
             android:name="com.google.android.gms.version"
             android:value="@integer/google_play_services_version" />
-
         <meta-data
             android:name="com.facebook.sdk.ApplicationId"
             android:value="@string/facebook_application_id" />
 
-        <meta-data
-            android:name="com.facebook.sdk.AutoLogAppEventsEnabled"
```
SUPERCILEX: We don't need this because the `facebook-login` module no longer includes it.

storage/build.gradle
```diff
@@ -3,28 +3,20 @@ apply from: '../library/quality/quality.gradle'
...
 }
 
 dependencies {
-    compile "com.android.support:appcompat-v7:$supportLibraryVersion"
-    compile "com.android.support:palette-v7:$supportLibraryVersion"
+    implementation "com.android.support:appcompat-v7:$supportLibraryVersion"
+    implementation "com.android.support:palette-v7:$supportLibraryVersion"
 
-    provided 'com.github.bumptech.glide:glide:4.1.1'
+    provided 'com.github.bumptech.glide:glide:4.2.0'
```
SUPERCILEX: I forgot, this should be `compileOnly`. (Why again, isn't it `api`?)

common/build.gradle
```diff
@@ -3,31 +3,16 @@ apply from: '../library/quality/quality.gradle'
...
-    // Common dependencies across database modules. Each module still has to provide an
-    // Architecture Components annotation processor and optionally, test dependencies.
-
-    compile "com.android.support:recyclerview-v7:$supportLibraryVersion"
     // Needed to override play services
-    compile "com.android.support:support-v4:$supportLibraryVersion"
-
-    compile "android.arch.lifecycle:runtime:$architectureVersion"
-    compile "android.arch.lifecycle:extensions:$architectureVersion"
+    api "com.android.support:support-v4:$supportLibraryVersion"
```
SUPERCILEX: Thinking back on this, I think we should get rid of the dependency block in the common module and add it to the DB and Firestore modules as an `implementation`. @samtstern what do you think?

samtstern: Done, although it has to be `api` since we expose `RecyclerView.Adapter`

samtstern:
Fixed some things in the POM generation.

Here's the old POM for `firebase-ui-database`:

```xml
  <dependencies>
    <dependency>
      <groupId>com.firebaseui</groupId>
      <artifactId>firebase-ui-common</artifactId>
      <version>3.1.0</version>
    </dependency>
    <dependency>
      <groupId>com.google.firebase</groupId>
      <artifactId>firebase-database</artifactId>
      <version>11.4.2</version>
    </dependency>
    <dependency>
      <groupId>android.arch.lifecycle</groupId>
      <artifactId>runtime</artifactId>
      <version>1.0.0-beta1</version>
    </dependency>
    <dependency>
      <groupId>android.arch.lifecycle</groupId>
      <artifactId>extensions</artifactId>
      <version>1.0.0-beta1</version>
    </dependency>
  </dependencies>
```

Here's the new one:

```xml
  <dependencies>
    <dependency>
      <groupId>com.firebaseui</groupId>
      <artifactId>firebase-ui-common</artifactId>
      <version>3.1.1-SNAPSHOT</version>
    </dependency>
    <dependency>
      <groupId>com.google.firebase</groupId>
      <artifactId>firebase-database</artifactId>
      <version>11.4.2</version>
    </dependency>
    <dependency>
      <groupId>com.android.support</groupId>
      <artifactId>support-v4</artifactId>
      <version>26.1.0</version>
    </dependency>
    <dependency>
      <groupId>com.android.support</groupId>
      <artifactId>recyclerview-v7</artifactId>
      <version>26.1.0</version>
    </dependency>
    <dependency>
      <groupId>android.arch.lifecycle</groupId>
      <artifactId>extensions</artifactId>
      <version>1.0.0-beta2</version>
    </dependency>
  </dependencies>
```

Now I'm not sure if this is an improvement or a regression in the dependencies!  Will need to think about it...

common/build.gradle
SUPERCILEX: You sure? The only reason we need that dependency is to override Play Services... We already include the recyclerview dep.

build.gradle
```diff
@@ -8,7 +8,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.3.1'
+        classpath 'com.android.tools.build:gradle:3.0.0-rc1'
         classpath 'com.google.gms:google-services:3.1.0'
```
SUPERCILEX: Should be `3.1.1`

build.gradle
```diff
@@ -8,7 +8,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.3.1'
+        classpath 'com.android.tools.build:gradle:3.0.0-rc1'
```
SUPERCILEX: Should be `3.0.0`.

samtstern:
@SUPERCILEX whoah big news!  Yeah I will take a look at all that, thanks for notifying me.

SUPERCILEX:
IKR!? The tools team has been working on this stuff for more than half a year now so it's great to see that finally come out. 😄

SUPERCILEX:
Awesome!

app/build.gradle
TheCraftKid: I believe this is actually the preferable action to take.

Issue #869, opened 2017-08-16
Welcome Back screen appears on email signup when saving by pressing enter on keyboard


harvitronix:
### Describe your environment

  * Android device: Emulator
  * Android OS version: 7.1.1
  * Google Play Services version: 11.0.4
  * Firebase/Play Services SDK version: 11.0.4
  * FirebaseUI version: 2.2.0
  
### Describe the problem:

When I create a new account using the email method, on the emulator, if I complete the sign up form by _pressing enter/return on my keyboard_, the "Welcome back!" modal appears on top of the signed in activity, even though this is a new signup. If I repeat the same steps but _click_ the Save button instead of pressing enter, the new user is created as expected, and the "Welcome back" model does not appear.

#### Steps to reproduce:

1. Create activities by following the example found here: https://github.com/firebase/FirebaseUI-Android/tree/master/app
3. Run the app in an emulator
4. Sign in with email
5. Fill out the email address input, click Next or press return
6. Fill out the First & last name and Password fields
7. Press ENTER (instead of clicking Save)

#### Observed Results:

User is created, and then immediately the "Welcome back!" screen appears. Entering the password we entered when creating the account causes a NPE:

```
E/AndroidRuntime: FATAL EXCEPTION: main
                                                                             Process: com.test.app, PID: 14783
                                                                             java.lang.NullPointerException: Activity must not be null
                                                                                 at com.google.android.gms.common.internal.zzbo.zzb(Unknown Source)
                                                                                 at com.google.android.gms.internal.zzbdr.<init>(Unknown Source)
                                                                                 at com.google.android.gms.common.api.GoogleApiClient$Builder.enableAutoManage(Unknown Source)
                                                                                 at com.firebase.ui.auth.util.signincontainer.SaveSmartLock.saveCredentialsOrFinish(SaveSmartLock.java:219)
                                                                                 at com.firebase.ui.auth.ui.HelperActivityBase.saveCredentialsOrFinish(HelperActivityBase.java:89)
                                                                                 at com.firebase.ui.auth.ui.accountlink.WelcomeBackPasswordPrompt$2.onSuccess(WelcomeBackPasswordPrompt.java:162)
                                                                                 at com.firebase.ui.auth.ui.accountlink.WelcomeBackPasswordPrompt$2.onSuccess(WelcomeBackPasswordPrompt.java:153)
                                                                                 at com.google.android.gms.tasks.zzj.run(Unknown Source)
[...]

PR #885, opened 2017-09-01
Fix duplicated callback events in DonePressedListener

SUPERCILEX:
See https://github.com/firebase/FirebaseUI-Android/issues/869#issuecomment-326655108

It crazy how little you know about X until you need to build a features that uses X to its full potential! 😆


samtstern:
@SUPERCILEX nice catch!  I actually knew about this but didn't think about it at the time.  Been bitten by this before, particularly the `return true` part.

PR #898, opened 2017-09-08
Lockdown firebase arrays and improve index add performance

SUPERCILEX:
Here's my reasoning behind the change:
1. Our classes aren't extensible anyway.
I'm personally a culprit here, I've never really bothered thinking about how devs would extend our firebase arrays. Take the `FirebaseIndexArray` for example. Let's say a developer wants to override `onKeyAdded`:
https://github.com/firebase/FirebaseUI-Android/blob/d871c7acd39d88ee8aba58739248804d5d34a898/database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java#L178-L185
All of the fields referenced in there are private so the developer would also have to override `onKey[Changed/Removed/Moved]`. Since they had to reimplement those methods, they also have to reimplement `DataRefListener` since it accesses private fields. And then all the other methods are private so at this point the developer has just reimplemented the whole class and there's no point in extending it in the first place. As for getting the callbacks, they should just be using our `ChangeEventListener` api.
2. It hinders improvements.
As you can see in this PR, I was able to improve performance by breaking backwards compatibility.
3. Favor composition over inheritance.
Using inheritance (on lists especially) is poor practice and we should encourage developers to use inheritance.

### DIFF
--- a/database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
+++ b/database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
@@ -28,7 +28,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-public class FirebaseIndexArray<T> extends CachingObservableSnapshotArray<T> implements ChangeEventListener {
+public final class FirebaseIndexArray<T> extends CachingObservableSnapshotArray<T>
+        implements ChangeEventListener {
     private static final String TAG = "FirebaseIndexArray";
 
     private DatabaseReference mDataRef;
@@ -107,7 +108,7 @@ public class FirebaseIndexArray<T> extends CachingObservableSnapshotArray<T> imp
     public void onChildChanged(EventType type, DataSnapshot snapshot, int index, int oldIndex) {
         switch (type) {
             case ADDED:
-                onKeyAdded(snapshot);
+                onKeyAdded(snapshot, newIndex);
                 break;
             case MOVED:
                 onKeyMoved(snapshot, index, oldIndex);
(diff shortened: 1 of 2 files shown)


SUPERCILEX:
@samtstern Could you make a `3.0` branch please?

To address your comment:
> Making methods private is probably a good thing, but making the class final means that you can't test this class with mocking, which is a pain. For that reason I don't really like using final except on fields.

You [_can_ mock final symbols](https://github.com/mockito/mockito/wiki/What's-new-in-Mockito-2#mock-the-unmockable-opt-in-mocking-of-final-classesmethods), it's just not a stable api yet. Does that work?

samtstern:
The mocking of final things does not play nice with Android, I tried really hard to get it to work before giving up :-(

And yep, will make that branch.  Will branch 3.0 from 2.3.1-dev to start

samtstern:
https://github.com/firebase/FirebaseUI-Android/tree/version-3.0.0-dev

SUPERCILEX:
@samtstern Oh well, that's too bad. 😢 I'll make the classes non-final and keep the other changes.

database/src/main/java/com/firebase/ui/database/FirebaseIndexArray.java
```diff
@@ -243,10 +244,14 @@ public String toString() {
     /**
      * A ValueEventListener attached to the joined child data.
      */
-    protected class DataRefListener implements ValueEventListener {
+    private final class DataRefListener implements ValueEventListener {
```
samtstern: Missed one!

SUPERCILEX: That shouldn't matter, right? We're making it private so it's not mockable anyway.

Eragon29:
I dont't want to bother you, but wanted to ask why this approach is more efficient?

SUPERCILEX:
@Eragon29 we don't have to query for the correct index on addition.

Eragon29:
@SUPERCILEX Ah, now I get it.
I think I am just not used to see code like
 ` int index = currentIndex = returnOrFindIndexForKey(currentIndex, key);`
 in Java. After that looking at the method `returnOrFindIndexForKey(index, key)` was enough.

SUPERCILEX:
👍

Issue #672, opened 2017-04-25
RecyclerView adapter does not work in fragment but works in activity


semenoh:
I've been using activity and firebase-ui in order to populate RecyclerView with data from Realtime Database. It works well in activity like this:
```Java
        @Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        setContentView(R.layout.fragment_transactions);
        list = (RecyclerView) findViewById(R.id.list);
        LinearLayoutManager layoutManager = new LinearLayoutManager(this);
        list.setLayoutManager(layoutManager);
        list.setAdapter(new TransactionsAdapter(
                DataManager.getTransactionsRef().orderByChild("date")
        ));
    }
```
but then I moved this code to fragment with the same layout:

```Java
    @Override
    public void onViewCreated(View view, @Nullable Bundle savedInstanceState) {
        super.onViewCreated(view, savedInstanceState);

        LinearLayoutManager layoutManager = new LinearLayoutManager(getActivity());
        list.setLayoutManager(layoutManager);
        list.setAdapter(new TransactionsAdapter(
                DataManager.getTransactionsRef().orderByChild("date")
        ));
    }
```

and it does not show any data. Currently I have both Activity and Fragment and it works in activity but it doesn't in Fragment.

Here is my dependencies:
```gradle
    compile 'com.google.firebase:firebase-core:10.2.1'
    compile 'com.google.firebase:firebase-database:10.2.1'
    compile 'com.google.firebase:firebase-auth:10.2.1'
    compile 'com.firebaseui:firebase-ui-auth:1.2.0'
    compile 'com.firebaseui:firebase-ui-database:1.2.0'
    compile 'com.android.support:design:25.3.1'
```

SUPERCILEX:
@semenoh This question would be better for Stack Overflow, but I'll give you some clues here anyway.

Ok, so you know how in your `AppCompatActivity` you use `setContentView(R.layout.something)`? Well, in Fragments you have to deal with that stuff yourself. I would recommend `Ctrl + B`ing the `super.onViewCreated(view, savedInstanceState);` to see what it does. If you do that, you'll see that it returns null and does nothing:
```java
@Nullable
public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container, @Nullable Bundle savedInstanceState) {
    return null;
}
```

TLDR: you have to manually inflate your layout like so:
```java
[...]

Issue #648, opened 2017-03-25
Firebase RecyclerView crash when closing an activity started by intent with 2 Fragments in FragmentStatePagerAdapter


lumleon:
Welcome to FirebaseUI and thanks for submitting an issue!

Please take a look at [open issues](https://github.com/firebase/FirebaseUI-Android/issues?q=is%3Aopen+is%3Aissue), as well as [resolved issues](https://github.com/firebase/FirebaseUI-Android/issues?q=is%3Aissue+is%3Aclosed), to see if your issue is either already being addressed, or has been solved by someone else.

If not, please feel free to fill in the following info so we can help faster!

### Step 1: Are you in the right place?

  * For issues or feature requests related to the code **in this repository** file a GitHub issue.
  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/) tagged appropriately.
  * For general Firebase discussion, use the [firebase-talk google group](https://groups.google.com/forum/#!forum/firebase-talk)
  * For help troubleshooting your application that does not fall under one of the above categories, reach out to the [personalized Firebase support channel](https://firebase.google.com/support/contact/troubleshooting/)

### Step 2: Describe your environment

  * Android device: _____ Android Emulator Nexus 5 API 25, Samsung I9505, Samsung P-605
  * Android OS version: __4.41, 5.01, 7.1___
  * Google Play Services version: _____ 10.2.1
  * Firebase/Play Services SDK version: _____10.2.1
  * FirebaseUI version: ____1.2.0
  
#### Steps to reproduce:

  1. create 2 fragments in FragmentStatePagerAdapter
  2. create FirebaseRecyclerAdapter in the 2 fragments and add clicklistener for the entire item
  3.  scroll to the end of the list view and click 
  4. start another activity by intent with the post key to show the child post details
  5.  click back and it will crash as the following log

  
#### Observed Results:

  * What happened?  This could be a description, `logcat` output, etc.
  03-25 22:48:54.560 8547-8583/com.anrex.raceblogger V/FA: Recording user engagement, ms: 4194
03-25 22:48:54.561 8547-8583/com.anrex.raceblogger V/FA: Activity paused, time: 2423311
03-25 22:48:54.574 8547-8583/com.anrex.raceblogger D/FA: Logging event (FE): _e, Bundle[{_o=auto, _et=4194, _sc=DailyRaceTipsActivity, _si=7506535637291765120}]
[...]

Issue #472, opened 2016-12-23
Firebase RecyclerView ClickListener Not Working


troy21688:
I followed the post [here][1]. I am essentially trying to set up a click listener in my `RecyclerView`. I am getting no response when I click an item in the `RecyclerView`. I have logged it and I'm not getting a call to the log when I click. Therefore, I know the `.onClick()` method is not firing. I initialize the adapter in `.onStart()`, as seen below:

    @Override
    public void onStart() {
        super.onStart();

        mFireAdapter = new FirebaseRecyclerAdapter<Poll, PollHolder>(Poll.class, R.layout.latest_item, PollHolder.class, mBaseRef.child("Polls")) {
            @Override
            protected void populateViewHolder(PollHolder viewHolder, Poll model, final int position) {
                viewHolder.mPollQuestion.setText(model.getQuestion());
                Picasso.with(getActivity().getApplicationContext())
                        .load(model.getImage_URL())
                        .fit()
                        .into(viewHolder.mPollImage);
                viewHolder.mView.setOnClickListener(new View.OnClickListener() {
                    @Override
                    public void onClick(View view) {
                        Log.v("TAG","ONCLICKED");
                        Intent toPoll = new Intent(getActivity(), PollHostActivity.class);
                        toPoll.putExtra("POLL_ID", mFireAdapter.getRef(position).getKey());
                        startActivity(toPoll);

                    }
                });
                Log.v("TAG", mFireAdapter.getRef(position).getKey());
                Log.v("TAG", String.valueOf(position));
                Log.v("TAG", model.getQuestion());
                Log.v("TAG", model.getImage_URL());
            }
        };

        mRecyclerview.setAdapter(mFireAdapter);

        mFireAdapter.registerAdapterDataObserver(new RecyclerView.AdapterDataObserver() {
            @Override
            public void onItemRangeInserted(int positionStart, int itemCount) {
                super.onItemRangeInserted(positionStart, itemCount);
                int pollCount = mFireAdapter.getItemCount();
                int lastVisiblePosition = mLayoutManager.findLastCompletelyVisibleItemPosition();

                // If the recycler view is initially being loaded or the user is at the bottom of the list, scroll
                // to the bottom of the list to show the newly added message.
                if (lastVisiblePosition == -1 ||
                        (positionStart >= (pollCount - 1) && lastVisiblePosition == (positionStart - 1))) {
                    mRecyclerview.scrollToPosition(positionStart);
                }
            }
        });
    }

[...]

PR #903, opened 2017-09-12
Revamp sample

SUPERCILEX:
- Only show relevant database code to setup adapter and send messages
- Reorganize strings
- Reorganize sample item order


samtstern:
@SUPERCILEX thanks for sending this over.  I see what you're trying to do with the base classes, it highlights only the FirebaseUI code and abstracts the boilerplate.

However I think that's a non-goal of this sample.  The README is where we can highlight code snippets without context.  The sample activity shows the code you actually need to wire this library into your app.  By adding base activities we sweep this complexity under the rug but the developer is going to need to look at it anyway to see how to actually put FirebaseUI in their app.

SUPERCILEX:
> I think that's a non-goal of this sample

Let me try to convince you otherwise. 😄

Over the time I've worked on FirebaseUI, I've noticed a fundamental disconnect for newbies between their understanding of the `RecyclerView` and the adapter. A significant portion of them seem to think that FirebaseUI-Database and the `RecyclerView` are the same thing. I actually feel like I thought something similar when I first started!

The problem as I see it is that we don't link to the support library documentation and instead try to teach all the `RecyclerView` goodness in our own README. In addition, we hide the normal `RecyclerView.Adapter` abstraction behind reflection and a weirdly named method like `populateViewHolder(...)`. While we're doing those other refactors 😉, I think we should actually just remove those methods and have the dev implement `onCreateViewHolder(...)` and `onBindViewHolder(..., T model)` themselves.

Going back to the sample, for newbies I think a clear separation between modular adapters and the actual `RecyclerView` is necessary. If we beef up our README to include tutorials from the support library about how to setup a RecyclerView, I think this will provide a very clear separation between "How do I setup a general `RecyclerView`?" and "How do I provide data to the `RecyclerView`?". Our current sample throws everything at newbies all at once whereas the new provides clear classes that answer the questions "How do I write data to the database and then read it back?" and "How do I prepare a `RecyclerView` for receiving data?". What I'm basically saying is that we need clear `RecyclerView` and `RV.Adapter` education that isn't confusingly mixed together.

@samtstern What do you think given the points above?

PS: if you want proof of the confusion, have a look at #648 where a dev crashed the `RecyclerView`, #672 where a dev didn't setup the `RecyclerView` height correctly, #556 and #482 where a dev didn't understand `RecyclerView` animations, #472 where a click listener wasn't working for some reason, #457 where a dev didn't know he had to call `RecyclerView#setAdapter(...)`, and a whole bunch more. There's evidence aplenty! :smile:

SUPERCILEX:
@samtstern went back to merging the sample after our internal discussion.

samtstern:
Note here: @SUPERCILEX and I are working on some API changes that will alleviate some of the pain points (which are enumerated above in detail) and then we will revisit this issue once those changes land.

There are still definitely some sample improvements we should take from this PR, but hopefully fewer once the API improves.

SUPERCILEX:
@samtstern I think this PR is ready for another round of review. The Git diff got messed up because of the name change, but a lot changed anyway:
- Merged improvements from intertal developeent
  - Use ButterKnife
  - Auth listener is tied to start/stop lifecycle instead of create/destroy
- Pressing enter on keyboard sends message
 - Other miscellaneous reorganization

app/src/main/java/com/firebase/uidemo/database/realtime/RealtimeDbChatActivity.java
```diff
@@ -0,0 +1,177 @@
...
+
+        if (isSignedIn()) {
+            attachRecyclerViewAdapter();
+        } else {
+            Toast.makeText(this, R.string.signing_in, Toast.LENGTH_SHORT).show();
+            auth.signInAnonymously()
+                    .addOnSuccessListener(this, new OnSuccessListener<AuthResult>() {
+                        @Override
+                        public void onSuccess(AuthResult result) {
+                            attachRecyclerViewAdapter();
```
samtstern: I think this success handler is not necessary, since when we sign in successfully the listener will trigger and isSignedIn will be true and then we'll attach the adapter.

app/src/main/java/com/firebase/uidemo/database/realtime/RealtimeDbChatActivity.java
```diff
@@ -0,0 +1,177 @@
...
+
+    private boolean isSignedIn() {
+        return FirebaseAuth.getInstance().getCurrentUser() != null;
+    }
+
+    private void attachRecyclerViewAdapter() {
+        final RecyclerView.Adapter adapter = newAdapter();
+
+        // Scroll to bottom on new messages
+        adapter.registerAdapterDataObserver(new RecyclerView.AdapterDataObserver() {
```
samtstern: Can we replace the non-scrolling pieces of this by watching onDataChanged?

app/src/main/java/com/firebase/uidemo/database/realtime/RealtimeDbChatActivity.java
SUPERCILEX: Whoops, refactor mistake. 😄

app/src/main/java/com/firebase/uidemo/database/realtime/RealtimeDbChatActivity.java
SUPERCILEX: Yep, old code from the abstraction.

samtstern:
@SUPERCILEX thank you!  This looks great now.  I really appreciate the back-and-forth here.  Once we're ready to land 3.0 we can revisit some of your other sample concerns and make sure they're all addressed.

SUPERCILEX:
Awesome, time to work on that internal sample revamp PR! 😄

Issue #887, opened 2017-09-04
How to have prefilled phone number from shared pref. in PhoneVerificationActivity?


pinkeshdarji:
Hi, I need to automatically fill phone number from shared pref. to firebase phone ui auth activity.
Is there any way that If I modify this
                        startActivityForResult(
                                AuthUI.getInstance().createSignInIntentBuilder().setAvailableProviders(
                                        Arrays.asList(new AuthUI.IdpConfig.Builder(AuthUI.PHONE_VERIFICATION_PROVIDER).build()))
                                        .build(), RC_SIGN_IN);

in such a way that, when verify phone activity gets open, instead of asking for a phone number I should have it from shared pref and user can continue instead of giving phone number again which we have already stored while registration.

samtstern:
@pinkeshdarji sorry this is not supported, as it's not a normal flow as far as FirebaseUI is concerned.  If you need to build custom flows like this, I suggest using [Firebase phone authentication](https://firebase.google.com/docs/auth/android/phone-auth) directly.

samtstern:
This issue has been fixed in version `2.4.0`:
https://github.com/firebase/FirebaseUI-Android/releases/tag/2.4.0

pinkeshdarji:
@samtstern can you please suggest how to use it? I am using firebase 3.0.0 version.

Issue #806, opened 2017-07-18
Centralize email configuration options


samtstern:
Right now we have an email option exposed on `SignInIntentBuilder` to allow new email accounts.  We have multiple feature requests for other email flags:

  * [467 - Don't require display name](https://github.com/firebase/FirebaseUI-Android/issues/467)
  * [312 - Require verified email address](https://github.com/firebase/FirebaseUI-Android/issues/312)

I'd like to see these implemented but don't think we should keep hanging random options off the sign-in intent builder.  Instead we'd like something like this (pseudocode):

```java
new AuthUI.IdpConfig.Builder(AuthUI.EMAIL_PROVIDER)
        .allowNewAccounts(false)
        .requireDisplayName(false)
        .requireEmailVerification(true)
        .build(),
```

This is a much more scalable way to add flags over time.

TheCraftKid:
I could make a PR if you're not working on this already.

samtstern:
@TheCraftKid thanks!  Not working on this already but do you want to quickly outline what you think the API in this issue before starting a PR?  That way we don't have to go back and forth on PR comments.

TheCraftKid:
This appears to be a good idea adding it to the `IdpConfig.Builder`, but would an option such as `requireEmailVerification()` on multiple identity providers make sense?

samtstern:
I am wondering if we could do something tricky.

Maybe `AuthUI.IdpConfig.Builder(AuthUI.EMAIL_PROVIDER)` would return a different `Builder` class that's a subclass of the generic one and includes email-specific options.

This pattern would also allow us to do the same for other methods.  For instance we could make `setPermissions` only appear on Google/Facebook/Twitter since it doesn't make sense on Phone/Email.

TheCraftKid:
Yeah, I was initially thinking generic classes would be the way to go. Just a bit more work. :smile:

TheCraftKid:
Would this be in 2.1.0, 2.2.0 or the next major release?

samtstern:
I think this can be done in the next release (which we are now calling 2.1.1 but would be renamed 2.2.0 if we had this change).

We can just mark the current email options as deprecated but not delete them, so it shouldn't be a breaking change.

TheCraftKid:
Okay, LGTM, I'll get to work.

TheCraftKid:
Okay, I want to squash #312, but the Auth module is so dense, I don't know where to verify/check if the email is verified. Thoughts?

[...]

PR #906, opened 2017-09-12
Default phone

yramanchuk:
Implement phone Auth flow Phone number configuration parameter.
Related issue https://github.com/firebase/FirebaseUI-Android/issues/887

### DIFF
--- a/app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
+++ b/app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
@@ -33,7 +33,6 @@ import android.widget.TextView;
 
 import com.firebase.ui.auth.AuthUI;
 import com.firebase.ui.auth.AuthUI.IdpConfig;
-import com.firebase.ui.auth.AuthUI.PhoneIdpConfig;
 import com.firebase.ui.auth.ErrorCodes;
 import com.firebase.ui.auth.IdpResponse;
 import com.firebase.uidemo.R;
@@ -206,7 +205,7 @@ public class AuthUiActivity extends AppCompatActivity {
                         .setTosUrl(getSelectedTosUrl())
                         .setPrivacyPolicyUrl(getSelectedPrivacyPolicyUrl())
                         .setIsSmartLockEnabled(mEnableCredentialSelector.isChecked(),
-                                               mEnableHintSelector.isChecked())
+                                mEnableHintSelector.isChecked())
                         .setAllowNewEmailAccounts(mAllowNewEmailAccounts.isChecked())
                         .build(),
                 RC_SIGN_IN);
@@ -338,8 +337,14 @@ public class AuthUiActivity extends AppCompatActivity {
         }
 
         if (mUsePhoneProvider.isChecked()) {
+            // Use the params Bundle to provide a default phone number that will populate the
+            // phone number field in PhoneVerificationActivity if the number is known in advance.
+            Bundle params = new Bundle();
+            params.putString(AuthUI.PHONE_VERIFICATION_DEFAULT_PHONE_EXTRA, "+13099126574");
             selectedProviders.add(
-                    new PhoneIdpConfig.Builder().setPhone("").build()); // Add default phone number.
+                    new IdpConfig.Builder(AuthUI.PHONE_VERIFICATION_PROVIDER)
+                            .setParams(params)
+                            .build());
         }
 
         return selectedProviders;
--- a/app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
+++ b/app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
@@ -337,13 +337,8 @@ public class AuthUiActivity extends AppCompatActivity {
         }
 
         if (mUsePhoneProvider.isChecked()) {
-            // Use the params Bundle to provide a default phone number that will populate the
-            // phone number field in PhoneVerificationActivity if the number is known in advance.
-            Bundle params = new Bundle();
-            params.putString(AuthUI.PHONE_VERIFICATION_DEFAULT_PHONE_EXTRA, "+13099126574");
             selectedProviders.add(
                     new IdpConfig.Builder(AuthUI.PHONE_VERIFICATION_PROVIDER)
-                            .setParams(params)
                             .build());
         }
 
--- a/app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
+++ b/app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
@@ -338,8 +338,7 @@ public class AuthUiActivity extends AppCompatActivity {
 
         if (mUsePhoneProvider.isChecked()) {
             selectedProviders.add(
-                    new IdpConfig.Builder(AuthUI.PHONE_VERIFICATION_PROVIDER)
-                            .build());
+                    new IdpConfig.Builder(AuthUI.PHONE_VERIFICATION_PROVIDER).build());
         }
 
         return selectedProviders;
--- a/app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
+++ b/app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
@@ -33,6 +33,7 @@ import android.widget.TextView;
 
 import com.firebase.ui.auth.AuthUI;
 import com.firebase.ui.auth.AuthUI.IdpConfig;
+import com.firebase.ui.auth.AuthUI.PhoneIdpConfig;
 import com.firebase.ui.auth.ErrorCodes;
 import com.firebase.ui.auth.IdpResponse;
 import com.firebase.uidemo.R;
@@ -338,7 +339,7 @@ public class AuthUiActivity extends AppCompatActivity {
 
         if (mUsePhoneProvider.isChecked()) {
             selectedProviders.add(
-                    new IdpConfig.Builder(AuthUI.PHONE_VERIFICATION_PROVIDER).build());
+                    new PhoneIdpConfig.Builder().setPhone("").build()); // User default phone number
         }
 
         return selectedProviders;
--- a/app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
+++ b/app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
@@ -339,7 +339,7 @@ public class AuthUiActivity extends AppCompatActivity {
 
         if (mUsePhoneProvider.isChecked()) {
             selectedProviders.add(
-                    new PhoneIdpConfig.Builder().setPhone("").build()); // User default phone number
+                    new PhoneIdpConfig.Builder().setPhone("").build()); // Add default phone number.
         }
 
         return selectedProviders;
--- a/auth/src/main/java/com/firebase/ui/auth/AuthUI.java
+++ b/auth/src/main/java/com/firebase/ui/auth/AuthUI.java
@@ -350,7 +350,7 @@ public class AuthUI {
             return "IdpConfig{" +
                     "mProviderId='" + mProviderId + '\'' +
                     ", mScopes=" + mScopes +
-                    ", mParams=" + mParams.toString() +
+                    ", mParams=" + mParams +
                     '}';
         }
 
--- a/auth/src/main/java/com/firebase/ui/auth/AuthUI.java
+++ b/auth/src/main/java/com/firebase/ui/auth/AuthUI.java
@@ -16,6 +16,7 @@ package com.firebase.ui.auth;
 
 import android.app.Activity;
 import android.content.Intent;
+import android.os.Bundle;
 import android.os.Parcel;
 import android.os.Parcelable;
 import android.support.annotation.CallSuper;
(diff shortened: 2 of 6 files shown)


SUPERCILEX:
@yramanchuk I believe the result of that issue was that FirebaseUI isn't going to support this kind of flow. Could you explain why you already have a phone number? The user should be giving that to FirebaseUI in the auth flow, not to you directly. Hope this clears things up! 😄

ashwinraghav:
@yramanchuk I think the issue that is linked to may not be completely comparable.

@SUPERCILEX The problem we are addressing is rooted in the fact that Firebase UI currently does not support 2FA (Say Facebook + Phone Auth). To work around this, apps that wish to use phone as the second factor can use Facebook login on their own outside Firebase UI and then use Firebase UI for the second factor phone auth.
This change helps apps that "know" the second factor phone number given a Facebook account, particularly when a user is logging into their secondary device like a tablet. Hope that helps.

yramanchuk:
@ashwinraghav Thanks for the clarification.
@SUPERCILEX here's another example: application may have multi-dialog account creation flow and they have already requested user to enter his phone number, so they would like to show phone verification dialog with pre-filled phone number.

Here's related FirebaseUI-iOS issue: https://github.com/firebase/FirebaseUI-iOS/issues/306 
This request is required for the feature parity: https://github.com/firebase/FirebaseUI-iOS/pull/334

auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
```diff
@@ -114,7 +115,8 @@ private void populateIdpList(List<IdpConfig> providers) {
                     mProviders.add(new EmailProvider(this, getFlowParams()));
                     break;
                 case AuthUI.PHONE_VERIFICATION_PROVIDER:
-                    mProviders.add(new PhoneProvider(this, getFlowParams()));
+                    PhoneIdpConfig phoneIdpConfig = (PhoneIdpConfig) idpConfig;
```
ashwinraghav: Would this break older clients, breaking backward compatibility?

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -378,6 +368,79 @@ public IdpConfig build() {
         }
     }
 
+    public static class PhoneIdpConfig extends IdpConfig {
```
ashwinraghav: I feel like introducing a config per for phone auth provider leaves me desiring symmetry for such a config in other providers. 

Also, the type safety that this implementation gives is hard to enforce in a backward compatible way. This leads to unsafe typecasting like the one made in https://github.com/firebase/FirebaseUI-Android/blob/d1da32786f33f7a5552d3a3c5c778eb58c24305a/auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java#L118

My suggestion would be to introduce a Bundle parameter into the existing IdpConfig that yields an API that would look like:

Bundle b = new Bundle();
b.putString(AuthUI.PHONE_VERIFICATION_PROVIDER.DEFAULT_PHONE_EXTRA, "+16478488383");

new AuthUI.IdpConfig.Builder(AuthUI.PHONE_PROVIDER).setParams(bundle).build()) and wire it into the providers from that point on.

yramanchuk: @ashwinraghav Thhank's for the suggestion. I was thinking about extending existing IdpConfig but in that case were are adding changes to all configs (Google, Facebook) and they they don't need that parameter.

Anyway I think your solution is more elegant. I just want to propose to use List of Bundles instead of just Bundle:

AuthUI.IdpConfig.Builder(AuthUI.PHONE_PROVIDER).setParams(List<Bundle>).build())

In that case we can pass multiple params.
@ashwinraghav what do you think?

samtstern:
@yramanchuk FYI I changed the target branch to the current development branch, `version-2.3.1-dev`.

samtstern:
Note that the issue of IDP-specific configuration has been discussed here as well:
https://github.com/firebase/FirebaseUI-Android/issues/806

app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
```diff
@@ -337,8 +337,14 @@ private int getSelectedLogo() {
         }
 
         if (mUsePhoneProvider.isChecked()) {
+            // Use the params Bundle to provide a default phone number that will populate the
+            // phone number field in PhoneVerificationActivity if the number is known in advance.
+            Bundle params = new Bundle();
+            params.putString(AuthUI.PHONE_VERIFICATION_DEFAULT_PHONE_EXTRA, "+13099126574");
```
samtstern: This is a minor point, but I think putting a random phone number into the sample will cause confusion.  Many people will want to use this sample to test phone auth with their real number.

This type of example snippet belongs in the README.

auth/src/main/java/com/firebase/ui/auth/util/signincontainer/SignInDelegate.java
```diff
@@ -248,8 +248,12 @@ private void startAuthMethodChoice() {
                     break;
                 case PhoneAuthProvider.PROVIDER_ID:
                     // Go directly to phone flow
+                    String phone = idpConfigs
+                            .get(0)
+                            .getParams()
+                            .getString(AuthUI.PHONE_VERIFICATION_DEFAULT_PHONE_EXTRA);
                     startActivityForResult(
-                            PhoneVerificationActivity.createIntent(getContext(), flowParams, null),
+                            PhoneVerificationActivity.createIntent(getContext(), flowParams, phone),
```
samtstern: This `createIntent` call uses `ExtraConstants.EXTRA_PHONE`.  We should unify the values between this and AuthUI.PHONE_VERIFICATION_DEFAULT_PHONE_EXTRA.

I'd suggest:

```java

class AuthUI {
  // ...
  public static final String EXTRA_DEFAULT_PHONE = ExtraConstants.EXTRA_PHONE;
  // ...
}
```

auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
samtstern: This class `PhoneIdpConfig` is used here but I don't see it defined in this PR anymore.  I think changes to this file need to be revisited.

samtstern:
@SUPERCILEX the main reason for supporting this flow is feature parity with what Digits used to offer.  We are telling all Digits developers to move to FirebaseUI and some of them noticed this missing feature.

You are right though, it was not a flow we originally intended to support.  Ashwin and Yury have provided some examples where it can be very useful though.

app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
jstimes: Moved to README, let me know if you don't think it's in an appropriate location.

auth/src/main/java/com/firebase/ui/auth/util/signincontainer/SignInDelegate.java
jstimes: That's reasonable, made your suggested change.

auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
jstimes: Oops, gone now.

app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
```diff
@@ -338,7 +338,8 @@ private int getSelectedLogo() {
 
         if (mUsePhoneProvider.isChecked()) {
             selectedProviders.add(
-                    new IdpConfig.Builder(AuthUI.PHONE_VERIFICATION_PROVIDER).build());
+                    new IdpConfig.Builder(AuthUI.PHONE_VERIFICATION_PROVIDER)
+                            .build());
```
SUPERCILEX: Might as well undo the changes in this file since they're leftovers from previous commits.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -329,15 +347,27 @@ public int hashCode() {
...
         public String toString() {
+            StringBuilder paramsValueBuilder = new StringBuilder();
+            paramsValueBuilder.append("{");
+            for (String key : mParams.keySet()) {
+                paramsValueBuilder.append(key);
+                paramsValueBuilder.append("=");
+                paramsValueBuilder.append(mParams.get(key).toString());
+                paramsValueBuilder.append(",");
+            }
+            paramsValueBuilder.append("}");
```
SUPERCILEX: Doesn't `Bundle` already have a `toString()` with very similar behavior?

auth/src/main/java/com/firebase/ui/auth/provider/PhoneProvider.java
```diff
@@ -34,8 +37,16 @@ public int getButtonLayout() {
 
     @Override
     public void startLogin(Activity activity) {
+
+        String phone = null;
+        for (AuthUI.IdpConfig idpConfig : mFlowParameters.providerInfo) {
+            if (idpConfig.getProviderId().equals(AuthUI.PHONE_VERIFICATION_PROVIDER)) {
+                phone = idpConfig.getParams().getString(EXTRA_DEFAULT_PHONE);
```
SUPERCILEX: To be consistent, we shouldn't statically import this and instead use `AuthUI.EXTRA_DEFAULT_PHONE`.

auth/src/main/java/com/firebase/ui/auth/util/signincontainer/SignInDelegate.java
```diff
@@ -248,8 +248,12 @@ private void startAuthMethodChoice() {
                     break;
                 case PhoneAuthProvider.PROVIDER_ID:
                     // Go directly to phone flow
+                    String phone = idpConfigs
+                            .get(0)
```
SUPERCILEX: This same logic is used above so we should probably extract a `firstIdpConfig` and replace `firstProvider` with the method call.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
SUPERCILEX: Yeah, just check the source and `Bundle#toString()` calls `mMap.toString()`. I'm assuming the formatting will be a little bit different, but the output should be effectively the same.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -269,15 +276,21 @@ public SignInIntentBuilder createSignInIntentBuilder() {
...
+                Bundle params) {
             mProviderId = providerId;
             mScopes = Collections.unmodifiableList(scopes);
+            mParams = params;
         }
 
         private IdpConfig(Parcel in) {
             mProviderId = in.readString();
             mScopes = Collections.unmodifiableList(in.createStringArrayList());
+            mParams = in.readBundle();
```
SUPERCILEX: This should be `in.readBundle(getClass().getClassLoader())` to make the build pass. Since we're just passing in strings right now we don't actually need this, but it'll ensure we're future proof (and make lint happy).

app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
jstimes: Reverted.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
jstimes: Ah, did not realize that. Switched to the Bundle toString.

auth/src/main/java/com/firebase/ui/auth/provider/PhoneProvider.java
jstimes: Fixed

auth/src/main/java/com/firebase/ui/auth/util/signincontainer/SignInDelegate.java
jstimes: Good point, done.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
jstimes: Ah, just noticed the CONTRIBUTING.md file. Apologies - just beginning contribution to the library and obviously wasn't paying full attention. Made the change and build should be passing now. Thanks for the review by the way!

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -332,12 +350,14 @@ public String toString() {
             return "IdpConfig{" +
                     "mProviderId='" + mProviderId + '\'' +
                     ", mScopes=" + mScopes +
+                    ", mParams=" + mParams.toString() +
```
SUPERCILEX: One last nit! 😄 We don't need the `toString()` call, `mParams` alone is just fine.

jstimes: Removed.

samtstern:
I will merge this once Travis is happy.

Issue #916, opened 2017-09-21
[Feature Request] Allow set default country code for phone auth


jhuleatt:
### Step 3: Describe the problem:

In [PhoneNumberUtils.java](https://github.com/firebase/FirebaseUI-Android/blob/f4494c4a5fcafc8b32351edcb02ed2834c6a32b4/auth/src/main/java/com/firebase/ui/auth/ui/phone/PhoneNumberUtils.java#L36), the default country code and locale are hardcoded as US. 

It does appear that the locale is [automatically set](https://github.com/firebase/FirebaseUI-Android/blob/f4494c4a5fcafc8b32351edcb02ed2834c6a32b4/auth/src/main/java/com/firebase/ui/auth/ui/phone/PhoneNumberUtils.java#L117) based on a user's SIM, but it would be helpful to be able to set a default locale too.

ashfaaaa:
is it possible to remove choose country spinner and set default country code? so users from the specified country can sign up.

SUPERCILEX:
@samtstern I believe this was fixed in FirebaseUI [`v2.4.0`](https://github.com/firebase/FirebaseUI-Android/releases/tag/2.4.0).

SUPERCILEX:
Nevermind, that was just to prefill the phone number. I'll add this feature to #1083.

samtstern:
This has been fixed and released in version `3.2.0`.

PR #919, opened 2017-09-25
Allowing default country code and national phone number

jstimes:
Allowing these to be provided similarly to how the default full phone number is provided.

Possible fix of issue #916, unless that is specifically asking for a setLocale method on PhoneUtils itself.

### DIFF
--- a/auth/README.md
+++ b/auth/README.md
@@ -263,8 +263,22 @@ country code and phone number input fields. The user is still able to edit the n
 ```java
 // Use a Bundle to hold the default number, and pass it to the Builder via setParams:
 Bundle params = new Bundle();
-params.putString(AuthUI.EXTRA_DEFAULT_PHONE, "+12345678901");
+params.putString(EXTRA_DEFAULT_PHONE, "+123456789");
 IdpConfig phoneConfigWithDefaultNumber =
+        new IdpConfig.Builder(AuthUI.PHONE_VERIFICATION_PROVIDER)
+                .setParams(params)
+                .build();
+```
+
+It is also possible to set a default country code along with a national number if a specific country
+is your app's target audience. This will take precedence over the full default phone number if both
+are provided.
+
+```java
+Bundle params = new Bundle();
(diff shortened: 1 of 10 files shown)


auth/README.md
```diff
@@ -263,13 +263,27 @@ country code and phone number input fields. The user is still able to edit the n
 ```java
 // Use a Bundle to hold the default number, and pass it to the Builder via setParams:
 Bundle params = new Bundle();
-params.putString(AuthUI.EXTRA_DEFAULT_PHONE, "+12345678901");
+params.putString(EXTRA_DEFAULT_PHONE, "+123456789");
```
samtstern: We should leave the `AuthUI.` otherwise developers won't know where this constant comes from

auth/README.md
```diff
@@ -263,13 +263,27 @@ country code and phone number input fields. The user is still able to edit the n
...
 ```
 
+It is also possible to set a default country code along with a national number if a specific country
+is your app's target audience. This will take precedence over the full default phone number if both
+are provided.
+
+```java
+Bundle params = new Bundle();
+params.putString(EXTRA_DEFAULT_COUNTRY_CODE, "ca");
+params.putString(EXTRA_DEFAULT_NATIONAL_NUMBER, "23456789");
```
samtstern: This naming is not consistent:
```
EXTRA_DEFAULT_PHONE
EXTRA_DEFAULT_NATIONAL_NUMBER
```

They should both be `*_NUMBER` or `*_PHONE`

auth/README.md
jstimes: Good point, updated now.

auth/README.md
jstimes: Using *_NUMBER now. 

Do you think EXTRA_DEFAULT_PHONE_NUMBER should be EXTRA_DEFAULT_FULL_PHONE_NUMBER for clarity?

samtstern:
@jstimes let's yet @yramanchuk or @ashwinraghav approve also before merging

samtstern:
Alright it's been two days, I am going to merge this.  Thanks @jstimes

Issue #921, opened 2017-09-25
NullPointerException in CompletableProgressDialog


samtstern:
Reported by partner.

```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'void android.widget.ProgressBar.setVisibility(int)' on a null object reference
       at com.firebase.ui.auth.ui.phone.CompletableProgressDialog.onComplete(CompletableProgressDialog.java:63)
       at com.firebase.ui.auth.ui.phone.PhoneVerificationActivity.completeLoadingDialog(PhoneVerificationActivity.java:369)
       at com.firebase.ui.auth.ui.phone.PhoneVerificationActivity.onCodeSent(PhoneVerificationActivity.java:184)
       at com.firebase.ui.auth.ui.phone.PhoneVerificationActivity.access$700(PhoneVerificationActivity.java:55)
       at com.firebase.ui.auth.ui.phone.PhoneVerificationActivity$2.onCodeSent(PhoneVerificationActivity.java:256)
```

TheCraftKid:
I thought we were getting rid of all the progress dialogs. :confused:

samtstern:
This issue has been fixed in version `2.4.0`:
https://github.com/firebase/FirebaseUI-Android/releases/tag/2.4.0

hackerutsav:
Hello,i have 2.3.0 and it's again coming here.It's not been fixed still :(

Issue #914, opened 2017-09-17
Exception java.lang.IllegalStateException: Can not perform this action after onSaveInstanceState in `CompletableProgressDialog`


yishai-glide:
* Android device: SM-J327P, E2115, vivo 1609, VS425PP
  * Android OS version: 19, 22,23, 24
  * Firebase/Play Services SDK version: 11.0.4
  * FirebaseUI version: 2.3.0
  
### Step 3: Describe the problem:
While monitoring our latest release, we've seen a crash that has originated from users in the wilde via Firebase Crash reporting.

seems like #817 is still hapening
  
#### Observed Results:

```
android.support.v4.app.FragmentManagerImpl.checkStateLoss (SourceFile:2044)
android.support.v4.app.FragmentManagerImpl.enqueueAction (SourceFile:2067)
android.support.v4.app.BackStackRecord.commitInternal (SourceFile:680)
android.support.v4.app.BackStackRecord.commit (SourceFile:634)
android.support.v4.app.DialogFragment.show (SourceFile:143)
com.firebase.ui.auth.ui.phone.CompletableProgressDialog.b (SourceFile:41)
com.firebase.ui.auth.ui.phone.PhoneVerificationActivity.af (SourceFile:377)
com.firebase.ui.auth.ui.phone.PhoneVerificationActivity.a (SourceFile:1174)
com.firebase.ui.auth.ui.phone.PhoneVerificationActivity$2.onVerificationCompleted (SourceFile:239)
com.google.android.gms.internal.kn.zza ()
com.google.android.gms.internal.kq.run ()
android.os.Handler.handleCallback (Handler.java:739)
android.os.Handler.dispatchMessage (Handler.java:95)
android.os.Looper.loop (Looper.java:145)
android.app.ActivityThread.main (ActivityThread.java:7007)
java.lang.reflect.Method.invoke (Method.java)
java.lang.reflect.Method.invoke (Method.java:372)
com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:1404)
com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1199)

```

samtstern:
This is the method that crashes:
https://github.com/firebase/FirebaseUI-Android/blob/9dc4c120f33d4f4acc42b2a0c2195e130e8b1076/auth/src/main/java/com/firebase/ui/auth/ui/phone/PhoneVerificationActivity.java#L377

It gets called from here:
[...]

PR #927, opened 2017-09-27
Fix dialog NPE and IllegalStateExceptions by guarding

samtstern:
Issues #914 and #921 addressed


samtstern:
cc @jstimes @ashwinraghav

Issue #922, opened 2017-09-25
IllegalArgumentException in PhoneVerificationActivity


samtstern:
https://github.com/firebase/FirebaseUI-Android/blob/f4494c4a5fcafc8b32351edcb02ed2834c6a32b4/auth/src/main/java/com/firebase/ui/auth/ui/phone/PhoneVerificationActivity.java#L161

Reported by a partner:
```
Fatal Exception: java.lang.IllegalArgumentException: Given String is empty or null
       at com.google.android.gms.common.internal.zzbo.zzcF(Unknown Source)
       at com.google.firebase.auth.PhoneAuthCredential.(Unknown Source)
       at com.google.firebase.auth.PhoneAuthProvider.getCredential(Unknown Source)
       at com.firebase.ui.auth.ui.phone.PhoneVerificationActivity.submitConfirmationCode(PhoneVerificationActivity.java:161)
       at com.firebase.ui.auth.ui.phone.SubmitConfirmationCodeFragment.submitConfirmationCode(SubmitConfirmationCodeFragment.java:175)
       at com.firebase.ui.auth.ui.phone.SubmitConfirmationCodeFragment.access$400(SubmitConfirmationCodeFragment.java:42)
       at com.firebase.ui.auth.ui.phone.SubmitConfirmationCodeFragment$2.onClick(SubmitConfirmationCodeFragment.java:169)
       at android.view.View.performClick(View.java:5706)
       at android.view.View$PerformClick.run(View.java:22834)
       at android.os.Handler.handleCallback(Handler.java:836)
       at android.os.Handler.dispatchMessage(Handler.java:103)
       at android.os.Looper.loop(Looper.java:203)
       at android.app.ActivityThread.main(ActivityThread.java:6339)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1084)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:945)
```

samtstern:
The method is called from `SubmitConfirmationCodeFragment`:

```java
private void submitConfirmationCode() {
        mVerifier.submitConfirmationCode(mConfirmationCodeEditText.getUnspacedText().toString());
    }
```

This is called in two places.  One is when the enter button is clicked on the keyboard:

```java
        ImeHelper.setImeOnDoneListener(mConfirmationCodeEditText,
[...]

PR #928, opened 2017-09-27
Fix IllegalArgumentException when submitting confirmation code

samtstern:
Issue #922


samtstern:
cc @ashwinraghav @jstimes

jstimes:
Is there a reason mVerificationId is not saved in onSaveInstanceState (in PhoneVerificationActivity)?

If a user leaves the app to look at the verification code then reopens it, onStart gets called and mVerificationId will be null until the onCodeSent callback gets fired. 

I agree that the null checks will prevent these crashes... just want to double-check that not saving mVerificationId is intentional.

samtstern:
@jstimes well that callback is re-entrant so it's not supposed to be necessary to save the verification Id.  But clearly we need to protect against not getting the callback before submission.

@ashwinraghav is putting together a new PR.

samtstern:
Closing in favor of #931

Issue #926, opened 2017-09-27
Class not found when unmarshalling: com.firebase.ui.auth.ui.FlowParameters


thailemon:
When using FirebaseUI Auth with Facebook, it crashes immediately after launching the Activity. The google-services.json file is updated. 

The Facebook app ID is set in strings.xml:

```xml
<string name="facebook_application_id" translatable="false">appid</string>
<string name="facebook_login_protocol_scheme" translatable="false">fb_appid</string>
```

Have tried versions of google play services, facebook sdk, firebase, firebase ui, to no avail.
When removing Facebook as provider, and only using Google Sign-in/Phone/Email, it works without any issues. 

Problem persists in debug and release APK.

Any suggestions on whats going on? 

```java
 startActivityForResult(
                AuthUI.getInstance().createSignInIntentBuilder()
                        .setTheme(getTheme())
                        .setLogo(R.drawable.logo)
                        .setAvailableProviders(getProviders())
                        .setTosUrl(getTosUrl())
                        .setPrivacyPolicyUrl(getPrivacyPolicyUrl())
                        .setIsSmartLockEnabled(false)
                        .setAllowNewEmailAccounts(false)
                        .build(),
                RC_SIGN_IN);
````

Log output:
======================================================

```
Class not found when unmarshalling: com.firebase.ui.auth.ui.FlowParameters
                                         java.lang.ClassNotFoundException: com.firebase.ui.auth.ui.FlowParameters
                                             at java.lang.Class.classForName(Native Method)
                                             at java.lang.Class.forName(Class.java:400)
                                             at android.os.Parcel.readParcelableCreator(Parcel.java:2517)
                                             at android.os.Parcel.readParcelable(Parcel.java:2471)
                                             at android.os.Parcel.readValue(Parcel.java:2374)
                                             at android.os.Parcel.readArrayMapInternal(Parcel.java:2727)
                                             at android.os.BaseBundle.unparcel(BaseBundle.java:269)
                                             at android.os.BaseBundle.getString(BaseBundle.java:992)
                                             at android.content.Intent.getStringExtra(Intent.java:6702)
                                             at com.android.server.am.ActivityStarter.startActivityLocked(ActivityStarter.java:336)
                                             at com.android.server.am.ActivityStarter.startActivityMayWait(ActivityStarter.java:1094)
                                             at com.android.server.am.ActivityManagerService.startActivityAsUser(ActivityManagerService.java:5836)
[...]

PR #929, opened 2017-09-28
Update and fix dependencies (cough cough, Facebook 😁)

SUPERCILEX:
See #926 

@samtstern I'm targeting `v2.3.1`... Do you still plan on releasing that? Or are we moving on to 3.0?

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -44,18 +44,18 @@ dependencies {
     compile "com.google.firebase:firebase-database:$firebaseVersion"
     compile "com.google.firebase:firebase-storage:$firebaseVersion"
 
-    compile('com.facebook.android:facebook-android-sdk:4.25.0')
+    compile('com.facebook.android:facebook-android-sdk:4.27.0')
     compile("com.twitter.sdk.android:twitter-core:3.0.0@aar") { transitive = true }
 
     // The following dependencies are not required to use the Firebase UI library.
     // They are used to make some aspects of the demo app implementation simpler for
     // demonstrative purposes, and you may find them useful in your own apps; YMMV.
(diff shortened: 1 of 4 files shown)


samtstern:
@SUPERCILEX this is the correct branch, may try and get a `2.4.0` out, `2.3.1` would not be the right version anymore since we added some functionality (default phone number args)

samtstern:
@SUPERCILEX what do we get by upgrading to `4.27.0`?

SUPERCILEX:
@samtstern Oh, yeah gotya. 😄 

> What do we get by upgrading to `4.27.0`?

Besides some bug fixes, we mostly benefit from Facebook's new module system to reduce APK size, hence why they broke compatibility. I'm guessing you're think of pushing this to 3.0 since it breaks compatibility for people who don't update the Facebook lib?

samtstern:
@SUPERCILEX does it break compatibility?  If someone consuming this specifies a lesser version Gradle will upgrade to ours.

SUPERCILEX:
@samtstern Not anymore, right? We're using provided deps for the auth stuff:
https://github.com/firebase/FirebaseUI-Android/blob/9dc4c120f33d4f4acc42b2a0c2195e130e8b1076/auth/build.gradle#L54-L55

And I'm pretty sure our Gradle deps don't touch the consumer's because we're using `provided`.

samtstern:
Ohhh yeah.  Well I think we can push this to 3.0.0 to be super-duper-safe.  Even though I bet it would be fine either way.

SUPERCILEX:
@samtstern K, I changed the base to 3.0.

Issue #941, opened 2017-10-03
Having trouble getting com.firebaseui:firebase-ui-firestore:3.0.0


bagintz:
When trying to use
`compile "com.firebaseui:firebase-ui:3.0.0"`
I get 
Error:Failed to resolve: firebaseui-android:common:unspecified

After trying to refresh gradle dependencies, I can see this too

Error:Failed to resolve: com.firebaseui:firebase-ui-firestore:3.0.0

Either way, I can't get it to work :(

bagintz:
https://jcenter.bintray.com/com/firebaseui/

Doesn't show the Firestore dir

samtstern:
D'oh!  There are a few things here.  First of all, I need to remove the dep on common from the POM.

Second, I forgot that because this is a new lib I have to get permission to add to jCenter.  Should be resolved by the end of the day.  For now I think this should work:

```groovy
repositories {
    maven {
      url 'https://dl.bintray.com/firebaseui/firebase-ui'
    }
}
```

samtstern:
Also @bagintz we really don't recommend ever using the combined target, use `-auth`, `-storage`, etc to get just the bits you need!

bagintz:
@samtstern will do, thank you!!

SUPERCILEX:
Yep, I'm on it.

SUPERCILEX:
Whoops, GitHub didn't update. Closing the PR.

bagintz:
@SUPERCILEX and @samtstern I am still getting 
`Error:Failed to resolve: firebaseui-android:common:unspecified`
with 3.0.0, both the combined target and separate

I tried both with the 
```
repositories {
    maven {
      url 'https://dl.bintray.com/firebaseui/firebase-ui'
    }
}
```
and without.  Any ideas?

samtstern:
@bagintz running this from my project root seemed to fix it for me:

```
./gradlew --refresh-dependencies clean cleanBuildCache
```

bagintz:
sigh, still not working for me :(, going to try a few different things

bagintz:
@samtstern ok, finally got it, had to clear out my .gradle folders

simmyy:
I used       compile "com.google.firebaseui:firebase-ui:3.0.0" and 
I'm getting
Error:Failed to resolve: com.google.firebaseui:firebase-ui:3.0.0

samtstern:
@simmyy it's `com.firebaseui` not `com.google.firebaseui`

PR #942, opened 2017-10-03
Fix DB and Firestore modules not working without common module

SUPERCILEX:
#941

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -29,8 +29,9 @@ android {
 
 dependencies {
     compile project(':auth')
-    compile project(':database')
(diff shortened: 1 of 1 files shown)


SUPERCILEX:
@samtstern Hmmm, JitPack is broken without the common module. I have no idea how the POM stuff works, but does just adding the common module to the library work? Or are you working on another solution?

PS: if the firebase-ui dep is broken without the common module, I think we should either kill it or publish the common module.

SUPERCILEX:
Nevermind, I've seen #944.

PR #946, opened 2017-10-05
Localizing verification messages

jstimes:
This change will localize SMS & email verification messages to match the Locale currently in use on the user's device. iOS and Web are adding this as well.


SUPERCILEX:
@isaisachen See #950

isaisachen:
Thanks @SUPERCILEX !

PR #968, opened 2017-10-12
Accept android licenses

samtstern:


samtstern:
@SUPERCILEX have you seen this issue on any of your projects?  Builds are stalling at license acceptance.  @kroikie just fixed it on `firebase/quickstart-android` but not working here.

hatboysam:
Actually seems to be working....

SUPERCILEX:
@samtstern Since this is working, LGTM!

I'm curious though, does it still work if we remove this:
https://github.com/firebase/FirebaseUI-Android/blob/ccc8a3385031af32c2fc579fbb91e263c37f9e93/.travis.yml#L13-L16

The more we can move away from Travis managed dependencies, the better IMO since they break stuff constantly.

samtstern:
@SUPERCILEX yep works without that!  Nice

Issue #985, opened 2017-10-18
NPE when dismissing loading dialog


abdulwasae:
The `dismissLoadingDialog()` method in PhoneVerificationActivity throws NPE at `            mProgressDialog.dismissAllowingStateLoss();
`

Here is the stacktrace:

`arrow_drop_down
Exception java.lang.NullPointerException: Attempt to invoke virtual method 'android.support.v4.app.FragmentTransaction android.support.v4.app.FragmentManager.beginTransaction()' on a null object reference
android.support.v4.app.DialogFragment.dismissInternal (DialogFragment.java:200)
android.support.v4.app.DialogFragment.dismissAllowingStateLoss (DialogFragment.java:181)
com.firebase.ui.auth.ui.phone.PhoneVerificationActivity.dismissLoadingDialog (PhoneVerificationActivity.java:398)
com.firebase.ui.auth.ui.phone.PhoneVerificationActivity.access$000 (PhoneVerificationActivity.java:55)
com.firebase.ui.auth.ui.phone.PhoneVerificationActivity$1.run (PhoneVerificationActivity.java:201)
android.os.Handler.handleCallback (Handler.java:739)
android.os.Handler.dispatchMessage (Handler.java:95)
android.os.Looper.loop (Looper.java:135)
android.app.ActivityThread.main (ActivityThread.java:5254)
java.lang.reflect.Method.invoke (Method.java)
java.lang.reflect.Method.invoke (Method.java:372)
com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:903)
com.android.internal.os.ZygoteInit.main (ZygoteInit.java:698)`

SUPERCILEX:
@abdulwasae Yep, we're [rewriting FirebaseUI](https://github.com/firebase/FirebaseUI-Android/pull/962) to fix the fundamental issues behind this.

@samtstern Lol, told ya it wasn't going to work. We tried so hard though! 😆 You could add a null check for the `FragmentManager` but it's kinda hopeless at this point.

quantumInfection:
@abdulwasae sitting at office. Fixing useless bugs.

ismail-khan:
@samtstern, @SUPERCILEX, any visibility on when this change will be released? Have a bunch of apps in production that are seeing this problem :(

samtstern:
@ismail-khan no fixed deadlines right now, we want to get a few other fixes in `3.1.1`.

[...]

PR #986, opened 2017-10-18
Fix NPE when dismissing loading dialog (#985)

abdulwasae:
Issue #985


abdulwasae:
I signed it!

samtstern:
@abdulwasae thanks for the contribution!  Are you able to reproduce the NPE and prove that this fixes it?  If so could you provide some steps for me to try?

abdulwasae:
@samtstern This does not happen always, but quite often. Steps that could reproduce this are as simple as using the phone verification feature again and again.

I have an app in production that utilises firebase phone verification, and a lot of the users are facing this NPE i posted above.

samtstern:
@abdulwasae can't hurt!  I will merge this into the `3.1.1-dev` branch so it will be part of that release.  Thanks for contributing to FirebaseUI!

PR #991, opened 2017-10-22
Fix adapter inconsistency crashes

SUPERCILEX:
@samtstern Fixes the same issue we had with the RTDB adapter: if a `FirestoreArray` has multiple listeners and the adapter stops listening, it isn't cleared properly so trying to add items again in `startListening` causes `IndexOutOfBoundsException: Inconsistency detected` errors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,10 @@ cache:
     - $HOME/.gradle/caches/
     - $HOME/.gradle/wrapper/
     - $HOME/.android/build-cache
+before_install:
+  - mkdir "$ANDROID_HOME/licenses" || true
+  - echo -e "\n8933bad161af4178b1185d1a37fbf41ea5269c55" > "$ANDROID_HOME/licenses/android-sdk-license"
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -10,6 +10,10 @@ cache:
     - $HOME/.gradle/caches/
     - $HOME/.gradle/wrapper/
     - $HOME/.android/build-cache
+before_install:
+  - mkdir "$ANDROID_HOME/licenses" || true
+  - echo -e "\n8933bad161af4178b1185d1a37fbf41ea5269c55" > "$ANDROID_HOME/licenses/android-sdk-license"
+  - echo -e "\n84831b9409646a918e30573bab4c9c91346d8abd" > "$ANDROID_HOME/licenses/android-sdk-preview-license"
```
SUPERCILEX: Turns out we _do_ need these licenses. 😢 I can split this out into another PR if necessary.

samtstern:
Nice catch, thanks1

PR #999, opened 2017-11-06
Improve Kotlin support

SUPERCILEX:
@samtstern This PR starts what will hopefully be an eventual transition to fully supporting Kotlin and then writing FirebaseUI in Kotlin. For starters, I've added nullability annotations to every single public API and also fixed some bad docs I noticed along the way.


common/src/main/java/com/firebase/ui/common/Adapter.java
```diff
@@ -0,0 +1,35 @@
+package com.firebase.ui.common;
+
+import android.arch.lifecycle.LifecycleObserver;
+import android.support.annotation.NonNull;
+import android.support.annotation.RestrictTo;
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+public interface Adapter<O extends BaseObservableSnapshotArray<S, E, ?, T>, S, E, T>
+        extends BaseChangeEventListener<S, E>, LifecycleObserver {
```
SUPERCILEX: We completely missed documentation for the `FirestoreRecyclerAdapter` so I feel pretty strongly towards having a base `Adapter` interface that we can inherit from to get shared documentation even with those ugly generics. @samtstern What do you think?

database/src/main/java/com/firebase/ui/database/FirebaseAdapter.java
```diff
@@ -1,26 +1,25 @@
...
 import com.google.firebase.database.DatabaseReference;
 
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-interface FirebaseAdapter<T> extends ChangeEventListener, LifecycleObserver {
+interface FirebaseAdapter<T> extends Adapter<ObservableSnapshotArray<T>,
+        DataSnapshot, DatabaseError,
+        T>, ChangeEventListener {
     /**
-     * If you need to do some setup before the adapter starts listening for change events in the
-     * database, do so it here and then call {@code super.startListening()}.
```
SUPERCILEX: The documentation in this class was all wrong! 😱 `startListening` is no longer called in the constructor, `stopListening` doesn't necessarily clear the array, and we were missing documentation for the other methods.

common/src/main/java/com/firebase/ui/common/Adapter.java
samtstern: Just looking at this class ... I can't imagine it meaning anything to anyone but the two of us.  Let's just solve the firestore docs issue by adding those docs directly, this abstraction doesn't seem necessary since we don't take `Adapter` as an argument to anything.

If you feel strongly about this it should be a separate PR anyway, this should stay focused on the `Null`/`NonNull` stuff for Kotlin.

database/src/main/java/com/firebase/ui/database/FirebaseAdapter.java
samtstern: Thanks for catching this!

samtstern:
@SUPERCILEX thanks for doing all this!  Is there some Android Studio refactor helper that assists the `Nullable` / `NonNull` adding or did you just do a manual sweep?

Also let's keep this PR simple and just cut it down to the annotation stuff, we can work on the docs issues you found in another PR.

common/src/main/java/com/firebase/ui/common/Adapter.java
SUPERCILEX: @samtstern I do prefer clarity over minimizing copypastas... since we both agree it's ugly and you think other people won't know what's going on, I'm happy to just copypasta. 😄

SUPERCILEX:
@samtstern Lol, sadly no. That's why I found documentation issues—I had to go through all of our public API... 😭😆

Yep, moving the docs to another PR.

samtstern:
@SUPERCILEX well thanks for going through all that!  Kotlin devs will thank you.

PR #1000, opened 2017-11-07
Update build.gradle

Gregcake:
Changed:
android.support.constraint-layout to follow architechtureVersion const


SUPERCILEX:
That's incorrect: those are two different libraries.

PR #1004, opened 2017-11-10
Fix broken publishing

SUPERCILEX:
@samtstern I was trying to get my custom build from JitPack, but it kept failing. I'm not sure if this is just an issue on my side, but this PR basically ensures all publishing tasks first build the modules. (I think the 3.0 gradle plugin might have broken the order somewhere.)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -45,6 +45,11 @@ allprojects { project ->
     def isSubmodule = submodules.contains(project.name)
 
     if (isLibrary || isSubmodule) {
+        tasks.whenTaskAdded { task ->
+            if (task.name.contains("publish") && task.name.contains("PublicationToMavenLocal")) {
+                task.dependsOn 'assemble'
+            }
+        }
 
         // So that we can resolve 'android' variable
         project.apply plugin: 'com.android.library'
--- a/build.gradle
+++ b/build.gradle
@@ -46,7 +46,7 @@ allprojects { project ->
 
     if (isLibrary || isSubmodule) {
         tasks.whenTaskAdded { task ->
-            if (task.name.contains("publish") && task.name.contains("PublicationToMavenLocal")) {
+            if (task.name.contains("publish") && task.name.toLowerCase().contains("publication")) {
                 task.dependsOn 'assemble'
             }
         }


build.gradle
```diff
@@ -45,6 +45,11 @@ allprojects { project ->
     def isSubmodule = submodules.contains(project.name)
 
     if (isLibrary || isSubmodule) {
+        tasks.whenTaskAdded { task ->
+            if (task.name.contains("publish") && task.name.contains("PublicationToMavenLocal")) {
```
samtstern: 1) I normally use the `library:prepareArtifacts` task before publishing, maybe we should add the assemble dependency there?

2) What about non-local publishing?

SUPERCILEX: @samtstern Nope, that didn't work for some reason. I deleted the above lines and replaced them with `publishToMavenLocal.dependsOn 'prepareArtifacts'` but that fails (BTW, `prepareArtifacts` already has a dependency on `assembleRelease`). The problem is that I don't control the tasks JitPack runs. If you'd rather keep these changes separate from the main repo since they seem to be a JitPack-ish issue, I'm happy to do so. Otherwise, grabbing the `PublicationToMavenLocal` deps is the best solution I came up with.

samtstern: Maybe something a little more general?  Like if the task name (lowercase) contains `"publication"` or `"bintray"`?

SUPERCILEX: @samtstern I'm trying to collect all [these tasks](https://github.com/SUPERCILEX/FirebaseUI-Android/blob/a8d9cfc5ddff444c3997aadf3d912c0cdd044afa/build.gradle#L117) which have the format `publish[Foo]PublicationToMavenLocal`. I could leave it at `task.name.contains("publish")`, but I wasn't sure if that would mess other stuff up. What do you think?

samtstern: I also think this would be useful for when I publish to non-local (for releases).  I think anything that includes both "publish" and "publication" would be a good heuristic.

SUPERCILEX:
@samtstern SGTM! Is this what you were looking for?
```groovy
task.name.contains("publish") && task.name.toLowerCase().contains("publication")
```

samtstern:
Yep!

SUPERCILEX:
Sweet!

PR #1006, opened 2017-11-13
Update to architecture components 1.0.0

samtstern:
cc @SUPERCILEX @CodingDoug

### DIFF
--- a/constants.gradle
+++ b/constants.gradle
@@ -10,5 +10,5 @@ project.ext {
 
     firebaseVersion = '11.4.2'
     supportLibraryVersion = '26.1.0'
-    architectureVersion = '1.0.0-rc1'
+    architectureVersion = '1.0.0'
 }


CodingDoug:
When I tried this in my own sandbox, I got lint errors on code that was generated by the architecture components annotation processor.  Did that not happen here?

SUPERCILEX:
Do you support Java 8? If so, you shouldn't be using the annotation processor because there's a interface with a bunch of empty defaults.

CodingDoug:
Not sure what you mean.  I'm talking about this annotation processor:
    annotationProcessor "android.arch.lifecycle:compiler:$architectureVersion"

That generates code which ./gradlew check complains about.

SUPERCILEX:
See https://developer.android.com/topic/libraries/architecture/adding-components.html.

```groovy
implementation "android.arch.lifecycle:common-java8:1.0.0"
```

SUPERCILEX:
Oh, I see. No, I haven't seen that with FirebaseUI, sorry! 😊

CodingDoug:
One other minor thing - there should be no need to include the runtime dependency here when using the extensions dependency.  extensions depends on runtime (which currently sits at 1.0.3).

Issue #1007, opened 2017-11-14
[Refactor] Move to GoogleSignInClient and CredentialsClient, remove GoogleApiClient


samtstern:
These are new in `11.6.0`:
https://developers.google.com/android/reference/com/google/android/gms/auth/api/credentials/CredentialsClient
https://developers.google.com/android/reference/com/google/android/gms/auth/api/signin/GoogleSignInClient

Moving to use these classes will allow better chaining (through `Task`) remove any `GoogleApiClient` helper classes and probably make things a lot simpler.

It's also worth nothing the one issue in the release:
"GoogleApi instances created with a Context in apps with targetSdkVersion 26 won't automatically prompt users to update Google Play services. Consider using GoogleApiAvailability.showErrorDialogFragment() under these conditions instead."

So we should make sure we have play services before starting the flow.  Not sure if we do that already.

samtstern:
(this is the beginning of breaking #962 up into issues which we can tackle one at a time)

samtstern:
@SUPERCILEX I'd like to take a crack at this but not sure how badly that would mess you up.

SUPERCILEX:
@samtstern I was actually planning to already start that refactor after the cleanup PRs are merged. Since it looks like that might happen tomorrow, we could already get that going by Friday. I think if it doesn't happen over the weekend at the latest, then yeah, you could start doing it. (Though it definitely will destroy my stuff since I deleted all the original Smart Lock files. 😄)

samtstern:
I don't want to do that, although I would like to find ways to contribute to this rewrite besides reviewing PRs.  This is one topic I understand particularly well, but maybe it's not the right place.

The breaking down of the rewrite is going well so far but it still kind of all has the shadow of the big rewrite PR hanging over it.  I wonder if we could break some of the tasks down into issues, agree on the APIs (even internal APIs) and then share the authoring load.

SUPERCILEX:
@samtstern Yeah, I totally agree with you, it would be awesome to work together on this and improve on my original ideas. 😄 Once that last refactor PR goes through, I was planning on opening a small and basic setup PR and use it in a PR refactoring the password reset activity with our new architecture. The plan was to use the password reset PR to dip our feet in the water and let me show you how I had done the other refactors.

Do you think I should follow through with that plan and then start opening a bunch of issues once you've seen the refactor style, or do you want to open the issues first and then do a "dip your feet in the water" PR?

samtstern:
[...]

Issue #997, opened 2017-11-02
Signout from Fragment or non-activity class


kidokimo:
Hi There, 

I am trying to use this sign out method just to be able to add an OnCompleteListener to the sign out process But it asks for a FragmentActivity and i am calling it from inside a fragment (this fragment is called by an AppCompatAcivity).. Here is the code

     AuthUI.getInstance() .signOut(getActivity()) // this is the issue
                            .addOnCompleteListener(new OnCompleteListener<Void>() {
                                public void onComplete(@NonNull Task<Void> task) {
                                    // user is now signed out
                                  
                                }
                            });

All what i need is one of the following:
1. How to pass the correct context knowing that i am calling this method inside a Fragment (this fragment is called by an AppCompatActivity)
2. is there any other method to have an OnCompleteListener to the sign out process from inside a fragment or a non-activity class ?

Thanks in advance

SUPERCILEX:
`AppCompatActvity` actually extends `FragmentActivity` so everything should compile just fine. Is something not working?

SUPERCILEX:
@samtstern @kidokimo This is fixed in FUI `v3.1.1`. Enjoy!

Issue #975, opened 2017-10-16
onActivityResult not triggered


CyberpunkPsychonaut:
I have a login activity calling 

`
    startActivityForResult(
                    AuthUI.getInstance()
                            .createSignInIntentBuilder()
                            .setIsSmartLockEnabled(!BuildConfig.DEBUG)
                            .setTheme(R.style.AuthUITheme)
                            .setLogo(R.drawable.logo_hi_res)
                            .setAvailableProviders(
                                    Arrays.asList(
                                            new AuthUI.IdpConfig.Builder(AuthUI.GOOGLE_PROVIDER).build(),
                                            new AuthUI.IdpConfig.Builder(AuthUI.EMAIL_PROVIDER).build(),
                                            new AuthUI.IdpConfig.Builder(AuthUI.PHONE_VERIFICATION_PROVIDER).build()))
                            .build(),
                    RC_SIGN_IN);
`

but none of the providers trigger @Override protected void onActivityResult(int requestCode, int resultCode, Intent data)..

Instead the activity simply acts as if it is recreated and goes through onCreate -> onStart after signing in from any of the providers

    implementation("com.firebaseui:firebase-ui-auth:3.0.0") {
        exclude module: 'support-v4'
    }

CyberpunkPsychonaut:
Any help here?

SUPERCILEX:
Do you have code that kills the activity in `onCreate` if the user is signed in? And are you calling `super.onActivityResult(...)` in your activities and fragments?

SUPERCILEX:
@dcalano I just had a revelation: are you finishing your activity in `on[Create/Start/Resume]` if the user is logged in? If so, check out https://github.com/firebase/FirebaseUI-Android/pull/1010#discussion_r151575726 which I still need to find a solution to.

SUPERCILEX:
@samtstern There hasn't been a response here in a while so we can probably close this issue.

rkuang:
Having this problem on v4.1.0. I've got a bunch of logs in onActivityResult that aren't being printed.

Additionally, whenever I press back from the Auth activity, a UserCancellationException gets thrown and I'm not sure what I should be doing there.

OyaCanli:
I'm having the same issue on the emulator with Android Pie. onActivityResult is not called for "cancel" case. Though the same code works fine on my real device which has Android Oreo.

itrevex:
[...]

PR #1010, opened 2017-11-16
Rewrite relevant parts of the public API

SUPERCILEX:
Built on top of #1009, starts work on #1007, `fixes` #997, preps for a future `fixing` #526.

This PR fixes some bugs in the sample app, fixes a bug where deleting a user doesn't sign it out from the idp providers, and migrates `IdpResponse` to using exceptions for its errors (#526).


auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -189,40 +195,27 @@ public static int getDefaultTheme() {
...
-        try {
-            LoginManager.getInstance().logOut();
-        } catch (NoClassDefFoundError e) {
-            // do nothing
+        CredentialsClient client;
+        if (context instanceof Activity) {
+            client = Credentials.getClient((Activity) context);
+        } else {
+            client = Credentials.getClient(context);
         }
```
SUPERCILEX: I looked at the PS code and it doesn't do this check for us which is too bad. We want to make sure to show a `Fix this Play Services problem` dialog if possible instead of just showing a notification.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -231,48 +224,87 @@ public static int getDefaultTheme() {
...
+
+        // Google sign out
+        return GoogleSignIn.getClient(context, GoogleSignInOptions.DEFAULT_SIGN_IN).signOut();
+    }
+
+    /**
+     * Make a list of {@link Credential} from a FirebaseUser. Useful for deleting Credentials, not
+     * for saving since we don't have access to the password.
+     */
+    private static List<Credential> credentialsFromFirebaseUser(@NonNull FirebaseUser user) {
```
SUPERCILEX: Moved this from that smart Lock class since it's going to die later.

auth/src/main/java/com/firebase/ui/auth/data/model/NetworkException.java
```diff
@@ -0,0 +1,15 @@
...
+
+import android.support.annotation.NonNull;
+
+import com.firebase.ui.auth.ErrorCodes;
+import com.google.firebase.auth.FirebaseAuthException;
+
+/**
+ * Represents an error connecting to the internet.
+ */
+public class NetworkException extends FirebaseAuthException {
```
SUPERCILEX: I'm not sure if we want to extend from Firebase auth exceptions or just do our own.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -231,48 +224,86 @@ public static int getDefaultTheme() {
...
         // Delete the Firebase user
-        Task<Void> deleteUserTask = firebaseUser.delete();
+        Task<Void> deleteUserTask = currentUser.delete();
 
         // Get all SmartLock credentials associated with the user
-        List<Credential> credentials = SmartLockBase.credentialsFromFirebaseUser(firebaseUser);
+        List<Credential> credentials = credentialsFromFirebaseUser(currentUser);
 
+        CredentialsClient client;
+        if (context instanceof Activity) {
```
samtstern: We're repeating this block, so we should probably move it out into a static method somewhere like:

```java
public static getCredentialsClient(@NonNull Context context) {
  if (context instanceof Activity) {
    return Credentials.getClient((Activity) context);
  } else {
    return Credentials.getClient(context);
  }
}
```

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -231,48 +224,86 @@ public static int getDefaultTheme() {
...
 
-        // Chain the Firebase Auth delete task with the combined Credentials task
-        // and return.
-        return deleteUserTask.continueWithTask(new Continuation<Void, Task<Void>>() {
-            @Override
-            public Task<Void> then(@NonNull Task<Void> task) throws Exception {
-                // Call getResult() to propagate failure by throwing an exception
-                // if there was one.
-                task.getResult(Exception.class);
+        return Tasks.whenAll(deleteUserTask, signOutIdps(context), combinedCredentialTask);
```
samtstern: This is a slight behavior change.  Before we tried to propagate user deletion failure exceptions explicitly by checking the result of that task first.  We used `continueWithTask` not `whenAll`.

Looking at this now, it may also be the case that it would be better to wait for the user `delete` to succeed before doing the credential deletion, but I don't feel strongly about that since I have never seen a `delete` fail.

auth/src/main/java/com/firebase/ui/auth/IdpResponse.java
```diff
@@ -33,20 +34,25 @@
...
+
     private final String mToken;
     private final String mSecret;
 
-    private final int mErrorCode;
+    @Nullable private final Exception mException;
 
-    private IdpResponse(int errorCode) {
-        this(null, null, null, errorCode);
+    private IdpResponse(Exception e) {
```
samtstern: Should this be `@NonNull`?

auth/src/main/java/com/firebase/ui/auth/IdpResponse.java
```diff
@@ -75,6 +96,11 @@ public Intent toIntent() {
     }
 
     @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+    public boolean isSuccessful() {
+        return mException == null;
```
samtstern: Should this check if exception is null **and** user is non-null?

auth/src/main/java/com/firebase/ui/auth/IdpResponse.java
```diff
@@ -124,7 +150,19 @@ public String getIdpSecret() {
      * Get the error code for a failed sign in
      */
     public int getErrorCode() {
-        return mErrorCode;
+        if (isSuccessful()) {
+            return Activity.RESULT_OK;
```
samtstern: Was it possible for this method to return `RESULT_OK` before?  Seems poorly named as `getErrorCode` since it's really `getResultCode`.

However if this was possible before it's not worth breaking the API to fix it.

auth/src/main/java/com/firebase/ui/auth/IdpResponse.java
```diff
@@ -33,20 +34,25 @@
  */
 public class IdpResponse implements Parcelable {
     private final User mUser;
+
     private final String mToken;
     private final String mSecret;
 
-    private final int mErrorCode;
+    @Nullable private final Exception mException;
```
samtstern: Maybe instead of having a raw `Exception` class here we could have `FirebaseUIException` that has a "cause" (the underlying `Exception`) and an `ErrorCode` field.

Then we wouldn't have to check `instanceof NetworkException` later and it gives us more room to expand.

auth/src/main/java/com/firebase/ui/auth/IdpResponse.java
```diff
@@ -179,20 +218,23 @@ public Builder setSecret(String secret) {
...
         public IdpResponse build() {
             String providerId = mUser.getProviderId();
-            if ((providerId.equalsIgnoreCase(GoogleAuthProvider.PROVIDER_ID)
-                    || providerId.equalsIgnoreCase(FacebookAuthProvider.PROVIDER_ID)
-                    || providerId.equalsIgnoreCase(TwitterAuthProvider.PROVIDER_ID))
+            if (!AuthUI.SUPPORTED_PROVIDERS.contains(providerId)) {
+                throw new IllegalStateException("Unknown provider: " + providerId);
+            }
+            if ((providerId.equals(GoogleAuthProvider.PROVIDER_ID)
+                    || providerId.equals(FacebookAuthProvider.PROVIDER_ID)
```
samtstern: Maybe we should make `AuthUI.SOCIAL_PROVIDERS` or something so we can keep this list in one place in case it grows in the future?

auth/src/main/java/com/firebase/ui/auth/data/model/NetworkException.java
samtstern: We should do our own, see comment above about how it might look.  Extending `FirebaseAuthException` doesn't really get us anything and is strange if we're propagating errors that might come from within our framework or a non-Firebase SDK.

app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
```diff
@@ -148,12 +149,20 @@ public void onCreate(Bundle savedInstanceState) {
...
-        FirebaseAuth auth = FirebaseAuth.getInstance();
-        if (auth.getCurrentUser() != null) {
-            startSignedInActivity(null);
-            finish();
-            return;
-        }
+        new Handler().post(new Runnable() {
+            @Override
+            public void run() {
+                // We need to post this to the main thread to ensure `onActivityResult` is called
```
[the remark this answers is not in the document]
SUPERCILEX: Yeah, so here's the issue I'm trying to solve: when going to a deeply nested activity (such as being 3 activities deep in the `EmailActivity`), the `AuthUiActivity` is going to die. When it's recreated, `onActivityResult` gets called _after_ `onCreate` which is too late by then because we've already started killing the activity.  I'm just posting this to the main thread as a hack to ensure that the `onActivityResult` message gets delivered first. Does that make sense?

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
SUPERCILEX: Yep, done.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
SUPERCILEX: Ohhhhhh, I see what was going on there: it prevents that "1 out of n underlying tasks failed" exception. Devs can still get the underlying exception with a `getCause`, but would you like to restore previous behavior?

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
samtstern: Don't we also have the same code to get a client in the `signOut()` method?  That's why I think it should be refactored out into a standalone getClient method that we can use in both places.

app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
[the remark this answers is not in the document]
samtstern: What if we moved to `onResume()`, isn't that always called after `onActivityResult` and `onCreate` have both finished?

auth/src/main/java/com/firebase/ui/auth/IdpResponse.java
SUPERCILEX: There we go, the `FirebaseUiException` thing is a much better way of doing it! In my PR, I have a bunch of exceptions extending Firebase SDK ones which felt really weird.

As for the error code thing, I was kinda hoping to move away from that. Ideally I would like us to make the exceptions part of our public API once they're stabilized so that we can provide specific, actionable errors. (We could check for developer errors and fail fast by returning an exception instead of logging stuff and hoping devs look at it.) Also, I don't want to deal with native Java serialization since we're passing this through activities 😊. What do you think?

auth/src/main/java/com/firebase/ui/auth/IdpResponse.java
SUPERCILEX: Yep, I've added another constructor so our nullability can be clearer.

auth/src/main/java/com/firebase/ui/auth/IdpResponse.java
SUPERCILEX: I don't that's necessary since those two are mutually exclusive. The builder throws an NPE if the user is null and `fromError` always puts in a nonnull exception and a null user.

auth/src/main/java/com/firebase/ui/auth/IdpResponse.java
SUPERCILEX: Nope, I did this to keep things backwards compatible. See https://github.com/firebase/FirebaseUI-Android/pull/1010#discussion_r151534213, I'd rather we not use error codes anyway.

Here's how we did it before (now it's clear how illogical that was 😆):
image

auth/src/main/java/com/firebase/ui/auth/IdpResponse.java
SUPERCILEX: Good idea, done.

auth/src/main/java/com/firebase/ui/auth/data/model/FirebaseUiNetworkException.java
```diff
@@ -0,0 +1,10 @@
+package com.firebase.ui.auth.data.model;
+
+/**
+ * Represents an error connecting to the internet.
+ */
+public class FirebaseUiNetworkException extends FirebaseUiException {
```
samtstern: I think just having a `code` on `FirebaseUiException` is probably better than subclasses.  Checking an `int` or `String` is easier than using `instanceof` and faster too.

I'd generally only use subclasses if extra functionality is needed, like accepting a more advanced parameter.

Wdyt?

SUPERCILEX: Oh, I see. Yep, that makes sense but do you agree that `IdpResponse` should eventually deprecate and remove the error codes getter in favor getting the exception (which has that info and more if relevant)?

SUPERCILEX: Oh sweet, it turns out `Exception` does some black magic to serialize subclass fields too so we don't lose our error code when passing the exception through an `Intent`. 👍

samtstern: Probably yes but for now we expose so little information in errors that we can stick with the code just to keep the API unchanged.

SUPERCILEX: Yes, yes of course. I keep forgetting that my PR was too big for you to take a look at in-depth. There are cool exceptions in that PR such as `MisconfiguredProviderException`, `ExternalProviderError`, etc... we'll be able to provide much more useful information! 😄

SUPERCILEX:
@samtstern I made some (more significant) changes in 87c0ebf: I removed a bunch of comments that were just [restating](https://testing.googleblog.com/2017/07/code-health-to-comment-or-not-to-comment.html) what the code did and I made sure the task order was correct for `delete(...)`. (You said that delete has never failed on you, but I just saw it fail when I came back to my desk and got a credential too old exception! 😆)

app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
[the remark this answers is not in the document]
SUPERCILEX: Ohhhhhhhhh, after a super long time of Googling around and trying to understand what was happening, something clicked and I finally figured it out. Everybody says that `onActivityResult` is guaranteed to be called _before_ `onResume`, but that's never what I've observed. Here's the thing though: the system also guarantees that `onResume` will be called if there's only a transparent activity on top. And guess what? We have a transparent `KickoffActivity`!

Do you remember [all](https://github.com/firebase/FirebaseUI-Android/issues/975) [those](https://github.com/firebase/FirebaseUI-Android/issues/553) [issues](https://github.com/firebase/FirebaseUI-Android/issues/182) where people are like "`onActivityResult` isn't being called" and we're like "well, we know it is." Those people probably copied from the sample and are running into the same "bug!"

I need to see if there's an API to check if an activity is on top of you or not, but otherwise I can't think of a better solution than the post.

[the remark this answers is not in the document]
samtstern: I just ran the sample app and I see `onResume` come after `onActivityResult` in `AuthUiActivity` every time.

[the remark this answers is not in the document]
SUPERCILEX: @samtstern Oh lol, you know what? It's a bug with the developer option "Don't keep activities" I've been using to test this stuff. I can confirm that the expected order of callbacks is correct with that dev option disabled. Anyway, I'm going to report that to AOSP, but glad to know I'm not completely losing my mind.

SUPERCILEX:
@samtstern Ok, just make sure 6002060 makes sense to you and then I think we're ready for a merge! 😄

samtstern:
This LGTM, thanks for the back-and-forth!

SUPERCILEX:
Awesome sauce!

PR #1013, opened 2017-11-16
Cleanup READMEs

SUPERCILEX:
@samtstern I noticed from this morning's PR that we forgot to use `implementation` in our docs instead of `compile`. This PR fixes that and cleans up stuff like how we show which dependencies devs should be using.

Includes a merge from master.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ A compatible FirebaseUI client is also available for [iOS](https://github.com/fi
 
 ## Usage
 
-FirebaseUI has separate modules for using Firebase Realtime Database, Cloud Firestore, 
+FirebaseUI has separate modules for using Firebase Realtime Database, Cloud Firestore,
 Firebase Auth, and Cloud Storage. To get started, see the individual instructions for each module:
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -38,17 +38,17 @@ libraries.
 
 ```groovy
 dependencies {
-    // FirebaseUI for Firebase Realtime Database
-    compile 'com.firebaseui:firebase-ui-database:3.1.0'
-    
-    // FirebaseUI for Cloud Firestore
-    compile 'com.firebaseui:firebase-ui-firestore:3.1.0'
+// FirebaseUI for Firebase Realtime Database
+implementation 'com.firebaseui:firebase-ui-database:3.1.0'
```
samtstern: nit: can we preserve the indentation in the `dependencies` block? Here and in the instance below.

SUPERCILEX: Are you sure? Everywhere else, we 0 indent even though the real code would be indented. Otherwise, the indents get copied which can mess things up when pasting the code in an IDE.

PR #1016, opened 2017-11-17
Rewrite layouts

SUPERCILEX:
This PR wraps every layout in a `ScrollLayout`, flattens all layouts with `ConstraintLayout`, fixes poor keyboard UX, and cleans some stuff up.


auth/src/main/res/layout/fui_phone_layout.xml
```diff
@@ -1,57 +1,53 @@
...
             android:id="@+id/country_list"
-            style="@style/FirebaseUI.CountrySpinner" />
+            style="@style/FirebaseUI.CountrySpinner"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintBottom_toBottomOf="@id/phone_layout" />
+
+        <android.support.design.widget.TextInputLayout
+            android:id="@+id/phone_layout"
+            style="@style/FirebaseUI.TextInputLayout.PhoneField"
+            android:layout_width="0dp"
```
samtstern: Is this normal with constraint layout?

auth/src/main/res/layout/fui_phone_layout.xml
```diff
@@ -1,57 +1,53 @@
...
         style="@style/FirebaseUI.WrapperStyle"
-        android:layout_width="match_parent"
         android:layout_height="wrap_content">
 
         <com.firebase.ui.auth.ui.phone.CountryListSpinner
             android:id="@+id/country_list"
-            style="@style/FirebaseUI.CountrySpinner" />
+            style="@style/FirebaseUI.CountrySpinner"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintBottom_toBottomOf="@id/phone_layout" />
```
samtstern: Before this was baseline aligned, I don't think bottom and baseline are the same.

auth/src/main/res/layout/fui_welcome_back_idp_prompt_layout.xml
```diff
@@ -1,29 +1,34 @@
 <?xml version="1.0" encoding="utf-8"?>
-<LinearLayout
+<ScrollView
```
samtstern: Same question here, this is a relatively "short" view that does not seem in danger of scrolling.

auth/src/main/res/layout/fui_check_email_layout.xml
```diff
@@ -1,33 +1,38 @@
 <?xml version="1.0" encoding="utf-8"?>
-<RelativeLayout
+<ScrollView
```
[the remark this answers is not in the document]
piotrek1543: I think It would be good to remove ScrollView and change LinearLayout with ConstraintLayout as you've already done in layouts below.

I'm pretty sure that adding `ScrollView` like adding any other layer of UI  to xml file would finally decrease app performance. Although in that situation, where layout is a relatively small, it would be pretty hard to notice. 

As @samtstern said, there's no reason to use `ScrollView` there, as it won't more space on small device.

auth/src/main/res/layout/fui_forgot_password_layout.xml
```diff
@@ -28,19 +28,11 @@
...
-            android:gravity="right"
-            tools:ignore="RtlHardcoded">
-
-            <Button
-                android:id="@+id/button_done"
-                style="@style/FirebaseUI.Button"
-                android:text="@string/fui_button_text_send" />
-
-        </LinearLayout>
+        <Button
```
piotrek1543: Good one!

auth/src/main/res/layout/fui_welcome_back_idp_prompt_layout.xml
piotrek1543: @SUPERCILEX have you noticed a problem with partially visible layout on any device? For me, it would be only reason to keep here `ScrollView`. If no, please remove it as it would decrease the app performance and increase app building process.

auth/src/main/res/layout/fui_confirmation_code_layout.xml
```diff
@@ -6,100 +6,106 @@
...
             android:layout_height="wrap_content"
-            android:layout_below="@id/dgts_titleText"
             android:textAppearance="?android:attr/textAppearanceMedium"
             android:textColor="@color/fui_linkColor"
+            app:layout_constraintTop_toBottomOf="@id/header"
             tools:ignore="SelectableText" />
 
-        <com.firebase.ui.auth.ui.phone.SpacedEditText
-            android:id="@+id/confirmation_code"
+        <android.support.design.widget.TextInputLayout
```
piotrek1543: @SUPERCILEX did you tested on any device with API lower than Lollipop? `TexInputLayout` is pretty UI improvement, bu it may looks weird on older devices :-)

auth/src/main/res/layout/fui_phone_layout.xml
SUPERCILEX: Yep, it means [`match_constraint`](https://developer.android.com/reference/android/support/constraint/ConstraintLayout.html#DimensionConstraints) which is essentially "go as far as your constraints let you" (or `match_parent` in this case since we don't have any side constraints).

auth/src/main/res/layout/fui_phone_layout.xml
SUPERCILEX: Good catch! 😄 When an error appears, the spinner follows the error which is wrong. I have to play around to see how to fix that.

auth/src/main/res/layout/fui_check_email_layout.xml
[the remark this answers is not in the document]
SUPERCILEX: @samtstern In an extreme case where a user has a small screen and is in multi-window mode, they could get to the point where they need to scroll:


That was on my Pixel 2 XL which has a pretty big aspect ratio so I can image the need to scroll on a normal screen. Then again, this is a pretty rare edge case.

@samtstern @piotrek1543 As for performance concerns, I've Googled around and didn't find anything. I think since it's just one top-level layout, performance is basically irrelevant. (The activity will take _way_ longer to start than the extra onMeasure calls from using a scrollview.)

But as I said, I'm not going to be nit picky here so it's your call @samtstern. 😄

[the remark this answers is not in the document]
samtstern: LGTM thanks for justifying.

PR #1017, opened 2017-11-17
Formalize new package structure and cleanup

SUPERCILEX:
This PR reorganizes our package structure into something that will scale with the PRs coming in the future. It also does some relatively minor cleanup. (The large number of LoC changed are because of a single phone class.)


SUPERCILEX:
Here's what the new structure looks like visually:
- data
  - client
  - model
  - remote
- ui
  - email
  - idp
  - phone
  - provider (Future providers will go here. The existing ones do a bit of everything.)
- util
  - data
  - ui
    - fieldvalidators

auth/src/main/java/com/firebase/ui/auth/util/data/PhoneNumberUtils.java
```diff
@@ -0,0 +1,658 @@
+package com.firebase.ui.auth.util.data;
```
SUPERCILEX: Here's the diff for this class: http://www.mergely.com/v0dlTATS/. Fixing it was _sooooo_ painful. I literally sat down with the ADB podcast and spent an hour simplifying it until I went insane and discovered this [unbelievably awesome IntelliJ feature](https://www.jetbrains.com/help/idea/structural-search-and-replace.html)—huge cudos to the JetBrains team. Anyway, how can I help you understand the diff?

SUPERCILEX: @samtstern Ok, I killed the `PhoneNumberUtils` changes and will move them to another PR so this diff can stay sane. 👍

samtstern:
@SUPERCILEX we can look at this one next if you want to.  But let's keep it to renames only.  If we're making any code changes (even formatting changes) let's do those in another PR.

SUPERCILEX:
@samtstern SGTM, done.

samtstern:
LGTM.

PR #1018, opened 2017-11-17
Remove boilerplate license headers

SUPERCILEX:
This is just an itty bitty nit I have, dunno if you want to go through with it. The Apache license doesn't require a header and we haven't been using one in new files for ages now, so I just wanted to make our code consistent and get rid of our current headers (except for the twitter code which has the special modifications license header). What do you think?


samtstern:
@SUPERCILEX we actually can't do this :-( technically we are running afoul of Google policy by not adding the header to new files.  So if we do anything about this we should probably use the Android Studio license adder thing to fix it across the board.

But let's save all that code churn for a more quiet time.

SUPERCILEX:
Haha, ok thanks for the info! 😄

PR #1019, opened 2017-11-17
Fix oversight introduced in #1010

SUPERCILEX:
@samtstern Oops! 😊

The code should explain what I mean and prevent future devs from making the same mistake.


samtstern:
Oops I should have caught this, thanks!

PR #1020, opened 2017-11-17
Fix user deletion possibly using wrong Auth instance

SUPERCILEX:
@samtstern Apparently, this bug has been in there [since the beginning](https://github.com/firebase/FirebaseUI-Android/commit/afbf1ed02c0f3477eb5cd43d895388e7a9dbabea#diff-3007df2d406e83be998849527d41ccedR322), but I just noticed it. Bug fixes! 🐛😄


samtstern:
SGTM, nice catch!

PR #1023, opened 2017-11-18
Phone utils rewrite

SUPERCILEX:
Moved (and built) from #1017


samtstern:
This PR seems to rely on the package-reorganization one so I will wait until that one happens before diving into this diff.

samtstern:
Ok let's rebase this one on top of what's done already to get a clean diff.

SUPERCILEX:
@samtstern Almost there, #1038 has to come first though.

auth/src/main/java/com/firebase/ui/auth/util/data/PhoneNumberUtils.java
```diff
@@ -46,21 +47,10 @@
...
-        CountryCodeToRegionCodeMap = Collections.unmodifiableMap(createCountryCodeToRegionCodeMap());
-        CountryCodeByIsoMap = Collections.unmodifiableMap(createCountryCodeByIsoMap());
-    }
-
-    /**
-     * This method may be used to force initialize the static members in the class. It is
-     * recommended to do this in the background since the HashMaps created above may be time
-     * consuming operations on some devices.
-     */
-    static void load() {
```
SUPERCILEX: We never used this anywhere and `PhoneNumberUtils.class` would work as well.

auth/src/main/java/com/firebase/ui/auth/util/data/PhoneNumberUtils.java
```diff
@@ -71,37 +61,25 @@ static void load() {
...
-                ? phoneNumber
-                : ("+" + String.valueOf(countryInfo.getCountryCode())
-                + phoneNumber.replaceAll("[^\\d.]", ""));
+    public static String format(@NonNull String phoneNumber, @NonNull CountryInfo countryInfo) {
+        if (phoneNumber.startsWith("+")) {
+            return phoneNumber;
+        } else {
+            return "+"
+                    + String.valueOf(countryInfo.getCountryCode())
+                    + phoneNumber.replaceAll("[^\\d.]", "");
```
SUPERCILEX: I felt like the `if` statement was clearer in this case.

auth/src/main/java/com/firebase/ui/auth/util/data/PhoneNumberUtils.java
```diff
@@ -71,37 +61,25 @@ static void load() {
...
-        final CountryInfo currentCountry = PhoneNumberUtils.getCurrentCountryInfo(context);
-
-        // TODO(samstern): Remove this call once the next version of Play servics is released
-        //                 estimated timelime August 10th. It is a known issue that phone numbers
-        //                 from the hint selector are susceptible to the false negatives of this
-        //                 method.
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
-            return android.telephony.PhoneNumberUtils
-                    .formatNumberToE164(phoneNumber, currentCountry.getLocale().getCountry());
-        }
```
SUPERCILEX: We can remove this right?

auth/src/main/java/com/firebase/ui/auth/util/data/PhoneNumberUtils.java
```diff
@@ -172,1187 +162,21 @@ private static String countryIsoForCountryCode(String countryCode) {
...
-            if (CountryCodeToRegionCodeMap.containsKey(countryCodeKey)) {
+            if (COUNTRY_TO_REGION_CODES.indexOfKey(countryCodeKey) >= 0) {
                 return potentialCountryCode;
             }
         }
 
         return DEFAULT_COUNTRY_CODE;
     }
 
-    private static Map<Integer, List<String>> createCountryCodeToRegionCodeMap() {
```
SUPERCILEX: These functions were kinda icky so I cleaned them up.

FYI: the cleanup is guaranteed to not include typos or missing entries since I discovered IntelliJ's structural search and replace which enables this kind of hardcore refactoring.

auth/src/main/java/com/firebase/ui/auth/util/data/PhoneNumberUtils.java
samtstern: Yep good call.

auth/src/main/java/com/firebase/ui/auth/util/data/PhoneNumberUtils.java
```diff
@@ -1362,13 +183,489 @@ private static String stripPlusSign(String phoneNumber) {
...
+        map.put(996, singletonList("KG"));
+        map.put(998, singletonList("UZ"));
+
+        return map;
+    }
+
+    private static synchronized Map<String, Integer> createCountryCodeByIsoMap() {
+        Map<String, Integer> map = new HashMap<>(MAX_COUNTRIES);
+
+        map.put("AF", 93);
```
samtstern: Could we single source this from the other map?

```java
SparseArray<List<String>> codeToRegion = createCountryCodeToRegionCodeMap();

for(int i = 0; i < codeToRegion.size(); i++) {
   int code = sparseArray.keyAt(i);
   List<Sttring> regions = sparseArray.get(code);

   for (String region : regions) {
      map.put(region, code);
    }
}
```

samtstern: Or maybe it's not the same map in reverse and that was just the case for the few I checked.  In which case, ignore me.

SUPERCILEX: @samtstern Thanks, I didn't notice that! 😄

The generated maps weren't exactly the same, but I finally managed to generate a decent diff. The hard coded map has `{ HM=672, GS=500, XK=381 }` additional values over the generated map whereas the latter has this extra value: `TA=290`. Could that actually be a mistake? If I added those 3 extra values to the map full of lists, it would make them perfect inverses. Should I do that?

samtstern: To be safe let's make them perfect inverses and then add those exceptions only where we have them now.  So basically each map is a view on a base map plus some exceptions.

Then we can file an issue to track down the source of those exceptions with Ashwin, but I don't want to block this PR on that investigation.

SUPERCILEX: SGTM, is 01d0adb what you were looking for?

auth/src/main/java/com/firebase/ui/auth/util/data/PhoneNumberUtils.java
```diff
@@ -1362,13 +183,259 @@ private static String stripPlusSign(String phoneNumber) {
...
+            List<String> regions = COUNTRY_TO_REGION_CODES.get(code);
+
+            for (String region : regions) {
+                map.put(region, code);
+            }
+        }
+
+        // TODO Figure out why these exceptions exist.
+        // This map used to be hardcoded so this is the diff from the generated version.
+        map.remove("001");
```
samtstern: Ok so it looks like in `COUNTRY_TO_REGION_CODES` we have:

```java
map.put(800, singletonList("001"));
map.put(808, singletonList("001"));
// ...
map.put(870, singletonList("001"));
map.put(878, singletonList("001"));
// ...
map.put(881, singletonList("001"));
map.put(882, singletonList("001"));
map.put(883, singletonList("001"));
// ...
map.put(888, singletonList("001"));
// ...
map.put(979, singletonList("001"));
 ```

So when we're looping through that map we would overwrite the `001` key with, `800` then `808` then `870` etc.  So while the end result is OK because we do `map.remove("001")` I am now worried about our map inversion technique.

Are there any other codes that appear multiple times that we would be overwriting?  If so isn't that almost definitely a problem?

Comment (samtstern) [reply auth/src/main/java/com/firebase/ui/auth/util/data/PhoneNumberUtils.java]: One way to find out and make sure it never happens:

```java
for (String region : regions) {
    if (map.containsKey("region")) {
      throw new IllegalStateException("Two country codes for region: " + region):
    }
    map.put(region, code);
}

```

SUPERCILEX: Oh shucks! Do you want to go back to the hardcoded version? For now, I've gone with your second comment.

samtstern:
@SUPERCILEX thanks for all the changes!

SUPERCILEX:
Awesome! Have you seen https://github.com/firebase/FirebaseUI-Android/issues/1007#issuecomment-350376671?

PR #1028, opened 2017-11-22
changed order of code to reflect autocreation

bytePro17124:
FirebaseRecyclerAdapter, swapped order of onBindViewHolder and onCreateViewHolder to better reflect how Android Studio currently auto-populates these.

### DIFF
--- a/database/README.md
+++ b/database/README.md
@@ -123,6 +123,12 @@ for displaying each item. In this case we will use a custom `ChatHolder` class:
 ```java
 FirebaseRecyclerAdapter adapter = new FirebaseRecyclerAdapter<Chat, ChatHolder>(options) {
     @Override
+    protected void onBindViewHolder(ChatHolder holder, int position, Chat model) {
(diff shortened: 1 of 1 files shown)


SUPERCILEX:
@bytePro17124 Thanks for the PR! However, I'd like to keep the existing order if possible. The reason IntelliJ puts `onBindViewHolder` on top is because it prioritizes interfaces at the bottom of the dependency tree vs the top. That is, we define `onBindViewHolder`, but `onCreateViewHolder` is defined by `RecyclerView.Adapter` which is one level up from the Firebase adapters—in `RecyclerView.Adapter`, `onCreateViewHolder` comes first. Technicalities aside, it's also just more logical to create the view holder first before binding it. 😄 

Does that make sense?

bytePro17124:
@SUPERCILEX That does make sense. Thanks for the explanation!

SUPERCILEX:
Awesome sauce!

PR #1038, opened 2017-12-06
Cleanup

SUPERCILEX:
The cleanup work that was pulled out of #1017


auth/src/main/java/com/firebase/ui/auth/util/PlayServicesHelper.java
```diff
@@ -11,13 +11,9 @@
  * client application
  */
 public class PlayServicesHelper {
-    private static GoogleApiAvailability mApiAvailability;
-
+    @Deprecated
```
samtstern: Any reason for this change?  Why not preserve the availability object?

auth/src/main/java/com/firebase/ui/auth/util/ui/PreambleHandler.java
```diff
@@ -41,7 +41,14 @@ public PreambleHandler(Context context, FlowParameters parameters, @StringRes in
         setupCreateAccountPreamble();
```
samtstern: Should we move this call out of the constructor and into the static `setup` method as well?

auth/src/main/java/com/firebase/ui/auth/util/PlayServicesHelper.java
SUPERCILEX: Isn't it kinda pointless to store a static field in a static field? 😆 I would guess this is old code from before you could get the availability singleton like so: `GoogleApiAvailability.getInstance()`.

Is there any reason to store the singleton's reference in our class?

auth/src/main/java/com/firebase/ui/auth/util/ui/PreambleHandler.java
SUPERCILEX: Good idea, done.

auth/src/main/java/com/firebase/ui/auth/util/PlayServicesHelper.java
samtstern: I see, I forgot that the underlying object was a singleton as well.  :+1:

SUPERCILEX: Haha, cool beans.

Issue #1044, opened 2017-12-11
FirestoreRecyclerAdapter: onBindViewHolder not called when item is changed AND moved


marzipanbrain:
Describe your environment: 
  * Android device: Emulator
  * Android OS version: Nougat
  * Google Play Services version: 11.7.43
  * FirebaseUI version: 3.1.0
  
Describe the problem:
I've written an adapter that extends the FirestoreRecyclerAdapter. My RecyclerView uses the adapter to display items. The adapter uses a query that is ordered by a boolean field named 'checkedIn'. The adapter also uses a custom parser that implements SnapshotParser. When I'm binding the data to my viewHolder, I use the value of checkedIn to set the background color of the item. I think everything that I'm doing is typical.
Here are the situations where I'm having trouble:
(1) When the value of checkedIn is changed on a document (for example: checkedIn goes from true to false), the FirestoreRecyclerAdapter 'hears' the change but treats it as a 'MOVE' instead of a both a 'MOVE' and a 'CHANGE'. The consequence of it being treated as only a 'MOVE' is that onBindViewHolder() never gets called and it never gets a chance to update the background color of the item. In the RecyclerView, the item does get moved. It's just the wrong color and contains the wrong data.
(2) The second situation relates to how the cache deals with moves that are also changes. If a MOVE ( with a CHANGE) comes through, the cache does not invalidate the old object in the cache. Consequently, the old cached object is passed to onBindViewHolder.

Expected Results:
I would expect onBindViewHolder to get called with the most current data whenever there is a change.

Relevant Code:
As a workaround on my local system, I've made the following changes:
For (1), I added a call to notifyItemChanged(newIndex) in the 'MOVED' case of FirestoreRecyclerAdapter.onChildChanged(). I placed it right after the existing call to  notifyItemMoved(oldIndex, newIndex);
For (2), I changed BaseObservableSnapshotArray.notifyOnChildChanged() to call mCachingParser.invalidate(snapshot) when the change type is MOVED. I realize that change undermines the effectiveness of the cache.
These two changes aren't ideal but they do resolve the problems I'm seeing. Could you recommend a better workaround or is this a bug?

SUPERCILEX:
Oops, that's an oversight on our part. I'm working on a fix right now.

samtstern:
This is fixed in version `3.1.1`.

marzipanbrain:
Thanks for taking care of this so quickly!

PR #1045, opened 2017-12-11
Fix Firestore move events ignoring changes

SUPERCILEX:
#1044


firestore/src/main/java/com/firebase/ui/firestore/FirestoreArray.java
```diff
@@ -110,13 +110,16 @@ private void onDocumentModified(DocumentChange change) {
...
+                    change.getNewIndex(), change.getNewIndex());
         } else {
             // Document moved and possibly also modified
             mSnapshots.remove(change.getOldIndex());
             mSnapshots.add(change.getNewIndex(), change.getDocument());
+
             notifyOnChildChanged(ChangeEventType.MOVED, change.getDocument(),
                     change.getNewIndex(), change.getOldIndex());
+            notifyOnChildChanged(ChangeEventType.CHANGED, change.getDocument(),
+                    change.getNewIndex(), change.getNewIndex());
```
SUPERCILEX: Should we try to no-op this? I can't think of a use case where a change isn't caused by the model also changing...

samtstern:
Missed this but thanks for the fix!

PR #1047, opened 2017-12-12
Travis fix

samtstern:
Travis is red on `master`.  Strangely enough the first red build was a change to only `README.md` ... *sigh*.  So gonna try a few things.

The build is already green on `version-3.1.1-dev` but having the red build on `master` is not a good look.


samtstern:
Well the build passes now ... still fails on my machine but I guess one thing at a time!

PR #1048, opened 2017-12-12
Refactor reset password flow to use new architecture

SUPERCILEX:
A "dip our feet in the water" PR to see what the new architecture is like.


auth/src/main/java/com/firebase/ui/auth/util/ui/FlowHolder.java
```diff
@@ -0,0 +1,75 @@
...
+                    INTENT_STARTER = new AutoClearSingleLiveEvent<>(this);
+                }
+            });
+    private static MutableLiveData<Pair<PendingIntent, Integer>> PENDING_INTENT_STARTER =
+            new AutoClearSingleLiveEvent<>(new Runnable() {
+                @Override
+                public void run() {
+                    PENDING_INTENT_STARTER = new AutoClearSingleLiveEvent<>(this);
+                }
+            });
```
SUPERCILEX: Here's where you get to admire the grossness of `AutoClearSingleLiveEvent`... oh well.

auth/src/test/java/com/firebase/ui/auth/ui/email/RecoverPasswordActivityTest.java
```diff
@@ -1,69 +0,0 @@
...
-import org.robolectric.annotation.Config;
-
-import java.util.Collections;
-
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
-
-@RunWith(RobolectricTestRunner.class)
-@Config(constants = BuildConfig.class, sdk = 25)
-public class RecoverPasswordActivityTest {
```
SUPERCILEX: Tests! I don't think this one is particularly useful, but I haven't written any new ones yet. I was think of writing tests for:
- `SingleLiveEvent`
- `AutoClearSingleLiveEvent`
- `ViewModelBase` for completness
- `RecoverPasswordHandler`?

Any thoughts?

auth/src/main/java/com/firebase/ui/auth/ui/HelperActivityBase.java
```diff
@@ -42,6 +51,40 @@ protected void onCreate(Bundle savedInstance) {
         super.onCreate(savedInstance);
         mAuthHelper = new AuthHelper(getFlowParams());
         mProgressDialogHolder = new ProgressDialogHolder(this);
+
+        getFlowHolder().getIntentStarter().observe(this, new Observer<Pair<Intent, Integer>>() {
```
samtstern: Since we're already using data types like `ActivityResult` let's create a custom object type like `IntentRequest` rather than use a `Pair` here.

auth/src/main/java/com/firebase/ui/auth/ui/HelperActivityBase.java
```diff
@@ -42,6 +51,40 @@ protected void onCreate(Bundle savedInstance) {
...
+            @Override
+            public void onChanged(@Nullable Pair<Intent, Integer> request) {
+                if (request == null) {
+                    throw new IllegalStateException("Cannot start null request");
+                }
+
+                startActivityForResult(request.first, request.second);
+            }
+        });
+        getFlowHolder().getPendingIntentStarter()
```
samtstern: How do we make sure we only observe the same `Pair` once?  Apologies, I am sort of an architecture components noob.

auth/src/main/java/com/firebase/ui/auth/ui/email/RecoverPasswordActivity.java
```diff
@@ -54,51 +56,57 @@ protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setContentView(R.layout.fui_forgot_password_layout);
 
-        mEmailFieldValidator =
-                new EmailFieldValidator((TextInputLayout) findViewById(R.id.email_layout));
+        mHandler = ViewModelProviders.of(this).get(RecoverPasswordHandler.class);
+        mHandler.init(getFlowHolder());
+        mHandler.getPasswordResetListener().observe(this, new Observer<Task<String>>() {
```
samtstern: Hmm this seems a little strange to me.  We call a `getListener` method and then `observe` a listener?  

I think what we want is something like an `Observable<PasswordResetState>` and the state can include properties like `boolean success  and `String errorMessage`.  Also avoids passing down the `Task` and lets us encapsulate the error `instanceof` logic higher up.

auth/src/main/java/com/firebase/ui/auth/ui/email/RecoverPasswordActivity.java
```diff
@@ -54,51 +56,57 @@ protected void onCreate(Bundle savedInstanceState) {
...
+                } else if (task.getException() instanceof FirebaseAuthInvalidUserException
+                        || task.getException() instanceof FirebaseAuthInvalidCredentialsException) {
+                    // No FirebaseUser exists with this email address, show error.
+                    mEmailInputLayout.setError(getString(R.string.fui_error_email_does_not_exist));
+                } else {
+                    mEmailInputLayout.setError(task.getException().getLocalizedMessage());
+                }
+            }
+        });
+        getFlowHolder().getProgressListener().observe(this, new Observer<Boolean>() {
```
samtstern: Again I have a preference for using a more abstracted type than `Boolean` maybe an object like `ProgressState` that has a boolean for if something is in progress?  Makes the purpose clearer.

auth/src/main/java/com/firebase/ui/auth/ui/email/RecoverPasswordHandler.java
```diff
@@ -0,0 +1,43 @@
...
+    public LiveData<Task<String>> getPasswordResetListener() {
+        return mPasswordResetListener;
+    }
+
+    public void startReset(final String email) {
+        if (mCachedPasswordReset == null || !TextUtils.equals(email, mCachedPasswordReset.first)) {
+            mCachedPasswordReset = Pair.create(email, mAuth.sendPasswordResetEmail(email));
+        }
+
+        mFlowHolder.getProgressListener().setValue(false);
```
samtstern: Can we expose `mFlowHolder` through a getter?  Accessing superclass variables directly when I can't see their definition always kinda freaks me out.

auth/src/main/java/com/firebase/ui/auth/util/data/AuthViewModelBase.java
```diff
@@ -0,0 +1,28 @@
...
+import android.app.Application;
+
+import com.firebase.ui.auth.util.ui.FlowHolder;
+import com.firebase.ui.auth.util.ui.ViewModelBase;
+import com.google.firebase.FirebaseApp;
+import com.google.firebase.auth.FirebaseAuth;
+import com.google.firebase.auth.PhoneAuthProvider;
+
+public class AuthViewModelBase extends ViewModelBase<FlowHolder> {
+    protected FlowHolder mFlowHolder;
```
samtstern: As in my other comment, would prefer to make each of these private and then access in child classes through getter.

auth/src/main/java/com/firebase/ui/auth/util/data/AuthViewModelBase.java
```diff
@@ -0,0 +1,28 @@
...
+    protected FirebaseAuth mAuth;
```
samtstern: Does putting `FirebaseAuth` in the `ViewModel` leak a `Context`?

auth/src/test/java/com/firebase/ui/auth/ui/email/RecoverPasswordActivityTest.java
samtstern: Those would be a great start, especially the `LiveEvent` classes.  I am fine if you want to make that a separate PR though.

auth/src/main/java/com/firebase/ui/auth/ui/HelperActivityBase.java
SUPERCILEX: Yep, that's just me being lazy. 😄

auth/src/main/java/com/firebase/ui/auth/ui/email/RecoverPasswordHandler.java
SUPERCILEX: SGTM, done.

auth/src/main/java/com/firebase/ui/auth/util/data/AuthViewModelBase.java
SUPERCILEX: Done.

auth/src/main/java/com/firebase/ui/auth/util/data/AuthViewModelBase.java
SUPERCILEX: No? I'm pretty sure `FirebaseAuth` only has access to app contexts.

samtstern: I think you're right.  It's only `FirebaseAnalytics` that takes a context when you get an instance.

SUPERCILEX: Even analytics doesn't use the context (dunno why):
image
image
image

auth/src/test/java/com/firebase/ui/auth/ui/email/RecoverPasswordActivityTest.java
SUPERCILEX: Yeah, I think a separate PR would be good so we can really focus on writing quality tests and setting a standard for the rest of the codebase.

auth/src/main/java/com/firebase/ui/auth/ui/HelperActivityBase.java
SUPERCILEX: That's what the `SingleLiveEvent` is for: it keeps track of which observers have received the update and then doesn't send it again if the observer already got it. In addition, lifecycle owners (activities) that aren't past `onStart` don't receive updates anyway.

No problemo about the questions, I'm happy to share what I've learned. 😄

auth/src/main/java/com/firebase/ui/auth/ui/email/RecoverPasswordActivity.java
SUPERCILEX: Yeah, this was poorly named.

I really like your idea of passing around a state object which includes everything needed to update the UI—I was always a little annoyed at having to observe two `LiveData`s. I'll play around with that in just a bit, thanks! 😄

SUPERCILEX: Ooooo, I'm super excited! Dunno why I didn't think of doing that from the start, but your idea makes things _waaaaaaaay_ nicer IMO. It also replaces the `DialogFragment` which is really cool and interesting. Whad'ya think? 😄

auth/src/main/java/com/firebase/ui/auth/data/model/ActivityResult.java
```diff
@@ -0,0 +1,58 @@
+package com.firebase.ui.auth.data.model;
+
+import android.content.Intent;
+import android.support.annotation.Nullable;
+
+public final class ActivityResult {
```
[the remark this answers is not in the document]
samtstern: SGTM.  Could we annotate/document each of these new public classes appropriately?  We definitely don't want anyone but us using them, and documentation couldn't hurt for future contributors reading the source.

auth/src/main/java/com/firebase/ui/auth/ui/email/RecoverPasswordHandler.java
```diff
@@ -0,0 +1,34 @@
...
+import android.support.annotation.NonNull;
+import android.text.TextUtils;
+import android.util.Pair;
+
+import com.firebase.ui.auth.util.data.AuthViewModelBase;
+import com.google.android.gms.tasks.OnCompleteListener;
+import com.google.android.gms.tasks.Task;
+
+public class RecoverPasswordHandler extends AuthViewModelBase {
+    private Pair<String, Task<Void>> mCachedPasswordReset;
```
samtstern: nit: what is the Pair doing for us here?  Why not just two fields?

auth/src/main/java/com/firebase/ui/auth/ui/email/RecoverPasswordHandler.java
```diff
@@ -0,0 +1,34 @@
...
+
+    public RecoverPasswordHandler(Application application) {
+        super(application);
+    }
+
+    public void startReset(final String email) {
+        if (mCachedPasswordReset == null || !TextUtils.equals(email, mCachedPasswordReset.first)) {
```
samtstern: Can you explain a little bit what's going on here?

auth/src/main/java/com/firebase/ui/auth/ui/email/RecoverPasswordActivity.java
```diff
@@ -54,51 +56,57 @@ protected void onCreate(Bundle savedInstanceState) {
...
+            @Override
+            public void onChanged(Boolean isDone) {
+                if (isDone) {
+                    getDialogHolder().dismissDialog();
+                } else {
+                    getDialogHolder().showLoadingDialog(R.string.fui_progress_dialog_sending);
+                }
+            }
+        });
```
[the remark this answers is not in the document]
samtstern: I like the direction but not so sure about this.  I don't like that we're subscribing to `ProgressState` but immediately checking for a cast to `RecoverPasswordProgressState`.

Why not just have the `RecoverPasswordHandler` expose the `RecoverPasswordProgressState` live data?

auth/src/main/java/com/firebase/ui/auth/ui/email/RecoverPasswordActivity.java
```diff
@@ -54,51 +60,69 @@ protected void onCreate(Bundle savedInstanceState) {
...
+    @Override
+    public void onDonePressed() {
+        mHandler.startReset(mEmailEditText.getText().toString());
+    }
+
+    private void showEmailSentDialog(String email) {
+        new AlertDialog.Builder(this)
+                .setTitle(R.string.fui_title_confirm_recover_password)
+                .setMessage(getString(R.string.fui_confirm_recovery_body, email))
+                .setOnDismissListener(this)
```
samtstern: nit: would it be clearer just to define the onDismissListener inline here?

auth/src/main/java/com/firebase/ui/auth/ui/email/RecoverPasswordProgressState.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public RecoverPasswordProgressState(boolean done,
+                                        @Nullable String email,
+                                        @Nullable Exception e) {
+        super(done, e);
+        mEmail = email;
+
+        if (isSuccessful() && TextUtils.isEmpty(email)) {
+            throw new IllegalStateException("Email cannot be null in the success state");
```
samtstern: nit: Why not expose two constructors?  One that takes a `NonNull` email and represents success and another that takes an `Exception` and represents failure?

auth/src/main/java/com/firebase/ui/auth/util/data/AutoClearSingleLiveEvent.java
```diff
@@ -0,0 +1,24 @@
...
+
+    public AutoClearSingleLiveEvent(Runnable resetter) {
+        mFieldResetter = resetter;
+    }
+
+    @Override
+    public void removeObserver(@NonNull Observer<T> observer) {
+        super.removeObserver(observer);
+        if (!hasObservers()) {
+            mFieldResetter.run();
```
samtstern: Yikes!  Why is this a runnable?  Could it be an abstract method that subclasses have to implement called `onClearFields` or something?

auth/src/main/java/com/firebase/ui/auth/util/data/SingleLiveEvent.java
```diff
@@ -0,0 +1,99 @@
...
+ * A LiveData that only notifies observers when new data arrives, used for events like navigation
+ * and SnackBar messages.
+ * <p>
+ * This avoids a common problem with events: on configuration change (like rotation) an update can
+ * be emitted if the observer is active.
+ * <p>
+ * This LiveData supports multiple observers per class definition, but works on a stack model. Only
+ * the newest observers get updates.
+ */
+public class SingleLiveEvent<T> extends MutableLiveData<T> {
```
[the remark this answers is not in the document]
samtstern: Ok I see what you are doing here but it took me a long time to sort it out so I'd suggest the following refactor:

  * Go back to the implementation from the sample that you linked to.  It is simple: no repeat updates.
  * Drop `EventFilterObserver`, I don't think we need it.
  * Make a class called like `SingleLiveEventRegistry<T>` or similar that keeps a map of `String --> SingleLiveEvent`
  * Rather than use classes for identification, have observers get this livedata by giving a `String` "TAG" to the registry.

So in your observer class you'd have something like:

```java
register.getLiveEvent("MyTag").observe(...)
```

I think that will simplify this a lot.   Correct me if I have missed the point of all this.

auth/src/main/java/com/firebase/ui/auth/ui/email/RecoverPasswordHandler.java
SUPERCILEX: The idea was to cache the task so that if the user mashed the "Send" button, we would only start the request once, but I thought about this and meh. It's extra complexity for something that would force the user to be stuck in an error state so I removed it.

auth/src/main/java/com/firebase/ui/auth/ui/email/RecoverPasswordActivity.java
SUPERCILEX: Yep, done.

auth/src/main/java/com/firebase/ui/auth/data/model/ActivityResult.java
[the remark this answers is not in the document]
SUPERCILEX: Done.

auth/src/main/java/com/firebase/ui/auth/ui/email/RecoverPasswordProgressState.java
SUPERCILEX: I was tired? 😆 Yep, done.

samtstern:
@SUPERCILEX thanks for the last round of commits!  Today's my last day in the office for 2017 so I am going to punt this one into the new year.  Happy holidays!

SUPERCILEX:
@samtstern SGTM, I've been back and forth between this PR and my fork to see how I can incorporate your ideas. I'm getting really close so something should be ready in 2018. (That sounds so weird. 😆)

Happy holidays! 🎉

auth/src/main/java/com/firebase/ui/auth/ui/email/RecoverPasswordActivity.java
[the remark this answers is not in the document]
SUPERCILEX: The reason I shared state everywhere was to support phone auth and email. For phone auth, when an auto-verification fails, I would set the state as failed with a custom `PhoneVerificationException`. Then, the activity would decide what to do with the UI and show the submit code fragment. It turns out that works just as well with a live data per activity which would get rid of the unpleasant instance of checks.

So TL;DR: great idea, thanks!

samtstern:
@SUPERCILEX do you think we should shift the milestone for this one to `3.2.0`?

auth/src/main/java/com/firebase/ui/auth/util/data/SingleLiveEvent.java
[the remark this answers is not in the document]
SUPERCILEX: @samtstern That would work, but I don't see how it will make things simpler since we'd just be moving the logic into another class (figuring out which generated live datas need to receive updates and keeping track of the real events etc...)

SUPERCILEX:
@samtstern 👍 done.

I've decided to pull the `SingleLiveData` stuff so we can add it later when it's actually being used. The main issue is that the future sign-in handler will subscribe to activity results which means the activity results live data can't be linked to an activity (must be static). That in and of itself is unpleasant and I'd much rather be bouncing ideas around with you when we get there than the current "head banging against the wall" stuff I'm doing 😆.

Anyway, I believe this PR should be very close to a merge since all the questionable changes were pulled. 😁 Any last comments?

auth/src/main/java/com/firebase/ui/auth/ui/email/RecoverPasswordHandler.java
```diff
@@ -0,0 +1,39 @@
...
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+public class RecoverPasswordHandler extends AuthViewModelBase {
+    private final MutableLiveData<RecoverPasswordProgressState> mProgressLiveData =
+            new MutableLiveData<>();
+
+    public RecoverPasswordHandler(Application application) {
+        super(application);
+    }
+
+    public MutableLiveData<RecoverPasswordProgressState> getProgressLiveData() {
```
samtstern: This method should return `LiveData` not `MutableLiveData` so that consumers don't mutate it.

auth/src/main/java/com/firebase/ui/auth/ui/email/RecoverPasswordActivity.java
```diff
@@ -54,51 +59,66 @@ protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setContentView(R.layout.fui_forgot_password_layout);
 
-        mEmailFieldValidator =
-                new EmailFieldValidator((TextInputLayout) findViewById(R.id.email_layout));
+        mHandler = ViewModelProviders.of(this).get(RecoverPasswordHandler.class);
+        mHandler.init(getFlowHolder());
```
samtstern: It does not seem like a good idea to pass one `ViewModel` into another.  If the handler needs the flow params maybe we should pass that?

auth/src/main/java/com/firebase/ui/auth/ui/email/RecoverPasswordProgressState.java
```diff
@@ -0,0 +1,67 @@
...
+
+import com.firebase.ui.auth.data.model.ProgressState;
+
+/**
+ * Password recovery state.
+ * <p>
+ * When this state is successful, the email who's password was reset can be retrieved.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+public final class RecoverPasswordProgressState extends ProgressState {
```
samtstern: Ok you're probably going to kill me for suggesting another direction here ... but I am learning a lot with each iteration!

I saw someone the other day define a class basically like this (simplified code):

```java
public class Resource<T> {

   State state; // some enum
   T value;
   Exception e;

}
```

So then recover password screen would watch a single live data that is `Resource<String>` where the `String` is the email.  And the `Observer` would get a series of `Resource` objects with states like `STATE_NONE`, `STATE_LOADING`, `STATE_SUCCESS/STATE_FAILURE`.  So it's kind of like `Task` but adapted for the world of `LiveData`.

In this way you don't need a subclass of `ProgressState` for each screen that loads something just a different `<T>`, which could be a compound object you define if you're waiting for a few results.

What do you think?  Cool?  Stupid?

auth/src/main/java/com/firebase/ui/auth/util/ui/FlowHolder.java
```diff
@@ -0,0 +1,24 @@
...
+    public FlowHolder(Application application) {
+        super(application);
+    }
+
+    @Override
+    protected void onCreate(FlowParameters args) {
+        mFlowParams = args;
+    }
+
+    public FlowParameters getParams() {
```
samtstern: nit: seems like this could be defined as `T getArguments()` in the `ViewModelBase` class?

auth/src/main/java/com/firebase/ui/auth/util/data/AuthViewModelBase.java
```diff
@@ -0,0 +1,40 @@
...
+    private FlowHolder mFlowHolder;
+    private FirebaseAuth mAuth;
+    private PhoneAuthProvider mPhoneAuth;
+
+    protected AuthViewModelBase(Application application) {
+        super(application);
+    }
+
+    @Override
+    protected void onCreate(FlowHolder args) {
```
samtstern: Similar to my previous comment, seems like we could pass `FlowParameters` here, right?  We only use the `FlowHolder` in `getFlowHolder()` but that is also defined by the base `Activity` class.

auth/src/main/java/com/firebase/ui/auth/util/data/AuthViewModelBase.java
```diff
@@ -0,0 +1,28 @@
...
+        mFlowHolder = args;
+
+        FirebaseApp app = FirebaseApp.getInstance(mFlowHolder.getParams().appName);
+        mAuth = FirebaseAuth.getInstance(app);
+        mPhoneAuth = PhoneAuthProvider.getInstance(mAuth);
```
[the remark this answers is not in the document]
samtstern: Would this leak if we kept it in the base activity class instead?  These are not really "model" objects so they don't need to be provided by a `ViewModel`

auth/src/main/java/com/firebase/ui/auth/ui/email/RecoverPasswordProgressState.java
SUPERCILEX: Ooooo, no, I really like that idea! 😁 It'll even make sure that we write proper model objects if we need to pass around a multi-arg result. (Let's agree not to use `Pair` 😉) Also, did you learn this from a talk or something I could look at too? Sounds really cool!

samtstern:
@SUPERCILEX well thanks for sticking with it, this looks great!  Merging this into `3.2.0-dev`

SUPERCILEX:
@samtstern Woot! This architecture is already waaaaaay better than what I had originally. Thanks! 😊

PR #1054, opened 2017-12-14
Add RestrictTo to all classes that need it

samtstern:
@SUPERCILEX making sure we have good coverage here and also making sure we can generate compatibility reports that ignore annotated classes.

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/data/client/CountryListLoadTask.java
+++ b/auth/src/main/java/com/firebase/ui/auth/data/client/CountryListLoadTask.java
@@ -19,6 +19,7 @@
 package com.firebase.ui.auth.data.client;
 
 import android.os.AsyncTask;
+import android.support.annotation.RestrictTo;
 
 import com.firebase.ui.auth.data.model.CountryInfo;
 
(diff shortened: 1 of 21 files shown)


samtstern:
Also after we get this in I would like to cut and release `3.1.1`.  It's been a while since we had a release and we have some good bugfixes in there.

SUPERCILEX:
@samtstern Yep, we definitely need a new release. 👍

As for the `RestrictTo` annotations, couldn't we put them on the package level to simplify things? I tried to make the entire `/data` package private for us, except for `FirebaseUiException` but that isn't public yet. (And we need a better package name for when we make it public anyway. 😄) Sadly, `/util` looks mixed.

What do you think?

samtstern:
I think in an ideal world we'd put them at the package level but the class level gains us a few things:

- Less worrying about package boundaries
- Clearly visible when reading the source code
- Compatibility with the compliance checker tool

So for now I'd like to get it on all of the classes, even if it's redundant.

Looks like `FirebaseUiException` would be included if we released today.  Do you want that marked restricted?

SUPERCILEX:
@samtstern SGTM about the package level annotations. And yeah, I think we should probably restrict `FirebaseUiException` since we haven't put enough thought into that API (and it's restricted in `IdpResponse`). Better safe than sorry! 😄

samtstern:
Ok SGTM I will do that and then merge this.  Thanks for the input!

Issue #1064, opened 2017-12-19
Can't build the project with Firebase UI Auth 3.1.1 / 3.1.2


jzurek:
### Step 2: Describe your environment
* Android OS version: N/A build time issue
* Firebase/Play Services SDK version: 11.6.2
* FirebaseUI version: 3.1.1, 3.1.2
  
### Step 3: Describe the problem:

Cannot build the project when the Firebase auth dependency is added.

#### Steps to reproduce:

  1. add following libraries 'com.firebaseui:firebase-ui-auth:3.1.1' to app/build.gradle (using `compile` as we're using older Gradle version)
```
compile 'com.android.support:appcompat-v7:27.0.2'
compile 'com.android.support:support-v4:27.0.2'
compile 'com.android.support:design:27.0.2'
compile 'com.android.support:customtabs:27.0.2'
compile 'com.android.support:cardview-v7:27.0.2'
compile 'com.android.support:recyclerview-v7:27.0.2'
compile 'com.google.android.gms:play-services-gcm:11.6.2'
compile 'com.google.firebase:firebase-core:11.6.2'
compile 'com.google.firebase:firebase-messaging:11.6.2'
compile 'com.google.firebase:firebase-auth:11.6.2'
compile 'com.firebaseui:firebase-ui-auth:3.1.2'
```

#### Observed Results:

```
The project build fails with following error:
(...)/app/build/intermediates/res/merged/dev/homologation/layout/fui_phone_layout.xml
  Error:(16, 57) No resource found that matches the given name (at 'layout_constraintBaseline_toBaselineOf' with value '@id/phone_layout').
```

#### Expected Results:

The app to build

#### Relevant Code:
`gradle-wrapper.properties`
```
#Mon Dec 18 18:10:48 BRST 2017
distributionBase=GRADLE_USER_HOME
distributionPath=wrapper/dists
zipStoreBase=GRADLE_USER_HOME
zipStorePath=wrapper/dists
distributionUrl=https\://services.gradle.org/distributions/gradle-4.4-all.zip
```

`build.gradle`
```buildscript {
    repositories {
        jcenter()
        mavenCentral()
        flatDir {
            dirs 'libs'
        }
    }

    dependencies {
        classpath 'com.android.tools.build:gradle:2.3.3'
        classpath 'me.tatarka:gradle-retrolambda:3.2.3'
        classpath 'com.neenbedankt.gradle.plugins:android-apt:1.8'
[...]

PR #1067, opened 2017-12-20
AAPT1 Compatibility

samtstern:
This fixes issue #1064, at least on my machine.

However I am not sure it's a good idea, will these just get reset the next time we modify a constraint?  And how can we prevent this in future layouts.


SUPERCILEX:
@samtstern ohhhhhhhhh, this isn't a CL thing, it's an aapt thing. The revamped aapt2 finally fixed the +id stuff so you no longer have to define them sequentially in order to be able to use them. @vanniktech that's wrong with aapt2: it creates a graph of ids first before processing the layouts which lets you create the IDs only where they are actually defined. (That makes code navigation better since you don't have a bunch of fake IDs to look through.)

samtstern:
@SUPERCILEX so basically the CL editor is relying on `aapt2` behavior?  Is this documented somewhere?  And what do you think of this PR ... should we do this or do you have another idea?

auth/src/main/res/layout/fui_phone_layout.xml
```diff
@@ -13,14 +13,14 @@
...
-            app:layout_constraintBaseline_toBaselineOf="@id/phone_layout" />
+            app:layout_constraintBaseline_toBaselineOf="@+id/phone_layout" />
 
         <com.firebase.ui.auth.util.ui.BaselineTextInputLayout
             android:id="@+id/phone_layout"
             style="@style/FirebaseUI.TextInputLayout.PhoneField"
             android:layout_width="0dp"
             android:layout_height="wrap_content"
-            app:layout_constraintStart_toEndOf="@id/country_list"
+            app:layout_constraintStart_toEndOf="@+id/country_list"
```
SUPERCILEX: The `+`s below this circular dependency can be removed.

auth/src/main/res/layout/fui_welcome_back_password_prompt_layout.xml
```diff
@@ -50,16 +50,16 @@
             android:layout_marginTop="@dimen/fui_field_padding_vert"
             android:text="@string/fui_trouble_signing_in"
             app:layout_constraintStart_toStartOf="parent"
-            app:layout_constraintEnd_toStartOf="@id/button_done"
-            app:layout_constraintTop_toBottomOf="@id/password_layout" />
+            app:layout_constraintEnd_toStartOf="@+id/button_done"
```
SUPERCILEX: Everything except for this `+` can be removed.

samtstern:
@SUPERCILEX what's the downside to just using `"@+id` everywhere as a project convention?  Seems like that will reduce the likelihood that we mess up in the future at the cost of some generated ID noise.

SUPERCILEX:
@samtstern We could add ids everywhere, but that seems counterproductive since the goal would be to remove them once everyone has upgraded. Adding ids everywhere isn't too bad, but the main downside is that when you `Ctrl + B` an id reference, it will show you _all_ the `@+id` declarations for that id which essentially renders the "Go to definition" useless since it's showing usages as well.

I personally prefer to prioritize the future over backcompat and I would hope most everyone will be on the 3.0 plugin within a few months, but I definitely see your maintenance concern so it's up to you. 😄

vanniktech:
Ideally lint would yell at you incorrectly use the id qualifier.

SUPERCILEX:
@vanniktech but again, what we are doing is correct on the 3.0 Gradle plugin, just not the 2.x ones... I would definitely recommend you upgrading to 3.x. 😀

vanniktech:
I'm not affected by this at all. Nonetheless I think having a Lint rule in place would be cool.

samtstern:
Ok I think for right now, until all of our developers are using build tools `3.0.0+`, the reasonable thing to to is take the hit and use `@+id` everywhere. 

And this may be overkill, but I implemented a custom lint check to enforce it!  Check out what happens when you mess up:
screenshot from 2017-12-20 15 44 57

This also lays a foundation for us to have custom lint in the future.  Which is pretty bad-ass.

lint/.gitignore
```diff
@@ -0,0 +1 @@
+/build
```
SUPERCILEX: This file should be unnecessary since we have a root-level ignore of all `build` folders.

vanniktech:
Oh wow. I meant that we'd file an issue to Google so they'd do it for us.

PR #1068, opened 2017-12-21
Better lint integration.

vanniktech:
We can do much better with Lint.

- Using Kotlin (precise and also half of the lint api is written in kotlin so we'll get extras like @Deprecated auto fix, extensions functions and all the other goodies)
- remove `'com.android.tools.lint:lint-checks:26.0.1'` since we don't use any of the android lint checks
- Use `LayoutDetector` so we don't need to check for the type
- Take effect of `visitElement` instead of visiting the entire document
- Unit test our lint check so we can enhance this further
- Take use of the quickfix api that was added in 3.0 so we can automatically from AS let lint fix the ids for us

Nice side effect, even with tests the LOC is the same 😀


lint/build.gradle
```diff
@@ -1,14 +1,12 @@
-apply plugin: 'java-library'
-
-configurations {
```
vanniktech: lintChecks comes out of the box with 3.0.0

lint/src/main/java/com/firebaseui/lint/NonGlobalIdDetector.kt
```diff
@@ -0,0 +1,52 @@
...
+import com.android.tools.lint.detector.api.Severity
+import com.android.tools.lint.detector.api.XmlContext
+import org.w3c.dom.Element
+
+/**
+ * Lint detector to find layout attributes that reference "@id" where they should reference
+ * "@+id" instead.
+ */
+class NonGlobalIdDetector : LayoutDetector() {
+    override fun getApplicableElements() = Detector.XmlScanner.ALL
```
SUPERCILEX: Let's specify the type here, `List<String>?` (I know it hurts 😄), since Kotlin is confused by opposing nullability annotations.

lint/src/main/java/com/firebaseui/lint/NonGlobalIdDetector.kt
```diff
@@ -0,0 +1,52 @@
...
+                .filter { it.nodeValue.contains(NON_GLOBAL_ID_PREFIX) }
+                .forEach {
+                    val quickFix = fix()
+                            .replace()
+                            .name("Fix id")
+                            .text(it.nodeValue)
+                            .with(it.nodeValue.replace(NON_GLOBAL_ID_PREFIX, GLOBAL_ID_PREFIX))
+                            .build()
+
+                    context.report(NON_GLOBAL_ID, it, context.getLocation(it), REPORT_MESSAGE, quickFix)
```
SUPERCILEX: I would prefer this like so, but it's up to you:
```kotlin
context.report(NON_GLOBAL_ID, it, context.getLocation(it), REPORT_MESSAGE, fix()
        .replace()
        .name("Fix id")
        .text(it.nodeValue)
        .with(it.nodeValue.replace(NON_GLOBAL_ID_PREFIX, GLOBAL_ID_PREFIX))
        .build()
)
```

lint/src/test/java/com/firebaseui/lint/NonGlobalIdDetectorTest.kt
```diff
@@ -0,0 +1,43 @@
+package com.firebaseui.lint;
```
SUPERCILEX: Naughty! 😄 Semicolons are so from the 70s.

samtstern:
@vanniktech thanks for sending this over!  I need to digest it but looks super cool.

I moved the base branch to `version-3.1.3-dev` which is the right target, please rebase when you get a chance.

lint/src/main/java/com/firebaseui/lint/NonGlobalIdDetector.kt
```diff
@@ -0,0 +1,52 @@
...
+ */
+class NonGlobalIdDetector : LayoutDetector() {
+    override fun getApplicableElements() = Detector.XmlScanner.ALL
+
+    override fun visitElement(context: XmlContext, element: Element) {
+        (0 until element.attributes.length)
+                .map { element.attributes.item(it) }
+                .filter { it.nodeValue.contains(NON_GLOBAL_ID_PREFIX) }
+                .forEach {
+                    val quickFix = fix()
```
samtstern: This is really cool!

lint/src/main/java/com/firebaseui/lint/NonGlobalIdDetector.kt
samtstern: Yeah I knew there had to be a better way than using `visitDocument`.  The JavaDoc for these detectors are ... not great.

lint/build.gradle
```diff
@@ -1,14 +1,12 @@
...
-    lintChecks
-}
+apply plugin: 'kotlin'
 
 dependencies {
     compileOnly 'com.android.tools.lint:lint-api:26.0.1'
-    compileOnly 'com.android.tools.lint:lint-checks:26.0.1'
-}
+    compileOnly 'org.jetbrains.kotlin:kotlin-stdlib-jre7:1.2.10'
```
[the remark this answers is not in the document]
samtstern: We had some trouble before when we started to use Java 8 internally, it made the library not work for developers on Java 7.

Are we 100% sure this won't bubble up in any way>

lint/src/main/java/com/firebaseui/lint/NonGlobalIdDetector.kt
SUPERCILEX: Can't you just walk over to Tor's desk and glare at him for a bit? 😆 Dunno if you saw his awesome talk, but if not here it is: https://youtu.be/p8yX5-lPS6o.

lint/build.gradle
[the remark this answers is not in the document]
SUPERCILEX: Yeah, I believe that was with EasyPermissions, right? Anyway, I have a hard time seeing how this could affect android app modules since this is a pure Java (or Kotlin) lib. I remember the issue occurring when we set the target compatibility in the `android` block whereas this is going to be run by IntelliJ. Also, if our `app` module doesn't compile, we'll know we messed up.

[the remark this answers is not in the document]
vanniktech: Do we really need `jre8`?

lint/src/test/java/com/firebaseui/lint/NonGlobalIdDetectorTest.kt
vanniktech: Oh this happened when converting the original file from java to kotlin. Will fix.

lint/src/main/java/com/firebaseui/lint/NonGlobalIdDetector.kt
vanniktech: Yeah the documentation is rough. There's the [google group](https://groups.google.com/forum/#!forum/lint-dev). And all of the other checks are open source so I started reading some of them and got to know the api through that.

lint/build.gradle
```diff
@@ -1,14 +1,12 @@
-apply plugin: 'java-library'
-
-configurations {
-    lintChecks
-}
+apply plugin: 'kotlin'
 
 dependencies {
     compileOnly 'com.android.tools.lint:lint-api:26.0.1'
```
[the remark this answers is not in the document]
vanniktech: It's [how](https://youtu.be/p8yX5-lPS6o?t=9m44s) you do it.

lint/src/main/java/com/firebaseui/lint/NonGlobalIdDetector.kt
vanniktech: I don't feel like it's necessary to add here.

lint/src/main/java/com/firebaseui/lint/NonGlobalIdDetector.kt
vanniktech: Not a strong opinion either

lint/build.gradle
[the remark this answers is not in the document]
samtstern: Yeah I honestly don't fully understand it but if you use `implementation` you get issues where there are multiple lint checkers conflicting.

build.gradle
```diff
@@ -8,9 +8,10 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.0.0'
+        classpath 'com.android.tools.build:gradle:3.0.1'
         classpath 'com.google.gms:google-services:3.1.1'
         classpath 'com.jfrog.bintray.gradle:gradle-bintray-plugin:1.7.3'
+        classpath 'org.jetbrains.kotlin:kotlin-gradle-plugin:1.2.10'
```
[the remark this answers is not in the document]
SUPERCILEX: Yeah, my B. `20` is in alpha. 😊

lint/src/main/java/com/firebaseui/lint/NonGlobalIdDetector.kt
SUPERCILEX: SGTM.

lint/src/main/java/com/firebaseui/lint/NonGlobalIdDetector.kt
SUPERCILEX: I hope I'm not being too much of a stickler, but here's my reasoning:
When it's a relatively small file like this, having a few platform dependent type warnings is fine, but should the file grow it just adds a whole bunch of unnecessary noise when you're trying to look at the scrollbar or `F2` through the file. Does that make sense?

PS: here's the error in case you turned it off:
untitled

vanniktech: Yeah I'm aware of that. I'll change it.

SUPERCILEX:
@vanniktech LGTM! Thanks for taking the first step on the road to Kotlintown! 😄

PR #1070, opened 2017-12-22
Update dependencies and set FUI version

### LINKED ISSUE #1065: Call setFirebaseUIVersion
In `11.8.0` we added a new method to the auth SDK:
https://firebase.google.com/docs/reference/android/com/google/firebase/auth/FirebaseAuth.html#setFirebaseUIVersion(java.lang.String)

Calling this method within FirebaseUI will help us better understand which auth patterns are coming from FirebaseUI and maybe unearth some error/usage patterns we didn't know about.

SUPERCILEX:
Fixes #1065


auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -166,6 +166,7 @@ private AuthUI(FirebaseApp app) {
         mAuth = FirebaseAuth.getInstance(mApp);
 
         mAuth.useAppLanguage();
+        mAuth.setFirebaseUIVersion(BuildConfig.VERSION_NAME);
```
SUPERCILEX: Is there anything else to this method? I've noticed that it returns a task (should we be listening for something?) and takes in a nullable parameter (is there a set/unset lifecycle?).

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -184,6 +184,20 @@ public static AuthUI getInstance() {
      */
     @NonNull
     public static AuthUI getInstance(@NonNull FirebaseApp app) {
+        AuthUI authUi;
+        synchronized (INSTANCES) {
+            authUi = INSTANCES.get(app);
+            if (authUi == null) {
+                authUi = new AuthUI(app);
+                authUi.mAuth.setFirebaseUIVersion(BuildConfig.VERSION_NAME);
```
SUPERCILEX: Is there anything else to this method? I've noticed that it returns a task (should we be listening for something?) and takes in a nullable parameter (is there a set/unset lifecycle?).

PS: Looks like there are conflicting annotations 😕:
image
image

samtstern: I think we can use it as fire-and-forget.

samtstern:
@SUPERCILEX thanks!

PR #1080, opened 2017-12-31
Use to Stale to simplify GitHub issue cleanup

SUPERCILEX:
@samtstern As you've probably noticed, I usually try to occasionally go through rounds where I ping you about a bunch of issues that can be closed now, but after working on Glide, I've noticed that Sam J. (yes, another awesome Sam! 😄) uses [Stale](https://github.com/apps/stale). The idea is very cool: until an issue is tagged with special tags (whitelisted), it's up for removal and will be closed (becomes stale) if there isn't a comment within a customizable timeout.

I've always been amazed at how few issues an incredibly popular library like Glide has, and it's all thanks to Stale. It would basically be a huge help in upkeeping our own issues. What do you think? If I had write access to the repo, I would totally set this up! 😜

If you do want to go through with this PR and Stale, you'll need to do a few things:
1. Add it to the `firebase` organization through the link above
1. Add a `feature` and `stale` label
1. That's it! Then, just make sure to add `feature`, `bug`, or `docs` tags once you're sure we're going to follow through on an issue.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -4,6 +4,7 @@ exemptLabels:
   - feature
   - enhancement
   - bug
+  - refactor
   - docs
   - fix-implemented
 staleLabel: stale
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,13 @@
+daysUntilStale: 7
+daysUntilClose: 7
+exemptLabels:
+  - feature
+  - bug
+  - docs
+  - fix-implemented
+staleLabel: stale
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. Thank you
+  for your contributions.
+closeComment: false
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -2,6 +2,7 @@ daysUntilStale: 7
 daysUntilClose: 7
 exemptLabels:
   - feature
+  - enhancement
   - bug
   - docs
   - fix-implemented


samtstern:
@SUPERCILEX thanks for thinking of this!  A few things:

- We actually have a home-grown bot (https://github.com/firebase/oss-bot) that can do something similar, but we have not yet enabled the auto-close feature because some people on our side thought that was a little harsh.  Not that home-grown == better (in fact it's the opposite I am sure) but this would at least eliminate the third point (see below).
- Can you show an example of some issues that would currently be marked as 'stale' if we turned this on today?
- I don't think I will be allowed to add this to the whole Firebase org, is there a way to do it on a per-reop basis?

Kurt29:
@samtstern 
It is possible to install it only for selected repositories
But I am not sure if you are able to do it when you don't have admin access to the whole company

SUPERCILEX:
@Kurt29 Thanks for the info!

@samtstern I think your second point hits home—I haven't really found any issues that could be close right now (that's a good thing! 😄). Anyway, I guess I got too excited in the moment. 😊 I'm going to close this PR, but hopefully it'll stay in the back of your mind if this type of bot could help out in a future repo.

samtstern:
@SUPERCILEX totally will keep it in mind if this repo ever gets out of hand again (like it was ~2 years ago).

I'm more interested in deploying these tools on repos without proactive contributors, making the Firebase github org healthier is a top priority for me.

PR #1081, opened 2017-12-31
Fix crash on API 19 when using email authentication.

clydebarrow:
Fixes issue # 1036

This fails but that's due to numerous pre-existing checkstyle validation failures.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,16 +39,16 @@ libraries.
 ```groovy
 dependencies {
 // FirebaseUI for Firebase Realtime Database
-implementation 'com.firebaseui:firebase-ui-database:3.1.1'
+implementation 'com.firebaseui:firebase-ui-database:3.1.2'
 
 // FirebaseUI for Cloud Firestore
-implementation 'com.firebaseui:firebase-ui-firestore:3.1.1'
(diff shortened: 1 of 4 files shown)


auth/src/main/java/com/firebase/ui/auth/ui/email/EmailActivity.java
```diff
@@ -123,7 +124,11 @@ public void onNewUser(User user) {
             FragmentTransaction ft = getSupportFragmentManager().beginTransaction()
                     .replace(R.id.fragment_register_email, fragment, RegisterEmailFragment.TAG);
 
-            if (emailLayout != null) ft.addSharedElement(emailLayout, "email_field");
+
+            if (emailLayout != null) {
+			    ViewCompat.setTransitionName(emailLayout, "email_field");
+			    ft.addSharedElement(emailLayout, "email_field");
+			}
```
SUPERCILEX: @clydebarrow Checkstyle is correct: please remove the new line above the if statement and pull back the indentation so that the end bracket is aligned with the `i` in `if` and the code is indented with 4 spaces.

Issue #1036, opened 2017-12-04
IllegalArgumentException in RegisterEmailActivity


guytavor:
```
Fatal Exception: java.lang.IllegalArgumentException: Unique transitionNames are required for all sharedElements
       at android.support.v4.app.BackStackRecord.addSharedElement(SourceFile:511)
       at com.firebase.ui.auth.ui.email.RegisterEmailActivity.onNewUser(SourceFile:127)
       at com.firebase.ui.auth.ui.email.CheckEmailFragment$2.onSuccess(SourceFile:208)
       at com.firebase.ui.auth.ui.email.CheckEmailFragment$2.onSuccess(SourceFile:204)
       at com.google.android.gms.tasks.zzj.run(Unknown Source)
       at android.os.Handler.handleCallback(Handler.java:733)
       at android.os.Handler.dispatchMessage(Handler.java:95)
       at android.os.Looper.loop(Looper.java:146)
       at android.app.ActivityThread.main(ActivityThread.java:5593)
       at java.lang.reflect.Method.invokeNative(Method.java)
       at java.lang.reflect.Method.invoke(Method.java:515)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1283)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1099)
       at dalvik.system.NativeStart.main(NativeStart.java)

```
We are using:
firebase-ui-auth:3.1.0
firebase-auth:11.4.2
GMS v11.4.2
gms:google-services:3.1.0
This seems to only happen on KitKat (api 19), multiple device manufacturers.

the entire crash information is available here: http://crashes.to/s/dfb96808883

We can't reproduce this, as we could not get our hands on an API19 device, but our guess is that this is an android sdk level incompatibility issue.

samtstern:
@guytavor thanks for the report and the API level information, this is definitely a strange crash and I agree with your initial assessment.

Lex74:
We have the same crash with API 17, 18, 19 after updating to
`com.google.firebase:firebase-auth:11.6.2` 
`com.android.support:appcompat-v7:27.0.2`

clydebarrow:
Seeing the same thing on API 19. It is reproduceable in the Android emulator.

samtstern:
This has been fixed and released in `3.1.3`.

thippesha:
I have updated dependency to
// FirebaseUI for Cloud Firestore
[...]

PR #1082, opened 2017-12-31
Fix crash using email authentication on API 19. Fixes issue #1036

clydebarrow:
Once more with style - using email authentication on Android API 19 results in a crash from the library. This PR resolves that issue (#1036).

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/ui/email/EmailActivity.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/email/EmailActivity.java
@@ -125,8 +125,9 @@ public class EmailActivity extends AppCompatBase implements
                     .replace(R.id.fragment_register_email, fragment, RegisterEmailFragment.TAG);
 
             if (emailLayout != null) {
-                ViewCompat.setTransitionName(emailLayout, "email_field");
-                ft.addSharedElement(emailLayout, "email_field");
+                String emailFieldName = getString(R.string.email_field_name);
+                ViewCompat.setTransitionName(emailLayout, emailFieldName);
+                ft.addSharedElement(emailLayout, emailFieldName);
             }
 
             ft.disallowAddToBackStack().commit();
--- a/auth/src/main/java/com/firebase/ui/auth/ui/email/EmailActivity.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/email/EmailActivity.java
@@ -125,7 +125,7 @@ public class EmailActivity extends AppCompatBase implements
                     .replace(R.id.fragment_register_email, fragment, RegisterEmailFragment.TAG);
 
             if (emailLayout != null) {
-                String emailFieldName = getString(R.string.email_field_name);
+                String emailFieldName = getString(R.string.fui_email_field_name);
                 ViewCompat.setTransitionName(emailLayout, emailFieldName);
                 ft.addSharedElement(emailLayout, emailFieldName);
             }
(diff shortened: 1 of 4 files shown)


SUPERCILEX:
Awesome, will check soon.

auth/src/main/java/com/firebase/ui/auth/ui/email/EmailActivity.java
```diff
@@ -123,7 +124,10 @@ public void onNewUser(User user) {
             FragmentTransaction ft = getSupportFragmentManager().beginTransaction()
                     .replace(R.id.fragment_register_email, fragment, RegisterEmailFragment.TAG);
 
-            if (emailLayout != null) ft.addSharedElement(emailLayout, "email_field");
+            if (emailLayout != null) {
```
piotrek1543: What about extracting '"email_field"' to local value as this hardcoded string was used twice

clydebarrow: Turns out it should never have been a local value as it is used in xml as well.

auth/src/main/java/com/firebase/ui/auth/ui/email/EmailActivity.java
```diff
@@ -123,7 +124,11 @@ public void onNewUser(User user) {
...
+                String emailFieldName = getString(R.string.fui_email_field_name);
+                ViewCompat.setTransitionName(emailLayout, emailFieldName);
+                ft.addSharedElement(emailLayout, emailFieldName);
```
SUPERCILEX: @clydebarrow Thanks again for the PR! I've tested your fix and it turns out view compat simply no-ops on API 19 and below. To make that clearer in _our_ code, it would be better to simply not add the transition on API 19 and below:
```java
if (Build.VERSION.SDK_INT > Build.VERSION_CODES.KITKAT && emailLayout != null) {
    ft.addSharedElement(emailLayout, getString(R.string.fui_email_field_name));
}
```

clydebarrow: According to the information I have, the support library (which is used here) provides the transition features back to API 14. I'm attempting to verify that it makes a visual difference on API 19 right now, but have been sidetracked by another issue.

SUPERCILEX: Yeah, I didn't notice a transition even with the view compat thing.

SUPERCILEX: Also, I couldn't find docs on the API 14 fragment transactions, would you mind sharing?

clydebarrow: The release notes for the support library 27.0.0 are here:
https://developer.android.com/topic/libraries/support-library/revisions.html#27-0-0

and say inter-alia "Fragment can use support library versions of Transition for fragment transitions, including shared-element transitions."

The support libraries have a minimum API level of 14.

I have confirmed that the transition on API 19 is present (at least it looks identical visually to running on API 25) so I suggest keeping the code as per my PR.

SUPERCILEX: SGTM, @samtstern should be back soon with actual push access.

samtstern:
@clydebarrow thank you for the PR and welcome to the contributors list!  LGTM, will merge into the branch.

Issue #1022, opened 2017-11-17
Allow passing full GoogleSignInOptions to auth


samtstern:
Requested in #757 and #1021 

Would allow more flexibility than just scopes.

samtstern:
Also related to #806

guillermo-varela:
Hi,

Are you guys working on this or do you need a PR?

Thanks.

samtstern:
@guillermo-varela we have not yet started working on this.  If you plan to submit a PR (which would be really cool!) could you show the proposed API here first?  As this would change the public API it needs a little more discussion than your average PR.

samtstern:
This has been fixed and released in version `3.2.0`.

PR #1083, opened 2017-12-31
Completely overhaul IDP config

SUPERCILEX:
@samtstern Ok, I am so pumped about this PR right now! 😄 The new APIs are really nice IMO and `fix` #806, #1022, and #916.

Here's the general idea for this PR: we tried going general and it didn't work. Now, we're going super specific and addressing each IDP individually which actually makes for a simple and cleaner API surface IMO. All the specific options are passed into a general `Bundle` which isn't great (I'm lazy 😄), but that's all hidden away from the consumer so we can easily change it later if someone wants to max out their Brain Processing Unit to 100%.

On top of all the huge API improvements, we're only making a _single_ very minor breaking change: if devs had set the `allowNewEmailAccounts` option _before_ setting the available providers, we unfortunately have to break them. All they have to do is move the method down! 👍

Anyway, I love the syntax and am pumped about all the new options we can add to each provider using it. I was also able to rewrite that phone params API that seemed like a quick fix. 😕

If you like the new syntax, ping me and I'll add documentation, update the README, and all those other goodies. 🚀


auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -168,6 +180,17 @@ private AuthUI(FirebaseApp app) {
         mAuth.useAppLanguage();
     }
 
+    @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+    public static void setApplicationContext(@NonNull Context context) {
```
SUPERCILEX: This is a tad bit ugly since we need our own `ContentProvider` to make it work, but it's needed to create the `GoogleSignInOptions` when retrieving an xml string. Then again, it could also be seen as a plus since the infrastructure will be there if we want to expand our magic.

SUPERCILEX:
We're even able to validate tokens/keys for IDPs in the builders now! No more weird sign-in issues complaining about incorrect setup for new devs. 😄

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -667,6 +710,25 @@ public FacebookBuilder setPermissions(@NonNull List<String> permissions) {
...
+                            "Twitter provider cannot be configured " +
+                                    "without dependency. Did you forget to add " +
+                                    "'com.twitter.sdk.android:twitter-core:VERSION' dependency?");
+                }
+
+                if (getApplicationContext().getString(R.string.twitter_consumer_key)
+                        .equals(UNCONFIGURED_CONFIG_VALUE)
+                        || getApplicationContext().getString(R.string.twitter_consumer_secret)
+                        .equals(UNCONFIGURED_CONFIG_VALUE)) {
+                    throw new RuntimeException("Twitter provider unconfigured. Make sure to add" +
```
piotrek1543: What about extracting hardcoded string values like this to `strings.xml`?

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -645,6 +666,28 @@ public IdpConfig build() {
...
+
+                if (getApplicationContext().getString(R.string.facebook_application_id)
+                        .equals(UNCONFIGURED_CONFIG_VALUE)) {
+                    throw new RuntimeException("Facebook provider unconfigured. Make sure to add" +
+                            " a `facebook_application_id` string. See the docs for more info:" +
+                            " https://github.com/firebase/FirebaseUI-Android/blob/master/auth/README.md#facebook");
+                }
+                if (getApplicationContext().getString(R.string.facebook_login_protocol_scheme)
+                        .equals("fbYOUR_APP_ID")) {
+                    Log.w("FacebookBuilder", "Facebook provider unconfigured for Chrome" +
```
piotrek1543: is there any purpose for this hardcoded `Log.w`?

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -572,22 +575,40 @@ public PhoneBuilder setDefaultPhoneNumber(String number) {
...
+            }
         }
 
         public static final class GoogleBuilder extends Builder {
             public GoogleBuilder() {
                 //noinspection deprecation taking a hit for the backcompat team
                 super(GoogleAuthProvider.PROVIDER_ID);
+                if (getApplicationContext().getString(R.string.default_web_client_id)
+                        .equals(UNCONFIGURED_CONFIG_VALUE)) {
+                    throw new RuntimeException("Check your google-services plugin configuration," +
```
piotrek1543: What about extracting hardcoded string values like this to strings.xml?

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
SUPERCILEX: Nah... you don't put developer error messages in user-facing strings. 😄

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
SUPERCILEX: Laziness? Fixed.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
SUPERCILEX: See https://github.com/firebase/FirebaseUI-Android/pull/1083#discussion_r159147678, nope.

lint/src/test/java/com/firebaseui/lint/NonGlobalIdDetectorTest.kt
```diff
@@ -40,4 +42,15 @@ class NonGlobalIdDetectorTest {
                         |+     android:id="@+id/invalid"/>
                         |""".trimMargin())
     }
+
+    companion object {
+        // Nasty hack to make lint tests pass on Windows
+        private val sdkPath = System.getProperty("java.library.path").split(';').find {
+            it.contains("SDK", true)
```
samtstern: Can you add some more comments explaining how this hack works?

lint/src/test/java/com/firebaseui/lint/NonGlobalIdDetectorTest.kt
```diff
@@ -1,14 +1,16 @@
...
 import com.android.tools.lint.checks.infrastructure.TestLintTask.lint
 import com.firebaseui.lint.NonGlobalIdDetector.Companion.NON_GLOBAL_ID
 import org.junit.Test
+import java.io.File
 
 class NonGlobalIdDetectorTest {
     @Test
-    fun validId() {
-        lint()
+    fun `Passes on valid view id`() {
```
samtstern: My thoughts on this Kotlin language feature:
https://www.youtube.com/watch?v=9nazm3_OXac

(but seems idiomatic so fine to leave it)

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -188,22 +215,7 @@ public static AuthUI getInstance(@NonNull FirebaseApp app) {
...
-
-    @RestrictTo(RestrictTo.Scope.TESTS)
-    static AuthUI getTestInstance(FirebaseApp app) {
-        AuthUI authUi;
-        synchronized (INSTANCES) {
-            authUi = INSTANCES.get(app);
-            if (authUi == null) {
-                authUi = new AuthUI(app);
-                INSTANCES.put(app, authUi);
+                INSTANCES.put(app, authUi = new AuthUI(app));
```
samtstern: nit: I find the old multi-line version much easier to understand.  I don't often see Java devs use the value of an assignment expression.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -379,28 +396,17 @@ public String getProviderId() {
             return mProviderId;
         }
 
+        @Deprecated
```
samtstern: Normally `@Deprecated` methods continue to work.  This is a breaking change for anyone using this.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -449,6 +452,7 @@ public String toString() {
              *                   AuthUI#GOOGLE_PROVIDER}. See {@link AuthUI#SUPPORTED_PROVIDERS} for
              *                   the complete list of supported Identity providers
              */
+            @Deprecated
```
samtstern: When deprecating we should always add JavaDoc pointing developers in the right direction.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -471,23 +475,247 @@ public Builder(@SupportedProvider @NonNull String providerId) {
...
+            }
+
+            @CallSuper
+            @NonNull
+            public IdpConfig build() {
+                return new IdpConfig(mProviderId, mParams);
+            }
+        }
+
+        public static final class EmailBuilder extends Builder {
```
samtstern: Public classes should have JavaDoc.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
SUPERCILEX: SGTM

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
SUPERCILEX: Of course, I haven't added any documentation anywhere. I want to make sure you like the new APIs first. 😄

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
SUPERCILEX: Yep, see https://github.com/firebase/FirebaseUI-Android/pull/1083#discussion_r159293138.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
SUPERCILEX: Fixed.

auth/src/main/java/com/firebase/ui/auth/ui/email/EmailActivity.java
```diff
@@ -116,7 +119,10 @@ public void onNewUser(User user) {
 
         TextInputLayout emailLayout = findViewById(R.id.email_layout);
 
-        if (getFlowParams().allowNewEmailAccounts) {
+        List<AuthUI.IdpConfig> providerConfig = getFlowParams().providerInfo;
+        int emailIndex = providerConfig.indexOf(new AuthUI.IdpConfig.EmailBuilder().build());
```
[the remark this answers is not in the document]
SUPERCILEX: Not really since that's the customizable order used to display the providers in the picker activity. However, I've cleaned this up to manually find the config with a for loop which is clearer.

auth/src/main/java/com/firebase/ui/auth/ui/phone/VerifyPhoneNumberFragment.java
```diff
@@ -143,15 +142,21 @@ public void onActivityCreated(@Nullable Bundle savedInstanceState) {
...
+            phone = params.getString(ExtraConstants.EXTRA_PHONE);
+            countryCode = params.getString(ExtraConstants.EXTRA_COUNTRY_CODE);
+            nationalNumber = params.getString(ExtraConstants.EXTRA_NATIONAL_NUMBER);
         }
         if (!TextUtils.isEmpty(countryCode) && !TextUtils.isEmpty(nationalNumber)) {
             // User supplied country code & national number
             PhoneNumber phoneNumber = PhoneNumberUtils.getPhoneNumber(countryCode, nationalNumber);
             setPhoneNumber(phoneNumber);
             setCountryCode(phoneNumber);
+        } else if (!TextUtils.isEmpty(countryCode)) {
```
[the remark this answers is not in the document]
SUPERCILEX: This is to fix #916, see c9300b4cb37d26d54a6bf818f0346eb5daa02740. I'd be happy to remove the new `GoogleSignInOptions` and phone builder options and add them back in another PR. I was using this PR as a proof-of-concept to ensure we can build new APIs with the new builders.

auth/src/main/java/com/firebase/ui/auth/util/data/PhoneNumberUtils.java
```diff
@@ -147,7 +147,7 @@ public static Integer getCountryCode(String countryIso) {
                 ? null : COUNTRY_TO_ISO_CODES.get(countryIso.toUpperCase(Locale.getDefault()));
     }
 
-    private static String countryIsoForCountryCode(String countryCode) {
+    public static String getCountryIsoForCountryCode(String countryCode) {
```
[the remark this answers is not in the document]
SUPERCILEX: See https://github.com/firebase/FirebaseUI-Android/pull/1083#discussion_r159302416.

auth/src/test/java/com/firebase/ui/auth/testhelpers/TestHelper.java
```diff
@@ -68,18 +73,35 @@ public static FirebaseUser getMockFirebaseUser() {
...
         return new FlowParameters(
                 FIREBASE_APP_NAME,
                 idpConfigs,
                 AuthUI.getDefaultTheme(),
                 AuthUI.NO_LOGO,
-                null  /* tosUrl */,
-                null  /* privacyPolicyUrl */,
-                true  /* credentialPickerEnabled */,
-                true  /* hintSelectorEnabled */,
-                true  /* allowNewEmailAccounts */);
```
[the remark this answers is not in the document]
SUPERCILEX: Oh, makes sense. I usually just click at the end of a line and look at the bottom right of the IDE for column numbers. You should try it, those hints are really helpful! 😄

lint/src/test/java/com/firebaseui/lint/NonGlobalIdDetectorTest.kt
SUPERCILEX: Haha, same. Yeah, this is taken straight from the Intellij [code style](https://kotlinlang.org/docs/reference/coding-conventions.html):
image

lint/src/test/java/com/firebaseui/lint/NonGlobalIdDetectorTest.kt
SUPERCILEX: Done.

auth/src/main/java/com/firebase/ui/auth/ui/phone/VerifyPhoneNumberFragment.java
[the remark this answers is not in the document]
samtstern: We can keep it here, I didn't see the connection but now I do.

auth/src/test/java/com/firebase/ui/auth/testhelpers/TestHelper.java
```diff
@@ -15,44 +15,105 @@
...
+            Object o;
+            try {
+                o = field.get(null);
+            } catch (IllegalAccessException e) {
+                throw new IllegalStateException(e);
+            } catch (NullPointerException e) {
+                continue;
+            }
+
+            if (o instanceof Map) {
```
samtstern: What is going on here ... I need some more comments and probably a justification for this extreme hackery haha

auth/src/test/java/com/firebase/ui/auth/testhelpers/TestHelper.java
```diff
@@ -15,44 +15,105 @@
...
+                FirebaseAuth.getInstance(app);
+                for (String id : instances.keySet()) {
+                    instances.put(id, mock(FirebaseAuth.class));
+                }
+
+                break;
+            }
         }
+
+        when(FirebaseAuth.getInstance().setFirebaseUIVersion(anyString())).thenReturn(null);
```
samtstern: nit: `thenReturn(Tasks.forResult(null))`

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -399,7 +400,23 @@ public String getProviderId() {
...
+                for (Scope scope : array) {
+                    scopes.add(scope.toString());
+                }
+                permissions = scopes;
+            } else if (mProviderId.equals(FacebookAuthProvider.PROVIDER_ID)) {
+                permissions = mParams.getStringArrayList(ExtraConstants.EXTRA_FACEBOOK_PERMISSIONS);
+            } else {
+                permissions = null;
+            }
+            return permissions == null ? Collections.<String>emptyList() : permissions;
```
samtstern: nit: why not just start out by making `permissions` the empty list?

samtstern:
https://github.com/firebase/FirebaseUI-Android/tree/version-3.2.0-dev

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
SUPERCILEX: I wish, but the Facebook permissions have always been nullable (so I kept it that way) and this null check at the end ensures we catch all the cases.

auth/src/test/java/com/firebase/ui/auth/testhelpers/TestHelper.java
SUPERCILEX: Whoops, yeah. This came about due to my frustration at not being able to properly spy/mock the auth instance. The method basically injects a mock FirebaseAuth into the instances map. I've added some documentation.

In case you were wondering, it's what lets us get rid of `AuthUI.getTestInstance`. I prefer clean production code and dirty tests than dirty prod code and clean tests. Hope that's cool with you! 😄

auth/src/test/java/com/firebase/ui/auth/testhelpers/TestHelper.java
SUPERCILEX: 👍 done.

auth/src/test/java/com/firebase/ui/auth/testhelpers/TestHelper.java
```diff
@@ -15,44 +15,120 @@
...
+    }
+
+    private static void initializeProviders() {
+        RuntimeEnvironment.application = spy(RuntimeEnvironment.application);
+        when(RuntimeEnvironment.application.getApplicationContext())
+                .thenReturn(RuntimeEnvironment.application);
+        Resources spiedResources = spy(RuntimeEnvironment.application.getResources());
+        when(RuntimeEnvironment.application.getResources()).thenReturn(spiedResources);
+
+        when(spiedResources.getString(R.string.default_web_client_id)).thenReturn("abc");
```
samtstern: nit: we might want to extract this into a test helper that:

  * Always spies the runtime application context
  * Provides a static method for resource override

Seems generally useful

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -469,25 +546,265 @@ public Builder(@SupportedProvider @NonNull String providerId) {
...
+
+            /**
+             * Set the default phone number that will be used to populate the phone verification
+             * sign-in flow.
+             *
+             * @param iso    the phone number's country code
+             * @param number the phone number in local format
+             */
+            @NonNull
+            public PhoneBuilder setDefaultNumber(@NonNull String iso, @NonNull String number) {
```
samtstern: We need some validation/documentation on choosing an ISO code that we will accept.  I wouldn't know where to start getting one.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -469,25 +546,265 @@ public Builder(@SupportedProvider @NonNull String providerId) {
...
+         * {@link IdpConfig} builder for the Google provider.
+         */
+        public static final class GoogleBuilder extends Builder {
+            public GoogleBuilder() {
+                //noinspection deprecation taking a hit for the backcompat team
+                super(GoogleAuthProvider.PROVIDER_ID);
+                if (getApplicationContext().getString(R.string.default_web_client_id)
+                        .equals(UNCONFIGURED_CONFIG_VALUE)) {
+                    throw new RuntimeException("Check your google-services plugin configuration," +
+                            " the client id string wasn't populated.");
```
samtstern: nit: error message should specifically mention `default_web_client_id`)

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -469,25 +546,265 @@ public Builder(@SupportedProvider @NonNull String providerId) {
...
+
+            /**
+             * Set the {@link GoogleSignInOptions} to be used for Google sign-in. Standard options
+             * like requesting the user's email will automatically be added.
+             *
+             * @param options sign-in options
+             */
+            @NonNull
+            public GoogleBuilder setSignInOptions(@NonNull GoogleSignInOptions options) {
+                if (getParams().containsKey(ExtraConstants.EXTRA_GOOGLE_SIGN_IN_OPTIONS)) {
```
samtstern: nit: Seems like we are doing checks of this style a lot, could be refactored into a `checkUnset(key, errorMsg)` method.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -469,25 +546,265 @@ public Builder(@SupportedProvider @NonNull String providerId) {
...
+                    //noinspection unused to possibly throw
+                    Class c = com.facebook.FacebookSdk.class;
+                } catch (NoClassDefFoundError e) {
+                    throw new RuntimeException(
+                            "Facebook provider cannot be configured " +
+                                    "without dependency. Did you forget to add " +
+                                    "'com.facebook.android:facebook-login:VERSION' dependency?");
+                }
+
+                if (getApplicationContext().getString(R.string.facebook_application_id)
```
samtstern: Same comment here, we should extract checking resources against the unconfigured value into a helper method.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -635,8 +930,10 @@ public T setProviders(@NonNull List<IdpConfig> idpConfigs) {
          * <p>SmartLock is enabled by default.
          *
          * @param enabled enables smartlock's credential selector and hint selector
+         * @deprecated use the more customizable {@link #setIsSmartLockEnabled(boolean, boolean)}
```
samtstern: I don't see a reason to deprecate this "disable all" convenience version.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -685,13 +982,35 @@ private SignInIntentBuilder() {
...
         }
 
+        @NonNull
+        @Override
+        public Intent build() {
+            if (mAllowNewEmailAccounts != null) {
+                // To ensure setAllowNewEmailAccounts backcompat
+                for (IdpConfig provider : mProviders) {
+                    if (provider.getProviderId().equals(EmailAuthProvider.PROVIDER_ID)) {
+                        mProviders.set(mProviders.indexOf(provider), new IdpConfig.EmailBuilder()
```
samtstern: If the loop contains `indexOf` it should probably be a loop over indexes, not a for-each loop.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
SUPERCILEX: Done.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
SUPERCILEX: Done.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
SUPERCILEX: Done.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
SUPERCILEX: Agreed, I don't remember why I deprecated it. 😕

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
SUPERCILEX: 👍

auth/src/main/java/com/firebase/ui/auth/ui/phone/VerifyPhoneNumberFragment.java
```diff
@@ -140,18 +139,24 @@ public void onActivityCreated(@Nullable Bundle savedInstanceState) {
...
         }
-        if (!TextUtils.isEmpty(countryCode) && !TextUtils.isEmpty(nationalNumber)) {
+        if (!TextUtils.isEmpty(countryIso) && !TextUtils.isEmpty(nationalNumber)) {
             // User supplied country code & national number
-            PhoneNumber phoneNumber = PhoneNumberUtils.getPhoneNumber(countryCode, nationalNumber);
+            PhoneNumber phoneNumber = PhoneNumberUtils.getPhoneNumber(countryIso, nationalNumber);
             setPhoneNumber(phoneNumber);
             setCountryCode(phoneNumber);
+        } else if (!TextUtils.isEmpty(countryIso)) {
+            // User supplied full phone number
```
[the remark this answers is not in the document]
SUPERCILEX: Oops, thx!

auth/src/test/java/com/firebase/ui/auth/testhelpers/TestHelper.java
SUPERCILEX: SGTM!

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
SUPERCILEX: Good point, we didn't have validation either before this PR. Let me see if I can cook something up. 😄

SUPERCILEX: Phone auth is so hard... I think I got it though. 😆

Issue #1095, opened 2018-01-09
Refactor welcome back password flow use new architecture


samtstern:
@SUPERCILEX I am going to take a stab at one of these refactors.

samtstern:
Note that this also means I will be converting `SaveSmartLock` which should be a fun challenge.

samtstern:
This will also encompass #1007

SUPERCILEX:
@samtstern awesome, can't wait to see how you solve this! I'll keep my fork up to date with the new changes so we can pool ideas. 💡

samtstern:
This has been fixed and released in version `3.2.0`.

PR #1096, opened 2018-01-09
Refactor welcome-back password flow, add ViewModel tests

samtstern:
This PR does a few things:

  * Removes almost all uses of `GoogleApiClient`
  * Refactors the WelcomeBackPasswordPrompt to the new architecture
  * Adds testing for the new viewmodels as well as the ones introduced in the previous refactor

This `fixes` #1095 and #1102


samtstern:
@SUPERCILEX take a look at what I've got so far ... as far as I can tell it works although I know I am missing some edge cases / need to clean up all my TODOs.  It's much easier to reason about and is the first step in killing `SaveSmartLock` (easily my least favorite class in the lib)

samtstern:
Testing is SO MUCH EASIER NOW

auth/src/main/java/com/firebase/ui/auth/ui/email/WelcomeBackPasswordPrompt.java
```diff
@@ -138,64 +202,18 @@ private void validateAndSignIn(final String email, final String password) {
...
+    public void onClick(View view) {
+        final int id = view.getId();
+        if (id == R.id.button_done) {
+            validateAndSignIn();
+        } else if (id == R.id.trouble_signing_in) {
+            onForgotPasswordClicked();
         }
-
-        // Sign in with known email and the password provided
-        getAuthHelper().getFirebaseAuth()
```
samtstern: Mwahahaha I love deleting these.

auth/src/main/java/com/firebase/ui/auth/util/data/AuthViewModelBase.java
```diff
@@ -30,4 +36,19 @@ protected FirebaseAuth getAuth() {
     protected PhoneAuthProvider getPhoneAuth() {
         return mPhoneAuth;
     }
+
+    protected CredentialsClient getCredentialsClient() {
+        return mCredentialsClient;
+    }
+
+    @VisibleForTesting
+    public void initializeForTesting(FlowParameters parameters,
```
samtstern: Would like to not need this but I can't see a way to test the ViewModels without it given how ViewModelBase handles initialization.

auth/src/main/java/com/firebase/ui/auth/viewmodel/email/WelcomeBackPasswordHandler.java
```diff
@@ -0,0 +1,183 @@
...
+                                    .continueWithTask(new ProfileMerger(outputResponse))
+                                    .addOnFailureListener(new TaskFailureLogger(TAG, "linkWithCredential+merge failed."));
+                        }
+                    }
+                })
+                .addOnCompleteListener(new OnCompleteListener<AuthResult>() {
+                    @Override
+                    public void onComplete(@NonNull Task<AuthResult> task) {
+                        if (!task.isSuccessful()) {
+                            mSignInLiveData.setValue(new Resource<IdpResponse>(task.getException()));
```
samtstern: It seems (from reading/ripping up SmartLockBase) we basically consider SmartLock operations to be best effort.

samtstern:
Adding more tests ... it's fun to add tests when you actually have separated business logic mwaahahaha

samtstern:
@SUPERCILEX thanks for the review!  I totally see what you mean, we're going to need a lot of this as common.

I still have not seen good, general advice on how to scope ViewModels.  Do you have one per View?  Do you organize them by function  and include more than one in a view? 

We also need to move towards the "repository" pattern where common data operations are handled in Repository objects that the ViewModel hold onto.  I think that's like `SignInHandler`.

Anyway I will address your feedback on Monday, enjoy the weekend!

SUPERCILEX:
Agreed, I'm not sure how to organize the view models either... maybe you could ask Yigit on the way to your desk? 😆 Also agreed for the repository pattern, that would definitely make `SignInHandler`-esk objects less bloated.

SGTM, see you on Tuesday and enjoy the weekend! 😄

auth/src/main/java/com/firebase/ui/auth/ui/email/WelcomeBackPasswordPrompt.java
```diff
@@ -106,21 +104,87 @@ protected void onCreate(Bundle savedInstanceState) {
...
+        }
+
+        try {
+            startIntentSenderForResult(resolution.getPendingIntent().getIntentSender(),
+                    resolution.getRequestCode(), null, 0, 0, 0);
+        } catch (IntentSender.SendIntentException e) {
+            Log.e(TAG, "Failed to send resolution.", e);
+
+            // TODO: Should this include the code and data?
+            finish();
```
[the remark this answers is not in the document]
samtstern: Done.  Do we use `RESULT_OK` with an error `IdpResponse` here?

samtstern:
Two remaining things:

1. Password in `IdpResponse`
2. Less logic in the ViewModel

For (1) I am not so sure it's a good idea just to make the internal code healthier.  I don't love giving developers back raw passwords, they may not handle it safely and nobody is asking for it.

For (2) I agree completely that a refactor will happen, but I would like to do it as-needed rather than doing it right now so that we get only the minimum level of abstraction.  Thanks for calling it out though.

WDYT?

auth/src/main/java/com/firebase/ui/auth/ui/email/WelcomeBackPasswordPrompt.java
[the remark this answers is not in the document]
SUPERCILEX: No? If we're going to extract this out later, I think we'd want it to return a failure code. Then again, I dunno what developers would want... 😊

auth/src/main/java/com/firebase/ui/auth/viewmodel/SingleLiveEvent.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @MainThread
+    public void observe(LifecycleOwner owner, final Observer<T> observer) {
+
+        if (hasActiveObservers()) {
+            Log.w(TAG, "Multiple observers registered but only one will be notified of changes.");
+        }
+
+        // Observe the internal MutableLiveData
+        super.observe(owner, new Observer<T>() {
```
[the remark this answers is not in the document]
samtstern: Oh good call!  I have added some `UnsupportedOperationException`s for those methods so that we can leave it simple for now without eventually shooting ourselves in the foot.

auth/src/main/java/com/firebase/ui/auth/viewmodel/SingleLiveEvent.java
```diff
@@ -46,6 +47,21 @@ public void onChanged(@Nullable T t) {
...
     }
 
+    @Override
+    public void observeForever(@NonNull Observer<T> observer) {
+        throw new UnsupportedOperationException("obseveForever not supported.");
+    }
+
+    @Override
+    public void removeObserver(@NonNull Observer<T> observer) {
+        throw new UnsupportedOperationException("removeObserver not supported.");
```
SUPERCILEX: Awww... I really wish this would work, but I'm pretty sure it'll crash if you rotate the activity (I could be wrong, so it would be a good idea to try it). Here's the source code:
```java
        public final LifecycleOwner owner;
        public final Observer<T> observer;

        LifecycleBoundObserver(LifecycleOwner owner, Observer<T> observer) {
            this.owner = owner;
            this.observer = observer;
        }

        @Override
        public void onStateChanged(LifecycleOwner source, Lifecycle.Event event) {
            if (owner.getLifecycle().getCurrentState() == DESTROYED) {
                removeObserver(observer); // <-- BOOM!
                return;
            }
        }
```

samtstern: You're right :-(

Ok in the overarching spirit of "deal with it when we get there" I have just added a class comment.

samtstern: I also overrode the methods so I could mark them `@Deprecated` which doesn't really help but can't hurt.

SUPERCILEX: Haha, SGTM 👍

Issue #1098, opened 2018-01-09
Wrong package name for ProviderUtils


lazyFloyd:
Just noticed that the [ProviderUtils](https://github.com/firebase/FirebaseUI-Android/blob/4b98a9853bf3cee9d0a32312ecf1897f740d1547/auth/src/main/java/com/firebase/ui/auth/provider/ProviderUtils.java) class has either the wrong package name or is located in the wrong place.

![firebase_ui](https://user-images.githubusercontent.com/23117238/34724439-177a4b5a-f54e-11e7-8950-08ae1cbad131.png)

samtstern:
@lazyFloyd nice catch!

samtstern:
This has been fixed and released in version `3.2.0`.

PR #1099, opened 2018-01-09
Move ProviderUtils to the correct package

samtstern:
See issue #1098


SUPERCILEX:
Shouldn't that throw a compiler error?

samtstern:
@SUPERCILEX I know ... but apparently we got away with it.  One more Java/Android mystery.

Issue #1103, opened 2018-01-10
Refactor/Remove GoogleSignInHelper


samtstern:
The class `GoogleSignInHelper` is mostly an abstraction over `GoogleApiClient` and its connection process.  With the new `GoogleSignInClient` this should not really be necessary and we will reduce our bug surface.

samtstern:
This has been fixed and released in version `3.2.0`.

PR #1109, opened 2018-01-19
Kill all instances of GoogleApiClient

samtstern:
This `fixes` #1103 and completes the `3.2.0` milestone.

It also makes me extremely happy.


aaronj1335:
um yes

samtstern:
(for context to @SUPERCILEX , @aaronj1335 was the lead for the "connectionless" effort that introduced the `{Foo}Client` pattern)

SUPERCILEX:
@samtstern what!?? No way! @aaronj1335 thank you thank you thank you!!! That's the best refactor Play Services has received in ages!

@samtstern not to be a party pooper, but do you remember when I complained about having to use that credentials builder reflection hack? Well, we need to use what it was trying to do (thanks to the PS team for removing the need for reflection!) to get our credential saving logic to work correctly. In O+, the save credentials dialog is [disabled by default](https://developers.google.com/android/reference/com/google/android/gms/auth/api/credentials/CredentialsOptions.Builder.html#forceEnableSaveDialog()) if autofill is available. However, this doesn't work with our UX since we check for saved emails before bringing the user to the email sign in/up page which means autofill will never be able to do its magic. Thus, we have to stick with Play Services autofill. Does that make sense?

samtstern:
@SUPERCILEX so you want to add this?

```java
CredentialsOptions options = new CredentialsOptions.Builder()
              .forceEnableSaveDialog()
              .build()
```

SUPERCILEX:
@samtstern Yup, that's what I'm doing. 😄

PR #1110, opened 2018-01-19
Version 3.2.0

samtstern:
This is the release PR for `3.2.0`.  Will probably merge Monday so as not to release-and-bail on a Friday afternoon.

PR #1111, opened 2018-01-20
Fix credential saving bugs and crashes

SUPERCILEX:
See https://github.com/firebase/FirebaseUI-Android/pull/1110#issuecomment-359128489, explantory comments coming in a review 👍.

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailFragment.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailFragment.java
@@ -26,7 +26,6 @@ import com.firebase.ui.auth.ui.TaskFailureLogger;
 import com.firebase.ui.auth.ui.idp.WelcomeBackIdpPrompt;
 import com.firebase.ui.auth.util.ExtraConstants;
 import com.firebase.ui.auth.util.data.ProviderUtils;
-import com.firebase.ui.auth.util.signincontainer.SaveSmartLock;
 import com.firebase.ui.auth.util.ui.ImeHelper;
 import com.firebase.ui.auth.util.ui.PreambleHandler;
 import com.firebase.ui.auth.util.ui.fieldvalidators.BaseValidator;
@@ -67,7 +66,6 @@ public class RegisterEmailFragment extends FragmentBase implements
     private EmailFieldValidator mEmailFieldValidator;
     private PasswordFieldValidator mPasswordFieldValidator;
     private BaseValidator mNameValidator;
-    private SaveSmartLock mSaveSmartLock;
 
     private User mUser;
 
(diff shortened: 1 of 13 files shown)


auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailFragment.java
```diff
@@ -267,9 +257,7 @@ private void registerUser(final String email, final String name, final String pa
                 .addOnSuccessListener(getActivity(), new OnSuccessListener<AuthResult>() {
                     @Override
                     public void onSuccess(AuthResult authResult) {
-                        SaveSmartLock saveSmartLock = getSaveSmartLock();
```
SUPERCILEX: Here's the fatal flaw that let us catch the issue: as you'll see below, in the other spots we do this, it works because we initialize the fragment _before_ starting a sign-in. Thus, the fragment has time to do its thing by the time it's actually used (since network calls are slow). However, in this case, we immediately use the fragment below and go BOOM! 😁

The solution used in this PR: add a lifecycle listener that kicks off the process in `onStart` when we are guaranteed to have an activity. Doing that fixes #856!

auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
```diff
@@ -175,7 +163,6 @@ public void onSuccess(IdpResponse response) {
                 .signInWithCredential(credential)
                 .addOnCompleteListener(new CredentialSignInHandler(
                         this,
-                        getSaveSmartLock(),
```
SUPERCILEX: Here's where it works because we initialize the fragment before Auth can give us a callback.

auth/src/main/java/com/firebase/ui/auth/util/signincontainer/SaveSmartLock.java
```diff
@@ -143,29 +142,27 @@ public void saveCredentialsOrFinish(FirebaseUser firebaseUser,
...
 
         if (credential == null) {
             Log.e(TAG, "Unable to save null credential!");
             finish();
             return;
         }
 
-        getCredentialsClient().save(credential)
-                .addOnCompleteListener(this);
+        getLifecycle().addObserver(new LifecycleObserver() {
```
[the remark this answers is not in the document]
SUPERCILEX: 1. Good point, I've fixed this.
1. Nope! Yigit is a genius: lifecycle listeners are brought up-to-date with the latest state so it doesn't matter when we register them.

samtstern:
@SUPERCILEX SGTM, thanks for catching and fixing!  Hopefully we can kill this fragment soon anyway.

Issue #1113, opened 2018-01-22
"Resend Code" button(turned blue i.e. enabled) in Phone Auth not working in 3.1.3


kshivang:
* Android device: MI4W (Problem is common for all devices)
  * Android OS version: 8.1.0
  * Google Play Services version: 3.1.1
  * Firebase/Play Services SDK version: 11.8.0
  * FirebaseUI version: 3.1.3
  
#### Steps to reproduce:

  1. Integrate signIn via phone authentication
  2. Fill your phone no and continue
  3. If you get OTP within 20 sec no problem but for some reason if you not receive OTP within 20 sec Resend Code button will turn to blue (from grey) but remain still unclickable.

#### Observed Results:

  * After updating from FirebaseUI version from 3.1.2(FirebaseVersion 11.6.2) to 3.1.3(FirebaseVersion 11.8.0), I noticed this error that "Resend Code" button turns blue(i.e. Enabled) but not working. In 3.1.2 on clicking "Resend Code" a dialog opens momentarily with message "OTP sent" and time counter reset  to 20 sec but in version 3.1.3 on clicking "Resend Code" button (enabled) nothing happens.
  
#### Relevant Code:

Code I used to integrate PhoneAuth

  ```
 startActivityForResult(
                AuthUI.getInstance()
                        .createSignInIntentBuilder()
                        .setAvailableProviders(
                                Collections.singletonList(new AuthUI.IdpConfig
                                        .Builder(AuthUI.PHONE_VERIFICATION_PROVIDER).build()))
                        .setTheme(R.style.FirebaseUITheme)
                        .build(),
                RC_SIGN_IN);
  ```

samtstern:
This has been fixed and released in version `3.2.0`.

PR #1114, opened 2018-01-22
Fix phone auth UI bugs

SUPERCILEX:
Fixes https://github.com/firebase/FirebaseUI-Android/pull/1110#issuecomment-359542203 and #1113. Already filed: https://issuetracker.google.com/issues/72325031

### DIFF
--- a/auth/build.gradle
+++ b/auth/build.gradle
@@ -36,7 +36,7 @@ android {
 dependencies {
     implementation "com.android.support:design:$supportLibraryVersion"
     implementation "com.android.support:customtabs:$supportLibraryVersion"
-    implementation 'com.android.support.constraint:constraint-layout:1.1.0-beta4'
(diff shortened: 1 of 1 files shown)


samtstern:
@SUPERCILEX wow I would have never figured out that we needed to downgrade.  Thanks for filing the issue as welll!

samtstern:
What's the relation between this and #1113?  Doesn't seem like it would affect button enablement?

SUPERCILEX:
@samtstern No problem! As for #1113, I just know that it didn't work with beta4 and does with beta3. If I had to guess, the TextView (without text since we had that bug) is invisible but still covering the resend button. When I clicked the button with beta4, nothing happened at all so while that bug report is about the button not updating itself, that's because it wasn't receiving the click in the first place. (I think 😊)

PR #1121, opened 2018-01-23
Don't show sign-in suggestions before picker activity

SUPERCILEX:
So it turns out resolutions can either be [`SIGN_IN_REQUIRED`](https://developers.google.com/android/reference/com/google/android/gms/common/api/CommonStatusCodes.html#SIGN_IN_REQUIRED) or [`RESOLUTION_REQUIRED`](https://developers.google.com/android/reference/com/google/android/gms/common/api/CommonStatusCodes.html#RESOLUTION_REQUIRED) and we _only_ want the `RESOLUTION_REQUIRED` one.


samtstern:
@SUPERCILEX the fact that I was the original author `ResolvableApiException` and **still** got this wrong makes me so sad.

Thanks for sorting that out!

SUPERCILEX:
Nooooooooo! Awww, sorry man! 😆

Issue #1123, opened 2018-01-23
Refactor (kill) SaveSmartLock


samtstern:
Just kinda marking my territory here so we don't do the same work, but I am going to kill `SaveSmartLock` and `SmartLockBase` once and for all.

samtstern:
This has been fixed and released in version `3.2.2`.

PR #1124, opened 2018-01-24
Kill smartlock fragments

samtstern:
WIP for #1123 💀


samtstern:
@SUPERCILEX ok I just got into exactly the situation you said would happen: I've got identical code in two ViewModels. 

How would you unify `SmartLockViewModel` and `WelcomeBackPasswordHandler` in your proposed design?

SUPERCILEX:
I basically had a God class that took in an `IdpResponse` and handled everything from there. The main issue was that it needed to stay alive even if activities on top of the stack forced lower ones to die. I think a solution to that (which would remove my gross static fields) would be to just store the `IdpResponse` in the saved state and then restore the God class to its correct state. To fix the God class, I think we could create bunch of repository classes that are in charge of individual stuff like idp, email, and phone auth and saving the credential. Then we would just need a view model to coordinate all that.

Then again, do you have any better ideas? 😊

samtstern:
@SUPERCILEX I got rid of the duplicated logic.  There aren't really any new abstractions here, the view just talks to multiple view models and calls one when the other is finished.

I'd like to remove all the logic from the view, but I couldn't think of a good way to do that right now.  At least this continues the refactor without yet having any duplicated logic.  Imo that's the minimum we can do at the moment.

samtstern:
@SUPERCILEX any thoughts on the recent changes?

SUPERCILEX:
Will take a look at this tomorrow. 👍

auth/src/main/java/com/firebase/ui/auth/data/model/Resource.java
```diff
@@ -18,10 +18,29 @@
     private final T mValue;
     private final Exception mException;
 
+    /**
+     * Creates a successful Resource&lt;Void&gt;.
```
SUPERCILEX: `Creates a successful, empty Resource.` would be clearer to read in the source code.

auth/src/main/java/com/firebase/ui/auth/data/model/Resource.java
```diff
@@ -43,12 +62,18 @@ public Resource(@NonNull T value) {
      *
      * @param exception error in computing the result
      */
-    public Resource(@NonNull Exception exception) {
+    Resource(@NonNull Exception exception) {
```
SUPERCILEX: Those static factory methods look great! Is there any reason to keep these constructors at all?

auth/src/main/java/com/firebase/ui/auth/data/model/Resource.java
```diff
@@ -43,12 +62,18 @@ public Resource(@NonNull T value) {
...
         mState = State.FAILURE;
         mValue = null;
         mException = Preconditions.checkNotNull(exception, "Failure state cannot have null error.");
     }
 
+    Resource(State state, T value, Exception exception) {
```
SUPERCILEX: Can this be private?

auth/src/main/java/com/firebase/ui/auth/ui/HelperActivityBase.java
```diff
@@ -84,19 +142,42 @@ public void finish(int resultCode, Intent intent) {
...
+
+    private void onPendingResolution(@NonNull PendingResolution resolution) {
+        try {
+            startIntentSenderForResult(
+                    resolution.getPendingIntent().getIntentSender(),
+                    resolution.getRequestCode(),
+                    null, 0, 0, 0);
+        } catch (IntentSender.SendIntentException e) {
+            Log.e(TAG, "STATUS: Failed to send resolution.", e);
+            finish(RESULT_OK, mPendingIdpResponse.toIntent());
```
SUPERCILEX: Because this implementation is specific to Smart Lock being failure tolerant, I feel like we should just inline it or add a comment specifying that it will finish the activity should an error occur. What do you think?

auth/src/main/java/com/firebase/ui/auth/ui/HelperActivityBase.java
```diff
@@ -84,19 +142,42 @@ public void finish(int resultCode, Intent intent) {
...
+    }
+
+    private void onSaveOperation(@NonNull Resource<Void> resource) {
+        switch (resource.getState()) {
+            case LOADING:
+                // No-op?
+                break;
+            case SUCCESS:
+            case FAILURE:
+                finish(RESULT_OK, mPendingIdpResponse.toIntent());
```
SUPERCILEX: Same here, this seems too specific to Smart Lock to be in `HelperActivityBase` 🤔.

auth/src/main/java/com/firebase/ui/auth/ui/email/WelcomeBackPasswordPrompt.java
```diff
@@ -79,6 +80,10 @@ protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setContentView(R.layout.fui_welcome_back_password_prompt_layout);
 
+        if (savedInstanceState != null) {
+            mPendingPassword = savedInstanceState.getString(ExtraConstants.EXTRA_PASSWORD, null);
+        }
```
SUPERCILEX: Maybe I'm not fully understanding the idea behind these pending data, but wouldn't it make more sense to keep those around in the `ViewModel`? That seems to be what you're trying to do... 🤷‍♂️

auth/src/main/java/com/firebase/ui/auth/ui/email/WelcomeBackPasswordPrompt.java
```diff
@@ -121,45 +126,44 @@ public void onChanged(@Nullable PendingResolution resolution) {
...
-            getDialogHolder().dismissDialog();
-            Log.d(TAG, "onAuthResult:SUCCESS:" + resource.getValue());
+        switch (resource.getState()) {
+            case LOADING:
+                getDialogHolder().showLoadingDialog(R.string.fui_progress_dialog_signing_in);
+                break;
+            case SUCCESS:
+                getDialogHolder().dismissDialog();
+
+                // TODO: Should this logic be in the View?  If not how do we link the view models?
```
SUPERCILEX: I think it makes sense to have this in the View since it's acting as the glue binding all the different UI components together: we're moving from sign-in to credential input  which are two different "screens".

auth/src/main/java/com/firebase/ui/auth/util/signincontainer/SaveSmartLock.java
```diff
@@ -1,168 +0,0 @@
...
-import com.firebase.ui.auth.util.CredentialsUtil;
-import com.firebase.ui.auth.util.GoogleApiUtils;
-import com.google.android.gms.auth.api.credentials.Credential;
-import com.google.android.gms.auth.api.credentials.CredentialsClient;
-import com.google.android.gms.common.api.ResolvableApiException;
-import com.google.android.gms.tasks.Task;
-import com.google.firebase.auth.FirebaseUser;
-
-@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-public class SaveSmartLock extends SmartLockBase<Void> {
```
SUPERCILEX: YAAAASSSSSSS! Thank you! 👏🙌🎉🥂😁

auth/src/main/java/com/firebase/ui/auth/util/signincontainer/SignInDelegate.java
```diff
@@ -54,7 +56,9 @@
  * email is supported, in which case the {@link EmailActivity} is started.
  */
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-public class SignInDelegate extends SmartLockBase<CredentialRequestResponse> {
+public class SignInDelegate extends FragmentBase
```
SUPERCILEX: Tick, tok, tick, tok, tick, tok... 😏

auth/src/main/java/com/firebase/ui/auth/viewmodel/smartlock/SmartLockHandler.java
```diff
@@ -0,0 +1,101 @@
...
+     * Forward the result of a resolution from the Activity to the ViewModel.
+     */
+    @Override
+    public boolean onActivityResult(int requestCode, int resultCode, Intent data) {
+        if (requestCode == RC_SAVE) {
+            if (resultCode == Activity.RESULT_OK) {
+                mResultLiveData.setValue(Resource.forVoidSuccess());
+            } else {
+                Log.e(TAG, "SAVE: Canceled by user.");
+                mResultLiveData.setValue(Resource.<Void>forFailure(new Exception("Save canceled by user.")));
```
SUPERCILEX: We should probably use `FirebaseUiException`.

auth/src/main/java/com/firebase/ui/auth/ui/email/WelcomeBackPasswordPrompt.java
samtstern: Doh. Thank you.  I still have some bad old instincts from the pre-ViewModel dark ages.

samtstern:
@SUPERCILEX thanks for taking a look.   Still trying to untangle all this.

You just made me think: should smartlock just actually be another view?  It could be a transparent Activity that we start for result whenever the flow needs to end in a save.

It seems like that would clean up everything, but want to get your thoughts before I do anything that crazy.

samtstern:
@SUPERCILEX I prototyped something here:
https://github.com/firebase/FirebaseUI-Android/pull/1151

SUPERCILEX:
@samtstern Actually, yeah, #1151 makes a lot of sense: if we're going to have everything be cleanly separated by component, then Smart Lock _should_ be in its own activity. #1151 certainly makes `HelperActivityBase` nicer! 😉

samtstern:
@SUPERCILEX ok I am going to go forward with that approach then, thanks for the sanity check.  And sorry this PR is getting so big :man_shrugging:

SUPERCILEX:
Haha, no prob!

samtstern:
@SUPERCILEX ok this one is **almost** done except for some rotational issues with the `CredentialSaveActivity`.  The save dialog currently stacks if you rotate it, which is not what we want.

auth/src/main/AndroidManifest.xml
```diff
@@ -79,7 +87,6 @@
             android:authorities="${applicationId}.authuiinitprovider"
             android:exported="false"
             android:initOrder="90" />
-
```
SUPERCILEX: Could you add these blank lines back?

auth/src/main/java/com/firebase/ui/auth/data/model/Resource.java
```diff
@@ -14,39 +12,47 @@
...
-        mException = Preconditions.checkNotNull(exception, "Failure state cannot have null error.");
+    @NonNull
+    public static <T> Resource<T> forLoading() {
+        return new Resource<>(State.LOADING, null, null);
+    }
+
+    private Resource(State state, T value, Exception exception) {
+        mState = state;
+        mValue = value;
+        mException = exception;
```
SUPERCILEX: Nit: the constructor should be above the static methods.

auth/src/main/AndroidManifest.xml
```diff
@@ -60,6 +61,13 @@
             android:configChanges="keyboard|keyboardHidden|screenLayout|screenSize|orientation"
             tools:ignore="MissingRegistered" />
 
+        <activity
+            android:name=".ui.credentials.CredentialSaveActivity"
+            android:label=""
+            android:exported="false"
+            android:theme="@style/FirebaseUI.Transparent">
+        </activity>
```
SUPERCILEX: Nit: this should just be ` />`.

auth/src/main/java/com/firebase/ui/auth/ui/credentials/CredentialSaveActivity.java
```diff
@@ -0,0 +1,129 @@
...
+
+    @NonNull
+    public static Intent createIntent(Context context,
+                                      FlowParameters flowParams,
+                                      FirebaseUser firebaseUser,
+                                      @Nullable String password,
+                                      IdpResponse response) {
+
+        String accountType = ProviderUtils.idpResponseToAccountType(response);
+        Credential credential = CredentialsUtil.buildCredential(firebaseUser, password, accountType);
```
SUPERCILEX: I think it makes more sense to pass in a credential indead of doing all that logic in here.

auth/src/main/java/com/firebase/ui/auth/ui/email/WelcomeBackPasswordPrompt.java
```diff
@@ -111,73 +107,39 @@ protected void onCreate(Bundle savedInstanceState) {
...
+            case SUCCESS:
+                getDialogHolder().dismissDialog();
+
+                // This logic remains in the view since SmartLock is effectively a different
+                // 'screen' after the sign-in process.
+                FirebaseUser user = getAuthHelper().getCurrentUser();
+                startSaveCredentials(user, mHandler.getPendingPassword(), resource.getValue());
+                break;
+            case FAILURE:
+                // TODO: Is this message what we want?
```
SUPERCILEX: Don't we have an "incorrect password" string? We could see if the exception is an instance of the `FirebaseAuthInvalidCredentialsException` and use that string or else show the localized message. That's better than what we have. 😁

auth/src/main/java/com/firebase/ui/auth/ui/idp/CredentialSignInHandler.java
```diff
@@ -53,37 +55,36 @@ public CredentialSignInHandler(
         mAccountLinkRequestCode = accountLinkRequestCode;
     }
 
+    // TODO:
+    // FirebaseUser firebaseUser = task.getResult().getUser();
+    // mActivity.saveCredentialsOrFinish(firebaseUser, null, mResponse);
```
SUPERCILEX: This can be removed, right?

auth/src/main/java/com/firebase/ui/auth/viewmodel/smartlock/SmartLockHandler.java
SUPERCILEX: Bump.

auth/src/main/java/com/firebase/ui/auth/ui/email/WelcomeBackPasswordPrompt.java
samtstern: We actually don't ... I added a TODO to make one since I don't want to add one here and get a red build for a non-translated string.

samtstern:
@SUPERCILEX I think I got it all, minus one `TODO` about getting a new translated error message.

Issue #779, opened 2017-06-28
IndexOutOfBoundsException SpacedEditText.setSelection(SpacedEditText.java:82)


krawa:
I get a lot of reports with this crash.

firebase-ui-auth:2.0.1

Android 6.0
Lenovo VIBE K10 (K10a40)
Lenovo A6600 Plus (A6600a40) 

Android 5.0
Lenovo A5000 (A5000)

```
java.lang.IndexOutOfBoundsException: 
  at android.text.SpannableStringBuilder.checkRange(SpannableStringBuilder.java:1108)
  at android.text.SpannableStringBuilder.setSpan(SpannableStringBuilder.java:680)
  at android.text.SpannableStringBuilder.setSpan(SpannableStringBuilder.java:673)
  at android.text.Selection.setSelection(Selection.java:76)
  at android.text.Selection.setSelection(Selection.java:87)
  at android.widget.EditText.setSelection(EditText.java:115)
  at com.firebase.ui.auth.ui.phone.SpacedEditText.setSelection(SpacedEditText.java:82)
  at android.widget.EditText.setText(EditText.java:99)
  at com.firebase.ui.auth.ui.phone.SpacedEditText.setText(SpacedEditText.java:62)
  at android.widget.TextView.setText(TextView.java:4222)
  at com.firebase.ui.auth.ui.phone.SubmitConfirmationCodeFragment.setupConfirmationCodeEditText(SubmitConfirmationCodeFragment.java:192)
  at com.firebase.ui.auth.ui.phone.SubmitConfirmationCodeFragment.onCreateView(SubmitConfirmationCodeFragment.java:88)
  at android.support.v4.app.Fragment.performCreateView(Fragment.java:2192)
  at android.support.v4.app.FragmentManagerImpl.moveToState(FragmentManagerImpl.java:1299)
  at android.support.v4.app.FragmentManagerImpl.moveFragmentToExpectedState(FragmentManagerImpl.java:1528)
  at android.support.v4.app.FragmentManagerImpl.moveToState(FragmentManagerImpl.java:1595)
  at android.support.v4.app.BackStackRecord.executeOps(BackStackRecord.java:758)
  at android.support.v4.app.FragmentManagerImpl.executeOps(FragmentManagerImpl.java:2363)
  at android.support.v4.app.FragmentManagerImpl.executeOpsTogether(FragmentManagerImpl.java:2149)
[...]

PR #1130, opened 2018-01-26
Naive fix for 779

samtstern:
`Fixes` #779 

I honestly can't figure out how this ever happens but we should just be able to cut it off before it ever does.  Maybe devices with a keyboard can select in different ways?  That's my only theory.

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/ui/phone/SpacedEditText.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/phone/SpacedEditText.java
@@ -65,27 +65,31 @@ public final class SpacedEditText extends TextInputEditText {
      */
     @Override
     public void setSelection(int index) {
-        // Prevent index from ever exceeding the bounds of the text (See #779)
-        if (index >= mOriginalText.length()) {
-            index = mOriginalText.length() - 1;
-        }
+        // Desired mapping:
+        // 0 --> 0
+        // 1 --> 1
+        // 2 --> 3
+        // 3 --> 5
+        // 4 --> 7
+        // 5 --> 9
+        // 6 --> 11
 
-        // If the index is the leading edge, there are no spaces before it.
-        // For all other cases, the index is preceded by index - 1 spaces.
-        int spacesUpToIndex;
-        if (index == 0) {
-            spacesUpToIndex = 0;
-        } else {
-            spacesUpToIndex = index - 1;
-        }
+        // Naive transformation
+        int newIndex = (index * 2) - 1;
+
+        // Lower bound is 0
+        newIndex = Math.max(newIndex, 0);
+
+        // Upper bound is original length * 2 - 1
+        newIndex = Math.min(newIndex, (mOriginalText.length() * 2) - 1);
 
         try {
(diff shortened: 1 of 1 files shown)


samtstern:
Here's a diagram of how we use this.

```
x _ x _ x _ x _ x _ x
0 1 2 3 4 5 6 7 8 9 10
```

`x` is a character (normally a `-`) and `_` is whitespace.

So valid selections should be `0..5`.  If we request 5 then `index=5` and `spacesUpToIndex = 4`.  Therfore we'd call `setSelection(9)` in our current implementation.

Is that even correct?

auth/src/main/java/com/firebase/ui/auth/ui/phone/SpacedEditText.java
```diff
@@ -65,6 +65,11 @@ public void setText(CharSequence text, BufferType type) {
      */
     @Override
     public void setSelection(int index) {
+        // Prevent index from ever exceeding the bounds of the text (See #779)
+        if (index >= mOriginalText.length()) {
```
SUPERCILEX: I don't think that's going to work: the index was 11 (should be 10) in the crash, but the original text is only 5 chars long. TBH, I'm confused as to how any of this works... how is the span 21 characters long? As a very ugly solution, we could just use the catch block to call super with index 0. Or we could multiply `mOriginalText.length()` by 2.

samtstern: In the crash report we see `originalText=------` which is a 6 character long string of dashes.  We insert a space in between each pair (but not before first or after last) which gives an 11-character long string of `-_-_-_-_-_-`.

I have no idea how we ever got to 21 though... I think the most faithful thing to do is just select the very end if we get too large of an argument.

auth/src/main/java/com/firebase/ui/auth/ui/phone/SpacedEditText.java
```diff
@@ -65,6 +65,11 @@ public void setText(CharSequence text, BufferType type) {
...
+            index = mOriginalText.length() - 1;
+        }
+
         // If the index is the leading edge, there are no spaces before it.
         // For all other cases, the index is preceded by index - 1 spaces.
         int spacesUpToIndex;
```
samtstern: Now thinking about this, we have:
``` 
index --> spacesUpToIndex:
0 --> 0
1 --> 0
2 --> 1
3 --> 2
4 --> 3
5 --> 4
```

But I actually think it should be:
``` 
index --> spacesUpToIndex:
0 --> 0
1 --> 1
2 --> 2
3 --> 3
4 --> 4
5 --> 5
```
Because index 5 is actually the 6th dash.

In the case where index == 5 (the max case) then this would call setSelection(10) which would select the final character of the 11-character string (zero-indexed)

samtstern: Thanks for that!  Implemented your suggestion but with some more comments and spacing for clarity.

SUPERCILEX:
Sweet, those comments look great! 👏

PR #1146, opened 2018-02-12
Enable artifactory publication

samtstern:
This enables publication of `-SNAPSHOT` builds on jFrog artifactory.  Once I get it configured, this can run from Travis (!!!)


build.gradle
```diff
@@ -206,16 +215,27 @@ allprojects { project ->
             }
         }
 
+        artifactory {
+            contextUrl = 'http://localhost:8081/artifactory'
```
samtstern: Need to replace this with real credentials for oss.jfrog

scripts/artifactory.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+./gradlew :library:prepareArtifacts
```
samtstern: I can't explain it, but this script is the only way I can get things to work. Gradle task dependencies randomly skips some of these.

samtstern:
I must have messed something up because I am getting a ton of lint errors like this:

```
... can only be called from within the same library group (groupId=artifacts)
```

samtstern:
Woo!
https://oss.jfrog.org/webapp/#/artifacts/browse/tree/PomView/oss-snapshot-local/com/firebaseui/firebase-ui-database/3.2.2-SNAPSHOT/firebase-ui-database-3.2.2-20180213.161706-1.pom

scripts/artifactory.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -e
```
SUPERCILEX: I believe this can be inlined: `#!/usr/bin/env bash -e`.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -2,4 +2,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.5-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.5-bin.zip
```
SUPERCILEX: Whoops, looks like this was a mistake.

scripts/artifactory.sh
samtstern: It didn't like that :man_shrugging:

gradle/wrapper/gradle-wrapper.properties
samtstern: :taco:

PR #1147, opened 2018-02-13
Update dependencies

SUPERCILEX:
This lets us use a smaller version of Architecture Components from `v1.1.x` and includes that ConstraintLayout fix.


common/build.gradle
```diff
@@ -13,7 +13,8 @@ android {
 }
 
 dependencies {
+    api "android.arch.lifecycle:runtime:$architectureVersion"
```
samtstern: This was `implementation` but is now two `api` dependencies.  Were we wrong before?

SUPERCILEX: No, I just moved this up the tree: we used to have it as `api` in both the `firestore` and `database` modules so I just pulled it up to the `common` module. Do you think we should leave it as it was in the lower modules?

samtstern: Ohhh I see.  SGTM!

PR #1151, opened 2018-02-15
Prototype separate save activity

samtstern:

### DIFF
--- a/auth/src/main/AndroidManifest.xml
+++ b/auth/src/main/AndroidManifest.xml
@@ -1,13 +1,12 @@
-<manifest
-    xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:tools="http://schemas.android.com/tools"
-    package="com.firebase.ui.auth">
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+          xmlns:tools="http://schemas.android.com/tools"
+          package="com.firebase.ui.auth">
 
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.INTERNET" />
 
     <application>
-
         <meta-data
             android:name="com.google.android.gms.version"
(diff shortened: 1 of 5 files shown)


samtstern:
This is just a basic implementation.  I could clean up some more and fix the tests if we choose to go this way.

auth/src/main/AndroidManifest.xml
```diff
@@ -1,13 +1,12 @@
-<manifest
-    xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:tools="http://schemas.android.com/tools"
-    package="com.firebase.ui.auth">
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+          xmlns:tools="http://schemas.android.com/tools"
+          package="com.firebase.ui.auth">
```
SUPERCILEX: Looks like some weird styling stuff got messed up here and below.

auth/src/main/java/com/firebase/ui/auth/ui/credentials/CredentialSaveActivity.java
```diff
@@ -0,0 +1,122 @@
...
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        mHandler = ViewModelProviders.of(this).get(SmartLockHandler.class);
+        mHandler.init(getFlowParams());
+
+        Credential credential = getIntent().getParcelableExtra(ExtraConstants.EXTRA_CREDENTIAL);
+        mIdpResponse = getIntent().getParcelableExtra(ExtraConstants.EXTRA_IDP_RESPONSE);
+
+        mHandler.saveCredentials(credential);
```
SUPERCILEX: This should be guarded by a `savedInstanceState == null` check. (Or should we make the activity resilient to config changes with `android:configChanges="keyboard|keyboardHidden|screenLayout|screenSize|orientation"`?)

samtstern:
@SUPERCILEX gonna move this back to #1124

Issue #901, opened 2017-09-11
Missing translations for signing in with email and password


jbahamon:
* Android device: Motorola G5 Plus
  * Android OS version: 7.0
  * Google Play Services version: 11.5.09
  * FirebaseUI version: 2.2.0

Using a device with Spanish (Chile) as the only language.
  
#### Steps to reproduce:

  1. Login using an email account
  2. Enter an existing email account
  3. Provide an incorrect password
  4. Press on the "Sign in" ("Acceder", in Spanish) button

  * The error text shows "The password is invalid or the user does not have a password." instead of translated text.

This might happen because the exception/error received in `WelcomeBackPasswordPrompt` is not localized, only showing the (non-localized) description from a `FirebaseAuthError`.

Additionally, the accesibility info for the password visibility toggling item (the eye icon) from the password `TextView` is not localized. This can probably be solved by setting the `app:passwordToggleContentDescription` field on the password element in `fui_register_email_layout.xml`.

samtstern:
@jbahamon thanks for reporting this.  This is probably a server error message that we are displaying directly, which is a bad idea.  And we also appreciate the note about accesibility!

SUPERCILEX:
@samtstern Doesn't #946 fix this?

jstimes:
@SUPERCILEX It seems like it would, but that method does not localize the server error messages, which is what this message is. We should log these server errors and have our own (localized) strings to show the user. I can work on this.

samtstern:
This has been fixed and released in version `3.2.2`.

PR #1163, opened 2018-02-21
Translate errors

samtstern:
See #901


samtstern:
I have kicked off the internal translation process for the new ones.

auth/src/main/res/values/strings.xml
```diff
@@ -75,6 +75,10 @@
     <string name="fui_progress_dialog_sending" translation_description="Progress dialog while password recovery email is being sent">Sending…</string>
     <string name="fui_error_email_does_not_exist" translation_description="Inline error when user attempt signin with invalid account">That email address doesn\'t match an existing account</string>
 
+    <!-- Other error messages -->
+    <string name="fui_error_unknown" translation_description="General error messages for an unknown failure.">An unknown error occurred.</string>
```
SUPERCILEX: Not sure if this is intentional or not, but we already have a very similar string:
```xml
<string name="fui_general_error" translation_description="Toast text flashing general error">An error has occurred.</string>
```

samtstern: True.  I will remove that one, I like it with 'unknown'

samtstern:
Ok so it looks like our translation system decided to remove the `translation_description` attribute from strings in the translated files which makes this diff kind of enormous.  I took a manual look at a few files and it looks like there are no other changes.

Issue #1154, opened 2018-02-16
[Discuss] Don't save credentials for non-email providers


samtstern:
The SmartLock team has learned over the years that IDP credentials that don't include a password or token actually do more to confuse than help.

We could change our strategy:
  * Only save email credentials
  * ^ and also save Google credentials since we can use them to get an ID token
  * ^ and also save Phone number credentials

But I don't think Facebook or Twitter credentials really have any value.

SUPERCILEX:
Huh, I thought I had responded to this, but I guess not. Anyway, I totally agree with you! 😄 In fact, I just ran into this a week or so ago: Smart Lock auto signed me in to Twitter which was really confusing since they aren't the best provider and use a Webview without autofill support. TL;DR: it was annoying and I would have much preferred Google or email sign-in which would have been faster. To compromise, I created #1166 which still saves the credentials so users know they have an account with an app, but it doesn't suggest using unpleasant IDPs to optimize the UX flow.

samtstern:
This issue has been fixed and released in `3.3.0`.

PR #1166, opened 2018-02-24
Don't show hard-to-use IDPs for Smart Lock sign-in

SUPERCILEX:
See #1154

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/util/signincontainer/SignInDelegate.java
+++ b/auth/src/main/java/com/firebase/ui/auth/util/signincontainer/SignInDelegate.java
@@ -109,7 +109,7 @@ public class SignInDelegate extends FragmentBase
                 supportPasswords = true;
             }
         }
-        List<String> accountTypes = getSupportedAccountTypes();
+        List<String> accountTypes = getCredentialAccountTypes();
 
         // If the request will be empty, avoid the step entirely
         boolean willRequestCredentials = supportPasswords || accountTypes.size() > 0;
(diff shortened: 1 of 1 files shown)


auth/src/main/java/com/firebase/ui/auth/util/signincontainer/SignInDelegate.java
```diff
@@ -224,9 +222,7 @@ public void finish(int resultCode, Intent resultIntent) {
         List<String> accounts = new ArrayList<>();
         for (AuthUI.IdpConfig idpConfig : getFlowParams().providerInfo) {
             @AuthUI.SupportedProvider String providerId = idpConfig.getProviderId();
-            if (providerId.equals(GoogleAuthProvider.PROVIDER_ID)
-                    || providerId.equals(FacebookAuthProvider.PROVIDER_ID)
-                    || providerId.equals(TwitterAuthProvider.PROVIDER_ID)) {
+            if (providerId.equals(GoogleAuthProvider.PROVIDER_ID)) {
```
samtstern: nit: can we change the name of this method to be "getCredentialAccountTypes" or something that implies it has to do with SmartLock?

samtstern:
@SUPERCILEX thanks for this!  I guess this would make the next release `3.3.0` since it's a behavior change?

auth/src/main/java/com/firebase/ui/auth/util/signincontainer/SignInDelegate.java
SUPERCILEX: 👍

SUPERCILEX:
Yup, SGTM

samtstern:
@SUPERCILEX I am actually going to bump this off of `3.2.2` and into the next release since I think we have some other low-hanging minor-version improvements that it would be nice to bundle together.

This way we can keep this bug-fix release as a true bug-fix release.

SUPERCILEX:
Yeah, SGTM. 👍

SUPERCILEX:
@samtstern Is this ready to roll?

samtstern:
@SUPERCILEX kept forgetting about this one, but good to go.

PR #1168, opened 2018-02-24
Give clients much more detailed errors

SUPERCILEX:
Fixes https://github.com/firebase/FirebaseUI-Android/issues/526


auth/src/main/java/com/firebase/ui/auth/ErrorCodes.java
```diff
@@ -1,20 +1,49 @@
...
+     * Valid codes that can be returned from {@link FirebaseUiException#getErrorCode()}.
+     */
+    @IntDef({
+                    UNKNOWN_ERROR,
+                    NO_NETWORK,
+                    PLAY_SERVICES_UPDATE_CANCELLED,
+                    DEVELOPER_ERROR
+            })
+    @Retention(RetentionPolicy.SOURCE)
+    public @interface All {}
```
SUPERCILEX: Not really happy with this name, any ideas?

auth/src/main/java/com/firebase/ui/auth/data/model/FirebaseUiException.java
```diff
@@ -1,33 +1,46 @@
 package com.firebase.ui.auth.data.model;
```
SUPERCILEX: Should we move this to the root package or leave it as is?

auth/src/main/java/com/firebase/ui/auth/ErrorCodes.java
samtstern: Maybe `Code`?

auth/src/main/java/com/firebase/ui/auth/data/model/FirebaseUiException.java
samtstern: Since this is `LIBRARY_GROUP` only that's not a breaking change, so if you want to move it to the root package that SGTM.

auth/src/main/java/com/firebase/ui/auth/provider/IdpProvider.java
```diff
@@ -14,13 +14,15 @@
...
+
 import com.firebase.ui.auth.IdpResponse;
 
 public interface IdpProvider extends Provider {
     interface IdpCallback {
-        void onSuccess(IdpResponse idpResponse);
+        void onSuccess(@NonNull IdpResponse idpResponse);
 
-        void onFailure();
+        void onFailure(@NonNull Exception e);
```
samtstern: Is it possible that anyone was relying on this interface?  If so this is a breaking change .... if this is meant to be internal let's annotate it.

samtstern:
@SUPERCILEX thanks!  Just left a few small comments.

auth/src/main/java/com/firebase/ui/auth/ErrorCodes.java
SUPERCILEX: SGTM, `ErrorCodes.Anything` feels weird, but yours makes more sense.

auth/src/main/java/com/firebase/ui/auth/provider/IdpProvider.java
SUPERCILEX: I don't think so, or at least, I know we've broken it before without notice. I'll annotate it.

PR #1169, opened 2018-02-24
Remove unnecessary @MainThread annotations

SUPERCILEX:
Not sure when you would ever be correct in calling methods inside an activity from a non-main thread... so might as well get rid of the noise.


samtstern:
@SUPERCILEX huh I can't believe I never noticed those before.  Thanks!

PR #1173, opened 2018-02-27
Remove deprecated references from README

SUPERCILEX:
Forgot to do this in #1168.


samtstern:
@SUPERCILEX good catch, thanks!

PR #1177, opened 2018-02-28
POJOize IdpResponse

SUPERCILEX:
Super minor, but I was helping another dev over on StackOverflow and realized that `IdpResponse`s look like `IdpResponse@546453`... not all that helpful. 😉 This PR makes `IdpResponse` a nice POJO.


samtstern:
@SUPERCILEX :+1:

PR #1178, opened 2018-03-01
Firestore paging adapter

samtstern:
See #17 

This is a WIP but I finally have the stability and general mechanics in a sane place.

### DIFF
--- a/app/src/main/java/com/firebase/uidemo/database/firestore/FirestorePagingActivity.java
+++ b/app/src/main/java/com/firebase/uidemo/database/firestore/FirestorePagingActivity.java
@@ -1,0 +1,130 @@
+package com.firebase.uidemo.database.firestore;
+
+import android.os.Bundle;
+import android.support.v7.app.AppCompatActivity;
+import android.support.v7.widget.LinearLayoutManager;
+import android.support.v7.widget.RecyclerView;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.TextView;
+
+import com.firebase.ui.firestore.FirestoreInfiniteScrollListener;
+import com.firebase.ui.firestore.FirestorePagingAdapter;
+import com.firebase.ui.firestore.FirestorePagingOptions;
+import com.firebase.uidemo.R;
+import com.google.android.gms.tasks.Task;
+import com.google.firebase.firestore.CollectionReference;
+import com.google.firebase.firestore.FirebaseFirestore;
+import com.google.firebase.firestore.Query;
+import com.google.firebase.firestore.WriteBatch;
+
+import java.util.Locale;
+
+import butterknife.BindView;
+import butterknife.ButterKnife;
+
+public class FirestorePagingActivity extends AppCompatActivity {
+
+    @BindView(R.id.paging_recycler)
+    RecyclerView mRecycler;
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_firestore_paging);
+        ButterKnife.bind(this);
+
+        setUpAdapter();
+    }
+
+    private void setUpAdapter() {
+        Query baseQuery = FirebaseFirestore.getInstance()
+                .collection("items")
+                .orderBy("value", Query.Direction.ASCENDING);
+
+        FirestorePagingOptions<Item> options = new FirestorePagingOptions.Builder<Item>()
+                .setQuery(baseQuery, Item.class)
+                .setPageSize(10)
+                .setLoadTriggerDistance(5)
+                .setMaxPages(3)
+                .build();
+
+        FirestorePagingAdapter<Item, ItemViewHolder> adapter =
+                new FirestorePagingAdapter<Item, ItemViewHolder>(options) {
+                    @Override
+                    public ItemViewHolder onCreateViewHolder(ViewGroup parent, int viewType) {
+                        View view = LayoutInflater.from(parent.getContext())
+                                .inflate(R.layout.item_item, parent, false);
+
+                        return new ItemViewHolder(view);
+                    }
+
+                    @Override
+                    protected void onBindViewHolder(ItemViewHolder holder, int position, Item model) {
+                        holder.bind(model);
+                    }
+                };
+
+        LinearLayoutManager manager = new LinearLayoutManager(this);
+
+        FirestoreInfiniteScrollListener listener =
+                new FirestoreInfiniteScrollListener(manager, adapter);
+
+        mRecycler.setLayoutManager(manager);
+        mRecycler.setAdapter(adapter);
+        mRecycler.addOnScrollListener(listener);
+    }
+
+    // TODO: Bind this to a menu item
+    public Task<Void> createItems() {
+
+        WriteBatch writeBatch = FirebaseFirestore.getInstance().batch();
+        CollectionReference collRef = FirebaseFirestore.getInstance().collection("items");
+
+        for (int i = 0; i < 500; i++) {
+            String title = "Item " + i;
+
+            String id = String.format(Locale.getDefault(), "item_%03d", i);
+            Item item = new Item(title, i);
+
+            writeBatch.set(collRef.document(id), item);
+        }
+
+        return writeBatch.commit();
+    }
+
+    public static class Item {
+
+        public String text;
+        public int value;
+
+        public Item() {}
+
+        public Item(String text, int value) {
+            this.text = text;
+            this.value = value;
+        }
+
+    }
+
+    public static class ItemViewHolder extends RecyclerView.ViewHolder {
+
+        @BindView(R.id.item_text)
+        TextView mTextView;
+
+        @BindView(R.id.item_value)
+        TextView mValueView;
+
+        ItemViewHolder(View itemView) {
+            super(itemView);
+            ButterKnife.bind(this, itemView);
+        }
+
+        void bind(Item item) {
+            mTextView.setText(item.text);
+            mValueView.setText(String.valueOf(item.value));
+        }
+    }
+
+}
--- a/app/src/main/java/com/firebase/uidemo/database/firestore/FirestorePagingActivity.java
+++ b/app/src/main/java/com/firebase/uidemo/database/firestore/FirestorePagingActivity.java
@@ -1,18 +1,25 @@
 package com.firebase.uidemo.database.firestore;
 
 import android.os.Bundle;
+import android.support.annotation.NonNull;
 import android.support.v7.app.AppCompatActivity;
 import android.support.v7.widget.LinearLayoutManager;
 import android.support.v7.widget.RecyclerView;
+import android.util.Log;
 import android.view.LayoutInflater;
+import android.view.Menu;
+import android.view.MenuItem;
 import android.view.View;
 import android.view.ViewGroup;
+import android.widget.ProgressBar;
 import android.widget.TextView;
+import android.widget.Toast;
 
 import com.firebase.ui.firestore.FirestoreInfiniteScrollListener;
 import com.firebase.ui.firestore.FirestorePagingAdapter;
 import com.firebase.ui.firestore.FirestorePagingOptions;
 import com.firebase.uidemo.R;
+import com.google.android.gms.tasks.OnCompleteListener;
 import com.google.android.gms.tasks.Task;
 import com.google.firebase.firestore.CollectionReference;
 import com.google.firebase.firestore.FirebaseFirestore;
@@ -26,8 +33,15 @@ import butterknife.ButterKnife;
 
 public class FirestorePagingActivity extends AppCompatActivity {
 
+    private static final String TAG = "FirestorePagingActivity";
+
     @BindView(R.id.paging_recycler)
     RecyclerView mRecycler;
+
+    @BindView(R.id.paging_loading)
+    ProgressBar mProgressBar;
+
+    private FirebaseFirestore mFirestore;
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
@@ -35,19 +49,25 @@ public class FirestorePagingActivity extends AppCompatActivity {
         setContentView(R.layout.activity_firestore_paging);
         ButterKnife.bind(this);
 
+        mFirestore = FirebaseFirestore.getInstance();
+        mProgressBar.setIndeterminate(true);
+
         setUpAdapter();
     }
 
     private void setUpAdapter() {
-        Query baseQuery = FirebaseFirestore.getInstance()
-                .collection("items")
+        Query baseQuery = mFirestore.collection("items")
                 .orderBy("value", Query.Direction.ASCENDING);
 
+        // Paging options:
+        //  * pageSize - number of items to load in each 'page'.
+        //  * loadTriggerDistance - how far from the bottom/top of the data set to trigger a load.
+        //  * maxPages - maximum number of pages to keep in memory at a time.
         FirestorePagingOptions<Item> options = new FirestorePagingOptions.Builder<Item>()
                 .setQuery(baseQuery, Item.class)
-                .setPageSize(10)
-                .setLoadTriggerDistance(5)
-                .setMaxPages(3)
+                .setPageSize(20)
+                .setLoadTriggerDistance(10)
+                .setMaxPages(5)
                 .build();
 
         FirestorePagingAdapter<Item, ItemViewHolder> adapter =
@@ -64,10 +84,22 @@ public class FirestorePagingActivity extends AppCompatActivity {
                     protected void onBindViewHolder(ItemViewHolder holder, int position, Item model) {
                         holder.bind(model);
                     }
+
+                    @Override
+                    protected void onLoadingStateChanged(boolean isLoading) {
+                        super.onLoadingStateChanged(isLoading);
+                        if (isLoading) {
+                            mProgressBar.setVisibility(View.VISIBLE);
+                        } else {
+                            mProgressBar.setVisibility(View.INVISIBLE);
+                        }
+                    }
                 };
 
         LinearLayoutManager manager = new LinearLayoutManager(this);
 
+        // The infinite scroll listener will instruct the adapter to page up and down at the
+        // appropriate times.
         FirestoreInfiniteScrollListener listener =
                 new FirestoreInfiniteScrollListener(manager, adapter);
 
@@ -76,11 +108,37 @@ public class FirestorePagingActivity extends AppCompatActivity {
         mRecycler.addOnScrollListener(listener);
     }
 
-    // TODO: Bind this to a menu item
-    public Task<Void> createItems() {
+    @Override
+    public boolean onCreateOptionsMenu(Menu menu) {
+        getMenuInflater().inflate(R.menu.menu_firestore_paging, menu);
+        return super.onCreateOptionsMenu(menu);
+    }
 
-        WriteBatch writeBatch = FirebaseFirestore.getInstance().batch();
-        CollectionReference collRef = FirebaseFirestore.getInstance().collection("items");
+    @Override
+    public boolean onOptionsItemSelected(MenuItem item) {
+        if (item.getItemId() == R.id.item_add_data) {
+            showToast("Adding data...");
+            createItems().addOnCompleteListener(this, new OnCompleteListener<Void>() {
+                @Override
+                public void onComplete(@NonNull Task<Void> task) {
+                    if (task.isSuccessful()) {
+                        showToast("Data added.");
+                    } else {
+                        Log.w(TAG, "addData", task.getException());
+                        showToast("Error adding data.");
+                    }
+                }
+            });
+
+            return true;
+        }
+        return super.onOptionsItemSelected(item);
+    }
+
+    private Task<Void> createItems() {
+
+        WriteBatch writeBatch = mFirestore.batch();
+        CollectionReference collRef = mFirestore.collection("items");
 
         for (int i = 0; i < 500; i++) {
             String title = "Item " + i;
@@ -92,6 +150,10 @@ public class FirestorePagingActivity extends AppCompatActivity {
         }
 
         return writeBatch.commit();
+    }
+
+    private void showToast(String message) {
+        Toast.makeText(this, message, Toast.LENGTH_SHORT).show();
     }
 
     public static class Item {
--- a/app/src/main/res/layout/activity_firestore_paging.xml
+++ b/app/src/main/res/layout/activity_firestore_paging.xml
@@ -6,14 +6,23 @@
     android:layout_height="match_parent"
     tools:context="com.firebase.uidemo.database.firestore.FirestorePagingActivity">
 
+    <ProgressBar
+        android:id="@+id/paging_loading"
+        style="?android:attr/progressBarStyleHorizontal"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="-6dp"
+        android:background="@android:color/transparent" />
+
     <android.support.v7.widget.RecyclerView
         android:id="@+id/paging_recycler"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
-        tools:listitem="@layout/item_item"
+        android:layout_below="@+id/paging_loading"
         android:paddingLeft="16dp"
         android:paddingRight="16dp"
         android:paddingTop="8dp"
-        android:clipToPadding="false" />
+        android:clipToPadding="false"
+        tools:listitem="@layout/item_item" />
 
 </RelativeLayout>
--- a/app/src/main/res/layout/activity_firestore_paging.xml
+++ b/app/src/main/res/layout/activity_firestore_paging.xml
@@ -1,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<RelativeLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    tools:context="com.firebase.uidemo.database.firestore.FirestorePagingActivity">
+
+    <android.support.v7.widget.RecyclerView
+        android:id="@+id/paging_recycler"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        tools:listitem="@layout/item_item"
+        android:paddingLeft="16dp"
+        android:paddingRight="16dp"
+        android:paddingTop="8dp"
+        android:clipToPadding="false" />
+
+</RelativeLayout>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -4,11 +4,13 @@
     <!-- Chooser -->
     <string name="title_auth_activity">Auth UI demo</string>
     <string name="title_firestore_activity">Cloud Firestore Demo</string>
+    <string name="title_firestore_paging_activity">Cloud Firestore Paging Demo</string>
     <string name="title_realtime_database_activity">Real-time database demo</string>
     <string name="title_storage_activity">Storage Image Demo</string>
 
     <string name="desc_auth">Demonstrates the Firebase Auth UI flow, with customization options.</string>
     <string name="desc_firestore">Demonstrates using a FirestoreRecyclerAdapter to load data from Cloud Firestore into a RecyclerView for a basic chat app.</string>
+    <string name="desc_firestore_paging">Demonstrates using a FirestorePagingAdapter to load/infinite scroll paging data from Cloud Firestore.</string>
     <string name="desc_realtime_database">Demonstrates using a FirebaseRecyclerAdapter to load data from Firebase Database into a RecyclerView for a basic chat app.</string>
     <string name="desc_storage">Demonstrates displaying an image from Cloud Storage using Glide.</string>
 
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -101,4 +101,5 @@
         Make sure your device is online and that Anonymous Auth is configured in your Firebase project
         (https://console.firebase.google.com/project/_/authentication/providers)
     </string>
+    <string name="menu_add_data">Add Data</string>
 </resources>
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -1,8 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest
-    xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:tools="http://schemas.android.com/tools"
-    package="com.firebase.uidemo">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+          xmlns:tools="http://schemas.android.com/tools"
+          package="com.firebase.uidemo">
 
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
@@ -17,7 +16,6 @@
         android:supportsRtl="true"
         android:theme="@style/AppTheme"
         tools:ignore="GoogleAppIndexingWarning">
-
         <activity android:name=".ChooserActivity">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
@@ -43,6 +41,11 @@
             android:name=".database.firestore.FirestoreChatActivity"
             android:label="@string/title_firestore_activity" />
 
+        <!-- Firestore paging demo -->
+        <activity
+            android:name=".database.firestore.FirestorePagingActivity"
+            android:label="@string/title_firestore_paging_activity" />
+
         <!-- Realtime database demo -->
         <activity
             android:name=".database.realtime.RealtimeDbChatActivity"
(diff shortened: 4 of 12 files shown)


samtstern:
cc @puf

samtstern:
@morganchen12 at a high level, how does this compare to what you've tried on iOS?

SUPERCILEX:
@samtstern Without looking at any code, my main concern is that the first issues people are going to submit will be something along the lines of "Paginated data not updating". Also, it'd make me so sad to see a Firebase app with the infamous refresh button. 😢😂 Have you looked at the [Paging Library](https://developer.android.com/topic/libraries/architecture/paging.html)? When I first played with it, I couldn't get things to work out nicely, but that was way back in alpha 1. It should remove a bunch of boilerplate and support changes by passing in new lists and doing diffs on a background thread. The focus is more directed towards supporting Room, but I think we could make it work. Do you want to investigate, or should I?

samtstern:
@SUPERCILEX I can definitely take a look there.  I thought it was Room-only so hadn't really looked.

I have a feeling that we're going to have so many issues if we try to maintain listeners on the paginated data ... but we can take another look at that.

samtstern:
@SUPERCILEX ok the Paging library is f***cking awesome!  I am at least not embarassed since my design was a lot like theirs, but theirs is way cleaner and best of all I don't have to maintain it.

Before I go that route entirely:

  * Do you have any idea if the paging library unloads old pages?  As far as I can tell the memory usage doesn't grow wildly as I scroll, but it doesn't seem to ever unload anything.
  * The paging library definitely won't allow us to update individual items as they change.  I am fine with that, but you seem concerned.  Any ideas?
  * I don't want to release until it hits at least beta, preferably stable.  I dunno if you feel differently.

SUPERCILEX:
@samtstern Haha, I thought you might like it! The Architecture Components team is doing an amazing job of perfecting and investing in all of their APIs so I'd expect nothing less from them. 😆 To answer your questions:
1. I've spent a bunch of time digging around the SDK and concluded that no, it doesn't drop pages [yet](https://medium.com/@chriscraik/thanks-f38b1216a88). If you poke around the `PagedStorage`, you'll find that `ArrayList<List<T>> mPages` is only ever added to, nothing is ever removed. I'd assume that'll get fixed before the beta.
1. Yeah, I see that they haven't addressed that issue yet either. Maybe they're holding off on #<span></span>1 and are going to pull off a brilliant move where you can get a callback for when old pages die which would let us keep listeners around and make the list real-time. Or not, but I think the first step is to use something like the Paging library so we can focus on building the Firebase features. Isn't that the whole slogan? "Focus on your users," not the infracture? 😆
1. **_Absolutely!_** This is like the second time they broken everybody, so we should definitely wait until the beta comes around (which should be soon?). Once it hits beta, they stop breaking stuff so we'd be good. Not sure where that leaves this PR though, so it's up to you.

samtstern:
@SUPERCILEX thanks for all the info.  Even if they don't get to `#1` very soon, I profiled my sample app and the memory usage doesn't get too crazy so I think we'll just say please don't use this with very large documents and wait for the paging library people to do the work for us.

As for `#2` and `#3` let's just say this whole thing is on hold until the paging library gets to at least `beta`.

samtstern:
The `-alpha7` release of the paging library just came out, and the release notes say `beta` is next.

samtstern:
@SUPERCILEX mind taking a look at this now that it's based on the paging library?  I assume that the `alpha7` --> `beta1` diff won't be so big so the review should not be a waste of time.

SUPERCILEX:
SGTM, will do tomorrow!

firestore/src/main/java/com/firebase/ui/firestore/paging/FirestorePagingAdapter.java
```diff
@@ -0,0 +1,118 @@
...
+        onBindViewHolder(holder, position, mParser.parseSnapshot(snapshot));
+    }
+
+    protected abstract void onBindViewHolder(@NonNull VH holder, int position, T model);
+
+    protected void onLoadingStateChanged(@NonNull LoadingState state) {
+        // For overriding
+    }
+
+    private void onListChanged(@NonNull PagedList<DocumentSnapshot> snapshots) {
```
SUPERCILEX: Couldn't we just inline this method? It just forwards a param to another method and confused me...

firestore/src/main/java/com/firebase/ui/firestore/paging/FirestorePagingOptions.java
```diff
@@ -0,0 +1,84 @@
...
+                                   @NonNull Class<T> modelClass) {
+            return setQuery(query, config, new ClassSnapshotParser<T>(modelClass));
+        }
+
+        @NonNull
+        public Builder<T> setQuery(@NonNull Query query,
+                                   @NonNull PagedList.Config config,
+                                   @NonNull SnapshotParser<T> parser) {
+
+
```
SUPERCILEX: Nit: lots of space! 🛰

firestore/src/main/java/com/firebase/ui/firestore/paging/FirestorePagingOptions.java
```diff
@@ -0,0 +1,84 @@
...
+            // Build paged list
+            FirestoreDataSource.Factory factory = new FirestoreDataSource.Factory(query);
+            mData = new PagingData(new LivePagedListBuilder<>(factory, config).build());
+
+            mParser = parser;
+            return this;
+        }
+
+        @NonNull
+        public Builder<T> setOwner(@NonNull LifecycleOwner owner) {
```
SUPERCILEX: Nit: I think we should call it `setLifecycleOwner` to be consistent with the other builders.

firestore/src/main/java/com/firebase/ui/firestore/paging/FirestorePagingOptions.java
```diff
@@ -0,0 +1,84 @@
...
+            mOwner = owner;
+            return this;
+        }
+
+        public FirestorePagingOptions<T> build() {
```
SUPERCILEX: Nit: `@NonNull`.

firestore/build.gradle
```diff
@@ -22,6 +22,10 @@ dependencies {
     api "com.android.support:recyclerview-v7:$supportLibraryVersion"
     annotationProcessor "android.arch.lifecycle:compiler:$architectureVersion"
 
+
+    api "android.arch.paging:common:1.0.0-alpha7"
```
SUPERCILEX: `runtime` already includes `common`.

SUPERCILEX:
Wow, GitHub is even hiding half my comments. Lol, sorry for the overload! 😉

firestore/src/main/java/com/firebase/ui/firestore/paging/FirestorePagingAdapter.java
```diff
@@ -0,0 +1,118 @@
...
+
+    private final Observer<LoadingState> mStateObserver =
+            new Observer<LoadingState>() {
+                @Override
+                public void onChanged(@Nullable LoadingState state) {
+                    if (state == null) {
+                        return;
+                    }
+
+                    onLoadingStateChanged(state);
```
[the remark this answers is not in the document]
samtstern: Yeah I even stole this pattern from the official sample.  I bothered the library authors about this and they agree that they want to handle this in future API versions, but not right now.

samtstern:
@SUPERCILEX thank you for all the great comments!  I will look at some today, then on vacation till next Weds so will be slow (on this all all other FirebaseUI things)

SUPERCILEX:
Gotya, SGTM. Have fun! 🏖🌴

app/src/main/res/layout/activity_firestore_paging.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="utf-8"?>
+<RelativeLayout
```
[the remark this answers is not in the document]
samtstern: `app` module doesn't currently have any ConstraintLayouts, didn't feel like introducing it for this simple layout.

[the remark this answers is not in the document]
SUPERCILEX: Well _technically_, `auth` already includes it so it wouldn't hurt... I guess if you don't want to that's fine—I love CL though so I gotta put in my plug whenever I can. 😆

samtstern:
From the release notes for Paging beta1 ([link](https://developer.android.com/topic/libraries/architecture/release-notes.html#april_5_2018))

> Paging will be in beta for a short time before progressing to release candidate. We are not planning further API changes for Paging 1.0, and the bar for any API changes is very high.

So I am comfortable still targeting this feature for `3.4.0`, assuming I can address all of your concerns,

firestore/src/main/java/com/firebase/ui/firestore/paging/FirestoreDataSource.java
```diff
@@ -0,0 +1,156 @@
...
+                    public void onFailure(@NonNull Exception e) {
+                        Log.w(TAG, "loadAfter:failure", e);
+
+                        // On error, return an empty page with the next page key being basically
+                        // equal to the initial query.
+                        callback.onResult(Collections.<DocumentSnapshot>emptyList(), key);
+
+                        mLoadingState.postValue(LoadingState.ERROR);
+                    }
+                });
```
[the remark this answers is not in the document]
samtstern: Ok simplifying as much as I can.  Can't be totally de-duped since the initial callback and the other callback are different (sigh)

firestore/src/main/java/com/firebase/ui/firestore/paging/FirestorePagingAdapter.java
```diff
@@ -0,0 +1,118 @@
...
+
+        if (options.getOwner() != null) {
+            options.getOwner().getLifecycle().addObserver(this);
+        }
+    }
+
+    @OnLifecycleEvent(Lifecycle.Event.ON_START)
+    public void startListening() {
+        mData.getSnapshots().observeForever(mDataObserver);
+        mData.getLoadingState().observeForever(mStateObserver);
```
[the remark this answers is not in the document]
samtstern: What should I do when they don't pass me an observer?

samtstern:
@SUPERCILEX what would you rather see as far as code splitting:

1. Make the paging library a "provided" dependency and treat it how we treat Twitter/Facebook
2. Make a whole new `firebase-ui-firestore-paging` target

I lean towards #1 since we've done it before and it has the same size savings for people who don't care about paging.

firestore/src/main/java/com/firebase/ui/firestore/paging/FirestoreDataSource.java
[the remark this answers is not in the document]
SUPERCILEX: SGTM

firestore/src/main/java/com/firebase/ui/firestore/paging/FirestorePagingAdapter.java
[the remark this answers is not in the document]
SUPERCILEX: Ohhhhh, I didn't see that this was a public API (I thought it was just copypasta from our other adapter), my bad then. 😊 It still feels like an anti-pattern though, but you're right, I'm not sure how to deal with it without the owner.

As a side note, _should_ we be providing the start/stop methods? Those seem to be the greatest source of confusion on Stack Overflow so maybe we should require an owner? Food for thought: we could always add start/stop methods but not remove them.

SUPERCILEX:
Yeah, I think you're right. My only concern is that people are going to be confused as to why we have 2 adapters. If I were just browsing the source code, I'd be like "why _wouldn't_ I want a paging adapter?". At some point if the AC team adds an unload callback we might be able to provide realtime updates and deprecate the old adapter or something, but for now, if we go with #<span></span>1, we need to have really clear docs. 📜😁 Yay, nay?

firestore/src/main/java/com/firebase/ui/firestore/paging/FirestorePagingAdapter.java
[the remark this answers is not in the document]
samtstern: I don't want to require an owner since I don't want to dictate lifecycle.  We've seen enough people  want to use create/destroy versus start/stop that we clearly can't make this decision for them.

You're right it's definitely a source of confusion, but I think sticking with our decision here will create less confusion rather than more (given that we can't walk back the methods on the non-paging adapter).

As for this being an antipattern ... I also don't love it but I can't think of anything else!  In the end it's no different than how we add a snapshot listener and remove it in the other adapter.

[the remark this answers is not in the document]
SUPERCILEX: Haha, yep, all of that makes sense. 👍

samtstern:
@SUPERCILEX yep I think careful documentation is going to be the main answer here.

As to "why wouldn't I want a paging adapter", that's a good question!  I think most people will want paging unless they need to observe non-added/removed changes to individual items in real time.

SUPERCILEX:
Huh, yeah, I guess it depends on the type of app you're building. I'm biased cuz my thing is cross-device collaboration, but I'd imagine most people are using Firestore for the database part primarily. Anyway, SGTM. 👍

I'll go through another round of review tomorrow or Friday. 👏

app/src/main/java/com/firebase/uidemo/database/firestore/FirestorePagingActivity.java
```diff
@@ -0,0 +1,194 @@
...
+    private FirebaseFirestore mFirestore;
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_firestore_paging);
+        ButterKnife.bind(this);
+
+        mFirestore = FirebaseFirestore.getInstance();
+        mProgressBar.setIndeterminate(true);
```
SUPERCILEX: Can we move this to the XML?

app/src/main/java/com/firebase/uidemo/database/firestore/FirestorePagingActivity.java
```diff
@@ -0,0 +1,194 @@
...
+
+    private FirebaseFirestore mFirestore;
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_firestore_paging);
+        ButterKnife.bind(this);
+
+        mFirestore = FirebaseFirestore.getInstance();
```
SUPERCILEX: I'm wondering what the best practices should be... I think we should store the collection as a field, but use the full `FirebaseFirestore.getInstance()` for other stuff like the batch. And make it final/inline the assignment? Yay, nay?

app/src/main/java/com/firebase/uidemo/database/firestore/FirestorePagingActivity.java
```diff
@@ -0,0 +1,194 @@
...
+                        switch (state) {
+                            case LOADING_INITIAL:
+                            case LOADING_MORE:
+                                mProgressBar.setVisibility(View.VISIBLE);
+                                break;
+                            case LOADED:
+                                mProgressBar.setVisibility(View.GONE);
+                                break;
+                            case ERROR:
+                                showToast("An error occurred.");
```
SUPERCILEX: Do we want to show how to use the retry logic?

app/src/main/java/com/firebase/uidemo/database/firestore/FirestorePagingActivity.java
```diff
@@ -0,0 +1,194 @@
...
+                };
+
+        mRecycler.setLayoutManager(new LinearLayoutManager(this));
+        mRecycler.setAdapter(adapter);
+    }
+
+    @Override
+    public boolean onCreateOptionsMenu(Menu menu) {
+        getMenuInflater().inflate(R.menu.menu_firestore_paging, menu);
+        return super.onCreateOptionsMenu(menu);
```
SUPERCILEX: We should return `true` here to ensure the menu is displayed.

app/src/main/java/com/firebase/uidemo/database/firestore/FirestorePagingActivity.java
```diff
@@ -0,0 +1,194 @@
...
+        }
+        return super.onOptionsItemSelected(item);
+    }
+
+    private Task<Void> createItems() {
+
+        WriteBatch writeBatch = mFirestore.batch();
+        CollectionReference collRef = mFirestore.collection("items");
+
+        for (int i = 0; i < 500; i++) {
```
SUPERCILEX: Isn't 500 a bit excessive for a page size of 20? Maybe like 250 or something?

Also, while I've snagged your attention, the pricing for write batches really isn't clear in the [docs](https://firebase.google.com/docs/firestore/pricing) (no mention, actually). You're still paying for 500 additions, right?

app/src/main/res/values/strings.xml
```diff
@@ -4,11 +4,13 @@
     <!-- Chooser -->
     <string name="title_auth_activity">Auth UI demo</string>
     <string name="title_firestore_activity">Cloud Firestore Demo</string>
+    <string name="title_firestore_paging_activity">Cloud Firestore Paging Demo</string>
     <string name="title_realtime_database_activity">Real-time database demo</string>
     <string name="title_storage_activity">Storage Image Demo</string>
 
     <string name="desc_auth">Demonstrates the Firebase Auth UI flow, with customization options.</string>
     <string name="desc_firestore">Demonstrates using a FirestoreRecyclerAdapter to load data from Cloud Firestore into a RecyclerView for a basic chat app.</string>
+    <string name="desc_firestore_paging">Demonstrates using a FirestorePagingAdapter to load/infinite scroll paging data from Cloud Firestore.</string>
```
SUPERCILEX: `... paging ...` -> `... paged ...`?

firestore/build.gradle
```diff
@@ -22,6 +22,8 @@ dependencies {
     api "com.android.support:recyclerview-v7:$supportLibraryVersion"
     annotationProcessor "android.arch.lifecycle:compiler:$architectureVersion"
 
+    api "android.arch.paging:runtime:1.0.0-beta1"
```
SUPERCILEX: We decided `compileOnly`, right?

firestore/src/main/java/com/firebase/ui/firestore/paging/FirestoreDataSource.java
```diff
@@ -0,0 +1,208 @@
...
+        };
+    }
+
+    /**
+     * Success listener that sets success state and nullifies the retry runnable.
+     */
+    private abstract class OnLoadSuccessListener implements OnSuccessListener<QuerySnapshot> {
+
+        @Override
+        public void onSuccess(QuerySnapshot snapshots) {
```
SUPERCILEX: Nit: `snapshot`

firestore/src/main/java/com/firebase/ui/firestore/paging/FirestorePagingAdapter.java
```diff
@@ -0,0 +1,103 @@
...
+        mData.getLoadingState().removeObserver(mStateObserver);
+    }
+
+    @Override
+    public void onBindViewHolder(@NonNull VH holder, int position) {
+        DocumentSnapshot snapshot = getItem(position);
+        onBindViewHolder(holder, position, mParser.parseSnapshot(snapshot));
+    }
+
+    protected abstract void onBindViewHolder(@NonNull VH holder, int position, T model);
```
SUPERCILEX: `@NonNull T model`

firestore/src/main/java/com/firebase/ui/firestore/paging/FirestorePagingOptions.java
```diff
@@ -0,0 +1,108 @@
...
+            // Build paged list
+            FirestoreDataSource.Factory factory = new FirestoreDataSource.Factory(query);
+            mData = new PagingData(new LivePagedListBuilder<>(factory, config).build());
+
+            mParser = parser;
+            return this;
+        }
+
+        @NonNull
+        public Builder<T> setLifecycleOwner(@NonNull LifecycleOwner owner) {
```
SUPERCILEX: Trying to keep the order consistent, this method should be below the diff callback one.

firestore/src/main/java/com/firebase/ui/firestore/paging/PageKey.java
```diff
@@ -0,0 +1,48 @@
...
+        } else {
+            pageQuery = pageQuery.limit(size);
+        }
+
+        return pageQuery;
+    }
+
+    @Override
+    public String toString() {
+        String startAfter = mStartAfter == null ? "null" : mStartAfter.getId();
```
SUPERCILEX: `String.valueOf(...)`? 😁 Or TBH, I'm pretty sure the compiler does this for you if you just inline everything like so:
```java
        return "PageKey{" +
                "StartAfter=" + mStartAfter.getId() +
                ", EndBefore=" + mEndBefore.getId() +
                '}';
```

firestore/src/main/java/com/firebase/ui/firestore/paging/PagingData.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public LiveData<PagedList<DocumentSnapshot>> getSnapshots() {
+        return mSnapshots;
+    }
+
+    public LiveData<LoadingState> getLoadingState() {
+        return mLoadingState;
+    }
+
+    public LiveData<FirestoreDataSource> getDataSource() { return mDataSource; }
```
SUPERCILEX: Since we're not doing this for the other getters/setters, I think we should use line breaks.

firestore/src/main/java/com/firebase/ui/firestore/paging/PagingData.java
```diff
@@ -0,0 +1,53 @@
...
+import android.support.annotation.NonNull;
+import android.support.annotation.RestrictTo;
+
+import com.google.firebase.firestore.DocumentSnapshot;
+
+/**
+ * All of the data the adapter needs.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+public class PagingData {
```
SUPERCILEX: I'm wondering if this class is necessary at all... isn't it just casting stuff and forwarding calls along? I feel like it would be simpler to just give our adapter a `PagedList` casted to the Firestore one and get the original live datas. Wha'd ya think?

firestore/src/main/java/com/firebase/ui/firestore/paging/SnapshotDiffCallback.java
```diff
@@ -0,0 +1,33 @@
...
+import android.support.v7.util.DiffUtil;
+
+import com.firebase.ui.firestore.SnapshotParser;
+import com.google.firebase.firestore.DocumentSnapshot;
+
+/**
+ * Default diff callback implementation for Firestore snapshots.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+public class SnapshotDiffCallback<T> extends DiffUtil.ItemCallback<DocumentSnapshot> {
```
SUPERCILEX: Maybe `DefaultSnapshotDiffCallback`?

app/src/main/java/com/firebase/uidemo/database/firestore/FirestorePagingActivity.java
samtstern: Re: pricing yeah they are priced per-write.

SUPERCILEX: Re: re: pricing thanks! 😁

app/src/main/java/com/firebase/uidemo/database/firestore/FirestorePagingActivity.java
samtstern: Gonna do all of that besides the inline assignment (not a fan of that in classes where you don't have a constructor, like `Activity` or `Fragment`)

firestore/src/main/java/com/firebase/ui/firestore/paging/FirestoreDataSource.java
samtstern: FYI the reason I keep using `snapshots` is because the IDE does that by default, probably because `QuerySnapshot implements Iterable<QueryDocumentSnapshot>`

firestore/src/main/java/com/firebase/ui/firestore/paging/PageKey.java
samtstern: It's not the ID I am worried about being null, it's `mStartAfter` and `mEndBefore` themselves which can be null. 

In Kotlin I could use `?.` but in Java I have to do this junk.

firestore/src/main/java/com/firebase/ui/firestore/paging/PagingData.java
samtstern: I was fooled by IDE trickery

firestore/src/main/java/com/firebase/ui/firestore/paging/PagingData.java
samtstern: You're absolutely right.  It was only necessary pre beta6

firestore/src/main/java/com/firebase/ui/firestore/paging/PageKey.java
SUPERCILEX: Whoops, yeah, I completely glossed over that one. What you're doing SGTM 👍

firestore/src/main/java/com/firebase/ui/firestore/paging/FirestoreDataSource.java
SUPERCILEX: Haha, good ol' IntelliJ! 😉

firestore/src/main/java/com/firebase/ui/firestore/paging/PagingData.java
SUPERCILEX: Aren't we all? 🤣

samtstern:
Build failure:

> Execution failed for task ':app:preDebugBuild'.
> > Android dependency 'android.arch.lifecycle:livedata-core' has different version for the compile (1.1.0) and runtime (1.1.1) classpath. You should manually set the same version via DependencyResolution

SUPERCILEX:
@samtstern You've got to add the paging dependency to our sample 😉

firestore/src/main/java/com/firebase/ui/firestore/paging/PagingData.java
SUPERCILEX: Ah, thanks for the explanation, I was just about to comment.

samtstern:
Ok @SUPERCILEX no more TODOs on my end (besides the red build, which I know how to fix).  I added tests and docs today.

firestore/README.md
```diff
@@ -109,6 +113,18 @@ updates with the `EventListener` on the `Query`.
 
 Fear not, FirebaseUI does all of this for you automatically!
 
+
+### Choosing an adapter
+
+FirebaseUI offers two types of RecyclerView adapters for Cloud Firestore:
+
+  * `FirestoreRecyclerAdapter` - bind a `Query` to a `RecyclerView` and respond to all real-time
```
SUPERCILEX: I fancy using `—` instead of `-` 😁. Also, should we change the tense to `binds [...] responds ...`?

firestore/src/main/java/com/firebase/ui/firestore/paging/FirestorePagingAdapter.java
```diff
@@ -0,0 +1,146 @@
...
+     * Start listening to paging / scrolling events and populating adapter data.
+     */
+    @OnLifecycleEvent(Lifecycle.Event.ON_START)
+    public void startListening() {
+        mSnapshots.observeForever(mDataObserver);
+        mLoadingState.observeForever(mStateObserver);
+    }
+
+    /**
+     * Unsubscribe from paging / scrolling events, no more data will be populated.
```
SUPERCILEX: Maybe `[...] no more data will be populated, but the existing data will remain.` since that's different from the other adapter.

firestore/README.md
```diff
@@ -227,7 +243,155 @@ FirestoreRecyclerAdapter adapter = new FirestoreRecyclerAdapter<Chat, ChatHolder
...
+
+The `FirestorePagingAdapter` binds a `Query` to a `RecyclerView` by loading documents in pages.
+This results in a time and memory efficient binding, however it gives up the real-time events
+afforted by the `FirestoreRecyclerAdaoter`.
+
+The `FirestorePagingAdapter` is built on top of the [Android Paging Support Library][paging-support].
+Before using the adapter in your application, you must add a dependency on the support library:
+
+```groovy
+implementation 'android.arch.paging:runtime:1.0.0-beta1'
```
SUPERCILEX: Should we use `1.x.x` instead of a specific version since we're going to forget to update it?

firestore/README.md
```diff
@@ -227,7 +243,155 @@ FirestoreRecyclerAdapter adapter = new FirestoreRecyclerAdapter<Chat, ChatHolder
...
+...setQuery(..., new SnapshotParser<Item>() {
+    @NonNull
+    @Override
+    public Item parseSnapshot(@NonNull DocumentSnapshot snapshot) {
+        return ...;
+    }
+});
+```
+
+Next create the `FirestorePagingAdapter` object. You should already have a `ViewHolder` subclass
```
SUPERCILEX: `Next, `

firestore/README.md
```diff
@@ -227,7 +243,155 @@ FirestoreRecyclerAdapter adapter = new FirestoreRecyclerAdapter<Chat, ChatHolder
...
+for displaying each item. In this case we will use a custom `ItemViewHolder` class:
+
+```java
+FirestorePagingAdapter<Item, ItemViewHolder> adapter =
+        new FirestorePagingAdapter<Item, ItemViewHolder>(options) {
+            @NonNull
+            @Override
+            public ItemViewHolder onCreateViewHolder(@NonNull ViewGroup parent, int viewType) {
+                // Create the ItemViewHolder
+                // ,,,
```
SUPERCILEX: Looks someone was a little too far left 😆, `...`.

firestore/README.md
```diff
@@ -227,7 +243,155 @@ FirestoreRecyclerAdapter adapter = new FirestoreRecyclerAdapter<Chat, ChatHolder
...
+            protected void onBindViewHolder(@NonNull ItemViewHolder holder,
+                                            int position,
+                                            @NonNull Item model) {
+                // Bind the item to the view holder
+                // ...
+            }
+        };
+```
+
+Finally attach the adapter to your `RecyclerView` with the `RecyclerView#setAdapter()`.
```
SUPERCILEX: `... method.`

firestore/README.md
```diff
@@ -227,7 +243,155 @@ FirestoreRecyclerAdapter adapter = new FirestoreRecyclerAdapter<Chat, ChatHolder
...
+
+```java
+@Override
+protected void onStart() {
+    super.onStart();
+    adapter.startListening();
+}
+```
+
+Similarly, the `stopListening()` call removes the snapshot listener and all data in the adapter.
```
SUPERCILEX: > removes all data in the adapter

Is this true? Pretty sure all the data is just going to stay there. Unless you meant to call `submitList(Collections.emptyList())` in `stopListening()`?

firestore/README.md
```diff
@@ -227,7 +243,155 @@ FirestoreRecyclerAdapter adapter = new FirestoreRecyclerAdapter<Chat, ChatHolder
...
+
+If you don't want to manually start/stop listening you can use
+[Android Architecture Components][arch-components] to automatically manage the lifecycle of the
+`FirestorePagingAdapter`. Pass a `LifecycleOwner` to
+`FirestorePagingOptions.Builder#setLifecycleOwner(...)` and FirebaseUI will automatically
+start and stop listening in `onStart()` and `onStop()`.
+
+#### Paging events
+
+When using the `FirestorePagingAdapter` you may want to perform some action every time data
```
SUPERCILEX: `... , you may ...`

morganchen12:
Whoops, I'm really late on this PR.

This is fundamentally the same approach that we'll use on iOS, since we can't have both realtime and paging/loading new elements with reasonable scaling performance. We don't have a PagingAdapter thing though, so the way Firestore's data is connected to the UI will be a little bit different. 

I'm not worried about that though. This LGTM

firestore/README.md
```diff
@@ -227,7 +243,157 @@ FirestoreRecyclerAdapter adapter = new FirestoreRecyclerAdapter<Chat, ChatHolder
 ```
 
 
+### Using the `FirestorePagingAdapter`
+
+The `FirestorePagingAdapter` binds a `Query` to a `RecyclerView` by loading documents in pages.
+This results in a time and memory efficient binding, however it gives up the real-time events
+afforted by the `FirestoreRecyclerAdaoter`.
```
jordond: Minor typo in `FirestoreRecyclerAdapter`

muammil:
**Is there any way to delete or update item in FirestorePagingAdapter locally?**

I have a story app in which users post stories of different languages and in my apps home screen I list all of them sort by date using FirestorePagingAdapter. Now I have added language filtering option that is for example if user opted to English, Dutch and Arabic stories app should only show that in main screen. Since firestore doesn't support OR queries I am unable to do that with query what I am currently doing is just hiding items by setting View.Gone from OnBindViewHolder of the FirestorePagingAdapter. What I need is an option to remove those from PagingAdapter list itself

PR #1184, opened 2018-03-11
Don't notify data changed if there were no document changes

SUPERCILEX:
I ran into issues with my loading animations being broken if Firestore sent an update event with no changes. Also, doesn't make sense to have an `onDataChanged` event callback when no data changed. 😉


SUPERCILEX:
Nevermind, I forgot that this will break stuff if the list is actually empty. The original reasoning was the when the device is offline, it will get an empty list if Firestore is unable to load any data which can mess with loading states.

Issue #1179, opened 2018-03-03
Firebase. Not get sms for users with migrate phone numbers


lmdic:
In my **app/build.gradle**:

```
compile 'com.firebaseui:firebase-ui-auth:2.3.0'
compile 'com.google.firebase:firebase-auth:11.0.4'
```

In my fragment when click my custom button I call Firebase authentication screen for SMS authentication

```
button.setOnClickListener(new View.OnClickListener() {
            @Override
            public void onClick(View v) {                   
                FirebaseAuth.getInstance().signOut();
                FirebaseAuth.getInstance().addAuthStateListener(((MainApp) getActivity().getApplication()).firebaseCallback);

                startActivityForResult(
                        AuthUI.getInstance()
                                .createSignInIntentBuilder()
                                .setTheme(R.style.SG_Firebase_Theme)
                                .setAvailableProviders(
                                        Collections.singletonList(new AuthUI.IdpConfig.Builder(AuthUI.PHONE_VERIFICATION_PROVIDER).build())
                                )
                                .build(),
                        RC_SIGN_IN);
            }
        });
```
Steps:

1.After click my button show Firbase screen
2.I input phone number
3.Forward to second Firebase screen
4.Success get SMS code
5.Input SMS
6.Success verify SMS

Nice. It's work fine... with one exception.

In our country we have 3 mobile operators. And any user can migrate from one operator to another **WITHOUT** change it's phone number. It is very convenient for users.

And if this kind of users input it's phone number in first Firebase screen and click button "Confirm phone number", they **NEVER** get SMS confirm code.

But the user that not mirgate to another mobile operator **SUCCESS** get sms.

Why?

samtstern:
@lmdic thanks for this report.

Unfortunately I can't help you here.  If you're having issues with SMS delivery please try Firebase Support:
firebase.google.com/support

Hope you get it figured out!

PR #1185, opened 2018-03-13
Account Linking Phase 1

samtstern:
See #1179 and #309 

TODO:

- [x] Handle register email flow
- [x] Handle the 'welcome back' email flow
- [x] Handle IDP flows
- [x] Handle phone flow
- [ ] Update docs
- [ ] Add comprehensive tests


samtstern:
@SUPERCILEX am I correct interpreting #309 to allow linking between any pre-signed-in account and not just anonymous?

samtstern:
I am also a little confused how to handle email.

Let's say after `CheckEmailFragment` we know that the email already exists.  Then we **know** that calling `linkWithCredential` will fail, so we are definitely going to have a merge conflict.

But, we want to make sure that in the merge conflict handler the `AuthCredential` object they get is always going to be valid ... and the only way to check if the email credential is valid is to attempt a sign-in which will destroy the Anonymous auth state.

Seems like in #309 you'd give the user a chance to read all data from the anon account before attempting the link, but not sure how it would work here.

app/src/main/java/com/firebase/uidemo/auth/AccountLinkActivity.java
```diff
@@ -0,0 +1,197 @@
+package com.firebase.uidemo.auth;
```
samtstern: Decided to do a new sample here since this is an "advanced" feature and I don't want to clutter the main sample.

auth/src/main/java/com/firebase/ui/auth/util/UpgradeFailureListener.java
```diff
@@ -0,0 +1,58 @@
...
+            onUpgradeFailure((FirebaseAuthUserCollisionException) e);
+        } else {
+            onNonUpgradeFailure(e);
+        }
+    }
+
+    protected void onUpgradeFailure(@NonNull FirebaseAuthUserCollisionException e) {
+        // In the case of Phone Auth the exception contains an updated credential we can
+        // user since the other credential is expired after one use.
+        AuthCredential credential = e.getUpdatedCredential() != null
```
samtstern: So this is a cool trick I learned: you get a collision exception from Phone Auth the exception gives you a try-again credential so you don't have to show the UI again.

samtstern:
Removing this from `3.3.0` as #1189 dropped a nuclear merge bomb on this :-)

SUPERCILEX:
Oh noes! 😞 I completely forgot about this PR, sorry man!

samtstern:
@SUPERCILEX haha no worries, I knew exactly what I was getting into.

samtstern:
@SUPERCILEX I've got the merge conflicts squared away.

The two places I am not sure how to apply the linking logic are in `SocialProviderResponseHandler` and `LinkingSocialProviderResponseHandler`.

Mind taking a look?

SUPERCILEX:
Will do! I just realized I forgot some loading stuff in the providers... so PR incoming for that first! 😊

SUPERCILEX:
Testing:
https://github.com/SUPERCILEX/FirebaseUI-Android/blob/460f4307089d3933e2517058758147d725eea567/auth/src/main/java/com/firebase/ui/auth/ui/idp/CredentialSignInHandler.java#L105-L127

samtstern:
@SUPERCILEX thanks for the comments.

The temporary `FirebaseApp` trick isn't to check for a collision, it's for when you **know** there's going to be a collision and want to make sure you're giving the user a valid `AuthCredential`.  So you can check if the credential is valid but without messing up the current sign-in state, so they still have a chance to transfer data.

As for why we're not doing the Service thing: basically because in the API discussion for doing this feature cross-platform the only thing we could agree on in the short term was something minimal and correct like this, while leaving room to improve the developer experience in the future.

app/src/main/java/com/firebase/uidemo/auth/AccountLinkActivity.java
samtstern: If we were going your way with this feature (service, data callbacks, etc) then I'd agree with merging the samples.  However in this case since you have to write extra code and really understand the merge-failure-upgrade flow I wanted to give it dedicated UI and keep the code separate so as not so confuse users of the 'basic' AuthUI flow with all the extra stuff.

If you try out the sample, you can see how it makes the various states extra clear.

auth/src/main/java/com/firebase/ui/auth/IdpResponse.java
```diff
@@ -241,18 +263,32 @@ public boolean equals(Object o) {
...
+        if (mToken != null ? !mToken.equals(response.mToken) : response.mToken != null) {
+            return false;
+        }
+        if (mSecret != null ? !mSecret.equals(response.mSecret) : response.mSecret != null) {
+            return false;
+        }
+        if (mException != null ? !mException.equals(response.mException) : response.mException != null) {
+            return false;
+        }
+        if (mPendingCredential != null ? !mPendingCredential.equals(response.mPendingCredential) : response.mPendingCredential != null) {
```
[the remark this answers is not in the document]
samtstern: To be honest I just let IntelliJ write this, and since these things are now marked `@Nullable` this is what it came up with.

But good call on the `AuthCredential` part.

auth/src/main/java/com/firebase/ui/auth/ui/FragmentBase.java
```diff
@@ -45,6 +46,10 @@ public AuthHelper getAuthHelper() {
         return mAuthHelper;
     }
 
+    public FirebaseAuth getFirebaseAuth() {
+        return mAuthHelper.getFirebaseAuth();
+    }
```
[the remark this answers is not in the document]
samtstern: That's a good law, I will refactor to preserve it.

auth/src/main/java/com/firebase/ui/auth/util/AnonymousUpgradeUtils.java
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * Utilities to help with Anonymous user upgrade.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+public class AnonymousUpgradeUtils {
+
+    /**
+     * Uses type system to enforce the proper failure listener.
+     */
+    public static class UpgradeTaskWrapper<T> {
```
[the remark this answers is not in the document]
samtstern: Yeah that was the plan, to make sure that we always handle that error case.  But you're right that it doesn't quite work, because in some places we have continuations.

app/src/main/java/com/firebase/uidemo/auth/AccountLinkActivity.java
SUPERCILEX: Gotya, that totally makes sense so 👍.

samtstern:
Closing this PR, it's very very conflicty and @lsirac is going to take over this task.

PR #1188, opened 2018-03-15
Prep for final refactors

SUPERCILEX:
@samtstern Alrighty, it's time to plow through that refactor!

This PR makes a few core changes:
- Lots of view models will need a way to start intents so that's now natively supported through custom exceptions. This way, we don't have to subscribe to 3 different `LiveData`s every time. I think that makes the most sense, what are your thoughts?
- Thanks to the above, the auth view model now exposes a single `LiveData` for all subclasses to interact with.
- Other random cleanup and changes I had in fork


auth/src/main/java/com/firebase/ui/auth/util/data/TaskFailureLogger.java
```diff
@@ -0,0 +1,38 @@
...
+import com.google.android.gms.tasks.OnFailureListener;
+
+public class TaskFailureLogger implements OnFailureListener {
+    private static final boolean LOG;
+
+    private String mTag;
+    private String mMessage;
+
+    static {
+        boolean log;
```
samtstern: Huh this is a neat but also kind of nasty trick.  Did you see this somewhere else?  Reflection in a static block seems like it could really slow down some startup times.

Maybe we should just add a FirebaseUI.setLogLevel method or something?

auth/src/main/java/com/firebase/ui/auth/viewmodel/AuthViewModelBase.java
```diff
@@ -1,27 +1,26 @@
...
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-public class AuthViewModelBase extends ViewModelBase<FlowParameters> {
-
+public abstract class AuthViewModelBase<T> extends ViewModelBase<FlowParameters> {
     private CredentialsClient mCredentialsClient;
     private FirebaseAuth mAuth;
     private PhoneAuthProvider mPhoneAuth;
 
-    private SingleLiveEvent<PendingResolution> mPendingResolution = new SingleLiveEvent<>();
+    private MutableLiveData<Resource<T>> mOperation = new MutableLiveData<>();
```
samtstern: Can you quickly explain how you're mitigating the risk that comes with removing the `SingleLiveEvent` for the errors?

auth/src/main/java/com/firebase/ui/auth/util/data/TaskFailureLogger.java
SUPERCILEX: Yeah, I just noticed that some exceptions had sensitive info like emails and whatnot that we shouldn't be logging in production. I'll put this on the TODO list for later to figure out. 👍

samtstern:
This looks really great!

I wonder: can/should we enforce that `Resource<>` goes through states in a sane order?  I am thinking that `SUCCESS` and `FAILURE` should be terminal and any attempt to go from one of those to the other or back to `LOADING` is some sort of problem.

But maybe that's not the case?  Either way it would be good to know/define.

auth/src/main/java/com/firebase/ui/auth/util/ui/FlowUtils.java
```diff
@@ -16,17 +17,20 @@ private FlowUtils() {
     }
 
     public static boolean handleError(@NonNull HelperActivityBase activity, @NonNull Exception e) {
+        if (!(e instanceof StatefulException) || ((StatefulException) e).isUsed()) {
+            return false;
+        }
+        ((StatefulException) e).setUsed(true);
```
samtstern: Maybe the `setUsed` could be automatically called inside `getPendingIntent()` or `getIntent()`  so that way you can really only get them once?

auth/src/main/java/com/firebase/ui/auth/viewmodel/AuthViewModelBase.java
SUPERCILEX: You got it below, it's the `setUsed` thing. I'm totally open to going back to our `SingleLiveEvent`, but that felt like the greater of the two evils. Thoughts?

auth/src/main/java/com/firebase/ui/auth/util/ui/FlowUtils.java
SUPERCILEX: Yeah, I actually did that to start but it felt kinda weird. If you had that idea too, then it can't be _that_ weird. 😆 SGTM!

SUPERCILEX:
@samtstern For the resource stuff, `FAILURE` and `LOADING` states can be called any number of times since the client can try to get around the failures or let the user try again, etc. However, it does make sense to me that we would block anything after `SUCCESS`, do you want to go with that?

auth/src/main/java/com/firebase/ui/auth/util/ui/FlowUtils.java
```diff
@@ -0,0 +1,42 @@
...
+import com.firebase.ui.auth.data.model.IntentRequiredException;
+import com.firebase.ui.auth.data.model.PendingIntentRequiredException;
+import com.firebase.ui.auth.ui.HelperActivityBase;
+
+public final class FlowUtils {
+    private FlowUtils() {
+        throw new AssertionError("No instance for you!");
+    }
+
+    public static boolean handleError(@NonNull HelperActivityBase activity, @NonNull Exception e) {
```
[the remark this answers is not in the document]
SUPERCILEX: I'd like to, but we already have code that sets the failure to `task.getException()`. I'm not sure if we'd want to wrap everything in a `FirebaseUiException`, but that would definitely make this stuff cleaner. Ideas?

auth/src/main/java/com/firebase/ui/auth/data/model/StatefulException.java
```diff
@@ -0,0 +1,10 @@
+package com.firebase.ui.auth.data.model;
+
+import com.firebase.ui.auth.ErrorCodes;
+import com.firebase.ui.auth.FirebaseUiException;
+
+public class StatefulException extends FirebaseUiException {
+    public StatefulException() {
+        super(ErrorCodes.UNKNOWN_ERROR);
```
samtstern: So do we need this class anymore if it doesn't actually hold any state?

auth/src/main/java/com/firebase/ui/auth/ui/credentials/CredentialSaveActivity.java
```diff
@@ -86,10 +71,8 @@ public void onChanged(@Nullable PendingResolution resolution) {
 
     @Override
     protected void onActivityResult(int requestCode, int resultCode, Intent data) {
-        // Forward activity results to the ViewModel
-        if (!mHandler.onActivityResult(requestCode, resultCode, data)) {
-            super.onActivityResult(requestCode, resultCode, data);
-        }
+        super.onActivityResult(requestCode, resultCode, data);
```
samtstern: Are we sure there's no issue unconditionally calling both?

auth/src/main/java/com/firebase/ui/auth/util/ui/FlowUtils.java
```diff
@@ -0,0 +1,43 @@
...
+import com.firebase.ui.auth.data.model.StatefulException;
+import com.firebase.ui.auth.ui.HelperActivityBase;
+
+public final class FlowUtils {
+    private FlowUtils() {
+        throw new AssertionError("No instance for you!");
+    }
+
+    public static boolean handleError(@NonNull HelperActivityBase activity, @NonNull Exception e) {
+        if (!(e instanceof StatefulException)) { return false; }
```
samtstern: Not sure that `StatefulException` is really a useful / meaningful base class (as stated in a comment above).  Why not just return `true` inside each of the `if` blocks and then return `false` at the end otherwise?

auth/src/test/java/com/firebase/ui/auth/viewmodel/SmartLockHandlerTest.java
```diff
@@ -86,20 +86,21 @@ public void testSaveCredentials_resolution() {
         mHandler.saveCredentials(TestHelper.getMockFirebaseUser(), TestConstants.PASSWORD, null);
 
         // Make sure we get a resolution
-        ArgumentCaptor<PendingResolution> resolveCaptor = ArgumentCaptor.forClass(PendingResolution.class);
-        verify(mResolutionObserver).onChanged(resolveCaptor.capture());
+        ArgumentCaptor<Resource<Void>> resolveCaptor = ArgumentCaptor.forClass(Resource.class);
+        verify(mResultObserver, times(2)).onChanged(resolveCaptor.capture());
```
samtstern: Should we change the test to actually verify the `onChanged` events separately?  Kind of confusing to verify that the second time is a `PendingIntentRequiredException` but we don't assert/care what the first event is.

auth/src/main/java/com/firebase/ui/auth/ui/credentials/CredentialSaveActivity.java
SUPERCILEX: Yeah, as long as we have different request codes (we do: 100 and 101). I felt like we'd get nipped in the behind later if `onActivityResult` was called non-deterministically. (Especially since fragments rely on that call.)

auth/src/main/java/com/firebase/ui/auth/util/ui/FlowUtils.java
SUPERCILEX: Totally agree, been there, done that. 👍😆

auth/src/test/java/com/firebase/ui/auth/viewmodel/SmartLockHandlerTest.java
SUPERCILEX: We do [below](https://github.com/firebase/FirebaseUI-Android/pull/1188/files/3acea2daf6aef96b14f3f99b22352b1433e5ee7b#diff-dc55ed1a9fc074c605f9a673fc79e118R98) but I got an exception without the `times(2)`. Is there a prettier way to do this with mockito?

samtstern: Just for learning purposes (I am fine with your code) you can tell Mockito you want to do events in order.

Here is some code from another project of mine which is basically the same (over there we call `Resource` a `Loadable`):

```java
    // Verify order of events
    InOrder inOrder = Mockito.inOrder(loadingObserver);

    // Start by loading
    liveData.postValue(Loadable.forLoading());
    inOrder.verify(loadingObserver).onChanged(true);

    // Move to success
    liveData.postValue(Loadable.forSuccess("Hello"));
    inOrder.verify(loadingObserver).onChanged(false);

    // Make sure that double-success does not make the observer get called again
    liveData.postValue(Loadable.forSuccess("This should have never been posted"));
    inOrder.verify(loadingObserver, never()).onChanged(anyBoolean());
```

SUPERCILEX: Whoa, that looks super cool!

> Just for learning purposes

Says you, I definitely using that! 😆 Thanks for tip 👍

samtstern: Want me to merge-as is or do you want to go and use `InOrder`?  Up to you.

SUPERCILEX: Sorry for the delay, my laptop died in the middle of running tests. 😭😆 Anyway, it's much prettier now!

samtstern:
image
:laughing:

PR #1189, opened 2018-03-16
Completely rewrite provider login with new architecture

### LINKED ISSUE #1200: NoClassDefFoundErrors logcat spam for optional providers
* Android device: Pixel 2
  * Android OS version: P
  * Google Play Services version: 
  * Firebase/Play Services SDK version: 
  * FirebaseUI version: 3.2.2
  
ART verification spams the logcat with NoClassDefFoundErrors, because it needs to resolve superclasses of TwitterProvider and FacebookProvider, which don't exist if you're not using these SDKs. 

Instead of extending/implementing Callbacks from the SDKs directly, these classes should have inner classes for callbacks, that will not be eagerly verified.

#### Steps to reproduce:

  1. Don't include Facebook and Twitter SDKs
  2. Try to sign in using Firebase Auth
  
#### Observed Results:

NoClassDefFoundErrors in logcat
  
#### Expected Results:

No errors

SUPERCILEX:
Built on top of #1188 

Fixes #1200, fixes #1201


auth/src/main/java/com/firebase/ui/auth/ui/idp/WelcomeBackIdpPrompt.java
```diff
@@ -14,71 +14,57 @@
...
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setContentView(R.layout.fui_welcome_back_idp_prompt_layout);
+        setupProvider();
+    }
 
-        IdpResponse newUserResponse = IdpResponse.fromResultIntent(getIntent());
-        if (newUserResponse != null) {
-            mPrevCredential = ProviderUtils.getAuthCredential(newUserResponse);
```
SUPERCILEX: TODO(me): add back account linking.

auth/src/main/java/com/firebase/ui/auth/data/remote/FacebookSignInHandler.java
```diff
@@ -0,0 +1,153 @@
+package com.firebase.ui.auth.data.remote;
```
SUPERCILEX: TODO(me): add tests

auth/src/main/java/com/firebase/ui/auth/data/remote/GoogleSignInHandler.java
```diff
@@ -0,0 +1,92 @@
+package com.firebase.ui.auth.data.remote;
```
SUPERCILEX: TODO(me): add tests

auth/src/main/java/com/firebase/ui/auth/data/remote/TwitterSignInHandler.java
```diff
@@ -0,0 +1,112 @@
+package com.firebase.ui.auth.data.remote;
```
SUPERCILEX: TODO(me): add tests

auth/src/main/java/com/firebase/ui/auth/viewmodel/idp/ProvidersHandler.java
```diff
@@ -0,0 +1,124 @@
+package com.firebase.ui.auth.viewmodel.idp;
```
SUPERCILEX: TODO(me): add tests

SUPERCILEX:
@samtstern This is going to be the biggest PR by far since it requires major refactors across a large portion of the FUI API surface. The basic ideas are as follows:
1. Providers are now middlemen to talk between the UI layer and their "handlers".
1. Each handler is in charge of processing whatever work needs to be done in the background or otherwise to generate an `IdpResponse` which is then reported back to the provider middleman who then reports it back to the activity scoped provider handler who's then in charge of signing the user in and reporting that result back to the activity. Whew! It's quite a long chain, but it makes sure everything is cleanly separated and no memory leaks can occur.
1. `IdpSignInContainer` is dead! 👏 It's been replaced with a dead-simple activity inspired from the credentials one you wrote.

samtstern:
@SUPERCILEX well this is up next, after merges are resolved.  Right?

SUPERCILEX:
@samtstern Yep, I'm working on tests right now.

auth/src/main/java/com/firebase/ui/auth/data/remote/FacebookSignInHandler.java
samtstern: I think just for review sanity you should do the tests in another PR.  This one is already 1k+ lines and tests will probably double it.

auth/src/main/java/com/firebase/ui/auth/data/remote/GoogleSignInHandler.java
```diff
@@ -0,0 +1,97 @@
...
+        mConfig = params.getConfig();
+        mEmail = params.getEmail();
+    }
+
+    public void start() {
+        mRequest.setValue(Resource.forSuccess(
+                GoogleSignIn.getClient(getApplication(), getSignInOptions()).getSignInIntent()));
+    }
+
+    public LiveData<Resource<Intent>> getRequest() {
```
samtstern: Any reason why this is a `Resource`?  It only ever seems to be in the `SUCCESS` state (or `null`).

auth/src/main/java/com/firebase/ui/auth/data/remote/GoogleSignInHandler.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    @Override
+    public void onActivityResult(int requestCode, int resultCode, @Nullable Intent data) {
+        try {
+            GoogleSignInAccount account = GoogleSignIn.getSignedInAccountFromIntent(data)
+                    .getResult(ApiException.class);
+            setResult(createIdpResponse(account));
+        } catch (ApiException e) {
+            if (e.getStatusCode() == CommonStatusCodes.INVALID_ACCOUNT) {
```
samtstern: This took me a minute to understand, so maybe add:

```
// If we get INVALID_ACCOUNT it means the pre-set account was not available on the device, set email to null
// and try to launch sign-in again.
```

auth/src/main/java/com/firebase/ui/auth/provider/EmailProvider.java
```diff
@@ -1,48 +0,0 @@
...
-import android.app.Activity;
-import android.content.Context;
-import android.content.Intent;
-import android.support.annotation.LayoutRes;
-
-import com.firebase.ui.auth.R;
-import com.firebase.ui.auth.data.model.FlowParameters;
-import com.firebase.ui.auth.ui.email.EmailActivity;
-
-public class EmailProvider implements Provider {
```
samtstern: We are deleting a formerly `public` class.  Should this have always been annotated with `@RestrictTo` or this is a real breaking change?

samtstern: Same comment for all providers.

auth/src/main/java/com/firebase/ui/auth/data/remote/FacebookSignInHandler.java
SUPERCILEX: SGTM for the individual provider handlers since they work AFAIK, but I've actually found a bunch of issues with my `ProvidersHandler` implementation thanks to some lovely TDD (so I'll just add that one for now).

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -276,7 +276,7 @@ public static int getDefaultTheme() {
                 .disableAutoSignIn()
                 .continueWithTask(new Continuation<Void, Task<Void>>() {
                     @Override
-                    public Task<Void> then(@NonNull Task<Void> task) throws Exception {
+                    public Task<Void> then(@NonNull Task<Void> task) {
```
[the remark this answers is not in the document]
SUPERCILEX: Yeah, IntelliJ has been removing these, not sure why, but it doesn't seem to matter.

auth/src/main/java/com/firebase/ui/auth/data/remote/FacebookParams.java
```diff
@@ -0,0 +1,18 @@
+package com.firebase.ui.auth.data.remote;
+
+import com.firebase.ui.auth.AuthUI;
+import com.firebase.ui.auth.viewmodel.idp.ProviderHandler;
+import com.firebase.ui.auth.viewmodel.idp.ProvidersHandler;
+
+public final class FacebookParams extends ProviderHandler.ParamsBase {
```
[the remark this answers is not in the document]
SUPERCILEX: Yeah, I'm actually hoping to get rid of these. They're in a separate class because the compiler won't let me make them a static inner...

auth/src/main/java/com/firebase/ui/auth/provider/EmailProvider.java
SUPERCILEX: This was from before we used the restrict annotations, but it's always been considered private API AFAIK:
https://github.com/firebase/FirebaseUI-Android/blob/a381e13957d9e9ea71da51a6ab03cd5f7e782d37/auth/src/main/java/com/firebase/ui/auth/provider/package-info.java

auth/src/main/java/com/firebase/ui/auth/data/remote/GoogleSignInHandler.java
SUPERCILEX: Done.

SUPERCILEX:
@samtstern Alrighty! Hope you had a great weekend, this PR is officially no longer a WIP. 👏 I've tidied a bunch of stuff up, finished implementing everything that needed to get done, and thoroughly tested everything until all those 🐛 were 💥 into oblivion.

The next PR is only a few hundred lines of code so once we get passed this, we'll be on a roll! 🥂 (My poor attempt at a motivational speech and apology for making this PR so big. 😆)

auth/src/main/java/com/firebase/ui/auth/data/remote/FacebookSignInHandler.java
```diff
@@ -0,0 +1,167 @@
...
+        private final LoginResult mResult;
+
+        public ProfileRequest(LoginResult result) {
+            mResult = result;
+        }
+
+        @Override
+        public void onCompleted(JSONObject object, GraphResponse response) {
+            FacebookRequestError error = response.getError();
+            if (error != null || object == null) {
```
samtstern: This `if` statement is a bit over packed.  It has conditions on `error` inside and outside.

Suggest separate if-return-error statements for the different error conditions.

auth/src/main/java/com/firebase/ui/auth/data/remote/FacebookSignInHandler.java
```diff
@@ -0,0 +1,167 @@
...
+            } catch (JSONException ignored) {}
+
+            setResult(createIdpResponse(mResult, email, name, photoUri));
+        }
+    }
+
+    public static final class Params extends ProviderHandlerParamsBase {
+        private final AuthUI.IdpConfig config;
+
+        public Params(ProvidersHandlerBase handler, AuthUI.IdpConfig config) {
```
samtstern: Strange to see a static inner class taking an instance of the parent class ... what's going on here?

auth/src/main/java/com/firebase/ui/auth/data/remote/GoogleSignInHandler.java
samtstern: Should this be a `Resource<Void>` with the `Intent` being part of an error like in other cases then?

auth/src/main/java/com/firebase/ui/auth/data/remote/TwitterSignInHandler.java
```diff
@@ -0,0 +1,119 @@
...
+            TwitterCore.getInstance();
+        } catch (IllegalStateException e) {
+            initialize();
+        }
+
+        TwitterCore.getInstance().getSessionManager().clearActiveSession();
+    }
+
+    private static void initialize() {
+        Context context = AuthUI.getApplicationContext();
```
samtstern: Could this be an instance method instead of a static method and use the `Context` passed into the constructor instead of the `AuthUI` application context?  I realize this would affect `signOut` as well but do we sign out from anywhere that we couldn't create a new instance?

samtstern: My main motivation for asking is the strangeness of a `static initialize()` method.

auth/src/main/java/com/firebase/ui/auth/data/remote/TwitterSignInHandler.java
```diff
@@ -0,0 +1,119 @@
...
+    public Callback getCallback() {
+        return mCallback;
+    }
+
+    @Override
+    public void onActivityResult(int requestCode, int resultCode, @Nullable Intent data) {
+        mClient.onActivityResult(requestCode, resultCode, data);
+    }
+
+    private class Callback extends com.twitter.sdk.android.core.Callback<TwitterSession> {
```
samtstern: Does this class have to be called `Callback`?  It conflicts with the `Twitter` class making this unnecessarily hard to read (as you can see)

auth/src/main/java/com/firebase/ui/auth/data/remote/TwitterSignInHandler.java
```diff
@@ -0,0 +1,119 @@
...
+                            setResult(createIdpResponse(
+                                    sessionResult.data,
+                                    user.email,
+                                    user.name,
+                                    Uri.parse(user.profileImageUrlHttps)));
+                        }
+
+                        @Override
+                        public void failure(TwitterException e) {
+                            setResult(IdpResponse.fromError(e));
```
samtstern: Should we be specifying the `PROVIDER_ERROR` code here (and below)

auth/src/main/java/com/firebase/ui/auth/provider/EmailProvider.java
samtstern: SGTM  let's just mark them all as restricted now

auth/src/main/java/com/firebase/ui/auth/provider/FacebookProvider.java
```diff
@@ -1,212 +0,0 @@
...
-    @Override
-    public void onActivityResult(int requestCode, int resultCode, Intent data) {
-        if (sCallbackManager != null) {
-            sCallbackManager.onActivityResult(requestCode, resultCode, data);
-        }
-    }
-
-    @Override
-    public void onSuccess(final LoginResult loginResult) {
-        GraphRequest request = GraphRequest.newMeRequest(
```
samtstern: Could you do me a favor and mark out anywhere in the new `Handler` classes that you've made a major logic change from the `Provider`?  For instance I notice this whole block is pretty much exactly the same in the new class but as a reviewer I still felt like I had to read it since Github preserved no diff.

auth/src/main/java/com/firebase/ui/auth/provider/TwitterProvider.java
```diff
@@ -1,137 +0,0 @@
...
-    }
-
-    public static AuthCredential createAuthCredential(IdpResponse response) {
-        if (!response.getProviderType().equalsIgnoreCase(TwitterAuthProvider.PROVIDER_ID)) {
-            return null;
-        }
-        return TwitterAuthProvider.getCredential(response.getIdpToken(), response.getIdpSecret());
-    }
-
-    private static void initialize(Context context) {
```
samtstern: I now see that the `initialize` bit I complained about is stolen from old code.  So I won't block this PR on that, but would be cool if we found a simple improvement!

auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
```diff
@@ -99,27 +93,51 @@ protected void onCreate(Bundle savedInstanceState) {
...
+        mHandler.getOperation().observe(this, new Observer<Resource<IdpResponse>>() {
+            @Override
+            public void onChanged(Resource<IdpResponse> resource) {
+                if (resource.getState() == State.LOADING) {
+                    getDialogHolder().showLoadingDialog(R.string.fui_progress_dialog_signing_in);
+                    return;
+                }
+                getDialogHolder().dismissDialog();
+
+                if (resource.isUsed()) { return; }
```
samtstern: Looking at this pattern, I wonder if this would be cleaner:

  * Introduce a new `UsableResource` class
  * Have it have the same three states are we have now, but also a `USED` state

Would that work?  Better or worse?

auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
```diff
@@ -99,27 +93,51 @@ protected void onCreate(Bundle savedInstanceState) {
...
+
+                if (resource.getState() == State.SUCCESS) {
+                    startSaveCredentials(mHandler.getCurrentUser(), null, resource.getValue());
+                } else {
+                    Exception e = resource.getException();
+                    if (!FlowUtils.handleError(AuthMethodPickerActivity.this, e)
+                            && e.getLocalizedMessage() != null) {
```
samtstern: Why do we care if it has a null message if we don't use the message?

auth/src/main/java/com/firebase/ui/auth/ui/provider/EmailProvider.java
```diff
@@ -0,0 +1,56 @@
...
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+public class EmailProvider extends ProviderBase {
+    public EmailProvider(ProvidersHandlerBase handler) {
+        super(handler);
+    }
+
+    @NonNull
+    @Override
+    public String getName() {
```
samtstern: Do we ever call `getName()` from somewhere that we don't have `Context`?  Would prefer to pass it in here rather than rely on `getApplicationContext()` more and more.  Maybe I'm being irrational though?

Other option would be to have this return `@StringRes int` rather than `String`.

auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
SUPERCILEX: I thought of doing something like that too, but then how do we use the `LOADING` state? Our problem is that actions sometimes need to happen every time the activity restarts and other times only once. So far, the terminal-ish states like `SUCCESS` and `FAILURE` seem to only be processed if the resource isn't used. Maybe we could make the `LOADING` state special somehow?

auth/src/main/java/com/firebase/ui/auth/provider/TwitterProvider.java
SUPERCILEX: https://github.com/SUPERCILEX/FirebaseUI-Android/commit/e79388fdfa7b43458cd43da6dbecfa41c095e4f8 makes it a little nicer, what do you think?

auth/src/main/java/com/firebase/ui/auth/ui/email/CheckEmailFragment.java
```diff
@@ -153,23 +150,15 @@ public void onSaveInstanceState(Bundle outState) {
...
-                        // Attempt to proceed
-                        validateAndProceed();
-                    }
-                }
-                break;
-            case RC_SIGN_IN:
-            case RC_WELCOME_BACK_IDP:
-                finish(resultCode, data);
-                break;
+        if (requestCode == RequestCodes.CRED_HINT && data != null) {
```
samtstern: So what happened to RC_SIGN_IN and RC_WELCOME_BACK_IDP?  I see we also dropped the handling for those in EmailActivity.

auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
samtstern: Wouldn't actions that need to be processed every time just not be `Usable` then?

auth/src/main/java/com/firebase/ui/auth/ui/idp/SingleSignInActivity.java
```diff
@@ -0,0 +1,111 @@
...
+                } else {
+                    Exception e = resource.getException();
+                    if (!FlowUtils.handleError(SingleSignInActivity.this, e)) {
+                        finish(RESULT_CANCELED, IdpResponse.getErrorIntent(e));
+                    }
+                }
+            }
+        });
+
+        if (savedInstanceState == null) {
```
samtstern: Is this something we could detect from the `Resource` state rather than from `savedInstanceState`?

auth/src/main/java/com/firebase/ui/auth/ui/idp/SingleSignInActivity.java
```diff
@@ -0,0 +1,111 @@
...
+            mProvider.startLogin(this);
+        }
+    }
+
+    private boolean setup() {
```
samtstern: Why have a setup method rather than just doing the work in `onCreate`?

auth/src/main/java/com/firebase/ui/auth/ui/idp/WelcomeBackIdpPrompt.java
```diff
@@ -111,93 +111,44 @@ protected void onCreate(Bundle savedInstanceState) {
...
+                }
+                getDialogHolder().dismissDialog();
 
-    @Override
-    public void onSuccess(@NonNull final IdpResponse idpResponse) {
-        AuthCredential newCredential = ProviderUtils.getAuthCredential(idpResponse);
-        if (newCredential == null) {
-            throw new IllegalStateException("Unknown provider: " + idpResponse.getProviderType());
-        }
+                if (resource.isUsed()) { return; }
```
samtstern: Maybe if we're going to have this logic everywhere we could instead make our own `Observer` subclass to handle it?

It could have separate callbacks for loading, and first-success or first-finish. 

Whatever we decide, I just want to get somewhere that we're not randomly adding this used check because I know one day we'll forget it and get some rotation bugs again

auth/src/main/java/com/firebase/ui/auth/ui/provider/ProviderBase.java
```diff
@@ -0,0 +1,20 @@
+package com.firebase.ui.auth.ui.provider;
+
+import android.support.annotation.NonNull;
+import android.support.annotation.RestrictTo;
+
+import com.firebase.ui.auth.viewmodel.idp.ProviderResponseHandlerBase;
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+public abstract class ProviderBase implements Provider {
```
samtstern: AbstractProvider?  Strange to see `FooBase` implements `Foo`.

auth/src/main/java/com/firebase/ui/auth/viewmodel/idp/LinkingProviderResponseHandler.java
```diff
@@ -0,0 +1,82 @@
...
+                        public void onComplete(@NonNull Task<AuthResult> task) {
+                            // I'm not sure why we ignore failures here, but this mirrors previous
+                            // behavior.
+                            setResult(Resource.forSuccess(response));
+                        }
+                    });
+        }
+    }
+
+    private class StartLink implements OnSuccessListener<AuthResult> {
```
samtstern: Would this be clearer if this was a `Continuation` and then you add an `onSuccessListener` to the whole chain (signin with credential and this) and call `setResult()` there?

auth/src/main/java/com/firebase/ui/auth/viewmodel/idp/ProviderHandlerBase.java
```diff
@@ -0,0 +1,23 @@
...
+import android.content.Intent;
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+import android.support.annotation.RestrictTo;
+
+import com.firebase.ui.auth.IdpResponse;
+import com.firebase.ui.auth.viewmodel.ViewModelBase;
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+public abstract class ProviderHandlerBase<T extends ProviderHandlerParamsBase> extends ViewModelBase<T> {
```
samtstern: I truly can't get my head around the `Provider`, `Providerbase`, `ProviderHandler`, `ProviderHandlerbase`, `ProviderHandlerParams`, and `ProviderHandlerParamsBase` classes.  The names are so similar and they all seem to be dependent on each other. 

I would love to find a way to simplify them.  I wish I had a more constructive idea and not just this comment.

auth/src/main/java/com/firebase/ui/auth/viewmodel/idp/ProviderHandlerParamsBase.java
```diff
@@ -0,0 +1,18 @@
+package com.firebase.ui.auth.viewmodel.idp;
+
+import android.support.annotation.NonNull;
+import android.support.annotation.RestrictTo;
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+public abstract class ProviderHandlerParamsBase {
+    private final ProviderResponseHandlerBase mHandler;
+
+    protected ProviderHandlerParamsBase(@NonNull ProviderResponseHandlerBase handler) {
```
samtstern: Why do we need a class that just holds a `ProviderResponseHandlerBase`?

auth/src/main/java/com/firebase/ui/auth/viewmodel/idp/ProviderResponseHandlerBase.java
```diff
@@ -0,0 +1,50 @@
...
+import com.firebase.ui.auth.IdpResponse;
+import com.firebase.ui.auth.data.model.Resource;
+import com.firebase.ui.auth.util.data.ProviderUtils;
+import com.firebase.ui.auth.viewmodel.AuthViewModelBase;
+import com.google.firebase.auth.AuthCredential;
+import com.google.firebase.auth.EmailAuthProvider;
+import com.google.firebase.auth.PhoneAuthProvider;
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+public abstract class ProviderResponseHandlerBase extends AuthViewModelBase<IdpResponse> {
```
samtstern: So the params to a viewmodel also contain another viewmodel?  That doesn't seem right to me (although I'm not an expert on arch components)

auth/src/main/java/com/firebase/ui/auth/viewmodel/idp/ProviderResponseHandlerBase.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public void startSignIn(@NonNull IdpResponse inputResponse) {
+        if (checkForInvalidSignInState(inputResponse)) return;
+        startSignIn(ProviderUtils.getAuthCredential(inputResponse), inputResponse);
+    }
+
+    /** Kick off the sign-in process. */
+    public void startSignIn(@NonNull AuthCredential credential,
+                            @NonNull final IdpResponse inputResponse) {
+        if (checkForInvalidSignInState(inputResponse)) return;
```
samtstern: nit: no brace-less `if` statements (here and above)

auth/src/main/res/values-de-rAT/strings.xml
```diff
@@ -1,4 +1,4 @@
-<resources xmlns:tools="http://schemas.android.com/tools" xmlns:xliff="urn:oasis:names:tc:xliff:document:1.2">
```
samtstern: Can we find a way to remove all these changes from the PR diff?  135 changed files makes me uncomfortable.  If you have to leave that one unused string in there, that's fine.

auth/src/main/java/com/firebase/ui/auth/ui/email/CheckEmailFragment.java
SUPERCILEX: Yeah, they've been [squished](https://github.com/firebase/FirebaseUI-Android/pull/1189/files/b1f235685dfe9135d0e51db999b80e8e21e83624#diff-958f7de5a044b2510b5bc0740b2f065bR86) into the `RequestCodes` class for our sanity's sake. So `RC_SIGN_IN` -> `RequestCodes.WELCOME_BACK_EMAIL_FLOW` and `RC_WELCOME_BACK_IDP` ->  `RequestCodes.WELCOME_BACK_IDP_FLOW`.

auth/src/main/java/com/firebase/ui/auth/ui/idp/SingleSignInActivity.java
SUPERCILEX: Legacy code? 😆 I had some reason for making separate methods, but I don't remember now.

auth/src/main/java/com/firebase/ui/auth/viewmodel/idp/LinkingProviderResponseHandler.java
SUPERCILEX: 👍

auth/src/main/res/values-de-rAT/strings.xml
SUPERCILEX: Yeah, I need to get rid of the string to make lint happy, though the rest was IntelliJ's doing. 😆 Time to commit stuff from the CLI! 😢

auth/src/main/java/com/firebase/ui/auth/viewmodel/idp/ProviderHandlerBase.java
SUPERCILEX: Yeah, I was trying to make it so a single `LiveData` could be used for all providers, but clearly, that's just made things confusing. I'll fix that and have the picker activity just subscribe to each provider's individual `LiveData`. Then I think things will quickly become clearer. 👍

SUPERCILEX:
Thanks for the great feedback, I don't mind at all! 😊 I'll address your flow questions here. Things should hopefully seem much clearer now:
- `ProviderBase` holds the context related stuff plus a `LiveData` with an `IdpResponse` so people can subscribe to "provider responses."
- `ProviderResponseHandler` takes in an `IdpResponse` and converts it to a Firebase Auth sign-in, then simply flushes the response through (though that can certainly change if need be in the future).
- `ProviderHandlerBase` is in charge of getting a response for a specific provider implementation from their backend.

I'd recommend looking at these classes from a package perspective:
- `ui` -> `ProviderBase`
- `viewmodel` -> `ProviderResponseHandler`
- `remote` -> `ProviderHandlerBase`

Once everything clicks, I think you'll appreciate the clean separation we have rockin! 😁

auth/src/main/java/com/firebase/ui/auth/viewmodel/idp/ProviderResponseHandlerBase.java
SUPERCILEX: Yup, fixed.

auth/src/main/java/com/firebase/ui/auth/ui/provider/ProviderBase.java
```diff
@@ -1,20 +1,30 @@
...
+import android.support.annotation.StringRes;
 
-import com.firebase.ui.auth.viewmodel.idp.ProviderResponseHandlerBase;
+import com.firebase.ui.auth.IdpResponse;
+import com.firebase.ui.auth.ui.HelperActivityBase;
 
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-public abstract class ProviderBase implements Provider {
-    private final ProviderResponseHandlerBase mHandler;
+public abstract class ProviderBase {
```
samtstern: Your github comments explaining the class hierarchy were very helpful!  Let's also add class-level JavaDoc comments to these new base/handler classes to explain it to future me when I inevitably confuse myself again.

auth/src/main/java/com/firebase/ui/auth/viewmodel/idp/LinkingProviderResponseHandler.java
```diff
@@ -33,11 +33,33 @@ protected void signIn(@NonNull AuthCredential credential,
...
             getAuth().signInWithCredential(credential)
-                    .addOnSuccessListener(new StartLink(response))
-                    .addOnFailureListener(new OnFailureListener() {
+                    .continueWithTask(new Continuation<AuthResult, Task<Void>>() {
                         @Override
-                        public void onFailure(@NonNull Exception e) {
-                            setResult(Resource.<IdpResponse>forFailure(e));
+                        public Task<Void> then(@NonNull Task<AuthResult> task) {
+                            if (mRequestedSignInCredential == null) {
+                                return Tasks.forResult(null);
```
samtstern: Can you add a comment here explaining why in this case we don't care at all if the `signInWithCredential` operation was successful or not?

auth/src/main/java/com/firebase/ui/auth/viewmodel/idp/LinkingProviderResponseHandler.java
```diff
@@ -33,11 +33,33 @@ protected void signIn(@NonNull AuthCredential credential,
...
+                                                return null;
+                                            }
+                                        });
+                            }
+                        }
+                    })
+                    .addOnCompleteListener(new OnCompleteListener<Void>() {
+                        @Override
+                        public void onComplete(@NonNull Task<Void> task) {
+                            if (task.isSuccessful()) {
```
samtstern: Seems like the way you have this set up the task is guaranteed to be successful (in the `if` branch you go straight to success and in the `else` branch you drop errors).  Is that your intention?

I think this class would be improved with a healthy dose of comments since I am a little lost in what you're trying to achieve and why some error cases are ignored.

auth/src/main/res/values-zh-rTW/strings.xml
```diff
@@ -47,7 +47,6 @@
     <string name="fui_error_unknown">發生不明錯誤。</string>
     <string name="fui_error_invalid_password">密碼不正確。</string>
     <string name="fui_accessibility_logo">應用程式標誌</string>
-    <string name="fui_signed_in_with_specific_account">以 %1$s 的身分登入</string>
```
samtstern: Thanks for reverting the `tools` declaration, but can we just revert all changes to the `strings` files?  Is there some lint about an unused resource you're trying to get around?

I would much rather ignore the lint for now than have all of these strings changes cluttering up this very important PR.

auth/src/main/java/com/firebase/ui/auth/viewmodel/idp/ProviderResponseHandlerBase.java
```diff
@@ -0,0 +1,50 @@
...
+        }
+        if (inputResponse.getProviderType().equals(EmailAuthProvider.PROVIDER_ID)
+                || inputResponse.getProviderType().equals(PhoneAuthProvider.PROVIDER_ID)) {
+            setResult(Resource.forSuccess(inputResponse));
+            return true;
+        }
+        return false;
+    }
+
+    protected abstract void signIn(
```
samtstern: Seems duplicative to have both this method and `startSignIn`?  Couldn't this method just check the state and set the loading state in the base class and be annotated with `@CallSuper`?

auth/src/main/java/com/firebase/ui/auth/ui/provider/ProviderBase.java
```diff
@@ -0,0 +1,30 @@
...
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+import android.support.annotation.RestrictTo;
+import android.support.annotation.StringRes;
+
+import com.firebase.ui.auth.IdpResponse;
+import com.firebase.ui.auth.ui.HelperActivityBase;
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+public abstract class ProviderBase {
```
samtstern: Since this class is only abstract methods ... could it be an `interface` instead?

auth/src/main/java/com/firebase/ui/auth/ui/provider/FacebookProvider.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @Override
+    @LayoutRes
+    public int getButtonLayout() {
+        return R.layout.fui_idp_button_facebook;
+    }
+
+    @Override
+    public void startLogin(@NonNull HelperActivityBase activity) {
```
samtstern: 1) nit, but for consistency this should probably be called `startSignIn` since we use the term 'sign in' not 'log in' pretty much everywhere.

2) I find it a bit strange that this class talks directly to `LoginManager`.  Shouldn't all of the Facebook SDK code be contained within the `FacebookSignInHandler` and this class merely acts as a pass-through?

samtstern:
Ok I have spent an hour staring at things and I think we can beat a little more sense into the class hierarchy still.  Here are the bits that still seem somewhat _off_ to me.

`ProviderBase`

  * This class does a few somewhat unrelated things.  The first thing it does is serve as a holder for a `StringRes` and a `LayoutRes`.  These are constants per-provider.  We could move these out into a separate class called `ProviderView` and store them in a simple static map somewhere from `String --> ProviderView` so you could just do `map.get(PROVIDER_ID).getNameRes()` and be done with it.

  * The second thing this class does is hold a `ProviderHandlerBase`, at least for IDPs.  The fact that the class holds one of these is not evident from reading the `ProviderBase` class definition (small problem in itself) but in practice, that's what it does.  It seems that the `startLogin`, `onActivityResult` and `getResponseListener` methods are little more than dumb proxies for the same methods in the `ProviderHandlerBase`.

  * The two classes that act differently here are `EmailProvider` and `PhoneProvider`, since rather than talking to a `Handler` they kick off some other activity and wait for the response.  So maybe this is the justification for the whole abstraction? 

`ProviderHandlerBase`

  * This is only used for social providers (Google, Facebook, etc) whereas Phone and Email are handled locally.  However this is not clear from the class hierarchy because `ProviderBase` does not make any distinction and the class names don't either.

  * Could the phone and email logic be pushed into a `ProviderHandlerBase`?  Or would it not fit that model?

`ProviderResponseHandlerBase`

  * This is basically an `AuthViewModel` with an extra sign in method.
  * Has only two subclasses, `SimpleProviderResponseHandler` and `LinkingProviderResponseHandler`.
  * There's only one place where we use the base class as either an argument or a return type (in `AuthMethodPickerActivity.populateIdpList)
  * I'm wondering if we can just remove the base layer of abstraction here at the expense of an extra `if` block in the auth method picker?

auth/src/main/java/com/firebase/ui/auth/data/model/Resource.java
```diff
@@ -64,20 +80,16 @@ public State getState() {
 
     @Nullable
     public final Exception getException() {
-        mUsed = true;
+        if (mUsable) { mState = State.USED; }
```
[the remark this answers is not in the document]
SUPERCILEX: Yeah, technically, the "resource" part is still immutable since the value and exception can't change. Also, the class is final so only we control the state's mutability, and for now, it can only go from `SUCCESS/FAILURE` -> `USED` so I don't think it'll be a problem unless we want to listen for `USED` events? I'm not sure when that would ever be applicable... I think of the state as more for helping us define our conditionals rather than defining the resource—it should be focused on the data, not the state. That's my opinion though, any digressions?

auth/src/main/java/com/firebase/ui/auth/ui/provider/ProviderBase.java
SUPERCILEX: Haha, yep!

auth/src/main/java/com/firebase/ui/auth/data/model/Resource.java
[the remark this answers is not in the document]
samtstern: Seems OK for now ... just something we may have to change in the future.

auth/src/main/java/com/firebase/ui/auth/viewmodel/idp/LinkingProviderResponseHandler.java
SUPERCILEX: Nope, that was a mistake, thanks for catching it!

auth/src/main/java/com/firebase/ui/auth/viewmodel/idp/LinkingProviderResponseHandler.java
SUPERCILEX: See above, my B.

auth/src/main/res/values-zh-rTW/strings.xml
SUPERCILEX: Oh gotya, I'll tell lint to keep calm and let the resources be unused. 😆

auth/src/main/java/com/firebase/ui/auth/ui/provider/ProviderBase.java
SUPERCILEX: Good catch, I didn't notice!

auth/src/main/java/com/firebase/ui/auth/viewmodel/idp/ProviderResponseHandlerBase.java
SUPERCILEX: Yeah, I've looked at this again and stuff has changed so whatever I was doing is now uncessessary. I'm going to simplify it and remove the abstractions.

auth/src/main/java/com/firebase/ui/auth/ui/provider/FacebookProvider.java
SUPERCILEX: Yup, fixed.

SUPERCILEX: For the second part, I'm not doing that so we don't send the activity over to the view model... do you think we should?

SUPERCILEX:
Thanks for the really thoughtful feedback! 

`ProviderBase`
TBH, this was just around because it was there before. I've looked at where some of the methods are being used and just inlined stuff. I've also moved the sign-in start process to the handler with a HUGE warning not to use the activity outside of that method.

`ProviderHandlerBase` has been squishified into `ProviderBase`.

`ProviderResponseHandlerBase` Yup, I did that in an earlier commit.

Ok, what do you think of https://github.com/SUPERCILEX/FirebaseUI-Android/commit/e9470309dbfb68caf0b45216afed3f13d6abe17b? I think we've finally reached critical mass and there's nothing left to remove. 😆

samtstern:
@SUPERCILEX wooo!  I have to head out a bit early today but on first pass :+1: 

I think this is really clear now, and also more lines deleted than added!  Not bad for a refactor that makes everything work better.

SUPERCILEX:
Alrighty, SGTM! Thanks!

samtstern:
@SUPERCILEX thanks for all the back and forth, all of this code is really clear now (or as clear as the tangled auth mess that is the AuthUI flow will ever be!).

I am going to merge this PR now, if you had any TODOs we can add them to your next one.

Congrats on making this happen, get yourself a beer (or two, or orange juice, or whatever, do you).

SUPERCILEX:
YASSSSSSSS! Stellar use of emojis BTW. 👌😆 You enjoy a beverage too, all of that reviewing must turn your brain into mush! 😳😊

PR #1197, opened 2018-03-20
Ensure auth code sample is clean, clear, and concise

SUPERCILEX:
I was playing with the Play Games provider and got lost in our sample (again 😆). This time, I've gone all out on organizing our auth sample: the strings, layout, and Java are now all perfect mirrors each other vs. the random haphazard mess they were before. Enjoy! 🥂 I also did a bit of general maintenance to hopefully make things clearer to other devs.


samtstern:
@SUPERCILEX thanks for taking the time to do this!  Looks good to me.

PR #1203, opened 2018-03-23
Update dependencies

SUPERCILEX:
12!

### DIFF
--- a/constants.gradle
+++ b/constants.gradle
@@ -8,7 +8,7 @@ project.ext {
     targetSdk = 27
     minSdk = 14
 
-    firebaseVersion = '11.8.0'
+    firebaseVersion = '12.0.0'
     supportLibraryVersion = '27.1.0'
     architectureVersion = '1.1.0'
     kotlinVersion = '1.2.30'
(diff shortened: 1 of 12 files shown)


constants.gradle
```diff
@@ -8,7 +8,7 @@ project.ext {
     targetSdk = 27
     minSdk = 14
 
-    firebaseVersion = '11.8.0'
+    firebaseVersion = '12.0.0'
```
samtstern: Let's wait for `12.0.1` ... `12.0.0` turned out to be kinda broken by some wayward ProGuard.

SUPERCILEX: Yeah, I just saw your other comment. SGTM!

SUPERCILEX: @samtstern Can you also unmark the `CredentialsClient` as `@Hide`? That makes lint think we're not allowed to use its methods... pretty sure you guys only meant to mark the constructors as `@Hide` 😄.

samtstern:
That's another 12.0.0 bug, it's getting fixed.

SUPERCILEX:
Awesome sauce!

samtstern:
@SUPERCILEX 12.0.1 is out, so we should be able to update this, merge it, and release `3.3.0`

samtstern:
@SUPERCILEX looks like `:app:lint` is not happy?  But I think that's my fault, I will merge this and fix that myself.

SUPERCILEX:
Sweet, I was just about to tell you I don't have my laptop with me, thanks! ❤️

PR #1204, opened 2018-03-23
Show loading dialog while providers are fetching extra information

SUPERCILEX:


auth/src/main/java/com/firebase/ui/auth/ui/idp/WelcomeBackIdpPrompt.java
```diff
@@ -128,7 +128,16 @@ protected void onCreate(Bundle savedInstanceState) {
         mProvider.getOperation().observe(this, new Observer<Resource<IdpResponse>>() {
             @Override
             public void onChanged(Resource<IdpResponse> resource) {
```
samtstern: Fine for now, but we should probably invest in our own `Observer` class that abstracts some of this boilerplate.  We almost always do the same things.

SUPERCILEX: Yeah, even IntelliJ is getting fussy:
image

samtstern: Damn you have a fancier IntelliJ than I have!  I guess you don't use Android Studio?

SUPERCILEX: Haha, yeah, my daily driver is IntelliJ EAP because of stuff like that. (Of course, I do have Android Studio Canary so I can hop over there if I need to. 😁)

PR #962, opened 2017-10-10
[FOR PROGRESS, DO NOT MERGE] Rewrite FirebaseUI auth

SUPERCILEX:
#856, #948, #526, future support for #724 and #661, easier #123 implementation. In addition, every config change bug you can think of that people haven't reported.

cc: @samtstern + another human?

I was going to name this PR something more specific like "Fix all config change bugs," but then I saw the LoC changed... sorry @samtstern! 😊 

## Intro

Basically, if you try rotating your screen on pre-PR FirebaseUI, either the user is half signed-in but the UI doesn't show it, or we crash, or we just break randomly. To be frank, we _suck_! For example, rotate the screen after clicking the email sign up button, the user will actually be signed-in, but we're now broken. Click the button again, and we'll say the user already has an existing account. Duh, they just signed up! 😄

If you can't already tell, I'm super excited about this one which is why I'm sharing it early. So far, it just works! Plain and simple, it _works_. On the emulator, I've been mashing the Ctrl + Left/Right buttons to create config change after config change and we nail the sign in every single time. Every. Single. Time!

Anyway, the post-PR world will be a new era of ultimate stability. 😄

TODO:
- [x] An event flow for progress changes and a `ProgressBar` under the `Toolbar`
- [x] Rewrite tests (shouldn't be testing UI, only business logic)
- [ ] Add an `AuthExcecutor` and run all `Task`s synchronously
- [ ] https://firebase.google.com/docs/reference/android/com/google/firebase/auth/FirebaseAuthUserCollisionException
- [x] Refactor phone auth, it's completely broken
- [x] Move `SignInHandler` to a service so we can get rid of the static fields? Nope. I'm open to reconsideration, but this seems like a lot of extra work for little benefit.

## The Problem

Fundamentally, most architectures assume that data doesn't talk back to UI. If we think of the auth process of signing-in and saving credentials etc. as "data," then we run into a problem: the data now needs to tell the UI to do stuff. This means both the user and the data are going to be fighting for control over the UI and need to somehow stay connected. In pre-PR FirebaseUI, this lead to nightmares where we just kinda gave up and ignored auth results if the user rotated their screen.

## The Solution

[...]

PR #1207, opened 2018-03-24
Remove a ton of redundant code

### LINKED ISSUE #1205: Create custom Observer class to abstract common Loading logic
We have a lot of code like this:

```java
mHandler.getOperation().observe(this, new Observer<Resource<IdpResponse>>() {
            @Override
            public void onChanged(Resource<IdpResponse> resource) {
                if (resource.getState() == State.LOADING) {
                    getDialogHolder().showLoadingDialog(R.string.fui_progress_dialog_signing_in);
                    return;
                }
                getDialogHolder().dismissDialog();

                if (resource.getState() == State.SUCCESS) {
                    // ...
                } else if (resource.getState() == State.FAILURE) {
                    // ...
                }
            }
        });
```

In particular the `LOADING` bits almost always look the same.  We could make something like:

```java
public class ResourceObserver<T> extends Observer<Resource<T>> {

          @Override
          public void onChanged(Resource<T> resource) {
                if (resource.getState() == State.LOADING) {
                    onLoading()
                } else {
                    onNotLoading(resource);
                }
          }


          protected void onLoading() {
             // .. Do standard stuf
          }

          protected abstract void onNotLoading(...);
}
```

SUPERCILEX:
@samtstern We're on the last mile for my [rewrite auth PR](https://github.com/firebase/FirebaseUI-Android/pull/962)! There's email and phone left, but I'm going to be busy next week again so I'm hoping get all my PRs open by the end of this weekend. Email will be easy, but phone literally required a complete rewrite. Sorry in advance! 😊 

This PR is starting to rip out all the leftover changes I haven't merged in yet. Also went ahead and fixed #1205 while I was at it.

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/viewmodel/ResourceObserver.java
+++ b/auth/src/main/java/com/firebase/ui/auth/viewmodel/ResourceObserver.java
@@ -1,0 +1,40 @@
+package com.firebase.ui.auth.viewmodel;
+
+import android.arch.lifecycle.Observer;
+import android.support.annotation.NonNull;
+import android.support.annotation.StringRes;
+
+import com.firebase.ui.auth.data.model.Resource;
+import com.firebase.ui.auth.data.model.State;
+import com.firebase.ui.auth.ui.HelperActivityBase;
+import com.firebase.ui.auth.util.ui.FlowUtils;
+
+public abstract class ResourceObserver<T> implements Observer<Resource<T>> {
+    private final HelperActivityBase mActivity;
+    private final int mLoadingMessage;
+
+    protected ResourceObserver(@NonNull HelperActivityBase activity, @StringRes int message) {
+        mActivity = activity;
+        mLoadingMessage = message;
+    }
+
+    @Override
+    public final void onChanged(Resource<T> resource) {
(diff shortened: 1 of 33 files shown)


auth/src/main/java/com/firebase/ui/auth/viewmodel/ResourceObserver.java
```diff
@@ -0,0 +1,42 @@
...
+    private final int mLoadingMessage;
+
+    protected ResourceObserver(@NonNull HelperActivityBase activity, @StringRes int message) {
+        mActivity = activity;
+        mLoadingMessage = message;
+    }
+
+    @Override
+    public final void onChanged(Resource<T> resource) {
+        if (resource.getState() == State.LOADING) {
```
samtstern: This will make it really easy to remove all the dialogs later! Since now we can just handle loading in some other way.

samtstern:
@SUPERCILEX nice!  Btw let's cut off `3.3.0` here.  Phone and Email can be in `3.3.1` (or whatever comes next).  Just in case this big refactor is not as wonderful as we think, I'd like to release it in pieces.

PR #1208, opened 2018-03-25
Fix rotation bugs

SUPERCILEX:
Built on top of #1207 

Darn it! I should have seen this coming, but the usable failure stuff doesn't cover success states or non-usable failure states! 🤦‍♂️

TLDR: I just made `Resource` immutable again and added back the `isUsed` method which is now checked only once in the `ResourceObserver`. The main beef you had with that method was that we were checking it everytime, right? Now we aren't. 😁

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/ui/credentials/CredentialSaveActivity.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/credentials/CredentialSaveActivity.java
@@ -1,6 +1,5 @@
 package com.firebase.ui.auth.ui.credentials;
 
-import android.arch.lifecycle.Observer;
 import android.arch.lifecycle.ViewModelProviders;
 import android.content.Context;
 import android.content.Intent;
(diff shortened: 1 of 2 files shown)


samtstern:
@SUPERCILEX this will be the final `3.3.0` PR after merge conflicts are resolved.

auth/src/main/java/com/firebase/ui/auth/ui/credentials/CredentialSaveActivity.java
```diff
@@ -55,7 +55,7 @@ public void onChanged(@Nullable Resource<Void> resource) {
                     return;
                 }
 
-                onSaveOperation(resource);
```
samtstern: 1) Why is `ResourceObserver` not good for this situation?
2) Having a usable `Resource<Void>` is kind of strange ... but maybe I should just get over it,

SUPERCILEX: That was you! 😆 Nah, I think we weren't doing it because we didn't want to show a loading dialog, but now that I'm looking at the code, we do have a task running in the background so we should be showing one. 🤷‍♂️, I'll use the `ResourceObserver` 😄.

SUPERCILEX: Huh, the void thing is kinda ugly: it messed with our nullability annotations. Not sure how to fix it though without getting rid of the `Resource` for this altogether.

Issue #1039, opened 2017-12-06
Add AuthUI method to perform silent Smart Lock sign-in


SUPERCILEX:
This should be a piece of cake with the new GPS APIs. (Is there a better acronym than GPS? 😄)

### Background
The beauty of Google Smart Lock is that it's able to magically sign users in when the app starts. Currently, FirebaseUI requires a full blown activity and loading UI to do anything with Smart Lock.

### Proposal
We should provide a task based method in `AuthUI` (or another class?) to sign the user in if they have saved credentials. This would work really nicely with Instant Apps where you're supposed to do the magic sign-in stuff.

### Problems
What to do with IDP providers or phone auth? It wouldn't be good UX to randomly start activities... so I guess we're stuck with email auth?

@samtstern What do you think?

samtstern:
Do you mean something that we'd expose in the 3P API or just a refactoring of our smartlock flow?

SUPERCILEX:
Probably a dumb question: what does `3P` mean?

I was thinking of adding a completely new flow that's designed to be started automatically (without user input). Our current flow is based off of the assumption that the user started the request (or at least they see some sort of UI indicating a sign-in is happening). The point was basically to enable the seamless Smart Lock blue banner at the bottom without the user having to initiate a sign-in request. Does that make sense?

samtstern:
3P == Third Party aka Public.  Sorry for the jargon.

Yes that does make sense, however it kind of strays from the core point of the library which is "make Firebase Auth easy" to "make SmartLock easy".  So I'd want to talk to some folks before jumping into this.

SUPERCILEX:
Oh, K. No problemo.

Yeah, I do see your point. That method would of course include using Firebase Auth stuff since it has to do email sign-in, but yeah, that decision is up to you. 😄

samtstern:
This has been fixed and released in `4.0.0`

PR #1212, opened 2018-03-25
Silent sign in

SUPERCILEX:
Proposal for #1039. @samtstern What do you think? Is this something that should go in FirebaseUI Auth? Personally, I think this is huge since it means a zero-effort way for developers to provide seamless sign-in after their users upgrade devices or reinstall their apps.

PS: to try it, just sign in with Google or email, clear app data, and then press the silent sign-in button.

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/AuthUI.java
+++ b/auth/src/main/java/com/firebase/ui/auth/AuthUI.java
@@ -271,14 +271,15 @@ public class AuthUI {
      * Signs the user in without any UI if possible. If this operation fails, you can safely start a
      * UI-based sign-in flow knowing it is required.
      *
-     * @param context the context requesting the user be signed in
+     * @param context requesting the user be signed in
+     * @param configs to use for silent sign in
      * @return a task which indicates whether or not the user was successfully signed in.
      */
     @NonNull
     public Task<AuthResult> silentSignIn(@NonNull final Context context,
                                          @NonNull List<IdpConfig> configs) {
         if (configs.isEmpty()) {
-            throw new IllegalArgumentException("Configs must not be empty.");
+            throw new IllegalArgumentException("At least one provider must be specified.");
         }
         for (IdpConfig config : configs) {
             String provider = config.getProviderId();
--- a/auth/src/main/java/com/firebase/ui/auth/AuthUI.java
+++ b/auth/src/main/java/com/firebase/ui/auth/AuthUI.java
@@ -267,7 +267,16 @@ public class AuthUI {
         return R.style.FirebaseUI;
     }
 
-    public Task<AuthResult> silentSignIn(@NonNull final Context context, List<IdpConfig> configs) {
+    /**
+     * Signs the user in without any UI if possible. If this operation fails, you can safely start a
+     * UI-based sign-in flow knowing it is required.
+     *
+     * @param context the context requesting the user be signed in
+     * @return a task which indicates whether or not the user was successfully signed in.
+     */
+    @NonNull
+    public Task<AuthResult> silentSignIn(@NonNull final Context context,
+                                         @NonNull List<IdpConfig> configs) {
         if (configs.isEmpty()) {
             throw new IllegalArgumentException("Configs must not be empty.");
         }
@@ -279,9 +288,8 @@ public class AuthUI {
                         "for silent sign-in.");
             }
         }
-
         if (mAuth.getCurrentUser() != null) {
-            return Tasks.forException(new IllegalStateException("User already signed in!"));
+            throw new IllegalArgumentException("User already signed in!");
         }
 
         final IdpConfig google =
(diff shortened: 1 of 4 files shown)


samtstern:
@SUPERCILEX this is really cool, but since it's cross-device in nature it's definitely something we'd want iOS and Web support for.  I know that iOS does not have SmartLock, but we could port as much of the idea as possible.

So just FYI this one could be on hold for a bit, as we have to drum up some excitement from the maintainers of those platforms.

SUPERCILEX:
Haha, SGTM!

SUPERCILEX:
Lol, I seem to have broken all of my PRs. Sorry for the spammy updates! ☺️

SUPERCILEX:
@samtstern Alright, created https://github.com/firebase/FirebaseUI-iOS/issues/426 and https://github.com/firebase/firebaseui-web/issues/358.

SUPERCILEX:
I still need to add README docs

SUPERCILEX:
@samtstern Alright, I think this PR is ready for review.

SUPERCILEX:
@samtstern Anything holding this one back?

samtstern:
@SUPERCILEX nothing philosophical holding this back, I was just focusing on `3.3.1` targeted things first.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -262,6 +267,93 @@ public static int getDefaultTheme() {
...
+            }
+        }
+
+        if (configs.isEmpty()) {
+            throw new IllegalArgumentException("No supported providers were supplied. " +
+                    "Add either Google or email support.");
+        }
+
+        final Context appContext = context.getApplicationContext();
+        final IdpConfig google =
```
samtstern: This is kind of a funny way to get here:

  * Run through desired configs, filter out everything besides `google` and `email`
  * Then use `getConfigFromIdps` to get `google` and `email` out of the filtered list

Why not just use `getConfigfromIdps` on `desiredConfigs` and throw the `IllegalArgumentException` if `google` and `email` both end up null?

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -262,6 +267,93 @@ public static int getDefaultTheme() {
...
+            if (last != null && last.getIdToken() != null) {
+                return mAuth.signInWithCredential(GoogleAuthProvider.getCredential(
+                        last.getIdToken(), null));
+            }
+
+            googleOptions = google.getParams()
+                    .getParcelable(ExtraConstants.GOOGLE_SIGN_IN_OPTIONS);
+        }
+
+        final GoogleSignInOptions finalGoogleOptions = googleOptions;
```
samtstern: Why do we need a separate variable to finalize?

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -262,6 +267,93 @@ public static int getDefaultTheme() {
...
+     * Signs the user in without any UI if possible. If this operation fails, you can safely start a
+     * UI-based sign-in flow knowing it is required.
+     *
+     * @param context        requesting the user be signed in
+     * @param desiredConfigs to use for silent sign in. Only Google and email are currently
+     *                       supported, the rest will be ignored.
+     * @return a task which indicates whether or not the user was successfully signed in.
+     */
+    @NonNull
+    public Task<AuthResult> silentSignIn(@NonNull Context context,
```
samtstern: I am somewhat uncomfortable with all this code and chained async operations happening in one huge method block in `AuthUI`.  Should we consider making this happen inside a ViewModel in `KickoffActivity` or something so that it's rotation-immune?  Obviously then we couldn't directly return a `Task` but I think `onActivityResult` would be appropriate (and more consistent with our general mechanism anyway)

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
SUPERCILEX: Lol, yeah, that's a much better way to do it.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
SUPERCILEX: Now we don't!

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
SUPERCILEX: Yeah, I don't like the number of chained tasks we have either, but I strongly disagree about the `onActivityResult` and `ViewModel` stuff. The point of this method is that it doesn't have any UI and should be done in the background (so it's up to the dev to deal with async callbacks as usual). As for rotation problems, they're up to the dev just like with any other `Task`. Also, I tried calling the method multiple times myself and it doesn't seem to matter besides being signed-in several times with the same credentials on the Firebase side (so the method is deterministic in that you'll just get the same result multiple times).

Do you see where I'm coming from? I literally call this method in my `Application#onCreate()` and I think other people will want to do something similar.

PR #1213, opened 2018-03-25
Rewrite email logic with new architecture

SUPERCILEX:
Built on top of #1207 and #1208 

We're so close! After this, all we've got to do is the phone stuff! 😊


SUPERCILEX:
@samtstern Closing in favor of #1214 then.

PR #1214, opened 2018-03-26
Rewrite email logic with new architecture

SUPERCILEX:
Built on top of #1213


auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailFragment.java
```diff
@@ -242,95 +232,13 @@ private void validateAndRegisterUser() {
...
-                    public void onFailure(@NonNull Exception e) {
-                        if (e instanceof FirebaseAuthWeakPasswordException) {
-                            // Password too weak
-                            mPasswordInput.setError(getResources().getQuantityString(
-                                    R.plurals.fui_error_weak_password, R.integer.fui_min_password_length));
-                        } else if (e instanceof FirebaseAuthInvalidCredentialsException) {
-                            // Email address is malformed
-                            mEmailInput.setError(getString(R.string.fui_invalid_email_address));
-                        } else if (e instanceof FirebaseAuthUserCollisionException) {
-                            // Collision with existing user email, it should be very hard for
```
samtstern: Are we changing our assumption from "very hard to get here" to "impossible to get here"?  Fine with me, just confirming.

auth/src/main/java/com/firebase/ui/auth/ui/email/WelcomeBackPasswordPrompt.java
```diff
@@ -141,10 +140,10 @@ public void onDonePressed() {
...
     private void validateAndSignIn() {
-        validateAndSignIn(mEmail, mPasswordField.getText().toString());
+        validateAndSignIn(mPasswordField.getText().toString());
     }
 
-    private void validateAndSignIn(final String email, final String password) {
+    private void validateAndSignIn(String password) {
         // Check for null or empty password
         if (TextUtils.isEmpty(password)) {
             mPasswordLayout.setError(getString(R.string.fui_required_field));
```
samtstern: Should we `return` if the password is empty?

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailFragment.java
samtstern: Oh nevermind, I see the `StartWelcomeBackFlow` in the `ViewModel`

auth/src/main/java/com/firebase/ui/auth/viewmodel/idp/EmailProviderResponseHandler.java
```diff
@@ -0,0 +1,103 @@
...
+                    }
+                })
+                .addOnFailureListener(new OnFailureListener() {
+                    @Override
+                    public void onFailure(@NonNull Exception e) {
+                        if (e instanceof FirebaseAuthUserCollisionException) {
+                            // Collision with existing user email, it should be very hard for
+                            // the user to even get to this error due to CheckEmailFragment.
+                            ProviderUtils.fetchTopProvider(getAuth(), email)
+                                    .addOnSuccessListener(new StartWelcomeBackFlow(email));
```
samtstern: What about when this call fails too?  Feels like we could get a case where the `Resource` never gets to a final state.

auth/src/main/java/com/firebase/ui/auth/ui/email/RegisterEmailFragment.java
SUPERCILEX: Yup, moved that logic there.

auth/src/main/java/com/firebase/ui/auth/ui/email/WelcomeBackPasswordPrompt.java
SUPERCILEX: We do, it looks like the diff hides it 😉

auth/src/main/java/com/firebase/ui/auth/viewmodel/idp/EmailProviderResponseHandler.java
SUPERCILEX: Whoops, yeah.

PR #1229, opened 2018-03-29
Fix email and Google providers creating duplicate credentials

SUPERCILEX:
Ideally this should be merged before #1212 

Since the Google provider upgrades email accounts we end up with two credentials which will pointlessly force Smart Lock UI.

In addition, this PR fixes:
- Email creds not always being deleted
- The void resource stuff
-  Kickstarter email sign-in not displaying a loading dialog


auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -384,11 +385,15 @@ public static int getDefaultTheme() {
...
+            if (type != null) {
+                credentials.add(CredentialsUtils.buildCredentialOrThrow(user, null, type));
+            }
+        }
 
-            credentials.add(new Credential.Builder(
-                    user.getEmail() == null ? user.getPhoneNumber() : user.getEmail())
-                    .setAccountType(type)
-                    .build());
+        // Also add an email credential to ensure password accounts get deleted.
```
samtstern: This confuses me a bit ... feels wrong that the `getCredentialsFromFirebaseUser` method will always return a password credential with a made-up password.

Seems like something that could surprise us down the road, and it makes the method contract kind of a lie.

auth/src/main/java/com/firebase/ui/auth/util/CredentialsUtils.java
```diff
@@ -0,0 +1,69 @@
...
+    @Nullable
+    public static Credential buildCredential(@NonNull FirebaseUser user,
+                                             @Nullable String password,
+                                             @Nullable String accountType) {
+        String email = user.getEmail();
+        String phone = user.getPhoneNumber();
+        Uri profilePictureUri =
+                user.getPhotoUrl() == null ? null : Uri.parse(user.getPhotoUrl().toString());
+
+        if (TextUtils.isEmpty(email) && TextUtils.isEmpty(phone)
```
samtstern: Nit: this would be clearer if separated into two if-return-null statements.

if (no phone and no email) {
  return null;
}

if (no password and no account type) {
  return null
}

auth/src/main/java/com/firebase/ui/auth/viewmodel/VoidResourceObserver.java
```diff
@@ -1,21 +0,0 @@
-package com.firebase.ui.auth.viewmodel;
-
-import android.support.annotation.NonNull;
-import android.support.annotation.RestrictTo;
-import android.support.annotation.StringRes;
-
-import com.firebase.ui.auth.ui.HelperActivityBase;
-
-@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-public abstract class VoidResourceObserver extends ResourceObserver<Void> {
```
samtstern: Byyeeeeee

auth/src/main/java/com/firebase/ui/auth/viewmodel/smartlock/SmartLockHandler.java
```diff
@@ -51,44 +61,54 @@ public void onActivityResult(int requestCode, int resultCode) {
...
-                            setResult(Resource.<Void>forFailure(exception));
+                                    task.getException())));
                         }
                     }
                 });
     }
+
+    private void deleteUnusedCredentials() {
+        if (mResponse.getProviderType().equals(GoogleAuthProvider.PROVIDER_ID)) {
+            // Since Google accounts upgrade email ones, we don't want to end up
```
samtstern: Ok so want to make sure what's going on here:
Whenever we sign in as a Google user we look for matching email credentials and delete them since the Google one is "superior"?

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
SUPERCILEX: Huge thank you on catching that one! Refactor mistake, I've fixed it.

auth/src/main/java/com/firebase/ui/auth/util/CredentialsUtils.java
SUPERCILEX: No probs.

auth/src/main/java/com/firebase/ui/auth/viewmodel/VoidResourceObserver.java
SUPERCILEX: 🤣👍

auth/src/main/java/com/firebase/ui/auth/viewmodel/smartlock/SmartLockHandler.java
SUPERCILEX: Close. You should try this yourself: the "bug" everyone has been complaining about is that when you use the Google provider, it basically overwrites existing providers and pretends that they never existed. AFAIK, this is expected behavior so this removes  the no-longer valid credential. Hope that clarifies things! 😄

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -388,11 +389,11 @@ public static int getDefaultTheme() {
             }
 
             String type = ProviderUtils.providerIdToAccountType(userInfo.getProviderId());
-
-            credentials.add(new Credential.Builder(
-                    TextUtils.isEmpty(user.getEmail()) ? user.getPhoneNumber() : user.getEmail())
-                    .setAccountType(type)
-                    .build());
+            if (type == null) {
```
samtstern: Let's just add a comment here for future-us explaining why we're doing this hack with the "pass" credential.

samtstern: Yeah just explaining why the password doesn't matter (since we're just creating a credential for deletion)

SUPERCILEX: Sweet, will do!

PR #1240, opened 2018-04-04
Fix bad sign out/delete behavior

SUPERCILEX:
Example: the user can be signed out/deleted even though the task failed. Here we ensure the task will be successful _only_ if the user was signed out/deleted.


SUPERCILEX:
@samtstern you're back! 😊💖

Issue #1239, opened 2018-04-03
3.3.0 NullPointerException


xuaninbox:
* Android device: Samsung
  * Android OS version: 8.0
  * Google Play Services version: 12.0.1
  * Firebase/Play Services SDK version: 12.0.1
  * FirebaseUI version: 3.3.0

#### Steps to reproduce:

reported by Crashlytics on 11 devices
  
#### Observed Results:

  crash with NullPointerException
  
#### Expected Results:

  not crash
  
#### Relevant Code:

  ```
Caused by java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String com.google.firebase.auth.FirebaseUser.getDisplayName()' on a null object reference
       at com.firebase.ui.auth.util.CredentialsUtil.buildCredential(CredentialsUtil.java:28)
       at com.firebase.ui.auth.ui.HelperActivityBase.startSaveCredentials(HelperActivityBase.java:107)
       at com.firebase.ui.auth.ui.idp.AuthMethodPickerActivity$1.onSuccess(AuthMethodPickerActivity.java:97)
       at com.firebase.ui.auth.ui.idp.AuthMethodPickerActivity$1.onSuccess(AuthMethodPickerActivity.java:94)
       at com.firebase.ui.auth.viewmodel.ResourceObserver.onChanged(ResourceObserver.java:34)
       at com.firebase.ui.auth.viewmodel.ResourceObserver.onChanged(ResourceObserver.java:13)
       at android.arch.lifecycle.LiveData.considerNotify(LiveData.java:109)
       at android.arch.lifecycle.LiveData.dispatchingValue(LiveData.java:121)
       at android.arch.lifecycle.LiveData.access$400(LiveData.java:59)
       at android.arch.lifecycle.LiveData$ObserverWrapper.activeStateChanged(LiveData.java:416)
       at android.arch.lifecycle.LiveData$LifecycleBoundObserver.onStateChanged(LiveData.java:368)
       at android.arch.lifecycle.LifecycleRegistry$ObserverWithState.dispatchEvent(LifecycleRegistry.java:354)
       at android.arch.lifecycle.LifecycleRegistry.forwardPass(LifecycleRegistry.java:292)
       at android.arch.lifecycle.LifecycleRegistry.sync(LifecycleRegistry.java:332)
       at android.arch.lifecycle.LifecycleRegistry.moveToState(LifecycleRegistry.java:137)
[...]

PR #1243, opened 2018-04-04
Eliminate unnecessary starting of credential save activity for anti-social providers

SUPERCILEX:
@samtstern Well I dunno exactly where the error is coming from, but this fixes some ugliness we had with credential saving and should fix <span></span>#1239 as a side effect.

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
@@ -171,10 +171,15 @@ public class AuthMethodPickerActivity extends AppCompatBase {
                 }
 
                 private void handleResponse(@NonNull IdpResponse response) {
-                    if (!response.isSuccessful()
-                            || AuthUI.SOCIAL_PROVIDERS.contains(response.getProviderType())) {
+                    if (!response.isSuccessful()) {
+                        // We have no idea what provider this error stemmed from so just forward
+                        // this along to the handler.
+                        handler.startSignIn(response);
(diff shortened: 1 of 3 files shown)


auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
```diff
@@ -162,12 +162,22 @@ private void populateIdpList(List<IdpConfig> providerConfigs,
...
                 }
 
                 @Override
                 protected void onFailure(@NonNull Exception e) {
-                    handler.startSignIn(IdpResponse.from(e));
+                    handleResponse(IdpResponse.from(e));
+                }
+
+                private void handleResponse(@NonNull IdpResponse response) {
+                    if (!response.isSuccessful()
```
samtstern: Can we simplify this if statement?  It's kinda melting my brain. 

```java
if (AuthUI.SOCIAL_PROVIDERS.contains(response.getProviderType())) {
  handler.startSignIn(response)
} else if (response.isSuccessful()) {
  finish(RESULT_OK, response.toIntent();
} else {
  finish(RESULT_CANCELED, response.toIntent());
}
```

SUPERCILEX: Unfortunately not, your thing will crash. The only reason I have the successful check is because the providerId will be null if it's a failure. Other ideas? (I agree that the statement is confusing, but that nullability is biting us in the behind.)

SUPERCILEX:
@samtstern changed the purpose and addressed your feedback, what do you think?

samtstern:
LGTM!

Issue #1250, opened 2018-04-11
App crash when calling signOut() if Facebook and Twitter SDK are not imported


btrautmann:
Welcome to FirebaseUI and thanks for submitting an issue!

Please take a look at [open issues](https://github.com/firebase/FirebaseUI-Android/issues?q=is%3Aopen+is%3Aissue), as well as [resolved issues](https://github.com/firebase/FirebaseUI-Android/issues?q=is%3Aissue+is%3Aclosed), to see if your issue is either already being addressed, or has been solved by someone else.

If not, please feel free to fill in the following info so we can help faster!

### Step 1: Are you in the right place?

  * For issues or feature requests related to the code **in this repository** file a GitHub issue.
  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/) tagged appropriately.
  * For general Firebase discussion, use the [firebase-talk google group](https://groups.google.com/forum/#!forum/firebase-talk)
  * For help troubleshooting your application that does not fall under one of the above categories, reach out to the [personalized Firebase support channel](https://firebase.google.com/support/contact/troubleshooting/)

### Step 2: Describe your environment

  * Android device: Android SDK built for x86 (Pixel 2 Emulator)
  * Android OS version: 8.1.0
  * Google Play Services version: 12.0.1
  * Firebase/Play Services SDK version: 3.3.0
  * FirebaseUI version: 3.3.0
  
#### Steps to reproduce:

  1. Set up library as mentioned in configuration
  2. Call following with an authenticated session
```
AuthUI.getInstance()
      .signOut(context)
```
  
#### Observed Results:

App crashes with stacktrace:
```
04-11 12:58:03.033 31405-31405/APP_PACKAGE E/AndroidRuntime: FATAL EXCEPTION: main
    Process: APP_PACKAGE, PID: 31405
    java.lang.NoClassDefFoundError: Failed resolution of: Lcom/facebook/login/LoginManager;
        at com.firebase.ui.auth.AuthUI.signOut(AuthUI.java:1363)
        at APP_PACKAGE.presentation.app.AuthenticationActivity.onCreate(AuthenticationActivity.kt:1040)
        at android.app.Activity.performCreate(Activity.java:7009)
        at android.app.Activity.performCreate(Activity.java:7000)
        at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1214)
        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2731)
[...]

PR #1252, opened 2018-04-12
Permanently fix sign out crashes

SUPERCILEX:
Fixes<span></span> #1250 

It even works on release builds with proguard!


samtstern:
Cool, thanks!

PR #1253, opened 2018-04-12
Rewrite phone auth with new architecture and finalize progress revamp

SUPERCILEX:
This is it! Half a year later and we're almost done! 👏🍾🎉


samtstern:
Not gonna look at this yet but +604 -1406 !!!!

SUPERCILEX:
@samtstern Alrighty, this one is ready for review! Gonna take a look a your PR next.

Some notes about this PR:
- Because of how the verification callbacks are implemented, I couldn't avoid some ugliness. Maybe you'll have some ideas, but the best I could come up with is a shared view model that's bound to the activity. As a side effect, our error states are propagated from the activity down to the correct fragment which seems weird to me. 🤷‍♂️
- I dunno what you've done in the progress dialog PR since I haven't looked at it yet, but I've taken the liberty to kill the completable progress dialog since it doesn't add value IMO. We could literally add a check with a delay to all of our progress dialogs, but we aren't in a sci-fi movie where you need "Access granted" visuals. 😂🛰
- I nuked those alert dialogs in favor of inlined errors in the `TextInputLayout`.

samtstern:
@SUPERCILEX excited to review this one!  But I've come down with a really strong case of I/O fever, so if I get to this at all this week it will be on Friday.  Sorry and appreciate the effort / patience!

auth/src/main/res/values-b+es+419/strings.xml
```diff
@@ -50,17 +50,11 @@
...
     <string name="fui_verify_your_phone_title">Verifica tu número de teléfono</string>
     <string name="fui_verifying">Verificando…</string>
-    <string name="fui_retrieving_sms">Recuperando SMS…</string>
     <string name="fui_incorrect_code_dialog_body">Código incorrecto. Vuelve a intentarlo.</string>
-    <string name="fui_incorrect_code_dialog_positive_button_text">Aceptar</string>
     <string name="fui_error_too_many_attempts">Este número de teléfono se usó demasiadas veces</string>
     <string name="fui_error_quota_exceeded">Ocurrió un problema durante la verificación de tu número de teléfono</string>
     <string name="fui_error_session_expired">Este código ya no es válido</string>
     <string name="fui_sign_in_with_phone_number">Acceder con el número de teléfono</string>
-    <string name="fui_done">Listo</string>
```
samtstern: This is definitely a highly annoying request, but could we split out the string removals into another PR?  That way this one doesn't have 177 files changed and thousands of lines.  I don't want an important behavioral change to get lost among all of this.

auth/src/main/java/com/firebase/ui/auth/data/remote/SignInKickstarter.java
```diff
@@ -127,6 +129,16 @@ private void redirectSignIn(String provider, String email) {
                         EmailActivity.createIntent(getApplication(), getArguments(), email),
                         RequestCodes.EMAIL_FLOW)));
                 break;
+            case PhoneAuthProvider.PROVIDER_ID:
+                Bundle args = new Bundle();
+                args.putString(ExtraConstants.PHONE, email);
```
samtstern: Should we rename the `email` arg to `identifier` or something?  Seems strange to see `put(PHONE, email)`

auth/src/main/java/com/firebase/ui/auth/ui/HelperActivityBase.java
```diff
@@ -70,10 +67,6 @@ public FlowParameters getFlowParams() {
         return mParams;
     }
 
-    public AuthHelper getAuthHelper() {
```
samtstern: Now the helper activity doesn't even have a helper!  Love seeing this class die haha.

auth/src/main/java/com/firebase/ui/auth/ui/phone/CheckPhoneNumberFragment.java
```diff
@@ -0,0 +1,197 @@
...
+        String phone = null;
+        String countryIso = null;
+        String nationalNumber = null;
+        if (params != null) {
+            phone = params.getString(ExtraConstants.PHONE);
+            countryIso = params.getString(ExtraConstants.COUNTRY_ISO);
+            nationalNumber = params.getString(ExtraConstants.NATIONAL_NUMBER);
+        }
+
+        if (!TextUtils.isEmpty(countryIso) && !TextUtils.isEmpty(nationalNumber)) {
```
samtstern: nit: This block is pretty much the heart of the decision making here so I think a comment or two would be good, especially for the final elseif block.

auth/src/main/java/com/firebase/ui/auth/ui/phone/CheckPhoneNumberFragment.java
```diff
@@ -0,0 +1,197 @@
...
+        onNext();
+    }
+
+    private void start(PhoneNumber number) {
+        if (PhoneNumber.isValid(number)) {
+            mPhoneEditText.setText(number.getPhoneNumber());
+            mPhoneEditText.setSelection(number.getPhoneNumber().length());
+        }
+        if (PhoneNumber.isCountryValid(number)) {
+            setCountryCode(number);
```
samtstern: If `isCountryValid` is `true` but `isValid` is `false` then it looks like we'll call `onNext()` with a situation that will cause `getPseudoValidPhoneNumber()` to return null and then put an error on screen.

auth/src/main/java/com/firebase/ui/auth/ui/phone/PhoneActivity.java
```diff
@@ -14,396 +14,182 @@
...
         super.onCreate(savedInstanceState);
         setContentView(R.layout.fui_activity_register_phone);
 
-        mHandler = new Handler();
-        mVerificationState = VerificationState.VERIFICATION_NOT_STARTED;
-        if (savedInstanceState != null && !savedInstanceState.isEmpty()) {
-            mPhoneNumber = savedInstanceState.getString(KEY_VERIFICATION_PHONE);
+        final PhoneProviderResponseHandler handler =
+                ViewModelProviders.of(this).get(PhoneProviderResponseHandler.class);
+        handler.init(getFlowParams());
```
samtstern: nit: name this something more specific than `handler`?  When it's used inside of `verificationHandler` it gets a bit confusing.

auth/src/main/java/com/firebase/ui/auth/ui/phone/PhoneActivity.java
```diff
@@ -14,396 +14,182 @@
...
+        final PhoneNumberVerificationHandler verificationHandler =
+                ViewModelProviders.of(this).get(PhoneNumberVerificationHandler.class);
+        verificationHandler.init(getFlowParams());
+        verificationHandler.getOperation().observe(this, new ResourceObserver<PhoneAuthCredential>(
+                this, R.string.fui_verifying) {
+            @Override
+            protected void onSuccess(@NonNull PhoneAuthCredential credential) {
+                handler.startSignIn(credential, new IdpResponse.Builder(
+                        new User.Builder(PhoneAuthProvider.PROVIDER_ID, null)
+                                .setPhoneNumber(getPhoneNumber())
```
samtstern: Ok looking at the two methods where we do fragment detection, `getErrorView()` and `getPhoneNumber()`, that's where we both agree we could improve.

Here I see that we use `getPhoneNumber()` in the verification handler to kick off the other handler.

What's the original reason why this flow benefits from being a multi-Fragment activity rather than a series of activities?  Would that help cut down on the patterns we don't like?

auth/src/main/res/values-b+es+419/strings.xml
SUPERCILEX: Not at all, I thought you might ask so I removed all the strings in one easily rebasable commit. 👍

auth/src/main/java/com/firebase/ui/auth/ui/HelperActivityBase.java
SUPERCILEX: Same here, that was the first class to die when I started the refactor—it killed me when I had to add it back for the staged PRs. 😆

auth/src/main/java/com/firebase/ui/auth/ui/phone/CheckPhoneNumberFragment.java
SUPERCILEX: Oops, you're right.

auth/src/main/java/com/firebase/ui/auth/ui/phone/PhoneActivity.java
SUPERCILEX: Renamed `verificationHandler` to `phoneVerifier`.

auth/src/main/java/com/firebase/ui/auth/ui/phone/PhoneActivity.java
SUPERCILEX: You actually just gave me a different idea here: I made the phone number pass through the verification view model—not sure why I didn't think of that before. 🤦‍♂️ It still doesn't address the error view thing though, but that's not as bad as the phone number bit IMO.

As for activities, I'm not sure how they would be solving our problem... they would make it harder to talk between them since we'd be relying on a bunch of bundles. Whad'ya think of bfb3b22?

samtstern: Looks better to me!

samtstern:
@SUPERCILEX do you think you can address the UX concerns and all other remaining issues in time for `4.0.0` (deadline for code complete 5/22 right now) or would you like to punt this to a `4.0.1` release?

SUPERCILEX:
@samtstern yeah, probably not going to get this done in time so 4.0.1 it is. (Unless I beat the tos translations and you have time to review—again, not likely. 😂)

auth/src/main/java/com/firebase/ui/auth/ui/phone/SubmitConfirmationCodeFragment.java
```diff
@@ -30,247 +29,187 @@
...
+    @Override
+    public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceState) {
+        mPhoneTextView = view.findViewById(R.id.edit_phone_number);
+        mCountDownTextView = view.findViewById(R.id.ticker);
+        mResendCodeTextView = view.findViewById(R.id.resend_code);
+        mConfirmationCodeEditText = view.findViewById(R.id.confirmation_code);
+        mSubmitConfirmationButton = view.findViewById(R.id.submit_confirmation_code);
+
+        getActivity().setTitle(getString(R.string.fui_verify_your_phone_title));
+        mCountdown.run();
```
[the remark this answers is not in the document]
ArtbyRere: Yes

auth/src/main/java/com/firebase/ui/auth/ui/phone/PhoneActivity.java
```diff
@@ -14,396 +14,167 @@
...
 
-import java.util.concurrent.TimeUnit;
-
 /**
  * Activity to control the entire phone verification flow. Plays host to {@link
- * VerifyPhoneNumberFragment} and {@link SubmitConfirmationCodeFragment}
+ * CheckPhoneNumberFragment} and {@link SubmitConfirmationCodeFragment}
  */
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 public class PhoneActivity extends AppCompatBase {
```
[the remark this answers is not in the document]
SUPERCILEX: That one isn't my fault, it's the Smart Lock dialog. 🤷‍♂️ Not much we can do there.

auth/src/main/java/com/firebase/ui/auth/ui/phone/PhoneActivity.java
```diff
@@ -14,401 +14,204 @@
...
+        final PhoneNumberVerificationHandler phoneVerifier =
+                ViewModelProviders.of(this).get(PhoneNumberVerificationHandler.class);
+        phoneVerifier.init(getFlowParams());
+        phoneVerifier.getOperation().observe(this, new ResourceObserver<PhoneVerification>(
+                this, R.string.fui_verifying) {
+            @Override
+            protected void onSuccess(@NonNull PhoneVerification verification) {
+                if (verification.isAutoVerified()) {
+                    Toast.makeText(
+                            PhoneActivity.this, R.string.fui_verified, Toast.LENGTH_LONG).show();
```
SUPERCILEX: Does this address your auto verified concerns?

auth/src/main/java/com/firebase/ui/auth/ui/phone/PhoneActivity.java
```diff
@@ -14,401 +14,204 @@
...
+                                .build())
+                        .build());
             }
-            return;
-        }
+
+            @Override
+            protected void onFailure(@NonNull Exception e) {
+                if (e instanceof PhoneNumberVerificationRequiredException) {
+                    // Ignore if resending verification code
```
samtstern: Can we add some more explanation for how this comment relates to the code below?

auth/src/main/java/com/firebase/ui/auth/ui/phone/PhoneActivity.java
```diff
@@ -14,401 +14,204 @@
...
-            FragmentTransaction t = getSupportFragmentManager().beginTransaction()
-                    .replace(R.id.fragment_verify_phone, f, SubmitConfirmationCodeFragment.TAG)
-                    .addToBackStack(null);
-
-            if (!isFinishing() && !mIsDestroyed) {
-                t.commitAllowingStateLoss();
-            }
+    @Override
+    public void showProgress(int message) {
+        FragmentBase fragment = (CheckPhoneNumberFragment)
```
samtstern: Could we DRY up the first few lines from `showProgress` and `hideProgress` into some `getActiveFragment` method or something?

auth/src/main/java/com/firebase/ui/auth/ui/phone/PhoneActivity.java
samtstern: Oh also I just went through and tried this, I think we should extend the `fui_verified` method to be something like "Phone number automatically verified"

auth/src/main/java/com/firebase/ui/auth/ui/phone/CheckPhoneNumberFragment.java
```diff
@@ -0,0 +1,239 @@
...
+            start(PhoneNumberUtils.getPhoneNumber(phone));
+        } else if (!TextUtils.isEmpty(countryIso) && !TextUtils.isEmpty(nationalNumber)) {
+            start(PhoneNumberUtils.getPhoneNumber(countryIso, nationalNumber));
+        } else if (!TextUtils.isEmpty(countryIso)) {
+            setCountryCode(new PhoneNumber(
+                    "",
+                    countryIso,
+                    String.valueOf(PhoneNumberUtils.getCountryCode(countryIso))));
+        } else if (getFlowParams().enableHints) {
+            FlowUtils.unhandled(this, new PendingIntentRequiredException(
```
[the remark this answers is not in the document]
SUPERCILEX: Agreed, I'm even using `mHandler.getApplication()` for the context so I must have forgotten this bit sometime while refactoring. 😊

auth/src/main/java/com/firebase/ui/auth/ui/HelperActivityBase.java
```diff
@@ -97,15 +74,13 @@ public void startSaveCredentials(
 
     @Override
     public void showProgress(@StringRes int message) {
-        getDialogHolder().showLoadingDialog(message);
+        throw new UnsupportedOperationException(
+                "Either a fragment or activity must handle progress updates.");
```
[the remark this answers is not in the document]
SUPERCILEX: Huh, I thought we still had some fragments that forwarded progress updates to the activity, but I guess the was in the `ProgressDialog` world. Everything is abstract now. 👍

SUPERCILEX:
@samtstern SGTM, though I had already merged from 4.1 dev. Do you want me to try and rebase those changes out to target 4.0.1?

samtstern:
Oh whoops.  That's my bad. Don't worry about it this can be in 4.1

samtstern:
@SUPERCILEX boom!  Your second big kahuna is now complete!

PR #1254, opened 2018-04-13
Update play services to 15.0.0

samtstern:
This is a big scary version number change but it's actually not a major change from our perspective.  It mostly represents some big internal changes to how we're building these SDKs (more communication to come soon).


samtstern:
@SUPERCILEX yeah the TL;DR on that was these changes were pending for a long time and we needed to just "create space" between this process and the old version numbers so that we could hard-code some assumptions into tools.  But like I said this will all be clearer in the next month or so once we publish some new comms.

SUPERCILEX:
@samtstern Hmmm, _month or so_? I smell some I/O announcements! 😆

PR #1264, opened 2018-04-21
Fix proguard errors when using firestore module without paging

SUPERCILEX:
Built on top of #1263

### DIFF
--- a/firestore/proguard-rules.pro
+++ b/firestore/proguard-rules.pro
@@ -1,0 +1,1 @@
+-dontwarn com.firebase.ui.firestore.paging.**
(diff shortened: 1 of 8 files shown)


firestore/proguard-rules.pro
```diff
@@ -0,0 +1 @@
+-dontwarn com.firebase.ui.firestore.paging.**
```
samtstern: Do we have some dependency from regular --> paging that is causing this?  I am surprised that we need to do this (but not opposed)

SUPERCILEX: Yeah, if someone doesn't use the paging library, proguard is going to complain about a bunch of missing classes.

samtstern: Ohh right.  Yeah that makes sense.

PR #1270, opened 2018-04-26
Scroll to end when keyboard is shown.

b-fg:
Scroll to the end of the chat when a new message is being written and the keyboard is shown.
Previous to this, when the keyboard popped up it was hiding the last messages of the chat.


b-fg:
I signed it!

PR #1271, opened 2018-04-27
Remove unnecessary Zygote class initialization activity when providers are absent

SUPERCILEX:


samtstern:
Oooh nice trick!  I'm pretty sure if I tried to solve this is would be more complicated and not work haha.  Thank you!

PR #1279, opened 2018-05-02
Fix issue 1278

samtstern:
This addresses issue 1278

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/data/remote/TwitterSignInHandler.java
+++ b/auth/src/main/java/com/firebase/ui/auth/data/remote/TwitterSignInHandler.java
@@ -33,12 +33,7 @@ import com.twitter.sdk.android.core.identity.TwitterAuthClient;
 public class TwitterSignInHandler extends ProviderSignInBase<Void> {
     static {
         if (ProviderAvailability.IS_TWITTER_AVAILABLE) {
-            Context context = AuthUI.getApplicationContext();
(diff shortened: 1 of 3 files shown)


auth/src/main/java/com/firebase/ui/auth/data/remote/TwitterSignInHandler.java
```diff
@@ -50,6 +45,18 @@ public TwitterSignInHandler(Application application) {
         mClient = new TwitterAuthClient();
     }
 
+    public static void initializeTwitter() {
+        Context context = AuthUI.getApplicationContext();
+
+        // Note: this has no effect if Twitter is already initialized so it's safe to call
+        // multiple times.
```
SUPERCILEX: I'm wondering if it makes sense to do what we had for the `PhoneNumberUtils`: move the initialization back to the static block and have the initialize method be empty with a comment saying it's forcing static init. What do you think? It just feel weird to be constantly initialising Twitter even though I know it doesn't do anything.

samtstern: If we did that, how would we make initialization happen before `signOut()` is called without getting a ton of bad-class log spam in the case where Twitter deps are not included?

samtstern: SGTM

PR #1301, opened 2018-05-13
Add rules to prevent provider availability false negatives

SUPERCILEX:


samtstern:
@SUPERCILEX good call

PR #1313, opened 2018-05-23
Update Firebase Versions

samtstern:
These changes look breaking but they're actually safe, the major version number was to escape the version-range hell created by the previous release.

### DIFF
--- a/auth/build.gradle
+++ b/auth/build.gradle
@@ -55,7 +55,7 @@ dependencies {
 
     testImplementation 'junit:junit:4.12'
     //noinspection GradleDynamicVersion
-    testImplementation 'org.mockito:mockito-core:2.15.+'
+    testImplementation 'org.mockito:mockito-core:2.18.3'
     testImplementation 'org.robolectric:robolectric:3.7'
     testImplementation 'com.facebook.android:facebook-login:4.32.0'
     testImplementation("com.twitter.sdk.android:twitter-core:3.1.1@aar") { transitive = true }
--- a/auth/src/main/java/com/firebase/ui/auth/AuthUI.java
+++ b/auth/src/main/java/com/firebase/ui/auth/AuthUI.java
@@ -26,6 +26,7 @@ import android.support.annotation.Nullable;
 import android.support.annotation.RestrictTo;
 import android.support.annotation.StringDef;
 import android.support.annotation.StyleRes;
+import android.support.annotation.VisibleForTesting;
 import android.text.TextUtils;
 import android.util.Log;
 
@@ -90,6 +91,9 @@ import java.util.Set;
  */
 public final class AuthUI {
 
+    @VisibleForTesting
+    protected static FirebaseAuth sDefaultAuth;
+
     @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
     public static final String TAG = "AuthUI";
 
@@ -142,7 +146,13 @@ public final class AuthUI {
 
     private AuthUI(FirebaseApp app) {
         mApp = app;
-        mAuth = FirebaseAuth.getInstance(mApp);
+
+        // TODO: This is a bad testing hack
+        if (sDefaultAuth != null) {
+            mAuth = sDefaultAuth;
+        } else {
+            mAuth = FirebaseAuth.getInstance(mApp);
+        }
 
         try {
             mAuth.setFirebaseUIVersion(BuildConfig.VERSION_NAME);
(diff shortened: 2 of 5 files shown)


samtstern:
@SUPERCILEX I am getting some very strange mockito failures on all tests.  Fixing...

samtstern:
Ok I have absolutely no idea what is going on.  And I don't actually think any of our tests were running before ...

SUPERCILEX:
Huh, we are using a dynamic version for mockito so maybe try using a fixed one that we know works?

samtstern:
Oh wait I know what it is.

This beauty definitely stopped working:

```java
    private static void injectMockFirebaseAuth(FirebaseApp app) {
        for (Field field : FirebaseAuth.class.getDeclaredFields()) {
            field.setAccessible(true);

            Object o;
            try {
                o = field.get(null);
            } catch (IllegalAccessException e) {
                throw new IllegalStateException(e);
            } catch (NullPointerException e) {
                continue;
            }

            Type genericType = field.getGenericType();
            if (o instanceof Map && genericType instanceof ParameterizedType) {
                Type[] parameterTypes = ((ParameterizedType) genericType).getActualTypeArguments();
                if (parameterTypes.length != 2 || parameterTypes[0] != String.class
                        || parameterTypes[1] != FirebaseAuth.class) {
                    continue;
                }

                //noinspection unchecked
                Map<String, FirebaseAuth> instances = (Map<String, FirebaseAuth>) o;

                FirebaseAuth.getInstance(app);
                for (String id : instances.keySet()) {
                    instances.put(id, mock(FirebaseAuth.class));
                }

                break;
            }
        }
```

SUPERCILEX:
Ouch, so is FirebaseAuth storing instances of itself differently? Do you want me to take a look or are you good?

samtstern:
I think I can figure it out.  I am now at the point where my tests are green in Android Studio but red on the command line ... man Android testing is a joke!

SUPERCILEX:
Haha, it's usually the other way around for me. 😉 Fingers crossed 🤞, right? 😂

samtstern:
Ok they are all failing like this:

```
org.mockito.exceptions.misusing.InvalidUseOfMatchersException: 
Invalid use of argument matchers!
0 matchers expected, 1 recorded:
-> at com.firebase.ui.auth.testhelpers.TestHelper.injectMockFirebaseAuth(TestHelper.java:135)

This exception may occur if matchers are combined with raw values:
    //incorrect:
    someMethod(anyObject(), "raw String");
When using matchers, all arguments have to be provided by matchers.
For example:
    //correct:
    someMethod(anyObject(), eq("String by matcher"));

For more info see javadoc for Matchers class.
```

Here's the line it doesn't like in `TestHelper`:
```java
        FirebaseAuth mockAuth = FirebaseAuth.getInstance(app);
        when(mockAuth.setFirebaseUIVersion(anyString()))  // <----- THIS ONE
                .thenReturn(Tasks.<Void>forResult(null));
```

Any ideas

SUPERCILEX:
Yup, as feared, the map is gone:
image

I'm looking into where it went.

SUPERCILEX:
Well darn, there's a new `ComponentContainer` interface which means we have to mess with that instead.

SUPERCILEX:
Alright, I think I got it.

SUPERCILEX:
@samtstern Ok, so things aren't looking too pretty. After investigating for ages how we could inject auth into the `final ComponentContainer` implementation, I realized that we don't even need to do it anymore since the `setFirebaseUIVersion` method is wrapped in a try catch anyway. (So that means you should be able to nuke the entire `injectMockFirebaseAuth` method.) However, now I can't get FirebaseAuth to not be null... it's kinda hard to understand how the components are added with minified and obfuscated source code. 😢 To me it looks like the component providers should be coming from the manifest, but I'm not sure.

samtstern:
Thanks for giving it a shot anyway.  I'll try and figure it out by
comparing the obfuscated code to the real source.

samtstern:
@SUPERCILEX after looking at how `FirebaseAuth.getInstance(FirebaseApp)` works now, it's a lot more complicated and not something we really want to be messing with via reflection, especially since a lot of the things we will depend on are obfuscated.

I have put in a very simple (and temporary) hack to get tests to pass: a `protected static FirebaseAuth` variable on `AuthUI` that can be overridden for testing.

Given that we really need to ship 4.0 to make the deadline for GDPR, I am inclined to go ahead with this testing hack and fix it in the next release by doing better injection of `FirebaseAuth`.

samtstern:
Wow squash and merge is very broken.

SUPERCILEX:
@samtstern #1337 addresses the testing hacks.

PR #1315, opened 2018-05-24
Include/exclude countries from the country selector list in phone authentication

lsirac:


lsirac:
I named everything "country codes" but I noticed you guys called them "country isos". I'll change that everywhere soon.

auth/src/main/java/com/firebase/ui/auth/data/client/CountryListLoadTask.java
```diff
@@ -38,267 +42,52 @@
...
+                               List<String> whitelistedCountryCodes,
+                               List<String> blacklistedCountryCodes) {
         mListener = listener;
+        this.whitelistedCountryCodes = whitelistedCountryCodes;
+        this.blacklistedCountryCodes = blacklistedCountryCodes;
     }
 
     @Override
     protected List<CountryInfo> doInBackground(Void... params) {
-        final List<CountryInfo> countryInfoList = new ArrayList<>(MAX_COUNTRIES);
```
samtstern: @SUPERCILEX I rememeber last time you touched this code you had a reason to keep this list separate from the map in `PhoneNumberUtils`.  Is it not exactly inverted?

auth/src/main/java/com/firebase/ui/auth/data/client/CountryListLoadTask.java
```diff
@@ -38,267 +42,52 @@
...
-        countryInfoList.add(new CountryInfo(new Locale("", "ZM"), 260));
-        countryInfoList.add(new CountryInfo(new Locale("", "ZW"), 263));
+        List<CountryInfo> countryInfoList = getAvailableCountryCodes();
         Collections.sort(countryInfoList);
         return countryInfoList;
     }
 
+    public List<CountryInfo> getAvailableCountryCodes() {
+        Map<String, Integer> countryInfoMap = PhoneNumberUtils.getImmutableCountryCodeMap();
+        if (whitelistedCountryCodes == null && blacklistedCountryCodes == null) {
```
samtstern: I think if you want this to be the default you should do this in the constructor and then let every other method in the class assume that the whitelist and blacklist are setup properly.

auth/src/main/java/com/firebase/ui/auth/data/client/CountryListLoadTask.java
```diff
@@ -38,267 +42,52 @@
...
+            whitelistedCountryCodes = new ArrayList<>(countryInfoMap.keySet());
+        }
+
+        List<CountryInfo> countryInfoList = new ArrayList<>();
+
+        Set<String> excludedCountries = new HashSet<>();
+        if (whitelistedCountryCodes == null) {
+            excludedCountries.addAll(blacklistedCountryCodes);
+        } else {
+            excludedCountries.addAll(countryInfoMap.keySet());
```
samtstern: Add some comments here, as the assumptions took me a minute to wrap my head around (I think you're saying that at this point you know only one of whitelist or blacklist will be non-null so you're just handling the two cases for whitelist and assuming the opposite about blacklist)

auth/src/main/java/com/firebase/ui/auth/data/client/CountryListLoadTask.java
samtstern: Might also want to throw an `IllegalStateException` in the constructor if they're both non-null.

auth/src/main/java/com/firebase/ui/auth/ui/phone/VerifyPhoneNumberFragment.java
```diff
@@ -134,6 +135,8 @@ public void onActivityCreated(@Nullable Bundle savedInstanceState) {
         // It is assumed that the phone number that are being wired in via Credential Selector
         // are e164 since we store it.
         Bundle params = getArguments().getBundle(ExtraConstants.PARAMS);
+        List<String> whitelistedCountryCodes = null;
```
samtstern: I think these two variables are not used?

auth/src/main/java/com/firebase/ui/auth/ui/phone/VerifyPhoneNumberFragment.java
```diff
@@ -219,6 +222,14 @@ private String getPseudoValidPhoneNumber() {
     }
 
     private void setupCountrySpinner() {
+        Bundle params = getArguments().getBundle(ExtraConstants.PARAMS);
+        if (params != null) {
+            mCountryListSpinner.setWhitelistedCountryCodes(
+                    params.getStringArrayList(ExtraConstants.WHITELISTED_COUNTRY_CODES));
```
samtstern: What's the default for `getStringArrayList`, is it `null`?

samtstern:
@lsirac as we talked about in chat, let's use `Isos` in the method names and remove `PhoneAuth` from them since that's implied by the `PhoneBuilder` bit.

auth/src/main/java/com/firebase/ui/auth/data/client/CountryListLoadTask.java
lsirac: I'm not sure if I agree with adding IllegalStateException. I moved that bit to the constructor though.

auth/src/main/java/com/firebase/ui/auth/data/client/CountryListLoadTask.java
lsirac: Done.

auth/src/main/java/com/firebase/ui/auth/ui/phone/VerifyPhoneNumberFragment.java
lsirac: Removed them in my last commit.

auth/src/main/java/com/firebase/ui/auth/ui/phone/VerifyPhoneNumberFragment.java
lsirac: Yes, it returns null if the key isn't present.

auth/src/main/java/com/firebase/ui/auth/data/client/CountryListLoadTask.java
samtstern: @lsirac btw holding my approval on this discussion.

samtstern:
I just took a look and, amazingly, it doesn't look like this will create any conflicts with #1253

auth/src/main/java/com/firebase/ui/auth/data/client/CountryListLoadTask.java
samtstern: Ok I went through and actually just printed out the map in each case.  And there is no diff in the map information:
http://www.mergely.com/MhLwmZdA/

auth/src/main/java/com/firebase/ui/auth/data/client/CountryListLoadTask.java
```diff
@@ -38,267 +43,74 @@
...
-    public CountryListLoadTask(Listener listener) {
+    Map<String, Integer> countryInfoMap;
+    List<String> whitelistedCountryIsos;
+    List<String> blacklistedCountryIsos;
+
+    public CountryListLoadTask(Listener listener,
+                               List<String> whitelistedCountries,
+                               List<String> blacklistedCountries) {
         mListener = listener;
+        countryInfoMap = PhoneNumberUtils.getImmutableCountryIsoMap();
```
samtstern: @lsirac would it mess you up if this work was moved to `doInBackground`? I think originally there was a concern that the construction of this map was expensive enough to be done in the background (motivates the whole existence of this class).

If you can move it to there, then I am confident this is a safe change.

samtstern: Looking around, I think that just means you will want to lazy-initialize the big static members of `PhoneNumberUtils` and expose getters for them.

lsirac: Done.

PR #1321, opened 2018-05-28
Join the cool kids club (aka migrate to Gradle Kotlin DSL)

SUPERCILEX:
Surprise, I found another way to use Kotlin in FUI! 🎉😁


app/build.gradle.kts
```diff
@@ -0,0 +1,59 @@
+android {
```
samtstern: Let's add a big comment on this file saying something like "hey we R cool kidz so we use the Kotlin DSL.  If you're writing an app that uses FirebaseUI check out README.md for better examples" or something like that (please not that verbatim)

auth/README.md
```diff
@@ -135,7 +135,7 @@ Twitter app as reported by the [Twitter application manager](https://apps.twitte
 In addition, you must enable the "Request email addresses from users" permission
 in the "Permissions" tab of your Twitter app.
 
-In order to resolve the Twitter SDK, add the following repository to your `build.gradle`:
+In order to resolve the Twitter SDK, add the following repository to your `build.gradle(.kts)`:
```
samtstern: I think we should use the longhand "`build.gradle` or `build.gradle.kts`" everywhere to minimize confusion.

auth/build.gradle.kts
```diff
@@ -0,0 +1,49 @@
...
+        getByName("release") {
+            postprocessing {
+                consumerProguardFiles("auth-proguard.pro")
+            }
+        }
+    }
+
+    lintOptions {
+        disable("UnusedQuantity")
+        disable("UnknownNullness") // Too much to deal with right now
```
samtstern: Fine with the disabling but can you explain what's going on?

build.gradle.kts
```diff
@@ -0,0 +1,340 @@
...
+            setupPublishing(isLibrary)
+        }
+    }
+}
+
+val Project.configDir get() = "$rootDir/library/quality"
+val Project.reportsDir get() = "$buildDir/reports"
+
+fun Project.configureAndroid() {
+    apply(plugin = "com.android.${if (name == "app" || name == "proguard-tests") {
```
samtstern: This is over my personal fanciness limit :-)

How about:

```
if (name == "app" || name == "proguard-tests") {
  // apply app plugin
} else {
  // apply library plugin
```

build.gradle.kts
```diff
@@ -0,0 +1,340 @@
...
+        configFile = file("$configDir/checkstyle.xml")
+        source("src")
+        include("**/*.java")
+        exclude("**/gen/**")
+        classpath = files()
+    }
+}
+
+fun Project.setupPublishing(isLibrary: Boolean) {
+    val publicationName = "${if (isLibrary) "monolith" else name}Library"
```
samtstern: I'd prefer not to have expressions in strings.  How about:

`val publicationName = if (isLibrary) "monolithLibrary" else "${name}Library"`

build.gradle.kts
```diff
@@ -0,0 +1,340 @@
...
+        if (name.contains("publish") && name.contains("publication", true)) {
+            dependsOn("assembleRelease")
+        }
+    }
+
+    afterEvaluate {
+        if (isLibrary) {
+            task("testAll") {
+                dependsOn(*Config.submodules.map {
+                    ":$it:testDebugUnitTest"
```
samtstern: Above we always do `${foo}` even when `$foo` would work and I think that's a good clarity standard, let's do that for all these `$it`s

buildSrc/src/main/kotlin/Config.kt
```diff
@@ -0,0 +1,104 @@
...
+        object PlayServices {
+            const val auth = "com.google.android.gms:play-services-auth:15.0.1"
+        }
+
+        object Provider {
+            const val facebook = "com.facebook.android:facebook-login:4.33.0"
+            const val twitter = "com.twitter.sdk.android:twitter-core:3.1.1@aar"
+        }
+
+        object Miscellaneous {
```
samtstern: nit nit: shorten to `Misc`?

firestore/src/main/java/com/firebase/ui/firestore/paging/PageKey.java
```diff
@@ -20,7 +21,8 @@ public PageKey(@Nullable DocumentSnapshot startAfter, @Nullable DocumentSnapshot
         mEndBefore = endBefore;
     }
 
-    public Query getPageQuery(Query baseQuery, int size) {
+    @NonNull
```
samtstern: Please update the PR title to reflect all the nullability stuff.

internal/lint/src/main/java/com/firebaseui/lint/LintIssueRegistry.kt
```diff
@@ -6,5 +6,5 @@ import com.android.tools.lint.client.api.IssueRegistry
  * Registry for custom FirebaseUI lint checks.
  */
 class LintIssueRegistry : IssueRegistry() {
-    override fun getIssues() = listOf(NonGlobalIdDetector.NON_GLOBAL_ID)
+    override val issues = listOf(NonGlobalIdDetector.NON_GLOBAL_ID)
```
samtstern: `override val` --> Kotlin is cool but also sometimes it's stupid.  This is not an actionable comment, just being grumpy Java guy.

library/build.gradle.kts
```diff
@@ -0,0 +1,14 @@
...
+    lintOptions {
+        isAbortOnError = false
+    }
+}
+
+dependencies {
+    api(project(":auth"))
+    api(project(":database"))
+    api(project(":firestore"))
+    api(project(":storage"))
```
samtstern: Do we need `common` here?

app/build.gradle.kts
SUPERCILEX: Haha, I'm definitely going to quote you on that one! 😁 Side note: do you think we should rename `:app` to `:sample`?

auth/README.md
SUPERCILEX: I've decided to simply remove those references since our samples won't compile in gradle.kts anyway. 🤷‍♂️

auth/build.gradle.kts
SUPERCILEX: Yeah, I realize the nullability changes account for a lot more thrashing than originally anticipated. I've rebased those out and will make another PR with a better explanation. (As a short explanation, Tor is continuing his eternal quest to keep us on our toes with new lint checks. 😆)

build.gradle.kts
SUPERCILEX: Haha, np. 😁😊

build.gradle.kts
SUPERCILEX: I think that just takes some getting used to. Also, it goes directly against the official style guide: https://kotlinlang.org/docs/reference/coding-conventions.html#string-templates.

If you have a strong opinion on this I won't mind, but I _am_ trying to wave an official document at you to scare you off. 😆

buildSrc/src/main/kotlin/Config.kt
SUPERCILEX: \*Goes back to his own project to fix that too\* 😆

firestore/src/main/java/com/firebase/ui/firestore/paging/PageKey.java
SUPERCILEX: Will be moved into another PR

library/build.gradle.kts
SUPERCILEX: We didn't have it before and Gradle should take care of downloading those kinds of transitive dependencies.

internal/lint/src/main/java/com/firebaseui/lint/LintIssueRegistry.kt
SUPERCILEX: Awww, c'mon!? How can you _not_ love overriding fields? It's so clean! ✨

SUPERCILEX:
@samtstern Yup, it fixes the lint issues and I've merged from 4.0.1. 👍

buildSrc/src/main/kotlin/Projects.kt
```diff
@@ -0,0 +1,9 @@
+import org.gradle.api.Project
+import org.gradle.api.Task
+
+fun Project.check(block: Task.() -> Unit) {
+    "check"(this, block)
+}
+
+private inline operator fun String.invoke(project: Project, crossinline block: Task.() -> Unit) =
```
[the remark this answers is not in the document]
SUPERCILEX: Yeah, I'm not super sold on this either, but I've seen similar DSL before: the idea is to override what happens when you call `"someString"()`. In this case, we're saying you can do this:
```kotlin
"someTask"(myProject) {
    // Do stuff with the task
}
```

I don't mind killing it if you're dumbfoundedly confused like how I felt the first time I saw that kind of DSL. 😁

samtstern:
@SUPERCILEX LGTM but the one part I don't agree with is `app` --> `sample`.  Especially since it makes this PR touch 100+ files.  Let's just rename it back to `app` and then we'll merge it and see if we can get `4.0.1` going.

SUPERCILEX:
SGTM, though do you think it should be done at some point? Personally, "app" has always felt vague.

SUPERCILEX:
Oh yeah, that makes sense since people will be familiar with it. 👍

SUPERCILEX:
@samtstern huh, I broke the snapshot builds. 😭 I'll look into it this afternoon. Is standard publishing working?

PR #1325, opened 2018-05-29
Force ToS & Pp to be added to align with FirebaseUI Web & iOS

lsirac:
Deprecated both #setTosUrl() and #setPrivacyPolicyUrl() and added #setTosAndPrivacyPolicyUrls(). 

If only one of ToS/Pp is provided nothing will be shown. Users must provide both.

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/AuthUI.java
+++ b/auth/src/main/java/com/firebase/ui/auth/AuthUI.java
@@ -835,6 +835,8 @@ public final class AuthUI {
         @NonNull
         public T setTosAndPrivacyPolicyUrls(@NonNull String tosUrl,
                                             @NonNull String privacyPolicyUrl) {
+            Preconditions.checkNotNull(tosUrl, "tosUrl cannot be null");
+            Preconditions.checkNotNull(privacyPolicyUrl, "privacyPolicyUrl cannot be null");
             mTosUrl = tosUrl;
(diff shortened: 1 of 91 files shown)


auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -808,22 +808,38 @@ public T setLogo(@DrawableRes int logo) {
...
         @NonNull
+        @Deprecated
         public T setPrivacyPolicyUrl(@Nullable String privacyPolicyUrl) {
             mPrivacyPolicyUrl = privacyPolicyUrl;
             return (T) this;
         }
 
+        @NonNull
+        public T setTosAndPrivacyPolicyUrls(@NonNull String tosUrl,
+                                            @NonNull String privacyPolicyUrl) {
```
samtstern: We should add a `Preconditions` check in here to check the non-null-ness since `@NonNull` is just an IDE hint.

samtstern:
Just one comment, also we need to update the docs here:
https://github.com/firebase/FirebaseUI-Android/blob/master/auth/README.md#adding-a-tos-and-privacy-policy

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
lsirac: Done.

PR #1326, opened 2018-05-30
Start adding nullability annotations to all public or protected symbols

SUPERCILEX:
> ALL HAIL THE LINT GOD, TOR!!! \*thunderous applause\*

Cough cough, ahem. Did you hear something? 😁

Built on top of #1321 which upgrades lint to a version that suggests adding nullability annotations to all public symbols. Also built on top of #1253 so I don't have to step all over myself.

### DIFF
--- a/auth/build.gradle.kts
+++ b/auth/build.gradle.kts
@@ -11,6 +11,7 @@ android {
 
     lintOptions {
         disable("UnusedQuantity")
+        disable("UnknownNullness") // TODO fix in future PR
     }
 
     testOptions {
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -55,6 +55,7 @@ dependencies {
     implementation(Config.Libs.Misc.butterKnife)
     annotationProcessor(Config.Libs.Misc.butterKnifeCompiler)
     debugImplementation(Config.Libs.Misc.leakCanary)
+    debugImplementation(Config.Libs.Misc.leakCanaryFragments)
     releaseImplementation(Config.Libs.Misc.leakCanaryNoop)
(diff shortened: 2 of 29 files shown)


samtstern:
@SUPERCILEX I targeted this for `4.1.0` but let me know if that's not right.

SUPERCILEX:
Nope, that seems right to me.

samtstern:
@SUPERCILEX this has ~1trillion merge conflicts.  Still want to try and get this in for `4.1.0` or should we punt it?

SUPERCILEX:
Eh, punt it. I'll get back to this later.

SUPERCILEX:
@samtstern This one should be fine to merge along with #1421 since it doesn't touch the auth module. 👍

TBH, I feel like the auth module has too many internal classes where nullability annotations don't matter for us to enable it. 🤷‍♂️

auth/build.gradle.kts
```diff
@@ -11,6 +11,7 @@ android {
 
     lintOptions {
         disable("UnusedQuantity")
+        disable("UnknownNullness") // TODO fix in future PR
```
samtstern: I wonder if there's a way we could disable the nullness check on classes that have `@RestrictTo` but leave it on other `public` classes?

samtstern:
@SUPERCILEX LGTM

auth/build.gradle.kts
SUPERCILEX: Damn, that's a great idea! I'll file an issue for Tor.

SUPERCILEX: https://issuetracker.google.com/issues/113439283

PR #1327, opened 2018-05-30
Make artifactory compatible with gradle 4.8

samtstern:
See https://github.com/JFrogDev/build-info/issues/166

Was getting the same issue, updating `build-info-extractor` to `4.7.3` works for me locally.


SUPERCILEX:
Oh sweet, whew 😔

PR #1329, opened 2018-05-31
Fix 4.1.0 lint

samtstern:

PR #1337, opened 2018-06-02
Remove testing hacks

SUPERCILEX:
This solution isn't great since it means we'll have to mock every `FirebaseApp` method we use, but it gives us absolute control over every Firebase API.


auth/src/test/java/com/firebase/ui/auth/testhelpers/TestHelper.java
```diff
@@ -72,18 +74,42 @@ private static void spyContextAndResources() {
...
+        for (Field field : FirebaseApp.class.getDeclaredFields()) {
+            field.setAccessible(true);
+
+            Object o;
+            try {
+                o = field.get(null);
+            } catch (IllegalAccessException e) {
+                throw new IllegalStateException(e);
+            } catch (NullPointerException e) {
+                continue; // Instance field, move on
```
piotrek1543: Seems like one hack is still here ;-)

SUPERCILEX: Lol, yeah. The more accurate title would be `Remove testing hacks in production in favor of hacks in testing code`.

SUPERCILEX:
@samtstern BTW, all 4.1 PRs are going to fail until #1329 is resolved.

samtstern:
@SUPERCILEX are you sure this is better?  It still relies on reflection that's likely to break one day.  May not be any less of a hack than my half-assed dependency injection.

SUPERCILEX:
Yeah, it's not great, but it moves the hacks to our tests instead of the production code. If you don't agree that that's better (I'm on the fence myself), I'll clean up your injection instead. 👍

samtstern:
If you have an idea to clean up the injection I'd be all for that instead!

SUPERCILEX:
@samtstern Voila! Yeah, you're right, this is _way_ better. Thanks! ❤️

samtstern:
@SUPERCILEX nice!

Issue #1339, opened 2018-06-03
Add Lint checks for FirestoreRecyclerAdapter lifecycle methods


pavlospt:
### Describe the problem:

Currently there are no Lint checks on whether someone has called `.startListening()` / `.stopListening()` methods when using a `FirestoreRecyclerAdapter`, or `.setLifecyclerOwner()` when using `FirestoreRecyclerOptions.Builder`. That way the `RecyclerView` using that adapter will never get any UI updates on items from real-time changes. 

#### Steps to reproduce:
Create a `FirestoreRecyclerAdapter` and do not call `.startListening()` / `.stopListening()` methods. 

Alternatively, use `FirestoreRecyclerOptions.Builder` and do not call `.setLifecycleOwner()`.

#### Observed Results:

RecyclerView using the adapter will not get any UI updates from real-time changes without any obvious reason, for those who haven't called `.startListening()` or `.setLifecycleOwner()`.

#### Expected Results:

Expected a Lint check in place, to inform on whether there is a missing call on `.startListening()` or `.setLifecycleOwner()` for `FirestoreRecyclerOptions.Builder`.
  
#### Relevant Code:

I have prepared a relevant PR: #1340

SUPERCILEX:
I like this idea, but I have trouble seeing how it will scale. People might be passing the builders around or managing lifecycle in another class. Basically, I feel like this issue is much more easily addressed by making our documentation clearer since there are tons of APIs out there that require something to kickstart them, us included.

Anyway, that's my initial reaction. @samtstern What do you think? If we do go forward, we'll have to move some modules around since our lint checks were meant to be internal.

samtstern:
@pavlospt first of all thanks for the very interesting suggestion and the extremely well-done PR.

My main concern here, as @SUPERCILEX said, is that we are not currently shipping any external lint checks.  So it would require some module shuffling, probably the creation of a `firebase-ui-lint` artifact or similar.

pavlospt:
Hey, @SUPERCILEX & @samtstern thank you for your opinions on the matter! I am open to any structural suggestion on the project and would be more than happy to contribute more once we end up with a proper solution :) A separate artifact sounds reasonable as well! I can hold off the PR and commit those checks once the structural changes have been applied to the project! 

[...]

PR #1340, opened 2018-06-03
Firestore adapter lifecycle lint checks

pavlospt:
This PR adds the following Lint checks:

1) Missing call on `startListening()` when using a `FirestoreRecyclerAdapter`.

2) Missing call on `stopListening()` when a `startListening()` method call has occurred previously.

3) Missing call on `setLifecycleOwner()` in case `FirestoreRecyclerOptions` are being used.

Relevant issue: https://github.com/firebase/FirebaseUI-Android/issues/1339

### DIFF
--- a/internal/lint/src/main/java/com/firebaseui/lint/FirestoreRecyclerAdapterLifecycleDetector.kt
+++ b/internal/lint/src/main/java/com/firebaseui/lint/FirestoreRecyclerAdapterLifecycleDetector.kt
@@ -1,0 +1,176 @@
+package com.firebaseui.lint
+
+import com.android.tools.lint.client.api.UElementHandler
+import com.android.tools.lint.detector.api.Category.Companion.CORRECTNESS
+import com.android.tools.lint.detector.api.Detector
+import com.android.tools.lint.detector.api.Implementation
+import com.android.tools.lint.detector.api.Issue
+import com.android.tools.lint.detector.api.JavaContext
+import com.android.tools.lint.detector.api.Scope
+import com.android.tools.lint.detector.api.Severity.WARNING
+import org.jetbrains.uast.UCallExpression
+import org.jetbrains.uast.UClass
+import org.jetbrains.uast.UField
+import org.jetbrains.uast.visitor.AbstractUastVisitor
+import java.util.EnumSet
+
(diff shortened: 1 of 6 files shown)


internal/lint/src/main/java/com/firebaseui/lint/FirestoreRecyclerAdapterLifecycleDetector.kt
```diff
@@ -0,0 +1,176 @@
...
+
+      adapterReferences.forEach {
+        if (it.hasCalledStart && !it.hasCalledStop) {
+          context.report(
+              ISSUE_MISSING_LISTENING_STOP_METHOD,
+              it.uField,
+              context.getLocation(it.uField),
+              "Have called .startListening() without .stopListening()."
+          )
+        } else if (!it.hasCalledStart) {
```
samtstern: Does this trigger if `startListening` is not called but `setLifecycleOwner` is called?

pavlospt: Nope have not taken that into account, but I was thinking it of it as well. I am not quite sure on the proper approach to solve this.

Currently `setLifecycleOwner` is applied on `FirestoreRecyclerOptions` which should be applied on `FirestoreRecyclerAdapter`, but there is nothing stopping the user of using both that method, as well as, the `start/stopListening` ones. So I am not quite sure how to prioritize those two together and chose to just add warnings based on the general usage and existence of lifecycle method invocations. That lets the user decide which warning to suppress, effectively acknowledging that it is the desired behavior and not just a method call that they forgot to invoke. 

@samtstern I am open to suggestions on the matter if you have anything to suggest :D

pavlospt: Sure I can do that :)

pavlospt: @samtstern I removed `setLifecyleOwner` related Lint checks and code :)

samtstern:
@pavlospt thanks for making the changes!  This looks good for merging into `internal` and then we'll have to schedule the work of making a public `lint` module.

pavlospt:
Glad to see so :D Thanks @samtstern !

PR #1345, opened 2018-06-08
Country selector fixes

lsirac:
Fixed a merge issue where another PR broke the selector (we need more tests.. :smiley:) 

Added some validation for when the default country is set in the selector via the two overloaded setDefaultNumber() methods. 

When the developer does not provide a default country, we get the user’s locale based on their SIM. If the resulting country is not whitelisted or has been blacklisted, we now choose the first valid country, where the valid countries are sorted in alphabetical order. Before this, an invalid country was displayed in the picker. 

When we successfully fetch the users' number, we'll input it, set the country code, and simulate a 'Verify Phone Number' button click. We now confirm that their country code is valid. If it isn't, we will still auto fill their number, but we don't continue to verify their number.


lsirac:
@SUPERCILEX I moved the logic away from the fragment but I decided against creating a manager/helper. We're not going to reuse the country spinner so I moved the logic there :smiley:. 

Also, based on the params needed & the changes I made I don't think it's easy to make a mistake editing the files now. Let me know if you don't agree.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -678,6 +678,12 @@ public PhoneBuilder setWhitelistedCountries(
                             "You can either whitelist or blacklist country codes for phone " +
                                     "authentication.");
                 }
+
+                String message = "Invalid argument: The whitelist is %s. For this case, " +
```
samtstern: Instead of: "For this case, don't use this method" how about  something like "Only non-null and non-empty whitelists are valid. To specify no whitelist, do not call this method."

lsirac: Much better. :+1:

auth/src/main/java/com/firebase/ui/auth/ui/phone/CheckPhoneNumberFragment.java
```diff
@@ -223,6 +192,53 @@ private void setCountryCode(PhoneNumber number) {
...
+        // Clear error when spinner is clicked on
+        mCountryListSpinner.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
+                mPhoneInputLayout.setError(null);
+            }
+        });
+    }
+
+    private void setDefaultCountryForSpinner() {
```
SUPERCILEX: This method needs to move back to where it was unfortunately (in `onActivityCreated`). Try clicking the phone number in your PR when you're at the submit code fragment. It will go back, but it will also show the phone picker again assuming you didn't pre-fill it in the AuthUI builder.

To see why that happens, read the comment where `mCalled` is used.

auth/src/main/java/com/firebase/ui/auth/ui/phone/CountryListSpinner.java
```diff
@@ -69,8 +73,101 @@ public CountryListSpinner(Context context, AttributeSet attrs, int defStyle) {
...
+    public void init(Bundle params) {
+        if (params != null) {
+            List<CountryInfo> countries = getCountriesToDisplayInSpinner(params);
+            setCountriesToDisplay(countries);
+            setDefaultCountryForSpinner(countries);
+        }
+    }
+
+    private void getCountrySpinnerIsosFromParams(Bundle params) {
+        if (params != null) {
```
SUPERCILEX: Let's:
- rename this method `init...` since it returns void
- remove the null check
- add nullability annotations.

auth/src/main/java/com/firebase/ui/auth/ui/phone/CountryListSpinner.java
```diff
@@ -105,19 +202,22 @@ private static void hideKeyboard(Context context, View view) {
...
+        if (isValidIso(locale.getCountry())) {
+            final String countryName = locale.getDisplayName();
+            if (!TextUtils.isEmpty(countryName) && !TextUtils.isEmpty(countryCode)) {
+                mSelectedCountryName = countryName;
+                setSelectedForCountry(Integer.parseInt(countryCode), locale);
+            }
         }
     }
 
+
```
SUPERCILEX: Nit: unnecessary new line

auth/src/main/java/com/firebase/ui/auth/ui/phone/CountryListSpinner.java
lsirac: Done.

auth/src/main/java/com/firebase/ui/auth/ui/phone/CountryListSpinner.java
lsirac: Done.

PR #1350, opened 2018-06-11
Upgrade to Gradle 4.8

SUPERCILEX:


build.gradle.kts
```diff
@@ -247,6 +251,50 @@ fun Project.setupPublishing() {
...
+                    developers {
+                        developer {
+                            id.set("samtstern")
+                            name.set("Sam Stern")
+                            email.set("samstern@google.com")
+                            organization.set("Firebase")
+                            organizationUrl.set("https://firebase.google.com")
+                            roles.set(listOf("Project-Administrator", "Developer"))
+                            timezone.set("-8")
+                        }
```
SUPERCILEX: @lsirac If you can send me a developer DSL like this, I'll add you to the list too. 👍

buildSrc/src/main/kotlin/Config.kt
```diff
@@ -11,11 +11,11 @@ object Config {
     }
 
     object Plugins {
-        const val android = "com.android.tools.build:gradle:3.2.0-alpha15"
```
samtstern: I didn't even notice that we had moved to `alpha` in the gradle --> kotlin switch.  Is there anything more stable that will work for us?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.7-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.8-all.zip
```
samtstern: So has the issue we had with 4.8 been fixed?

build.gradle.kts
samtstern: BTW @lsirac totally optional, if you don't want to expose your email address or name you can skip.

gradle/wrapper/gradle-wrapper.properties
SUPERCILEX: Yup, in bintray 1.8.1.

buildSrc/src/main/kotlin/Config.kt
SUPERCILEX: That was for lint to work. Also, we already shipped v4.0.1 without any problems so I wouldn't worry about it.

SUPERCILEX: Ohhhhhhhhh, but I forgot that it will make it hard for people to use the sample 😕. Well unless we want to undo a few weeks of PRs, there's no going back. Sorry about that! 😊

samtstern: NP let's just get onto a beta-or-better train as soon as one is available.

samtstern:
I am going to merge and let @lsirac send a PR later for adding his credits to the POM (completely optional)

Issue #1242, opened 2018-04-04
Expose isNewUser information in IDPResponse


samtstern:
See:
https://github.com/firebase/FirebaseUI-Android/issues/1209#issuecomment-378708048

Currently we tell developers to rely on the created / signed in timestamps when checking to see if a `FirebaseUser` is new or not.  However these timestamps have some limitations that make them occasionally unreliable.

If we could expose this information in the response we provide at the end of the AuthUI flow, that would help developers launch follow-up flows.

curiousily:
Is there any **reliable way** to workaround this using the current version of the library?

samtstern:
@curiousily the timestamps are reliable with the edge case issue that if a new user signs up, signs out, and signs in within two minutes they will still look "new" because the timestamps wont change.

j2emanue:
you guys can also send a flag in onActivityResult intent. like a boolean to represent if its new or existing.

j2emanue:
guys, its me again. so i noticed one strange thing.  my business has both website and app but use same firebase authentication table.   so i signed up for a new account on our website as an email provider.  checked firebase console and saw the account there as email provider. waited almost 2 hours..... then logged into app with email provider, but to my surprise it found it as a new signup.  so i checked and the following was true, why ?

**metadata.creationTimestamp == metadata.lastSignInTimestamp**

after waiting almost 2 hours should this not be false.  is this check only local to the mobile application ?

curiousily:
Hey @samtstern,

Any progress on this issue? This really degrades the experience for our users.

samtstern:
This has been released in version `4.1.0`

AndrazP:
Hi @samtstern 
I found this old issue after noticing that comparing timestamps
`metadata.getCreationTimestamp() == metadata.getLastSignInTimestamp()`
is not reliable. `isNewUser` seems to fix it.
Are there any drawbacks of using this method? Why is comparing timestamps still recommended in README?

samtstern:
@AndrazP I think I just forgot to update the README!  `isNewUser` is preferred.   Want to send a PR and fix the README?

PR #1355, opened 2018-06-13
Add new user information

SUPERCILEX:
Fixes <span></span>#1242


samtstern:
@SUPERCILEX thanks for doing this!  Code LGTM.

Just wondering: what criteria did you use to find all the places where you care about the `AuthResult`?  Is there any way we can generalize some of what you did so that we don't forget to pass through information from the `AuthResult` in future additions / refactors?

SUPERCILEX:
@samtstern Yeah, I did a find on all calls to `Resource#setResult` and then looked for ones in a Firebase auth callback. I agree that we should generalize this so I added a base class with a helper method, but I don't see a way of enforcing that it's called instead of `setResult` directly. 🤷‍♂️

samtstern:
@SUPERCILEX this looks better to me, thanks for the refactor!

Issue #1180, opened 2018-03-05
Facebook provider overwritten by Google provider


felipe-gouveia:
### Step 1: Are you in the right place?
   Yes. :)
  
### Step 2: Describe your environment

  * Android device: Google Pixel
  * Android OS version: 8.1.0
  * Google Play Services version: 11.8.0
  * Firebase/Play Services SDK version: 11.8.0
  * FirebaseUI version: 3.2.2

  Would like to point out that this issue is not new. Has been noticed for a couple months now.
  
### Step 3: Describe the problem:

Facebook provider is overwritten by Google provider when Google login is used after Facebook login. This issue happens even if the Facebook login is already email verified.

#### Steps to reproduce:

  1. Login using Facebook.
  2. Validate email of Facebook login (although this step is not necessary)
  3. At this point, check provider in Firebase Console Authentication > Users. It shows Facebook provider.
  4. Logout user.
  5. Login using Google.
  6. Refresh Firebase Console Authentication > Users page (here you should refresh the whole page using F5. The change will not show up when using the reload button in console). Facebook provider is now gone, only showing the Google provider.
  7. Logout user.
  8. Try login with Facebook again. It will prompt that there is already a google login, **as if the Facebook 
  login never existed.**
  9. Login using Facebook.
  10. Now both providers show up. No issue when using Facebook login after Google login.

  I would also like to point out that this issue does not happen when using email provider. If the user logs with email and then with Google it will work as intended, mantaining both providers.
  
#### Observed Results:

  * Google provider overwrites Facebook provider if user had previous logged in with Facebook provider.
  
#### Expected Results:

  * When logging with Google provider after the Facebook provider, both providers should be kept (like 
  how it works with email provider).

samtstern:
@felipe-gouveia thank you for the very detailed reproduction steps!  I need to talk to some other people to decide if this is:

  * A bug in FirebaseUI
  * A bug in Firebase Auth
  * "Working as intended" since we have some special provisions for Google accounts

bojeil-google:
Hey @felipe-gouveia, this is currently work as intended.

To learn more why the Facebook provider is overwritten, check the following posts:
https://groups.google.com/forum/#!searchin/firebase-talk/liu/firebase-talk/ms_NVQem_Cw/8g7BFk1IAAAJ

Please check this post on Facebook not being considered a `verified` email provider:
https://github.com/firebase/firebase-js-sdk/issues/340

[...]

PR #1362, opened 2018-06-19
Rank providers by usability when picking top provider

SUPERCILEX:
As a side effect, this fixes the bug where linking is broken if the developer removes support for a provider.


samtstern:
@SUPERCILEX do you think developers expect what they passed in to to be a priority?  In my opinion it's just a UI concern, devs may be choosing this order based on color combinations / other aesthetic concerns.

Also is it safe (with linking and stuff) to change this ordering now?

SUPERCILEX:
@samtstern Huh, I hadn't thought of it that way. Basically, my gripe is that if I'm already linked to Google and then add Facebook, we're always going to try and use Facebook—even if I sign in later with Google again. And we all know that no sign-in method can compete with Google's simplicity. That's why we stopped showing the others before the picker screen.

I'd be happy to just force Google to be on top, but I feel like we had decided against that for some reason. What do you think?

samtstern:
@SUPERCILEX I do see what you're saying but I think we want a way to solve this that's not tied to the order things are presented in the UI ... let's punt to 4.2.0 and think about it more.

SUPERCILEX:
SGTM!

SUPERCILEX:
@samtstern Alrighty, I redid this PR so it has these changes now:
- Fixes the bug where a provider that's no longer allowed is still returned.
- If the Google provider is present, make it the priority. I think this is fair not only because Firebase is a Google product, but also because Google sign-in is objectively a way better experience for the user than any other provider.

WDYT?

ArtbyRere:
Great

auth/src/main/java/com/firebase/ui/auth/util/data/ProviderUtils.java
```diff
@@ -154,41 +155,60 @@ public static String accountTypeToProviderId(@NonNull String accountType) {
...
+
+                        List<String> lastSignedInProviders = new ArrayList<>(methods.size());
+                        for (String method : methods) {
+                            String id = signInMethodToProviderId(method);
+                            if (allowedProviders.contains(id)) {
+                                lastSignedInProviders.add(0, id);
+                            }
+                        }
+
+                        if (lastSignedInProviders.contains(GoogleAuthProvider.PROVIDER_ID)) {
```
samtstern: Add a comment here explaining what you're doing (making sure Google is always first) and why?

auth/src/main/java/com/firebase/ui/auth/viewmodel/idp/SocialProviderResponseHandler.java
```diff
@@ -80,20 +79,24 @@ public void onFailure(@NonNull final Exception e) {
...
+                                        public void onSuccess(List<String> providers) {
+                                            if (providers.contains(response.getProviderType())) {
                                                 // Case 1
                                                 handleMergeFailure(credential);
                                             } else {
-                                                // Case 2 & 3 - we need to link
-                                                startWelcomeBackFlowForLinking(
-                                                        ProviderUtils.getTopProvider(providers),
-                                                        response);
+                                                if (providers.isEmpty()) {
```
samtstern: nit: instead of `else { if {} else {} }` how about `else if { } else { }`?

samtstern:
(also apologies for being so slow to get to this!  I like the rewrite though)

SUPERCILEX:
@samtstern The reasoning behind this change is that the Google provider is simply a better experience for the user (less clicks). Devs complain about [other providers being removed in favor of Google](https://github.com/firebase/FirebaseUI-Android/issues/1180) (which isn't great), but this is only a user-facing change. All this does is make it so users get the 1-click experience when signing back in (and doesn't change the sign-up experience). Is that reasonable?

samtstern:
@SUPERCILEX ok I am convinced, your judgement on these things is almost always good.  LGTM.

Issue #1338, opened 2018-06-03
Checkbox to accept Terms


baudev:
Currently on the auth UI, there is only the phrase "by continuing you accept..." the Terms of Use and Privacy Policy. Although this sentence has legal value in most countries of the world, it is not the case everywhere.

Take the case of French jurisdiction. It considers the general conditions of use as a contract between the publisher and the Internet user. However, this jurisdiction considers a contract valid if only the two parties exchange consents. 

> Le contrat est consensuel lorsqu'il se forme par le seul échange des consentements quel qu'en soit le mode d'expression. [Code civil - Article 1109](https://www.legifrance.gouv.fr/affichCodeArticle.do?idArticle=LEGIARTI000006436120&cidTexte=LEGITEXT000006070721)

This implies that the user must accept the general terms of use explicitly (to be considered as a consent) and not implicitly as does the phrase "by continuing you accept".
It is therefore necessary to set up a checkbox, **not pre-checked**, with the sentence "by checking this box, I acknowledge that I have read...".

As it depends of the country jurisdiction, it could be interesting having one additional parameter as : 
``` Java
AuthUI.getInstance()
        .createSignInIntentBuilder()
        .setAvailableProviders(providers)
        .showCheckboxTerms(boolean)
        .build(),
    RC_SIGN_IN);
```

iamyaoxi:
I think this is pretty important feature. As in some parts of the world, companies are obligated to have user's agreement first.

Having a simple checkbox to agree or no will be very useful for them.

I'd like to try work on it. :)

samtstern:
@iamyaoxi if you're serious about working on it we'll happily take the contribution!  The only thing we need to do first is agree on the API for turning this on/off and then the rest is implementation.

iamyaoxi:
@samtstern Ya sure! My idea is similar like that of @baudev 

- Developers can choose if they have ToS and/or PP
- Developers can choose if they want to have checkbox or not.

If they have ToS / PP, they go with current implementation. "By continuing, you are indicating...".
But if they have checkbox enabled, a checkbox will appear and the text might change into: "I have read and accept..."

I'm still new in contributing to big open source project like this, but I do want to try. (FYI, I already read and do CONTRIBUTING.md) Would you mind to help me out? Like when you said is to agree with API. Do I make pull request for the API that I suggested and then we go work on the implementation?

samtstern:
Happy to help you get started, let's pick this up on Monday (I'm in Pacific
[...]

PR #1370, opened 2018-06-23
Propose API for showing checkbox to agree for terms #1338

iamyaoxi:
See #1338


samtstern:
@iamyaoxi thanks for this proposal!  This is a minor point, but maybe we should name the method something like `setRequireExplicitAcceptance()` or something so that it's not about the form of the UI (could be a checkbox, a dialog, etc) but rather about the user action.

Either way, the name of the method is not super important.  I agree with your general form for this: a new boolean method on the intent builder.

You can either:
  a) Fix the conflicts in this PR and begin implementing here.
  b) Close this PR and start a new one for implementation.

iamyaoxi:
I'll just create new one. :)

PR #1375, opened 2018-06-26
Adds continue as guest functionality

lsirac:
image

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/AuthUI.java
+++ b/auth/src/main/java/com/firebase/ui/auth/AuthUI.java
@@ -101,7 +101,8 @@ public final class AuthUI {
                        TwitterAuthProvider.PROVIDER_ID,
                        GithubAuthProvider.PROVIDER_ID,
                        EmailAuthProvider.PROVIDER_ID,
-                       PhoneAuthProvider.PROVIDER_ID
+                       PhoneAuthProvider.PROVIDER_ID,
+                       AnonymousAuthProvider.PROVIDER_ID
                })
     @Retention(RetentionPolicy.SOURCE)
     public @interface SupportedProvider {}
@@ -121,7 +122,8 @@ public final class AuthUI {
                     TwitterAuthProvider.PROVIDER_ID,
                     GithubAuthProvider.PROVIDER_ID,
                     EmailAuthProvider.PROVIDER_ID,
-                    PhoneAuthProvider.PROVIDER_ID
+                    PhoneAuthProvider.PROVIDER_ID,
+                    AnonymousAuthProvider.PROVIDER_ID
             )));
 
     /**
@@ -990,6 +992,16 @@ public final class AuthUI {
                 return this;
             }
         }
+
+        /**
+         * {@link IdpConfig} builder for the Anonymous provider.
+         */
+
+        public static final class AnonymousBuilder extends Builder {
+            public AnonymousBuilder() {
+                super(AnonymousAuthProvider.PROVIDER_ID);
+            }
+        }
     }
 
     /**
@@ -1066,7 +1078,8 @@ public final class AuthUI {
 
         /**
          * Specified the set of supported authentication providers. At least one provider must
-         * be specified. There may only be one instance of each provider.
+         * be specified. There may only be one instance of each provider. Anonymous provider cannot
+         * be the only provider specified.
          * <p>
          * <p>If no providers are explicitly specified by calling this method, then the email
          * provider is the default supported provider.
@@ -1074,9 +1087,19 @@ public final class AuthUI {
          * @param idpConfigs a list of {@link IdpConfig}s, where each {@link IdpConfig} contains the
          *                   configuration parameters for the IDP.
          * @see IdpConfig
+         *
+         * @throws IllegalStateException if anonymous provider is the only specified provider.
          */
         @NonNull
         public T setAvailableProviders(@NonNull List<IdpConfig> idpConfigs) {
+            Preconditions.checkNotNull(idpConfigs, "idpConfigs cannot be null");
+            if (idpConfigs.size() == 1 &&
+                    idpConfigs.get(0).getProviderId().equals(AnonymousAuthProvider.PROVIDER_ID)) {
+                throw new IllegalStateException("Sign in as guest cannot be the only sign in " +
+                        "method. In this case, sign the user in anonymously your self; " +
+                        "no UI is needed.");
+            }
+
             mProviders.clear();
 
             for (IdpConfig config : idpConfigs) {
@@ -1169,5 +1192,9 @@ public final class AuthUI {
                     mEnableHints,
                     mEnableAnonymousUpgrade);
         }
+    }
+
+    public class AnonymousAuthProvider {
+        public static final String PROVIDER_ID = "anonymous";
     }
 }
--- a/auth/src/main/java/com/firebase/ui/auth/data/remote/AnonymousSignInHandler.java
+++ b/auth/src/main/java/com/firebase/ui/auth/data/remote/AnonymousSignInHandler.java
@@ -1,0 +1,81 @@
+package com.firebase.ui.auth.data.remote;
+
+import android.app.Application;
+import android.content.Intent;
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+import android.support.annotation.RestrictTo;
+import android.support.annotation.VisibleForTesting;
+
+import com.firebase.ui.auth.AuthUI;
+import com.firebase.ui.auth.IdpResponse;
+import com.firebase.ui.auth.data.model.FlowParameters;
+import com.firebase.ui.auth.data.model.Resource;
+import com.firebase.ui.auth.data.model.User;
+import com.firebase.ui.auth.ui.HelperActivityBase;
+import com.firebase.ui.auth.viewmodel.ProviderSignInBase;
+import com.google.android.gms.tasks.OnFailureListener;
+import com.google.android.gms.tasks.OnSuccessListener;
+import com.google.firebase.FirebaseApp;
+import com.google.firebase.auth.AuthResult;
+import com.google.firebase.auth.FirebaseAuth;
+
+
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+public class AnonymousSignInHandler extends ProviderSignInBase<FlowParameters> {
+
+    @VisibleForTesting
+    public FirebaseAuth mAuth;
+
+    public AnonymousSignInHandler(Application application) {
+        super(application);
+    }
+
+    @Override
+    protected void onCreate() {
+        mAuth = getAuth();
+    }
+
+    @Override
+    public void startSignIn(@NonNull HelperActivityBase activity) {
+        showTopProgressBar();
+
+        // Calling signInAnonymously() will always return the same anonymous user if already
+        // available. This is enforced by the client SDK.
+        mAuth.signInAnonymously()
+                .addOnSuccessListener(new OnSuccessListener<AuthResult>() {
+                    @Override
+                    public void onSuccess(AuthResult result) {
+                        setResult(Resource.<IdpResponse>forSuccess(getResponse(
+                                result.getAdditionalUserInfo().isNewUser())));
+                    }
+                })
+                .addOnFailureListener(new OnFailureListener() {
+                    @Override
+                    public void onFailure(@NonNull Exception e) {
+                        setResult(Resource.<IdpResponse>forFailure(e));
+                    }
+                });
+
+    }
+
+    public IdpResponse getResponse(boolean isNewUser) {
+        return new IdpResponse.Builder(
+                new User.Builder(AuthUI.AnonymousAuthProvider.PROVIDER_ID, null)
+                        .build())
+                .setNewUser(isNewUser)
+                .build();
+    }
+
+    private void showTopProgressBar() {
+        setResult(Resource.<IdpResponse>forLoading());
+    }
+
+    private FirebaseAuth getAuth() {
+        FirebaseApp app = FirebaseApp.getInstance(getArguments().appName);
+        return FirebaseAuth.getInstance(app);
+    }
+
+    @Override
+    public void onActivityResult(int requestCode, int resultCode, @Nullable Intent data) {}
+}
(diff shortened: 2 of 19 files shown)


auth/src/main/java/com/firebase/ui/auth/data/remote/AnonymousSignInHandler.java
```diff
@@ -0,0 +1,81 @@
...
+
+    public IdpResponse getResponse(boolean isNewUser) {
+        return new IdpResponse.Builder(
+                new User.Builder(AuthUI.AnonymousAuthProvider.PROVIDER_ID, null)
+                        .build())
+                .setNewUser(isNewUser)
+                .build();
+    }
+
+    private void showTopProgressBar() {
```
samtstern: I think we should either rename this to `setLoading()` or just inline it at all call sites.  The handlers should not have UI logic, it's up to the observer to decide how to display (or ignore) progress events.  Top progress bar is our current favorite pattern but it was a dialog before that and could easily change again.

auth/src/main/java/com/firebase/ui/auth/data/remote/AnonymousSignInHandler.java
```diff
@@ -0,0 +1,81 @@
...
+                .addOnFailureListener(new OnFailureListener() {
+                    @Override
+                    public void onFailure(@NonNull Exception e) {
+                        setResult(Resource.<IdpResponse>forFailure(e));
+                    }
+                });
+
+    }
+
+    public IdpResponse getResponse(boolean isNewUser) {
```
samtstern: Is this used in more than one place?  Could we just inline it?  Also does it need to be `public`?

auth/src/main/java/com/firebase/ui/auth/data/remote/AnonymousSignInHandler.java
```diff
@@ -0,0 +1,81 @@
...
+                        .build())
+                .setNewUser(isNewUser)
+                .build();
+    }
+
+    private void showTopProgressBar() {
+        setResult(Resource.<IdpResponse>forLoading());
+    }
+
+    private FirebaseAuth getAuth() {
```
samtstern: I think you can get this from `AuthUI` right?  The other handlers don't duplicate this logic afaik.

auth/src/main/java/com/firebase/ui/auth/data/remote/AnonymousSignInHandler.java
```diff
@@ -0,0 +1,81 @@
...
+        FirebaseApp app = FirebaseApp.getInstance(getArguments().appName);
+        return FirebaseAuth.getInstance(app);
+    }
+
+    @Override
+    public void onActivityResult(int requestCode, int resultCode, @Nullable Intent data) {}
```
samtstern: nit: move this up with the other `@Override` methods.

build.gradle.kts
```diff
@@ -101,7 +101,8 @@ fun Project.configureAndroid() {
                     "IconExpectedSize",
                     "InvalidPackage", // Firestore uses GRPC which makes lint mad
                     "NewerVersionAvailable", "GradleDependency", // For reproducible builds
-                    "SelectableText", "SyntheticAccessor" // We almost never care about this
+                    "SelectableText", "SyntheticAccessor", // We almost never care about this
+                    "InvalidVectorPath" // false positive for fui_ic_anonymous_white_24dp
```
samtstern: I think we may just want to add a single exception for this to the lint baseline rather than ignoring all `InvalidVectorPath` errors since they could be serious in the future.

auth/src/main/java/com/firebase/ui/auth/data/remote/AnonymousSignInHandler.java
lsirac: Definitely should be private.

auth/src/main/java/com/firebase/ui/auth/data/remote/AnonymousSignInHandler.java
lsirac: Done.

auth/src/main/java/com/firebase/ui/auth/data/remote/AnonymousSignInHandler.java
lsirac: Other handlers extend from AuthViewModelBase. Added a todo for this.

auth/src/main/java/com/firebase/ui/auth/data/remote/AnonymousSignInHandler.java
lsirac: Done.

build.gradle.kts
lsirac: Done.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -990,6 +992,16 @@ public GitHubBuilder setPermissions(@NonNull List<String> permissions) {
                 return this;
             }
         }
+
+        /**
+         * {@link IdpConfig} builder for the Anonymous provider.
+         */
+
```
SUPERCILEX: Nit: remove line

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -1170,4 +1193,8 @@ protected FlowParameters getFlowParams() {
                     mEnableAnonymousUpgrade);
         }
     }
+
+    public class AnonymousAuthProvider {
+        public static final String PROVIDER_ID = "anonymous";
+    }
```
SUPERCILEX: Not quite sure I see why we need this... Can't we just use `AnonymousAuthProvider.PROVIDER_ID`?

auth/src/main/res/layout/fui_provider_button_anonymous.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<com.firebase.ui.auth.util.ui.SupportVectorDrawablesButton
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:id="@+id/anonymous_button"
+    style="@style/FirebaseUI.Button.AccountChooser.AnonymousButton"
+    android:text="@string/fui_sign_in_anonymously">
+</com.firebase.ui.auth.util.ui.SupportVectorDrawablesButton>
```
SUPERCILEX: Nit: unnecessary tag (could just be `/>`)

library/quality/lint-baseline.xml
```diff
@@ -8,4 +8,9 @@
         <location file="src/main/AndroidManifest.xml" />
     </issue>
 
+    <issue
+        id="InvalidVectorPath"
+        message="Use 0.94 instead of .94 to avoid crashes on some devices">
+        <location file="src/main/res/drawable/fui_ic_anonymous_white_24dp.xml"></location>
+    </issue>
```
SUPERCILEX: I don't think we want this to be in our baseline since we're purposefully ignoring it. The baseline is meant to be a baseline (for lack of a better word 😆) that you're trying to slowly shrink over time.

auth/src/main/java/com/firebase/ui/auth/data/remote/AnonymousSignInHandler.java
```diff
@@ -0,0 +1,78 @@
...
+
+    private IdpResponse initResponse(boolean isNewUser) {
+        return new IdpResponse.Builder(
+                new User.Builder(AuthUI.AnonymousAuthProvider.PROVIDER_ID, null)
+                        .build())
+                .setNewUser(isNewUser)
+                .build();
+    }
+
+    // TODO: Change ProviderSignInBase to extend from AuthViewModelBase
```
SUPERCILEX: `ProviderSignInBase` classes are meant to only retrieve remote provider data, hence making it difficult to get an auth instance as you've noticed. It was also supposed to be UI focused which is why there's a bunch of activity related stuff.

I feel like we should make this class be a bare bones middle man that sends an `IdpResponse` to an `AnonymousProviderResponseHandler` with the auth stuff. @samtstern What do you think?

samtstern:
@SUPERCILEX agree, but enough devs have asked for this (across all platforms) that we just decided to give them the option.  Apparently some people like the UX of the "continue as guest" button.  I am not one of those people but hey, free choice!

library/quality/lint-baseline.xml
samtstern: Do you know how to purposefully ignore a warning like this?  Can we add something in the vector XML itself?

SUPERCILEX:
@samtstern Seems weird to me, but alrighty then! 👍

library/quality/lint-baseline.xml
SUPERCILEX: Yup, should be the same `tools:ignore="InvalidVectorPath"` we all know and love. 😆

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
lsirac: Done.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
lsirac: Where is this AnonymousAuthProvider that you speak of? :smile:

auth/src/main/res/layout/fui_provider_button_anonymous.xml
lsirac: Done.

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
SUPERCILEX: 😆❤️

library/quality/lint-baseline.xml
lsirac: Done.

auth/src/main/java/com/firebase/ui/auth/data/remote/AnonymousSignInHandler.java
samtstern: @SUPERCILEX in this version which class actually would call `signInAnonymously`?

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
SUPERCILEX: Oh lol, you named it so well I thought you were joking. 😊 Anyway... since Firebase _doesn't_ have a built-in `AnonymousAuthProvider`, shouldn't the [`FirebaseAuthProvider`](https://firebase.google.com/docs/reference/android/com/google/firebase/auth/FirebaseAuthProvider) essentially be the same thing? @samtstern Am I mixing it up or is `FirebaseAuthProvider` intended to represent anon auth?

samtstern: No idea, @lsirac might know better than I do!

auth/src/main/java/com/firebase/ui/auth/data/remote/AnonymousSignInHandler.java
SUPERCILEX: `signInAnonymously` would be called in the `AnonymousProviderResponseHandler`. And `AnonymousSignInHandler` would be something as simple as:
```java
// ...

@Override
public void startSignIn(@NonNull HelperActivityBase activity) {
    setResult(Resource.<IdpResponse>forSuccess(new IdpResponse.Builder(
                new User.Builder(AuthUI.AnonymousAuthProvider.PROVIDER_ID, null).build())
                .build()));
}
```

Not sure if that's actually better though since having `AnonymousSignInHandler` becomes kinda pointless. I'm just not a huge fan of doing auth stuff in a non-auth view model. 🤷‍♂️

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
lsirac: I don't think FirebaseAuthProvider is intended to represent anon auth. Every signed in user has a "firebase" provider.

SUPERCILEX: Gotya, then I think we should stick to a field in AuthUI directly. We used to have stuff like this:
```
-    /**
-     * Provider identifier for email and password credentials, for use with {@link
-     * SignInIntentBuilder#setAvailableProviders(List)}.
-     */
-    public static final String EMAIL_PROVIDER = EmailAuthProvider.PROVIDER_ID;
```

So we could do `ANONYMOUS_PROVIDER`. Also, does it need to be publicly accessible? I think we could mark it with `@RestrictTo`.

lsirac: So we'd do `AuthUI.ANONYMOUS_PROVIDER` instead of `AuthUI.AnonymousAuthProvider.PROVIDER_ID`? and yeah we can mark it with `@RestrictTo`

SUPERCILEX: Yup! 😊

auth/src/main/java/com/firebase/ui/auth/data/remote/AnonymousSignInHandler.java
lsirac: @samtstern thoughts?

samtstern: Maybe we compromise and **only** have the `ResponseHandler` but call it immediately, rather than having the useless `ViewModel`.  I agree we shouldn't do the signing in in the non-auth view model if we can avoid it.

samtstern:
@SUPERCILEX after talking to Leo, we think the right move is to stick with the original organization and have the anon sign-in code be part of the `SignInHandler` rather than a `ResponseHandler`.

I know this breaks your class hierarchy a bit, but I don't think adding a useless class is any better.  If we just pretend that the Anonymous sign in is some IDP called "Anonymous, Incorporated" it's not hard to imagine why the Firebase Auth code belongs there.

Maybe we can find a future way to bring the two ViewModel class hierarchy trees closer together in the future?

samtstern:
@austinhodak could you file a separate bug to track that issue? Thanks!

PR #1402, opened 2018-07-28
Fix Turkish translation for 'fui_name_hint'

gaeshi:
This is a simple fix for Turkish translation of "Name and Surname" in FirebaseUI-Android/auth. "Ad ve Soyadı" should be "Ad ve Soyad".


gaeshi:
I signed it!

samtstern:
@gaeshi Google Translate seems to think they both translate the same.  I am happy to merge this but can you explain the difference?  Is the current translation a typo or it has the wrong meaning?

gaeshi:
@samtstern it's just a typo. In this case you don't need to add "ı" at the end of the word.

samtstern:
@gaeshi thanks!

PR #1405, opened 2018-08-05
Fix GitHub provider NPE when no perms are supplied

SUPERCILEX:


samtstern:
The triumphant return of @SUPERCILEX !

PR #1419, opened 2018-08-27
Update deps

SUPERCILEX:


SUPERCILEX:
@samtstern And I'm back! 😁 Sort of. It looks like I'll be swamped for who knows how long, but I'm planning spending a few days every month on FUI. It was the first project I ever contributed to, so it holds a special place in my heart. ❤️ TL;DR: I won't be as active as I'd like to be, but I'll be around for a long time.

All that was a long way of saying I've got a bunch of PRs to submit today and tomorrow. 🎉

SUPERCILEX:
@samtstern Thanks for the kind words!

Issue #1412, opened 2018-08-14
GitHub IDP brings unwanted dependencies even if not used - Retrofit and Retrofit GSON


curiousily:
The (somewhat) recently introduced GitHub IDP adds dependencies to Retrofit and Retrofit GSON. I am not aware of the number of users of Firebase Auth that will use this exact IDP but bringing those libraries as required seems strange. Our app does not use any of these libraries.

Please, make the GitHub IDP optional or make optional including the Retrofit, Retrofit Gson and GSON (not sure used elsewhere in the library).

Reference: [https://github.com/firebase/FirebaseUI-Android/blob/master/auth/build.gradle.kts](https://github.com/firebase/FirebaseUI-Android/blob/master/auth/build.gradle.kts)

samtstern:
@curiousily thanks for the feedback.  When you ship your app with ProGuard, are you seeing a lot of these classes still hanging around?  I realize that we may be bloating your debug builds but hopefully that's not a blocking issue.

ArtbyRere:
K can you help

SUPERCILEX:
Ok, so only the Twitter dep uses retrofit. If you're not using Twitter, then it is an unnecessary dependency. @samtstern What do you think of making it `compileOnly`? Would that count as a breaking change?

@curiousily BTW, you can always exclude it in the meantime:
```kt
implementation("com.firebaseui:firebase-ui-auth:4.1.0") {
    exclude(module = "retrofit") // In Groovy, I think this would be `exclude module: 'retrofit'
    exclude(module = "converter-gson")
}
```

samtstern:
@SUPERCILEX I think we probably need to change how GitHub works due to the other bug about `CHANGE-ME` showing up in system settings when it's not configured.

I think in theory it's a breaking change but in practice I don't think we should update the major version number for this when I can't see anyone getting "accidentally" broken by it.

SUPERCILEX:
Yeah, I'm thinking we need to pull out a sub-module with the manifest entry and those deps. 👍

curiousily:
Hey guys,

Yes, one can remove the deps using ProGuard or something similar, but that ain't a real solution, as you've probably guessed.

@SUPERCILEX even though we don't use Retrofit (we're hardcore - using OkHttp 😃) I tend to agree with you about it. However, GSON is definitely not the only reasonable JSON lib option 🙂. That said, I will use the proposed workaround for now (exclude the deps).

[...]

Issue #1389, opened 2018-07-08
CHANGE-ME link appears in system settings


javmarina:
### Step 2: Describe your environment

  * Android device: Nokia 7 plus
  * Android OS version: 8.1.0
  * Google Play Services version: 15.0.1 - 16.0.0
  * Firebase/Play Services SDK version: 16.0.0
  * FirebaseUI version: 4.1.0
  
#### Steps to reproduce:

  1. Include FirebaseUI (auth) in a project
  2. Build and install
  3. Go to Settings > Apps > installed testing app > Open by default
  
#### Observed Results:

  * What happened?
In the "Supported links" line, you can see CHANGE-ME as a link. This is because Firebase UI includes this in the manifest. CHANGE-ME  is used for Github sign in, I think.
  
#### Expected Results:

  * What did you expect to happen?
I didn't expect this to appear, as it is a string related to a part of the library that I don't use. Moreover, users can see it too.
  
![screenshot_20180709-015538](https://user-images.githubusercontent.com/2737599/42425300-97d892ce-831b-11e8-8660-6affebc08dd7.png)

SUPERCILEX:
Ouch, that was a miscalculation on my part, sorry! 😊 For a more permanent solution, I think we might want to make people opt-in to the GitHub stuff by enabling the manifest entry, but for now, this should fix it:
```xml
<activity
    android:name=".ui.provider.GitHubLoginActivity"
    tools:node="remove" />
```

javmarina:
This is enough for now, thanks!

noelmansour:
I'm not sure if this affects priority at all, but this bug breaks app links (like those used for passwordless sign up).

Also, I had to put the fully qualified activity name for the workaround to work:

```
<activity
    android:name="com.firebase.ui.auth.ui.provider.GitHubLoginActivity"
    tools:node="remove" />
```

samtstern:
This was fixed and released in version `4.2.0`

javmarina:
Thanks!

PR #1421, opened 2018-08-27
Split GitHub provider out into its own module

SUPERCILEX:
Fixes<span></span> #1389 and #1412


auth-github/src/main/java/com/firebase/ui/auth/github/GitHubSignInHandler.java
```diff
@@ -1,4 +1,4 @@
-package com.firebase.ui.auth.data.remote;
+package com.firebase.ui.auth.github;
```
samtstern: Should this be in `.github.data.remote` just to be consistent?

auth/src/main/res/values/config.xml
```diff
@@ -1,44 +1,13 @@
...
-     To enable Chrome Custom Tabs for Facebook Login on devices without the Facebook app
-     installed, change this value to your Facebook App ID prefixed with 'fb'. This should be
-     'fb' + the value of the facebook_application_id string above
-
-     See:
-     https://developers.facebook.com/docs/facebook-login/android#chrome_custom_tabs
-    -->
     <string name="facebook_login_protocol_scheme" translatable="false">fbYOUR_APP_ID</string>
 
-    <!--
```
samtstern: Any particular reason to kill these comments?

SUPERCILEX:
Regarding 5.0, I don't think so? Everything will still compile as before, it'll "just" be a runtime crash. I think we don't necessarily need to bump to 5.0 because people need to update their deps anyway to get this change (so it's not super hard to add another dep).

> I am being overly sentimental about versioning

🤣

TODO(me):
- [ ] Check release builds
- [ ] Update documentation

auth-github/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,35 @@
...
+            <intent-filter
+                android:autoVerify="true"
+                tools:ignore="UnusedAttribute">
+                <action android:name="android.intent.action.VIEW" />
+
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+
+                <data
+                    android:host="@string/firebase_web_host"
```
samtstern: Ok two questions about this string:

  1. I see we have duplicate `strings.xml` entries for `firebase_web_host` in both `auth/.../strings.xml` and `auth-github/.../strings.xml` ... is there a reason it needs to exist outside of the GitHub module?  I feel like we only use it there but could be forgetting something.
  1. Now that we're sure you want to use GitHub auth when you include the `-github` module, should we explode at runtime if this string is still equal to `CHANGE-ME`?  I think that might be helpful.

auth/build.gradle.kts
```diff
@@ -33,14 +33,9 @@ dependencies {
     api(Config.Libs.PlayServices.auth)
 
     compileOnly(Config.Libs.Provider.facebook)
-    // Needed to override Facebook
-    implementation(Config.Libs.Support.cardView)
-    implementation(Config.Libs.Support.customTabs)
+    implementation(Config.Libs.Support.cardView) // Needed to override Facebook
```
samtstern: Did something change about how Facebook uses custom tabs?

auth-github/src/main/AndroidManifest.xml
SUPERCILEX: Yup, both those birds were killed with the same stone in the original implementation 😁: https://github.com/firebase/FirebaseUI-Android/blob/f83c6219a32456b24c9a78cd597f97943c938669/auth/src/main/java/com/firebase/ui/auth/AuthUI.java#L973-L977

1. We need `firebase_web_host` in the auth module to do that check
1. We do that check 😂

auth/build.gradle.kts
SUPERCILEX: Nope, but we use custom tabs ourselves (for the ToS and PP) so it was a duplicate dependency:
https://github.com/firebase/FirebaseUI-Android/blob/df554e2f8cec7ae594ea5d05e5ce61e39281f4d2/auth/build.gradle.kts#L25

auth-github/src/main/AndroidManifest.xml
samtstern: D'oh

PR #1422, opened 2018-08-27
Optimize build performance

SUPERCILEX:


SUPERCILEX:
Done! BTW, you can read up on the configuration avoidance stuff here: https://docs.gradle.org/current/userguide/task_configuration_avoidance.html

Issue #1351, opened 2018-06-12
Create a publicly usable lint module


samtstern:
See #1340 which implements new lint checks but is not yet exposed because it's in `internal`.

samtstern:
cc @pavlospt

One possible roadblock here: the lint checks are written in Kotlin and we have not yet shipped any Kotlin to developers.  I am not sure how this will impact their builds...

pavlospt:
Isn't just a jar shipped as a dependency for Lint modules? Would this also bring the whole Kotlin stdlib in?

samtstern:
This was fixed and released in version `4.2.0`

PR #1427, opened 2018-08-29
Publicly usable lint checks

samtstern:
See #1351

I wonder if we should create an empty `firebaseui-lint` android library that only exists to have the `lintChecks` call.  I say that because I am not sure what happens if we had multiple (for example auth and firestore) depending on the `lint` module as `lintChecks`


samtstern:
cc @pavlospt wdyt?

samtstern:
Remaining error:
```
Errors found:
  
  /usr/local/google/home/samstern/Projects/firebase/firebaseui-android/auth/src/main/res/values/com_crashlytics_export_strings.xml:9: Error: Resource named 'com.crashlytics.android.build_id' does not start with the project's resource prefix 'fui_'; rename to 'fui_com.crashlytics.android.build_id' ? [ResourceName]
      <string tools:ignore="UnusedResources,TypographyDashes" name="com.crashlytics.android.build_id" translatable="false">77252c51-646a-4395-affc-28655847aec9</string>
```

I know we hit this before but I forget why.

pavlospt:
If there was a firebaseui-lint library it could be served as a dependency as well for standalone usage! Otherwise the separation looks good to me!

proguard-tests/build.gradle.kts
```diff
@@ -4,6 +4,14 @@ android {
     }
 
     buildTypes {
+        named("debug").configure {
```
SUPERCILEX: Why'd we add this back? It adds like 5+ mins to a full debug build. Shouldn't we only run it in CI?

samtstern: Oh damn I forgot why we removed it.  Android Studio gets pretty angry when not all modules have matching configurations and this made it happy.

Let me see if I can appease Android Studio without compromising time.

SUPERCILEX: Ohhh, yeah, I remember seeing those warnings. Let's use a variant filter then: https://developer.android.com/studio/build/build-variants#filter-variants

samtstern: Yep it looks like that's already working with `setIgnore`.  I pushed a commit to simplify the `debug` config to the minimum that makes AS happy.

I did a `time` of some clean builds and the extra config doesn't cost us anything.  Will confirm via Travis logs.

SUPERCILEX: Awesome, SGTM!

samtstern: Travis confirmed: no slowdown.

PR #1428, opened 2018-08-30
Add proguard rules suggested by R8 team

samtstern:
See https://buganizer.corp.google.com/issues/113513217 for context

@SUPERCILEX do you think we should add these rules as the team suggested?


samtstern:
(note: going to wait until that issue is resolved before doing anything, just opening this PR as a place to do the work)

SUPERCILEX:
I don't think we should do that since it's adding a hack for a bug in pre-release software: https://issuetracker.google.com/issues/113513217#comment7

samtstern:
Closing for now.

Issue #1349, opened 2018-06-11
Android studio does not recognize the project as android project


abdurahmanadilovic:
I cloned the repository and I tried to open it with Android studio and got nowhere, what is a proper way to open this project with android studio 3.1.3?

samtstern:
@abdurahmanadilovic I am also using Studio `3.1.x` but (obviously) I can't see the same problem.  Please provide more information, otherwise I think this is an issue with your machine/configuration.

abdurahmanadilovic:
I tried to clone the project then open the cloned folder in the android studio and also I tried to clone the repo through the Android studio, using the create new project from source control option and I am unable to build the project with both methods.

In the project explorer if I select android nothing shows up if I select Project view [this is only what appears in the explorer](https://imgur.com/a/jQmc8wI)

I guess the build.gradle.kts might cause this issues since android studio is looking for build.gradle without the .kts extension?

SUPERCILEX:
Oh, that's because you didn't import it from Gradle properly. `Ctrl + E` to get to the Gradle sidebar. Then click the + button to add the root module and start a sync. If that doesn't work, close the project and delete the `.idea` folder. Then choose import from existing sources and go through the Gradle flow.

abdurahmanadilovic:
So [this happens if](https://imgur.com/a/bJe0n6K) I try to import a gradle project. It detects multiple android / java modules and none of them are runnable. If I understood you correctly I should select import project then select gradle, instead of "Open an existing Android Studio project"?

SUPERCILEX:
In the second step, you should have clicked the `Import project from external model` option. Be sure to delete the `.idea` folder again before importing.

abdurahmanadilovic:
I managed to build the project by selecting "import project from external model", now everything works, thanks for helping me out.

sushihangover:
FYI: I am using AS 3.2C18 and since there is an existing repo-based file within .idea:

* .idea/codeStyleSettings.xml

Importing via Gradle fails to import the project properly. 

Deleting the `.idea` directory and re-importing from external model works fine.

sugoireed:
What worked for me was 

On Android Studio 

Open the settings.gradle file

include your project

include ':myproject'

ashomokdev:
To open the project you need to have Gradle installed on your PC. Then 
Download repo FirebaseUI-Android from github. 
[...]

PR #1438, opened 2018-09-06
Add small troubleshooting for importing sample app

leonjoosse:
I would like to clarify this issue in the README: #1349

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,6 +193,14 @@ and Gradle while trying to run the sample app, try disabling the Instant
 Run feature of Android Studio. Alternatively, update Android Studio and
 Gradle to their latest versions.
 
+A note on importing the project using Android Studio: Using 'Project from 
+Version Control' will not automatically link the project with Gradle 
+(issue [#1349](https://github.com/firebase/FirebaseUI-Android/issues/1349)). 
+When doing so and opening any `build.gradle.kts` file, an error shows up: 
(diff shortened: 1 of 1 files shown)


leonjoosse:
I signed it!

samtstern:
@leonjoosse thanks for this clarification!  LGTM, @SUPERCILEX wdyt?

SUPERCILEX:
@samtstern This is kinda Studio's fault, but I don't see why not. 😃

trigal2012:
I'm new to Android and Firebase and I'm trying to use this project. I clone the entire project locally, then opened AS, selected import, then I chose "import project from external model", clicked on Next. The form is then looking for  the "Gradle Home". "Use local gradle distribution" is selected. What value do I enter here? I've tried pointing it to the "gradle" folder located at "FirebaseUI-android/gradle" but that did not work. I also tried pointing to  "FirebaseUI-android/app". I also removed the .idea folder. any help is appreciated.

SUPERCILEX:
@trigal2012 Yeah, that dialog is annoying. Just select the first option to use the Gradle wrapper and click next. I think AS 3.2 will be able to handle the project better.

trigal2012:
@SUPERCILEX thanks for the quick response. looks like I may have another issue because the option you mentioned is grayed out. I'll dig around. Thanks!

Issue #1404, opened 2018-08-01
Cancelling Google login gives displays an error


richjhart:
### Step 2: Describe your environment

  * Android device: OnePlus 5
  * Android OS version: 8.1
  * Google Play Services version: 12.8.74 (040400-204998136)
  * Firebase/Play Services SDK version: 
  * FirebaseUI version: 
    implementation 'com.google.firebase:firebase-core:16.0.1'
    implementation 'com.google.firebase:firebase-auth:16.0.2'
    implementation 'com.google.android.gms:play-services-auth:15.0.1'
    implementation 'com.firebaseui:firebase-ui-auth:4.1.0'
  
### Step 3: Describe the problem:

When logging in using "Sign in with Google", if the user cancels it, an error pops up - when no error has occurred

#### Steps to reproduce:

  1. Begin the "Sign in with Google" flow
  2. Wait for the account select dialog to show
  3. Press back to cancel it
  
#### Observed Results:

  * The dialog is dismissed (correct)
  * A Toast appears saying: "Code: 13, message: 13,"
  
#### Expected Results:

  * After the dialog is dismissed, either no message, or something more along the lines of "Login Cancelled"

samtstern:
@richjhart this is a known issue in Google Sign In, you can follow along here:
https://github.com/googlesamples/google-services/issues/374

PR #1443, opened 2018-09-12
Avoid error when canceling google signin

diedu89:
Currently if user cancel signin with google an error is shown as a toast instead of the message of user cancellation #1404, to avoid this play-services-auth is upgraded
Also the handle of the error code 12501 is added in order to set an UserCancellationException instead of FirebaseUiException


diedu89:
I signed it

samtstern:
@diedu89 thank you!

samtstern:
Merged into the `4.2.0` development branch

Issue #1416, opened 2018-08-24
java.lang.ClassNotFoundException: com.firebase.ui.auth.data.model.FlowParameters


dimipaun:
### Step 1: 

When trying to get the Auth UI going:
```
        Intent intent = AuthUI.getInstance()
                    .createSignInIntentBuilder()
                    .setTheme(R.style.SplashTheme)
                    .setAvailableProviders(idpConfigs)
                    .setTosAndPrivacyPolicyUrls(getString(R.string.terms_url), getString(R.string.privacy_url))
                    .setIsSmartLockEnabled(!BuildConfig.DEBUG)
                    .build();
        startActivityForResult(intent, RC_FIREBASE_UI_SIGNIN);
```

I see the following behaviour:
  * if `idpConfigs` contains exactly one item, it correctly proceeds to the one login method
  * if `idpConfigs` contains more than one item, the `AuthMethodPickerActivity` is displayed WITHOUT any of the login buttons (but it correctly shows the ToS/PP at the bottom)

Looking through the logs, I see this error:
```
08-23 21:01:07.343 1358-2025/? E/Parcel: Class not found when unmarshalling: com.firebase.ui.auth.data.model.FlowParameters
    java.lang.ClassNotFoundException: com.firebase.ui.auth.data.model.FlowParameters
        at java.lang.Class.classForName(Native Method)
        at java.lang.Class.forName(Class.java:453)
        at android.os.Parcel.readParcelableCreator(Parcel.java:2835)
        at android.os.Parcel.readParcelable(Parcel.java:2789)
        at android.os.Parcel.readValue(Parcel.java:2692)
        at android.os.Parcel.readArrayMapInternal(Parcel.java:3059)
        at android.os.BaseBundle.unparcel(BaseBundle.java:257)
        at android.os.BaseBundle.getString(BaseBundle.java:1086)
        at android.content.Intent.getStringExtra(Intent.java:7706)
        at com.android.server.am.ActivityStarter.startActivity(ActivityStarter.java:458)
        at com.android.server.am.ActivityStarter.startActivityLocked(ActivityStarter.java:409)
        at com.android.server.am.ActivityStarter.startActivityMayWait(ActivityStarter.java:1270)
        at com.android.server.am.ActivityManagerService.startActivityAsUser(ActivityManagerService.java:6053)
        at com.android.server.am.ActivityManagerService.startActivity(ActivityManagerService.java:5906)
[...]

PR #1452, opened 2018-09-20
Attempt another fix at Samsung parcelable unmarshalling errors

SUPERCILEX:
See #1416


samtstern:
Looks like we have a competing PR in #1453

Issue #1434, opened 2018-09-04
FirebaseUI not adjusted according to the rotation and numbers of providers


JitenderSha:
### Describe your environment

  * Android device:  Nexus 5
  * Android OS version: 6
  * Google Play Services version: 12.8.74
  * Firebase/Play Services SDK version: firebase-auth:16.0.3, play-services-location:15.0.1
  * FirebaseUI version: 4.1.0
  
### Step 3: Describe the problem:

1. In Portrait mode 
      As in screenshot 1, the blank space is leaved if used two providers.

2. In Landscape mode
     As in screenshot 2, the blank space is leaved at the bottom of the logo and logo is not in vertical center.

#### Steps to reproduce:

  1. Use FirebaseUI using
startActivityForResult(
                AuthUI.getInstance()
                        .createSignInIntentBuilder()
                        .setAvailableProviders(providers)
                        .setLogo(R.mipmap.tb_logo)      // Set logo drawable
                        .setTheme(R.style.AppTheme)      // Set theme
                        .build(),
                RC_SIGN_IN);
  2. Run the app
  3. See in both orientations
  
#### Observed Results:

  * All is in screenshots
  
#### Expected Results:

  * What did you expect to happen?
In Portrait mode, if two providers are used the space should evenly distributed between the logo and buttons.

In Landscape mode, the logo should be placed in vertical center of the screen.
  
#### Relevant Code:
Just copy paste the following :

  // Choose authentication providers
       
       List<AuthUI.IdpConfig> providers = Arrays.asList(
                new AuthUI.IdpConfig.PhoneBuilder().build(),
                new AuthUI.IdpConfig.FacebookBuilder().build());

        startActivityForResult(
                AuthUI.getInstance()
                        .createSignInIntentBuilder()
                        .setAvailableProviders(providers)
                        .setLogo(R.mipmap.tb_logo)      // Set logo drawable
                        .setTheme(R.style.AppTheme)      // Set theme
                        .build(),
                RC_SIGN_IN);

![screenshot_1](https://user-images.githubusercontent.com/38163139/45026564-4ff90600-b05c-11e8-8fd4-c54e1d072f6a.png)
![screenshot_2](https://user-images.githubusercontent.com/38163139/45026577-56877d80-b05c-11e8-8167-9de1ed06a4da.png)

samtstern:
@JitenderSha thanks for reporting.  What is the bug in portrait mode?  That looks correct to me.

[...]

PR #1458, opened 2018-09-24
Fix incorrect horizontal logo padding

SUPERCILEX:
Fixes <span><span/>#1434

Old | New
--- | ---
screenshot_20180924-162454 | screenshot_20180924-162424

### DIFF
--- a/auth/src/main/res/layout-land/fui_auth_method_picker_layout.xml
+++ b/auth/src/main/res/layout-land/fui_auth_method_picker_layout.xml
@@ -22,6 +22,7 @@
         app:layout_constraintEnd_toStartOf="@+id/container"
         app:layout_constraintTop_toTopOf="parent"
         app:layout_constraintBottom_toTopOf="@+id/main_tos_and_pp"
+        app:layout_constraintVertical_weight="2"
         tools:ignore="ContentDescription" />  <!-- TODO remove once the bug is fixed: https://issuetracker.google.com/issues/38281866-->
 
     <TextView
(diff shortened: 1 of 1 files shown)


samtstern:
@SUPERCILEX I was just working on this, but you beat me to it!

auth/src/main/res/layout-land/fui_auth_method_picker_layout.xml
```diff
@@ -22,6 +22,7 @@
         app:layout_constraintEnd_toStartOf="@+id/container"
         app:layout_constraintTop_toTopOf="parent"
         app:layout_constraintBottom_toTopOf="@+id/main_tos_and_pp"
+        app:layout_constraintVertical_weight="2"
```
samtstern: This works, but I was also thinking what we might want is to center the logo between the top of the parent and the top of the ToS/PP.  What do you think about that?

SUPERCILEX: I think it's already doing that, right? At least, that's what I'm getting from this screenshot. The issue seems to be that there's a lot more padding than we're used to.
screenshot_20180924-165834

samtstern:
@SUPERCILEX I took a slightly different approach here:
https://github.com/firebase/FirebaseUI-Android/pull/1460

WDYT?  Not trying to step on your toes, I was already working on a fix when you sent your PR so I figured I might as well finish it and then we can decide.

SUPERCILEX:
@samtstern well I guess it's step-over-each-others-toes week 😂. Anyway, I think we should go with your PR since it has nice sample changes, but you'll probably have to use this PR's layout changes. (I'll post the bug on your PR.)

PR #1460, opened 2018-09-25
Fix landscape NASCAR layout

samtstern:
See #1434 

I also updated the sample app in two ways:
  * Combine the ToS and PP selection radio box now that we make you pass them together anyway
  * Allow a 'None' option

Here's how it looks now:
fui_landscape_notos
fui_landscape_withtos

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
@@ -132,6 +132,12 @@ public class AuthMethodPickerActivity extends AppCompatBase {
         PrivacyDisclosureUtils.setupTermsOfServiceAndPrivacyPolicyText(this,
                 getFlowParams(),
                 termsText);
+
+        // No ToS or PP provided, so we should hide the view entirely
+        if (!getFlowParams().isPrivacyPolicyUrlProvided() &&
+                !getFlowParams().isTermsOfServiceUrlProvided()) {
+            termsText.setVisibility(View.GONE);
+        }
     }
 
     private void populateIdpList(List<IdpConfig> providerConfigs,
--- a/auth/src/main/res/layout-land/fui_auth_method_picker_layout.xml
+++ b/auth/src/main/res/layout-land/fui_auth_method_picker_layout.xml
@@ -22,16 +22,17 @@
         app:layout_constraintEnd_toStartOf="@+id/container"
         app:layout_constraintTop_toTopOf="parent"
         app:layout_constraintBottom_toTopOf="@+id/main_tos_and_pp"
+        android:layout_marginTop="@dimen/fui_field_padding_vert"
+        android:layout_marginBottom="@dimen/fui_field_padding_vert"
         tools:ignore="ContentDescription" />  <!-- TODO remove once the bug is fixed: https://issuetracker.google.com/issues/38281866-->
 
     <TextView
         android:id="@+id/main_tos_and_pp"
         style="@style/FirebaseUI.Text.BodyText"
         android:layout_width="0dp"
-        android:layout_height="0dp"
+        android:layout_height="wrap_content"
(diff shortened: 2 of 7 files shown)


auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
```diff
@@ -132,6 +132,12 @@ protected void onFailure(@NonNull Exception e) {
         PrivacyDisclosureUtils.setupTermsOfServiceAndPrivacyPolicyText(this,
                 getFlowParams(),
                 termsText);
+
+        // No ToS or PP provided, so we should hide the view entirely
+        if (!getFlowParams().isPrivacyPolicyUrlProvided() &&
+                !getFlowParams().isTermsOfServiceUrlProvided()) {
+            termsText.setVisibility(View.GONE);
```
samtstern: Just realized this doesn't look so nice in portrait :man_facepalming:

auth/src/main/res/layout-land/fui_auth_method_picker_layout.xml
```diff
@@ -22,16 +22,17 @@
...
+        android:layout_marginTop="@dimen/fui_field_padding_vert"
+        android:layout_marginBottom="@dimen/fui_field_padding_vert"
         tools:ignore="ContentDescription" />  <!-- TODO remove once the bug is fixed: https://issuetracker.google.com/issues/38281866-->
 
     <TextView
         android:id="@+id/main_tos_and_pp"
         style="@style/FirebaseUI.Text.BodyText"
         android:layout_width="0dp"
-        android:layout_height="0dp"
+        android:layout_height="wrap_content"
```
SUPERCILEX: I wanted to do that too, but then this happens. Maybe you know more ConstraintLayout jitsu? 😆 

screenshot_20180925-103634

samtstern: Hmmm.  I probably need to mess with the vertical chain type then.  All the "padding" on the ToS/PP text isn't actually padding, it's a result of the view bounds stretching to meet its constraints.

samtstern:
@SUPERCILEX I think at some point with split screen there's no good layout.  Here's what I have now at half-screen in both portrait and landscape.  IMO it looks as good as we can hope for.

Plus while it's a bummer visually, I think if you have a terms/privacy policy you need them to be visible at all costs.

**Half of Lanscape**
fui_half_land

**Half of Portrait**
fui_half_portrait

SUPERCILEX:
A few thoughts:
- Is the large layout-land fixed? Since you reverted `wrap_content` and didn't add the contraint weights, I'm pretty sure this doesn't fix the original issue 😂
- The ToS actually scrolls so I don't think it's _too_ big a deal if not all of it is visible at once
- Small layouts are hard! 🤪

samtstern:
What do you mean by "large layout-land"?  I think that's what I showed in the very first comment, but maybe there's something else?

Also how does the ToS scroll?

samtstern:
Yep the landscape (the original bug) is still working with the latest commits:

android_emulator_-_pixel_2_api_28_5554_and_firebaseui-android____projects_firebaseui-android__-_____buildsrc_src_main_kotlin_config_kt__buildsrc_

SUPERCILEX:
Ohhh, nevermind you made the `ScrollView` `0dp`. So that means the screenshot below is still a bug on your branch. Were you saying that's ok? That's why I was pointing out that the ToS is scrollable.

I've [created a patch](https://github.com/firebase/FirebaseUI-Android/files/2422278/fix.zip) with my fix to clear confusion, but it _does_ mean you have to scroll the ToS. WDYT? Is that a reasonable tradeoff?

screenshot_20180926-194021

samtstern:
@SUPERCILEX thanks for the patch!  I applied it and it definitely helps in the super-small situation.

From what I can tell, this now looks passable in any configuration of options/orientation/size.

Issue #1462, opened 2018-09-26
[Internal] Make our gradle build consistent


samtstern:
I have been fighting the build all day so I just want to log some of the issues here so that I can try and fix them.

This is what we run for our builds;
`./gradlew clean assembleDebug proguard-tests:build check`

It should always give the same result, on any machine.  However I have run it about 100x today and run into sporadic issues like:

  * `Cannot invoke method containsKey() on null object` (from various tasks)
  * `Task :lint:compileKotlin FAILED`
  * `Task :common:generateDebugRFile FAILED`
  * `Task :library:packageDebugResources FAILED`

I'd say of the 100 builds I tried, maybe 25 succeded :-(

samtstern:
Also might be a good opportunity to get ready for Gradle 5.0

SUPERCILEX:
Ouch, yeah I haven't seen those kinds of issues 😞.

samtstern:
@SUPERCILEX do you know of any way to make importing the project easier?  Could we have like a root `build.gradle` that just references a `kts` file with all the meat?

samtstern:
Yeah maybe just waiting will be the thing to do.  I'm surprised that more of the Android community hasn't jumped to Kotlin scripts.  

Also still can't believe you don't use AS!

SUPERCILEX:
Yeah, I dunno.

I do use AS Canary for Android specific stuff, but IntelliJ has the latest and greatest in terms of pure coding tools. 😉

PR #1463, opened 2018-09-26
Stabilize Build

samtstern:
See #1462

These are some *very* simple changes but on my machine they have an impact on stability :man_shrugging:

### DIFF
--- a/proguard-tests/build.gradle.kts
+++ b/proguard-tests/build.gradle.kts
@@ -1,3 +1,6 @@
+// This is always set to 'true' on Travis CI
+val inCiBuild = System.getenv("CI") == "true"
+
 android {
     defaultConfig {
         multiDexEnabled = true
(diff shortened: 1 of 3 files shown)


proguard-tests/build.gradle.kts
```diff
@@ -1,3 +1,6 @@
+// This is always set to 'true' on Travis CI
+val inCiBuild = System.getenv("CI") == "true"
```
SUPERCILEX: I would check for nullity instead so we're resilient to possible changes: https://github.com/SUPERCILEX/Robot-Scouter/blob/2da4e7eed01da413eb151ce8cb09c75e33e7d53d/buildSrc/src/main/kotlin/com/supercilex/robotscouter/build/internal/Constants.kt#L8

SUPERCILEX:
Oooooh, just a second too late 😂. Shouldn't really matter so 🤷‍♂️👍

Issue #1465, opened 2018-09-28
Firebase Auth - Facebook signin violates facebook app requirements


ryanzor:
### Environment

  * Android device: _____Pixel
  * Android OS version: _____P
  * Firebase/Play Services SDK version: _____16.0.3 (firebase core)
  * FirebaseUI version: ____4.2.0
  
### The problem:

Facebook is getting very stringent on their app review process.  I recently had an app rejected from submission because my login button does not match Facebook guidelines (https://developers.facebook.com/docs/facebook-login/best-practices#brandedlogin).  Specifically with FirebaseUI if Facebook is the only login option firebaseUI goes ahead and starts the facebook login flow, this seems to not be okay with Facebook app approvers and will be a cause for an app not being approved for Facebook login.

#### Steps to reproduce:

  1. _____Make Facebook the only login option
  2. _____Iniate login flow
  3. _____Notice you don't see the "Login with Facebook" button that matches Facebook branding guidelines
 
  
#### Relevant Code:

In sign in kickstarter the below method should have an "or facebook is the provider" in the if statement.  Alternatively, it may make sense to add a flag to force showing the auth method choice.  This would not change the behavior for existing clients and allow users to make their own "Login with Facebook" button if they wanted.  The main downside would be that Facebook review process is now 1-2 weeks with little feedback so I'd expect more people trying out FirebaseUI to lose those weeks without realizing they need to set the flag.

  ```
   private void startAuthMethodChoice() {
        // If there is only one provider selected, launch the flow directly
        if (getArguments().isSingleProviderFlow()) {
  ```

samtstern:
@ryanzor thanks for filing this!  This is an interesting and important development.

The relevant policy (from [this page](https://developers.facebook.com/policy/#login)) seems to be:

> Use a clearly branded log in button that follows the guidelines in the [Facebook Brand Resource Center](https://www.facebookbrand.com/#).

samtstern:
@ryanzor what UI does your app show before launching into FirebaseUI?  It seems like another solution would be for you to put a "Log in with Facebook" button into your own app and then launch the single-provider FirebaseUI flow from there.

SUPERCILEX:
On our side, it shouldn't be too hard to force the picker activity to show just Facebook. But yeah, if you have a login button that only shows 1 IDP, then might as well be branded

samtstern:
@ryanzor in version `4.2.1` we have added the following option:

[...]

PR #1467, opened 2018-10-02
Ss allow force nascar

samtstern:
See #1465


auth/src/main/java/com/firebase/ui/auth/data/model/FlowParameters.java
```diff
@@ -162,4 +167,11 @@ public boolean isAnonymousUpgradeEnabled() {
         return enableAnonymousUpgrade;
     }
 
+    public boolean isAlwaysShowProviderChoice() {
```
SUPERCILEX: Unused method?

Issue #1476, opened 2018-10-09
Light shadow when using anonymous auth option and dark theme.


austinhodak:
```
<style name="SignInTheme" parent="Theme.AppCompat">
        <item name="colorPrimary">@color/timelineBG</item>
        <item name="colorPrimaryDark">@color/timelineDark</item>
        <item name="colorAccent">@color/md_orange_500</item>

        <item name="colorControlNormal">@color/md_orange_500</item>
        <item name="colorControlActivated">@color/md_orange_A700</item>
        <item name="colorControlHighlight">@color/md_orange_A200</item>
        <item name="android:windowBackground">@color/timelineBG</item>
    </style>
```

As you can see the rest of the button shadows are styled corrected other than the new anonymous one.

![smartselect_20181009-122203_battle buddy](https://user-images.githubusercontent.com/5241478/46683814-25e8b580-cbbf-11e8-9c0e-a09aef6c1c4e.jpg)

samtstern:
Thanks @austinhodak for catching this!

cc @lsirac

samtstern:
This has been fixed and released in version `4.2.1`.

Issue #1415, opened 2018-08-23
Touch area issue on some UI elements (38dp -> 48dp)


peterhav:
* Android device: all
  * Android OS version: all
  * FirebaseUI version: 4.1.0
  
#### Steps to reproduce:

After uploading the app to Google Play, the automatically created pre-launch reports state that some clickable parts of the Firebase UI have a too small size (38dp, should be 48dp).
  
#### Observed Results:

![firebase_ui_toucharea](https://user-images.githubusercontent.com/1724223/44546471-53b19200-a718-11e8-9195-df2dca234dcb.png)
  
#### Expected Results:

Firebase-UI should not give any problems in the pre-launch reports of Google Play.

samtstern:
This has been fixed and released in version `4.2.1`.

Issue #1192, opened 2018-03-17
FirebaseAuth UI country code selection scrollbar not implemented right. 


srujanb:
### Step 2: Describe your environment

  * Android device: OnePlus
  * Android OS version: 8.0
  * Google Play Services version: 11.8.0
  * Firebase/Play Services SDK version: 11.8.0
  * FirebaseUI version: 3.2.1
  
### Step 3: Describe the problem:

The scrollbar should not go beyond the screen when scrolling happens on the dialog box. See the **scrollbar** in the video below. 

https://youtu.be/t2xaA-M0t-E

#### Steps to reproduce:

1.  Implement firebaseUI auth. 
2.  Include SMS auth builder
3.  Click on the list of countries to select country code.
4.  Scroll 
5.  Get annoyed with the scrollbar going all the way to the bottom, out of the screen. 
  
#### Observed Results:

See the **Scrollbar** in the video.

https://youtu.be/t2xaA-M0t-E
  
#### Expected Results:

  Scrollbar not going beyond screen
  
#### Relevant Code:

None.

samtstern:
This has been fixed and released in version `4.2.1`.

PR #1495, opened 2018-10-22
Fix small issues left in version 4.2.1

samtstern:
Fixes:
  * #1192 
  * #1376 
  * #1415 
  * #1476 

Other changes:
  * Bump some dependencies
  * Add some `.gradle` files to make Android Studio pick up the build easier.  I got a new computer so I was able to experience the first-timer pain.


auth/src/main/java/com/firebase/ui/auth/ui/phone/CountryListAdapter.java
```diff
@@ -92,6 +102,16 @@ public int getPositionForSection(int index) {
 
     @Override
     public int getSectionForPosition(int position) {
+        if (sections == null) {
```
samtstern: Since we were using "fast scrolling" this method had to be implemented to get the scroll bar to work correctly.

auth/src/main/res/drawable-v21/fui_idp_button_background_anonymous.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
```
samtstern: This fixes the issue with the anonymous button not working in certain themes.

auth/src/main/res/drawable/fui_ic_anonymous_white_24dp.xml
```diff
@@ -1,12 +1,10 @@
...
     android:width="24dp"
     android:height="24dp"
     android:viewportHeight="24.0"
-    android:viewportWidth="24.0"
-    tools:ignore="InvalidVectorPath">
+    android:viewportWidth="24.0">
     <path
         android:fillColor="#FFFFFF"
-        android:pathData="M12 5.9c1.16 0 2.1.94 2.1 2.1s-0.94 2.1-2.1 2.1S9.9 9.16 9.9 8s0.94-2.1 2.1-2.1m0 9c2.97 0 6.1 1.46 6.1 2.1v1.1H5.9V17c0-0.64 3.13-2.1 6.1-2.1M12 4C9.79 4 8 5.79 8 8s1.79 4 4 4 4-1.79 4-4-1.79-4-4-4zm0 9c-2.67 0-8 1.34-8 4v3h16v-3c0-2.66-5.33-4-8-4z" />
+        android:pathData="M12,5.9c1.16,0 2.1,0.94 2.1,2.1s-0.94,2.1 -2.1,2.1S9.9,9.16 9.9,8s0.94,-2.1 2.1,-2.1m0,9c2.97,0 6.1,1.46 6.1,2.1v1.1L5.9,18.1L5.9,17c0,-0.64 3.13,-2.1 6.1,-2.1M12,4C9.79,4 8,5.79 8,8s1.79,4 4,4 4,-1.79 4,-4 -1.79,-4 -4,-4zM12,13c-2.67,0 -8,1.34 -8,4v3h16v-3c0,-2.66 -5.33,-4 -8,-4z" />
```
samtstern: I used Android-studio's built in version of this same icon, which I assume will be valid on all devices (not the internal svg version)

auth/src/test/java/com/firebase/ui/auth/testhelpers/TestHelper.java
```diff
@@ -44,13 +46,17 @@
 import static org.mockito.Mockito.when;
 
 public final class TestHelper {
+
+    private static final String TAG = "TestHelper";
+    private static final String DEFAULT_APP_NAME = "[DEFAULT]";
```
samtstern: FirebaseApp no longer exposes this constant.

auth/src/test/java/com/firebase/ui/auth/viewmodel/PhoneProviderResponseHandlerTest.java
```diff
@@ -104,9 +105,13 @@ public void testSignIn_autoUpgradeAnonymousEnabledWithExistingUser_expectMergeFa
         mHandler.getOperation().observeForever(mResponseObserver);
         setupAnonymousUpgrade();
 
+        FirebaseAuthUserCollisionException ex =
+                new FirebaseAuthUserCollisionException("foo", "bar");
+        TestHelper.setPrivateField(ex, FirebaseAuthUserCollisionException.class,
```
samtstern: @lsirac you removed the constructor of `FirebaseAuthUserCollision` that we were using, so I have to use a reflection hack to make this test pass.

Can't use a mock because the exception class is `final`

lsirac: I added a #setAuthCredential method, can we just use that? Or will the symbols change on each release?

samtstern: Yeah unfortunately that's proguarded away for 3p:

```java
    public final FirebaseAuthUserCollisionException zza(AuthCredential var1) {
        this.zzhi = var1;
        return this;
    }
```

auth/src/main/java/com/firebase/ui/auth/ui/phone/CountryListAdapter.java
samtstern: Yeah won't show up till you start scrolling, that's normal.  There's nothing I can figure out to fix the jumpiness, will maybe look into it again later if people report it.

Issue #1501, opened 2018-10-24
Add translations and documentation for email link sign in (Phase 1)


samtstern:
See #1486

samtstern:
Fixed by #1521

PR #1503, opened 2018-10-25
Email Link Sign In Documentation + Backstack fixes

lsirac:
https://github.com/firebase/FirebaseUI-Android/issues/1501

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/ui/email/EmailActivity.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/email/EmailActivity.java
@@ -173,7 +173,7 @@ public class EmailActivity extends AppCompatBase implements CheckEmailFragment.C
     public void onTroubleSigningIn(String email) {
         TroubleSigningInFragment troubleSigningInFragment = TroubleSigningInFragment.newInstance
                 (email);
-        switchFragment(troubleSigningInFragment, TroubleSigningInFragment.TAG, true);
+        switchFragment(troubleSigningInFragment, TroubleSigningInFragment.TAG, true, true);
     }
 
     @Override
@@ -214,16 +214,22 @@ public class EmailActivity extends AppCompatBase implements CheckEmailFragment.C
     }
 
 
-    private void switchFragment(Fragment fragment, String tag, boolean withTransition) {
+    private void switchFragment(Fragment fragment, String tag, boolean withTransition,
+                                boolean addToBackStack) {
         FragmentTransaction ft = getSupportFragmentManager().beginTransaction();
         if (withTransition) {
             ft.setCustomAnimations(R.anim.fui_slide_in_right, R.anim.fui_slide_out_left);
         }
-        ft.replace(R.id.fragment_register_email, fragment, tag).disallowAddToBackStack().commit();
+        ft.replace(R.id.fragment_register_email, fragment, tag);
+        if (addToBackStack) {
+            ft.addToBackStack(null).commit();
+        } else {
+            ft.disallowAddToBackStack().commit();
+        }
     }
 
     private void switchFragment(Fragment fragment, String tag) {
-        switchFragment(fragment, tag, false);
+        switchFragment(fragment, tag, false, false);
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


auth/src/main/java/com/firebase/ui/auth/ui/email/EmailLinkFragment.java
```diff
@@ -85,7 +89,8 @@ public void onActivityCreated(@Nullable Bundle savedInstanceState) {
         ActionCodeSettings actionCodeSettings = getArguments().getParcelable(ExtraConstants
                 .ACTION_CODE_SETTINGS);
 
-        if (savedInstanceState == null || !savedInstanceState.getBoolean(EMAIL_SENT)) {
+        if (!mEmailSent && (savedInstanceState == null || !savedInstanceState.getBoolean
```
samtstern: Your crazy formatter is at it again!

auth/src/main/java/com/firebase/ui/auth/ui/email/EmailLinkFragment.java
```diff
@@ -34,6 +34,9 @@
     private TroubleSigningInListener mListener;
     private ScrollView mTopLevelView;
 
+    // Used to avoid sending a new email when popping off the fragment backstack
+    private boolean mEmailSent;
```
samtstern: Does this survive rotation?

auth/src/main/java/com/firebase/ui/auth/ui/email/EmailActivity.java
```diff
@@ -214,16 +214,22 @@ private void showRegisterEmailLinkFragment(AuthUI.IdpConfig emailConfig,
     }
 
 
-    private void switchFragment(Fragment fragment, String tag, boolean withTransition) {
+    private void switchFragment(Fragment fragment, String tag, boolean withTransition,
```
samtstern: Can we make an overload without the new param that defaults it to false?

auth/src/main/java/com/firebase/ui/auth/ui/email/EmailLinkFragment.java
lsirac: :cry: fixed!

auth/src/main/java/com/firebase/ui/auth/ui/email/EmailLinkFragment.java
lsirac: Nope, my bad :-1: . I also fixed the case where we go from EmailLinkFragment -> TroubleSigningInFragment -> EmailLinkFragment (by clicking resend), we'd be at EmailLinkFragment and have to press the back twice to get back to the nascar screen.

auth/src/main/java/com/firebase/ui/auth/ui/email/EmailActivity.java
lsirac: We have one already that sets both booleans to false.

auth/src/main/java/com/firebase/ui/auth/ui/email/EmailActivity.java
```diff
@@ -173,11 +173,14 @@ public void onNewUser(User user) {
     public void onTroubleSigningIn(String email) {
         TroubleSigningInFragment troubleSigningInFragment = TroubleSigningInFragment.newInstance
                 (email);
-        switchFragment(troubleSigningInFragment, TroubleSigningInFragment.TAG, true);
+        switchFragment(troubleSigningInFragment, TroubleSigningInFragment.TAG, true, true);
     }
 
     @Override
     public void onClickResendEmail(String email) {
+        if (getSupportFragmentManager().getBackStackEntryCount() > 0) {
```
samtstern: Can you just add a comment here detailing the assumption you're making?  As I understand it, you're assuming that if there's any back stack at all here it's due to the trouble sending flow and you want to pop from the stack.

lsirac: Good idea. Done. :smiley:

PR #1514, opened 2018-10-31
Basque translation

landa95:
I added basque translations.


samtstern:
@landa95 thanks for this contribution!

Historically we have not accepted user contributions for new languages since we'd be pretty stuck when we want to add new strings if we don't have our own translators.

But let me see if I can get general Basque support in FirebaseUI and then add your translations as a "seed".

nanangsugianto:
Semangat, walau pun saya pemula dan masih binggung tapi kami slalu mendukung untuk terus maju😍

PR #1541, opened 2018-11-20
Fix Issue 1538

samtstern:
Sets a loading state to prevent the button from being clicked twice in a row.


samtstern:
Hate to self-merge actual code but ... it's one line.

samtstern:
🦃

PR #1558, opened 2018-12-30
Automatically submit phone verification codes when possible

### LINKED ISSUE #1526: Improve phone auth UX
I've used several other phone auth systems and found a few improvements we can do:
- If the user is typing, automatically start the auth process once all numbers have been typed
   - We'll probably have to update the ToS text
- In `onResume` check the clipboard and automatically start the auth process if a 6 digit number is present

SUPERCILEX:
Fixes #1526

@samtstern WDYT of the new UX?

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/ui/phone/PhoneActivity.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/phone/PhoneActivity.java
@@ -22,6 +22,7 @@ import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.support.annotation.RestrictTo;
 import android.support.design.widget.TextInputLayout;
+import android.support.v4.app.FragmentManager;
 import android.widget.Toast;
 
 import com.firebase.ui.auth.ErrorCodes;
@@ -83,7 +84,18 @@ public class PhoneActivity extends AppCompatBase {
             protected void onSuccess(@NonNull PhoneVerification verification) {
                 if (verification.isAutoVerified()) {
                     Toast.makeText(
-                            PhoneActivity.this, R.string.fui_auto_verified, Toast.LENGTH_LONG).show();
+                            PhoneActivity.this,
+                            R.string.fui_auto_verified,
+                            Toast.LENGTH_LONG
+                    ).show();
+
+                    FragmentManager manager = getSupportFragmentManager();
+                    if (manager.findFragmentByTag(SubmitConfirmationCodeFragment.TAG) != null) {
+                        // Ensure the submit code screen isn't visible if there's no code to submit.
+                        // It's possible to get into this state when an SMS is sent, but then
+                        // automatically retrieved.
+                        manager.popBackStack();
+                    }
                 }
 
                 handler.startSignIn(verification.getCredential(), new IdpResponse.Builder(
--- a/auth/src/main/java/com/firebase/ui/auth/ui/phone/SubmitConfirmationCodeFragment.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/phone/SubmitConfirmationCodeFragment.java
@@ -15,12 +15,15 @@
 package com.firebase.ui.auth.ui.phone;
 
 import android.arch.lifecycle.ViewModelProviders;
+import android.content.ClipData;
+import android.content.ClipboardManager;
 import android.content.Context;
 import android.os.Bundle;
 import android.os.Handler;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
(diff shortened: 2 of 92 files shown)


auth/src/main/java/com/firebase/ui/auth/ui/phone/SubmitConfirmationCodeFragment.java
```diff
@@ -139,41 +166,19 @@ public void onDestroy() {
...
     private void setupConfirmationCodeEditText() {
         mConfirmationCodeEditText.setText("------");
         mConfirmationCodeEditText.addTextChangedListener(new BucketedTextChangeListener(
-                mConfirmationCodeEditText, 6, "-",
+                mConfirmationCodeEditText, VERIFICATION_CODE_LENGTH, "-",
                 new BucketedTextChangeListener.ContentChangeCallback() {
                     @Override
-                    public void whileComplete() {
-                        mSubmitConfirmationButton.setEnabled(true);
+                    public void whenComplete() {
```
samtstern: My only concern here is that this is less typo-friendly.  If you accidentally type the wrong 6 digits (maybe you fat-finger a duplicate digit in the middle) then it's going to submit automatically.

For instance I just entered the code and got into this state:
image

There are two UX issues here:
  * The error message position is strange
  * It's not obvious that entering that last digit was "submitting" and I should back out and proceed.

I think the second issue could be mitigated by deleting the digits entered when an error happens?  Although that could be jarring.

SUPERCILEX: Yeah, you're right, we're optimizing for the successful case (which should be more common).

For the error message, I think that's out of scope for this PR since we'll have to use some [ugly hacks](https://stackoverflow.com/questions/42575825/textinputlayout-error-right-align). As for typos, I like the idea of nuking the text. 👍

samtstern:
@SUPERCILEX thanks for the new changes!  FYI I am on Jury Duty this week so will be a few days before I can properly test this, but the changes look good to me at a glance.

auth/src/main/java/com/firebase/ui/auth/ui/phone/PhoneActivity.java
```diff
@@ -83,7 +84,18 @@ protected void onFailure(@NonNull Exception e) {
...
+                            R.string.fui_auto_verified,
+                            Toast.LENGTH_LONG
+                    ).show();
+
+                    FragmentManager manager = getSupportFragmentManager();
+                    if (manager.findFragmentByTag(SubmitConfirmationCodeFragment.TAG) != null) {
+                        // Ensure the submit code screen isn't visible if there's no code to submit.
+                        // It's possible to get into this state when an SMS is sent, but then
+                        // automatically retrieved.
+                        manager.popBackStack();
```
samtstern: This works for auto-verification (where no SMS is sent) but does not work for the case where the code is sent but then auto-retrieved.  In that case we get `onCodeSent` first and then when this happens the flow goes a little crazy, I couldn't capture a video of it.

auth/src/main/java/com/firebase/ui/auth/ui/phone/SubmitConfirmationCodeFragment.java
```diff
@@ -117,6 +123,22 @@ public void onViewCreated(@NonNull View view, @Nullable Bundle savedInstanceStat
...
+    public void onActivityCreated(@Nullable Bundle savedInstanceState) {
+        super.onActivityCreated(savedInstanceState);
+        ViewModelProviders.of(requireActivity())
+                .get(PhoneProviderResponseHandler.class)
+                .getOperation()
+                .observe(this, new Observer<Resource<IdpResponse>>() {
+                    @Override
+                    public void onChanged(@Nullable Resource<IdpResponse> resource) {
+                        if (resource.getState() == State.FAILURE) {
+                            mConfirmationCodeEditText.setText("");
```
samtstern: This works well :-)

auth/src/main/java/com/firebase/ui/auth/ui/phone/PhoneActivity.java
SUPERCILEX: I think I fixed the craziness, but I'm still not super happy with the UX. It's either show the Smart Lock dialog on top of a code-less edittext or show it over the phone number entry fragment. I know we have a toast, but maybe we need a brief bottom sheet or something? Either way, I doubt I'll have the time to play with that. 😢

samtstern:
Sorry I took a while getting here but LGTM!

Issue #1555, opened 2018-12-18
Phone auth UI area code wrong position


jeffreyliu8:
![7891877299955355863](https://user-images.githubusercontent.com/12398645/50184161-170e3580-02c9-11e9-9f90-af2acb8630b5.png)

samtstern:
@jeffreyliu8 that looks bad!  But a screenshot is not enough information to debug this.  A few questions:

- If you run the sample app provided in the `app` module of this repo, can you reproduce the issue?
- What version of FIrebaseUI are you using?  What version of other dependencies are you using?

iamdeadman:
Hi @samtstern 

Facing the same issue here, Haven't tried with sample firebase android auth project but just did a fresh install of FirebaseAuthUi 4.3.0 and please find below the other firebase dependencies I am using if that could help debug and resolve this issue. 

    ```
implementation 'com.google.firebase:firebase-auth:16.1.0'
    implementation 'com.firebaseui:firebase-ui-auth:4.3.0'
    implementation 'com.google.firebase:firebase-dynamic-links:16.1.5'
    implementation 'com.google.firebase:firebase-messaging:17.3.4'
    implementation 'com.google.firebase:firebase-core:16.0.6'
    implementation 'com.google.firebase:firebase-appindexing:17.1.0'
    implementation 'com.google.firebase:firebase-crash:16.2.1'
    implementation 'com.google.firebase:firebase-config:16.1.2'
    implementation 'com.firebase:firebase-jobdispatcher:0.8.2'
    implementation 'com.crashlytics.sdk.android:crashlytics:2.9.7'
    implementation 'com.google.android.gms:play-services-auth:16.0.1'
```

Please let me know if you need anymore information from my side in order to debug this issue further. 


Thanks, 
Harshit

samtstern:
Thanks Harshit.  What versions of the support/androidx libraries are you
using?  Specifically constraint layout.

iamdeadman:
Hi,    supportLibraryVersion = '28.0.0' for all support libraries and 1.1.3 for constraint layout
   // android support libraries
    //noinspection GradleDependency,GradleCompatible
    implementation "com.android.support:support-v13:$rootProject.supportLibraryVersion"
    implementation "com.android.support:support-core-utils:$rootProject.supportLibraryVersion"
    implementation "com.android.support:support-core-ui:$rootProject.supportLibraryVersion"
[...]

PR #1562, opened 2019-01-02
Revert baseline textview change

samtstern:
See issue #1555, it looks like removing our baseline text input layout class was premature, it's not in the support design lib yet.


samtstern:
Since this is a revert I view this as safe.

Issue #1559, opened 2019-01-01
No "data rates may apply" message when using only Phone Auth method


nabax:
Using FirebaseUI Auth to manage authentication only with phone method.
If I set only "PhoneBuilder" as AvailableProvider, the "phone number verification" screen does not show the text "By tapping "Verify Phone number", an SMS may be sent. Message & data rates may apply".
See here: 
![2019-01-01_171702](https://user-images.githubusercontent.com/46299438/50574481-c1b64600-0de9-11e9-8340-63faff05bd5c.jpg)

If I add any other provider (so that at least 2 methods are available and thus the Authentication Method Picker is presented at first), clicking on "phone authentication" takes me to the "phone number verification" page where the text correctly appears.
See here:
![2019-01-01_171821](https://user-images.githubusercontent.com/46299438/50574483-ced33500-0de9-11e9-8a39-aadaec130d5c.jpg)

Here's my list of dependencies (already had to roll back to firebase-ui-auth 4.1.0 to avoid the alignement issue described in https://github.com/firebase/FirebaseUI-Android/issues/1555): 

    implementation 'com.android.support:appcompat-v7:28.0.0'
    implementation 'com.android.support.constraint:constraint-layout:1.1.3'
    implementation 'com.android.support:design:28.0.0'
    implementation 'com.google.firebase:firebase-core:16.0.6'
    implementation 'com.google.firebase:firebase-messaging:17.3.4'
    implementation 'com.firebaseui:firebase-ui-auth:4.1.0' 

Do you know how to get the message to appears also when only Phone method is used?
Thank you

samtstern:
@nabax can you show your code where you are building up the `AuthUI.SignInIntentBuilder`?  Specifically the part where you set the ToS and Privacy policy?  I don't see this behavior in my sample app.

samtstern:
Ah OK I see the conditions where this happens now:
  * No terms or privacy policy URL is set
  * The only available provider is the phone number

In this case, there is no text shown.

nabax:
Yes, that's correct: no ToS/PP is set.
Thank you.

PR #1563, opened 2019-01-02
Show the SMS rates warning even without ToS/PP

samtstern:
See #1559


samtstern:
@lsirac pinged me to say LGTM since he is locked out of GitHub.

PR #1565, opened 2019-01-03
Version 4.3.1

samtstern:


samtstern:
Not sure why the commit log has some dupes? But the change log looks sane.

Issue #1585, opened 2019-02-21
Firebase Auth UI not showing sign in flow when offline


nerdyadventurer:
First I asked [a question on SO](https://stackoverflow.com/questions/54741573/firebase-authui-not-showing-sign-in-flow) but there was no solution. This may be a bug or my mistake.

### Step 2: Describe your environment

  * Android device: LG Zero
  * Android OS version: 6.0
  * Google Play Services version: play-services-auth : 16.0.1
  * Firebase/Play Services SDK version: , firebase-core:16.0.6
  * FirebaseUI version: 4.3.1
  
### Step 3: Describe the problem:

When running the app firebase auth ui sign in flow is not show, instead shows the main activity, which in this case a blank constraint layout.

#### Steps to reproduce:

  1. Clone the repository at [https://github.com/nerdyadventurer/firebase-auth-ui](url)
  2. open project using android studio and setup firebase authentication(enable Google and Email authentication providers from the firebase console)
  3. Run the app
  
#### Observed Results:

  * What happened?  This could be a description, `logcat` output, etc.
   Logcat is empty.
  Shows the main activity (which is black in this case)

#### Expected Results:

  * What did you expect to happen?
      Showing the sign in flow with Google sign in, email sign in buttons and  etc.

#### Relevant Code:

**dependencies**
  ```
 dependencies {
    implementation fileTree(dir: 'libs', include: ['*.jar'])
    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
    implementation 'androidx.appcompat:appcompat:1.1.0-alpha02'

    implementation 'androidx.constraintlayout:constraintlayout:1.1.3'

    // firebase
    implementation 'com.firebaseui:firebase-ui-auth:4.3.1'
    implementation 'com.google.firebase:firebase-core:16.0.6'
    implementation 'com.google.firebase:firebase-auth:16.1.0'
    implementation 'com.google.android.gms:play-services-auth:16.0.1'

    // testing
    testImplementation 'junit:junit:4.12'
    androidTestImplementation 'androidx.test:runner:1.1.2-alpha01'
    androidTestImplementation 'androidx.test.espresso:espresso-core:3.1.2-alpha01'
}

  ```
**MainActivity.kt**

```
package example.com.firebaseauthui

import android.app.Activity
import android.content.Intent
import android.os.Bundle
import android.widget.Toast
import androidx.appcompat.app.AppCompatActivity
[...]

PR #1586, opened 2019-02-22
Offline Auth method picker

samtstern:
See #1585

### DIFF
--- a/auth/build.gradle.kts
+++ b/auth/build.gradle.kts
@@ -13,6 +13,7 @@ android {
         disable("UnusedQuantity")
         disable("UnknownNullness")  // TODO fix in future PR
         disable("TypographyQuotes") // Straight versus directional quotes
+        disable("MissingTranslation") // STOPSHIP: Get translations
     }
 
     testOptions {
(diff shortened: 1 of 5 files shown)


samtstern:
The idea here is to allow launching the auth method picker while offline in many cases.  If your app uses FirebaseUI as the "home" screen for new users this can be essential.

samtstern:
@SUPERCILEX curious if you see any reasons to check for offline before showing the picker screen other than SmartLock?

SUPERCILEX:
I'm actually curious, what happens if we don't do the check at all? `getActiveNetworkInfo()` is kind of a hack anyway since it will say you're connected even if the WiFi is behind a paywall or something like that (think Gogo Inflight).

It'll probably be hard to do, but I almost feel like each network request failure should be handled separately. We could map all the separate exceptions to a FUI network exception, similar to the user cancellation one. Then, we only fail when we truly couldn't connect to the network. It'll also improve the UX if a user was connected before the Smart Lock screen but then lost their internet juices.

In any case, I'm a fan of letting the user access the picker screen without an internet connection. 👍

samtstern:
@SUPERCILEX good call, removing the offline check produces the same behavior (at the start of the flow).

Later in the flow some things go badly if we're offline:

- Email flow just basically does nothing.
- Google sign in flow hangs forever

We should handle all of this, as you said, but for now I just want to keep this change small and straighforward.

Issue #1524, opened 2018-11-06
Pagination with realtime data in FireStore


dovahkiin98:
In the FireStore Adapter, there are 2 options: No paging, and paging.
As said, the paging option doesn't offer realtime data, but the no paging option becomes time expensive when the data becomes large in an app like chatting apps.

Is there a way to get the best of both worlds?

samtstern:
@dovahkiin98 no, right now there is not.  Can you describe your use case?  Do you need:

1. Just notifications for new items
2. (1) and notifications when an existing item changes
3. (1) and (2) and the ability for items to be re-ordered dynamically?

Or something else?

dovahkiin98:
Notifications for when items are added, changes, or removed. But also the ability to get pages, for example, receive each 50 items alone, not the whole block.
I'm guessing the only way is to make a custom listener and change the limit dynamically, tho that didn't work well with me in Realtime database.

samtstern:
Can you explain what you mean by "But also the ability to get pages, for example, receive each 50 items alone, not the whole block." ?

dovahkiin98:
Getting all the data at once would take time, I want to be able to take pages instead of all the data. 50 messages a time, then add them to the list. But all the pages are real time. Listening to edits and deletions, while one page is listening to addition.
Can the query for the adapter be updated after the items are laid out?

For example:
I make the view listen for changes and deletion with a pagination adapter, while a snapshot listener listens for additions. But the pagination adapter isn't real time, so it only reads the data once right? If changes or deletions happen, the adapter won't be notified.

johnnie90:
As @dovahkiin98 wrote, it could be great a  a way to get the best of both worlds.

I would appreciate it if this improvement is made in the future.

At the moment my solution was reload the adapter to my recyclerview.

RubyLichtenstein:
please check my implementation for real time pagination 

https://github.com/RubyLichtenstein/FireCat 

@dovahkiin98

sabeehshah:
@jonthn90 hey, can you explain if you found a solution on how to accomplish this while still using FirestorePagingAdapter?

jonthn90:
hi @sabeehshah. This was some time ago so I don't remember exactly how was solve :(. But I remember    I implemented the pagination without the Real Time feature. And every time a user adds/removes an item, refresh the adaptar to reload all the collection again. This was not the most eficcient but with the pagination don't reload all the colleccion just the number of items that you configurate in your pagination.

[...]

Issue #1388, opened 2018-07-06
Refactor to AndroidX


SUPERCILEX:
AndroidX just hit beta (stable APIs) a few days ago which means we should all get ready to migrate, but the Android team hasn't left us a whole lot of guidance on how we're all supposed to move forward. For public APIs that take in old support components (like our recycler view adapters), they won't compile with the new stuff last time I tried to do so in alpha. That means we have to release a 5.0, I guess? But if we do that, then devs have to upgrade to AndroidX themselves and other libraries might not have upgraded. But if we don't, devs can't upgrade. 😕 Seems like a catch-22 unless I'm missing something.

Anyway, wanted to start this discussion with everyone. @samtstern What are your thoughts?

PS: [this comment](https://github.com/google/dagger/issues/1188#issuecomment-399341700) suggests the public API stuff not compiling was a bug. Will update after I try refactoring again.

SUPERCILEX:
Update: turns out the Jetifier does work now. 👏 Stuff will compile, it's just IntelliJ that gets mad. 👍

iamyaoxi:
I really love this idea! I've been using AndroidX for most of my projects. And so far, integrating Firebase with my AndroidX-based project turns out to be well. 

Looking forward for this feature to be implemented.

SUPERCILEX:
@iamyaoxi Funny that you mention it, I'm actually in the middle of the refactor right now. 😁 The tooling is _waaaaaaay_ better than it was back in the early alphas. I'm actually really impressed—haven't had any compile time issues so far. (We'll see about runtime issues 😉)

Anyway, another thought: we'll also want to look into the new `com.google.android.material` components to see what UI stuff we can spice up.

iamyaoxi:
Yes, that'd be awesome! 👍 

And the new Material Components for Android makes it very easy to Theme and Style for even older Android version!

krwilson8:
Android X is in beta and claims to be stable for production use with minor bugs. 

I went ahead and moved to Android X not realizing this would affect firebase ui. I had no problems using a pixel 2 xl and galaxy s8 as test devices.

Some users (about 3 percent from what I saw - could be higher but I had it fixed quickly) could not use the firebase ui sign in. The app would just crash with resource not found - including some users using the same devices I tested on. 

I have no idea why this happens but adding my own Google sign in option was my work around for the firebase ui login screen not working. 

The recycler adapter appears to work just fine and has for the last two weeks.

[...]

PR #1599, opened 2019-03-13
Feature FirestoreItemKeyedDatasource

abond:
Hello!
I want to share my implementation of DataSourse for Firestore that support data refresh on real time events.

Related issue #1524

### DIFF
--- a/firestore/build.gradle.kts
+++ b/firestore/build.gradle.kts
@@ -36,6 +36,7 @@ dependencies {
 }
 apply {
     plugin("kotlin-android")
+    plugin("kotlin-kapt")
     plugin("kotlin-android-extensions")
 }
 repositories {
--- a/firestore/build.gradle.kts
+++ b/firestore/build.gradle.kts
@@ -1,3 +1,4 @@
+val kotlin_version: String = "1.3.21"
 tasks.named("check").configure { dependsOn("compileDebugAndroidTestJavaWithJavac") }
 
 android {
@@ -31,4 +32,12 @@ dependencies {
     androidTestImplementation(Config.Libs.Test.rules)
     androidTestImplementation(Config.Libs.Test.mockito)
     androidTestImplementation(Config.Libs.Arch.paging)
+    implementation(kotlin("stdlib-jdk7", kotlin_version))
+}
+apply {
+    plugin("kotlin-android")
+    plugin("kotlin-android-extensions")
+}
+repositories {
+    mavenCentral()
 }
--- a/firestore/src/main/java/com/firebase/ui/firestore/paging/FirestoreItemKeyedDataSource.kt
+++ b/firestore/src/main/java/com/firebase/ui/firestore/paging/FirestoreItemKeyedDataSource.kt
@@ -1,0 +1,235 @@
+package com.firebase.ui.firestore.paging
+
+
+import android.arch.paging.DataSource
+import android.arch.paging.ItemKeyedDataSource
+import android.arch.paging.PagedList
+import com.google.android.gms.tasks.Task
+import com.google.android.gms.tasks.TaskCompletionSource
+import com.google.android.gms.tasks.Tasks
+import com.google.firebase.firestore.*
+import com.google.firebase.firestore.Source.CACHE
+import java.util.concurrent.atomic.AtomicBoolean
+
+/**
+ * Implementation of [ItemKeyedDataSource] for Firebase Firestore.
+ *
+ * [errorCallback] is optional and if not provided the [defaultErrorCallback] will be used.
+ *
+ * Note: If user provide [errorCallback] he should control data source invalidation manually.
+ *
+ * @property directQuery The direct query.
+ * @property reverseQuery The reverse query.
+ * @property errorCallback The error callback.
+ */
+@Suppress("NOTHING_TO_INLINE")
+class FirestoreItemKeyedDataSource(
+        private val directQuery: Query,
+        private val reverseQuery: Query,
+        private val errorCallback: ErrorCallback?
+) : ItemKeyedDataSource<DocumentSnapshot, DocumentSnapshot>() {
+
+    private val registrations = mutableListOf<ListenerRegistration>()
+
+    /**
+     * Factory for [FirestoreItemKeyedDataSource].
+     *
+     * @property queryFactory The query factory.
+     * @property errorCallback The error callback.
+     */
+    class Factory @JvmOverloads constructor(
+            private val queryFactory: QueryFactory,
+            private val errorCallback: ErrorCallback? = null
+    ) : DataSource.Factory<DocumentSnapshot, DocumentSnapshot>() {
+
+        /**
+         * Creates and returns [FirestoreItemKeyedDataSource].
+         *
+         * @return The created data source.
+         */
+        override fun create(): DataSource<DocumentSnapshot, DocumentSnapshot> =
+                FirestoreItemKeyedDataSource(queryFactory.create(), queryFactory.createReverse(), errorCallback)
+    }
+
+    override fun loadInitial(
+            params: LoadInitialParams<DocumentSnapshot>,
+            callback: LoadInitialCallback<DocumentSnapshot>
+    ) {
+        val key = params.requestedInitialKey
+        val loadSize = params.requestedLoadSize.toLong()
+
+        if (key == null) {  // Initial data load could be asynchronous.
+            loadFromBeginning(loadSize, callback)
+        } else {    // Stale data refresh. Should be synchronous to prevent RecyclerView items disappearing and loss of current position.
+            loadAround(key, loadSize, callback).await()
+        }
+    }
+
+    /**
+     * Tries to load some items before and after provided [key].
+     *
+     * @param key The key to load around.
+     * @param loadSize Requested number of items to load.
+     *
+     * @return A Task that will be resolved with the results of the Query.
+     */
+    private fun loadAround(
+            key: DocumentSnapshot,
+            loadSize: Long,
+            callback: LoadInitialCallback<DocumentSnapshot>
+    ) = tryGetPriorKey(key, loadSize / 2).continueWithTask { previousTask ->
+        loadStartAt(previousTask.result!!, loadSize, callback)
+    }
+
+    /**
+     * Tries to get key before provided [key] to start load page from it.
+     *
+     * @param key The key to load before.
+     * @param preloadSize Requested number of items to preload.
+     *
+     * @return A Task that will be resolved with the results of the Query.
+     */
+    private fun tryGetPriorKey(
+            key: DocumentSnapshot,
+            preloadSize: Long
+    ): Task<DocumentSnapshot> = getItemsBeforeKey(key, preloadSize).continueWith { task ->
+        if (task.isSuccessful) {
+            val documents = task.result!!.documents
+            documents.getOrElse(documents.lastIndex) { key }
+        } else {
+            key
+        }
+    }
+
+    private fun getItemsBeforeKey(
+            key: DocumentSnapshot,
+            preloadSize: Long
+    ) = endAtQuery(key, preloadSize).get(CACHE)
+
+    private fun loadStartAt(
+            key: DocumentSnapshot,
+            loadSize: Long,
+            callback: LoadInitialCallback<DocumentSnapshot>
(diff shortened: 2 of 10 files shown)


samtstern:
@abond thank you for sending this PR!  Before I even begin to take a look I want to warn you that we can't ship Kotlin in this library yet as it will force our many Java developers to make the switch.  However I am happy to look over the implementation and decide if we want to pursue converting it to Java and including it in the library.

Also thanks for bringing `ItemKeyedDataSource` to my attention!  It looks like something new in the paging library that would definitely have been useful to us back when we were first implementing pagination support.

abond:
But Kotlin fully interoperablewith java. Nor users of library won't notice anything, nor other developers will be forced to use kotlin.

samtstern:
I don't know of a way to ship a library developed in Kotlin in a way that
Java developer can consume it without enabling Kotlin in their tool chain
but I'd love to be wrong!

Do you know of anyone doing this?

abond:
Users of library written in kotlin can use it in java project like any other jvm lib. They don't need to enable kotlin in their project. Kotlin was developed with this goal in mind. It also developed to support step by step migration in project that start using it. That means that u can write new code in kotlin without conversion all Java files to kotlin. Bite code is fully interoperable. There is some features that you can't use from Java, but I was trying my best to implement this Java friendly.

firestore/src/main/java/com/firebase/ui/firestore/paging/QueryFactory.kt
```diff
@@ -0,0 +1,261 @@
...
+     */
+    fun createReverse(): Query =
+            filters.fold(collectionReference as Query) { query, filter -> filter.applyReverse(query) }
+
+    /**
+     * Base class for filters.
+     *
+     * @property fieldPath The field to sort by.
+     */
+    private abstract class Filter(
```
samtstern: One of the informal rules we have on the team is "don't re-implement the query interface".  The reason is that this library will always be chasing the official SDK as queries are added / changed.  Plus as you can see by your implementation, it's a lot of code.

From what I can tell this system is used to be able to create a reversible query.  Have you considered just asking the developer to pass in both a forward and reverse Query object (with no `limit()`, `startAt()` or `endAt()` clauses) when they create the recycler adapter?

firestore/src/main/java/com/firebase/ui/firestore/paging/FirestoreItemKeyedDataSource.kt
```diff
@@ -0,0 +1,235 @@
...
+     * Callback to listen to data loading errors.
+     */
+    interface ErrorCallback {
+        fun onError(exception: FirebaseFirestoreException)
+    }
+
+    private inline fun <TResult> Task<TResult>.await() {
+        try {
+            Tasks.await(this)
+        } catch (e: Throwable) {  }
```
samtstern: What sorts of errors are you ignoring here and why?

abond:
> First of all I really appreciate your sending this PR! I know this is a very complex topic.
> 
> Before I go and review the details I have some high-level questions about your implementation and I also left one or two comments:
> 
> 1. How does this deal with items that move?  For example if my query is `collection(users).orderBy(score)` and users scores are changing in real-time, what happens?  What if the user at the "page boundary" which is used to load the next page changes score dramatically?
> 2. How many pages are kept in memory at once?  When are they loaded/unloaded?

1. All this "load when" logic is done by Paging library. I just implement DataSource by recomendations that is given in Paging Library documentation. 
Here https://developer.android.com/reference/android/arch/paging/DataSource#updating-paged-data in last paragraph written how to implement DS for services like firestore:

> If you have more granular update signals, such as a network API signaling an update to a single item in the list, it's recommended to load data from network into memory. Then present that data to the PagedList via a DataSource that wraps an in-memory snapshot. Each time the in-memory copy changes, invalidate the previous DataSource, and a new one wrapping the new state of the snapshot can be created.

So my DS add data listener which treat first data load event as fresh page and second as data change and invalidate DS. DS invalidation leads to recreation of PagedList+DataSource pair. During recreation PagedList provide last accessed document and requestedLoadSize(by default pageSize*3, but you can change by PagedList.Config) as a parameters to DS. DS tries to load new portion of data around provided document and pass it back to PL which submit it to PagedAdapter. PagedAdapter compares old list with new and generete and apply update sequense which gently updates RecyclerView.

2. You can't control number of pages kept in memory before androidx library if I'm not wrong. In androidx PagedList.Config theree is maxSize paramerer which how many items PL holds in memory. 
And as I said even  before androidx pages are dropped on each data source change.

firestore/src/main/java/com/firebase/ui/firestore/paging/QueryFactory.kt
abond: Yep, we can ask for backward query, but as I understand only a few people understand that to query documents before provided you should create reverced query, not just call endAt + limit. 
To tell the truth I myself find out this only while working on this feature. =)

firestore/src/main/java/com/firebase/ui/firestore/paging/FirestoreItemKeyedDataSource.kt
abond: It waits for completion of the Task that tries to get key befor one that is provided on DS recreation after invalidation(that happens after any document change, creadion or delition, if this change is in the range of existing listener thar previously load page). If it fails DS just start load new portion of documents beginning from provided key.

abond: Sorry! It's not true right now. Described in the previous comment is how it was working at first. Now it wait for whole initial data load finished. And error handling occurred in ResultsListener.onResults method. This code only needed because if we don't wait for this load synchronously Page library erase Recyclerview by submitting empty results.

samtstern:
@abond thanks for your explanations!  I have to apologize I am super busy right now and will need a few more days until I can get back to this PR. 

While I'm being slow, maybe @SUPERCILEX or @puf has some thoughts on this.

firestore/build.gradle.kts
```diff
@@ -31,4 +32,13 @@ dependencies {
     androidTestImplementation(Config.Libs.Test.rules)
     androidTestImplementation(Config.Libs.Test.mockito)
     androidTestImplementation(Config.Libs.Arch.paging)
+    implementation(kotlin("stdlib-jdk7", kotlin_version))
```
SUPERCILEX: Whoa there! @samtstern Is FUI accepting Kotlin now? From the [FAQ](https://developer.android.com/kotlin/faq):
image

Typically, libs will be written _entirely_ in Kotlin (not postfixed like this) and only if it offers some sort of syntactic benefit that isn't available in Java (like coroutines or multi-platform stuff, see https://github.com/square/okio). Here, I have trouble seeing what's worth the 1MB and cognitive overhead of ensuring our APIs stay idiomatic in both Java _and_ Kotlin.

firestore/src/main/java/com/firebase/ui/firestore/paging/FirestoreItemKeyedDataSource.kt
```diff
@@ -0,0 +1,263 @@
...
+    override fun getKey(item: DocumentSnapshot) = item
+
+    override fun invalidate() {
+        super.invalidate()
+        removeAllPageLoaders()
+    }
+
+    private fun removeAllPageLoaders() {
+        registrations.forEach { it.remove() }
+        registrations.clear()
```
SUPERCILEX: I believe one of the reasons this hasn't been done yet is because of how difficult it is to properly subscribe and unsub from these listeners. Can you explain a bit more how and when `invalidate` is called?

firestore/src/main/java/com/firebase/ui/firestore/paging/QueryFactory.kt
SUPERCILEX: Yeah, this whole file should probably be removed. 🙂

firestore/build.gradle.kts
samtstern: Yeah I am pushing the Kotlin question to the end.  If the PR is worth having we can translate it back to Java at the last step.

SUPERCILEX: SGTM 👍

firestore/src/main/java/com/firebase/ui/firestore/paging/FirestoreItemKeyedDataSource.kt
abond: No, I just forget to unsubscribe.

I wrote it previously. By design of Paging library to refresh data in UI you should invalidate DS.
Here quote from Paging library documentation:

> A PagedList / DataSource pair are a snapshot of the data set. A new pair of PagedList / DataSource must be created if an update occurs, such as a reorder, insert, delete, or content update occurs. A DataSource must detect that it cannot continue loading its snapshot (for instance, when Database query notices a table being invalidated), and call invalidate(). Then a new PagedList / DataSource pair would be created to load data from the new state of the Database query.

SUPERCILEX: Ok, I think I understand what's going on now. 👍

I'm a little concerned about cost. Correct me if I'm wrong, but here's what I think is going on: when a page is being loaded, it adds a snapshot listener which submits the data the first time it gets triggered. Then, on the second trigger, it invalidates the entire data set and removes all the listeners. This causes the adapter to requery everything from scratch and start the process anew. Is that correct?

If so, this feature would be extremely expensive to users. They wouldn't be paying for the diff—every single update would cost them the entire query. Coming from an official Firebase library, that seems a little abusive. @samtstern WDYT?

abond: Yes, you right that it requery data on every chnge, but not all the data only recent pageSize*3 items by default or can be specified by PagedList.Config.initialLoadSizeHint. What is more when you finish #1388 there will be PagedList.Config.maxSize which limits number of item currently kept in memory and which will be involved in old with new data comparison.

What about requerying this 3 resent pages most likely will be loaded from cache if I understand correctly how firebase works.

abond: After some testing I come to conclusion that I was wrong - every requery leeds to extra charges.
I was wrongly assumed that since offline access is enabled by default recently accessed documents will be fetched from local cache and do not count as priced reads. But unfortunately thats not true.

samtstern: If you issue the **exact same** query multiple times within about 30 minutes, you shouldn't be charged for the entire doc set multiple times.  You'll only be charged for the new docs.

abond: Yes, you right, but problem is that exactly same means with the same limit and start/end conditions.
But in this implementation we can't avoid creating new queries starting from different positions so I think it is expensive to use. 
I have an idea how to solve this issue. If I succeed I send new PR.

nlgtuankiet: Assume that the query result is identical

> execute exact same query multiple times doesn't charge for the entire doc set multiple times

Is this still true? @samtstern

samtstern: Yes that's true, identical queries issued within 30m of each other don't double-charge unless there are new results.

nlgtuankiet: What if we have 1 new document in the query result, for example, the first query return 3 documents, the second query return the same 3 documents + 1 new document. In this case, how many documents read will be charged?

samtstern: In that case 4 total reads.  3 for the initial result.  1 for the update later.

samtstern:
Since this PR is very old and we just decided to move to Paging 3 for Firebase UI v8.0.0 I am going to close this. @abond sorry we never resolved all the outstanding questions, we still really appreciate this contribution!

Issue #1601, opened 2019-03-18
Firebase Database Paging Support - [New Feature-Pagination]


PatilShreyas:
Firebase Realtime Database has **FirebaseRecyclerAdapter** to show Data Items in RecyclerView. But, the issue is it has not Paging support added until now. I have developed a solution on Pagination of Firebase Database in RecyclerView on below repository with library code and sample app.
https://github.com/PatilShreyas/FirebaseRecyclerPagination

I referred source code of FirestorePagingAdapter while developing the solution.
FirebaseRecyclerPagingAdapter is built on top of Android Paging Support Library.
So, I request you to please permit me to add this feature in the official FirebaseUI library to help developers with this paging library to load data in pages.
Can I fork this repository and make this feature available?

samtstern:
@PatilShreyas thanks for filing this!  So since you based this on the Firestore adapter, does that mean the tradeoffs are the same?  In particular, does the adapter ignore realtime updates?

PatilShreyas:
Yes. Till now, it does not support realtime updates.

PatilShreyas:
@samtstern Thank you !
I will do it.

PatilShreyas:
@samtstern I have updated code as you suggested here #1603 
Implemented the same as Firestore version.

PatilShreyas:
Hi @samtstern 
I go through [FirestorePagingAdapter](https://github.com/firebase/FirebaseUI-Android/blob/master/firestore/src/main/java/com/firebase/ui/firestore/paging/FirestorePagingAdapter.java) and there is a method in it.

```java
public void retry() {
        FirestoreDataSource source = mDataSource.getValue();
        if (source == null) {
            Log.w(TAG, "Called retry() when FirestoreDataSource is null!");
            return;
        }

        source.retry();
    }
```
I tried to implement an exact feature in our `FirebaseRecyclerPagingAdapter`.
But, I observed that `source` always remains null. Every time it shows Log...
_"Called retry() when FirestoreDataSource is null!"_. 
I also tested calling this method from [FirestorePagingActivity](https://github.com/firebase/FirebaseUI-Android/blob/master/app/src/main/java/com/firebase/uidemo/database/firestore/FirestorePagingActivity.java) after caught in error. Still, it failed.

How do we implement `retry()` in `FirebaseRecyclerPagingAdapter`

samtstern:
This has been fixed and released in `5.0.0`

SerggioC:
[...]

Issue #1592, opened 2019-03-01
FirestoreDataSource - Please add the ability to get an occurred exception


Zeliret:
Right now I can only get an error status and call retry. But I need to get the exact exception to provide an explanation to my users.
I have found the place where the error state is populated and It seems that I need to override a lot of code to get the error message. 
Is there a good way to workaround this?

samtstern:
@Zeliret can you show an example of the code you're using now and maybe some comments on how you'd want it to improve?

Zeliret:
@samtstern sure. Thanks for the answer.
So, I have a Query instance and configured PagedList builder.
I create a new adapter:
```kotlin
val options = FirestorePagingOptions.Builder<Trip>()
                        .setLifecycleOwner(this)
                        .setQuery(result.data.query, Source.DEFAULT, pagingConfig) {
                            result.data.converter(it)
                        }
                        .build()

                    binding.tripsList.swapAdapter(TripsAdapter(options), false)
```
Let's assume I get an error while I'm loading a page. I can check it here:
```kotlin
override fun onLoadingStateChanged(state: LoadingState) {
            if( state == LoadingState.ERROR ) {
                retry()
            }
        }
```
But I can get only the flag. How can I get the actual error? As far as I can see from the source code:
```java
/**
     * Error listener that logs, sets the error state, and sets up retry.
     */
    private abstract class OnLoadFailureListener implements OnFailureListener {

        @Override
        public void onFailure(@NonNull Exception e) {
            Log.w(TAG, "load:onFailure", e);

            // On error we do NOT post any value to the PagedList, we just tell
            // the developer that we are now in the error state.
            mLoadingState.postValue(LoadingState.ERROR);

            // Set the retry action
            mRetryRunnable = getRetryRunnable();
        }

        protected abstract Runnable getRetryRunnable();
    }
```
Here the exception just dumped into the log. 
So I can not check if it is a network error or show it to the user.

samtstern:
@Zeliret ah I see that's a good feature request!  You're right, there's no good way to get it right now in the Paginated adapters.

samtstern:
How would you feel about something like a `getLastError()` call in the paginated adapter classes?

Zeliret:
> How would you feel about something like a `getLastError()` call in the paginated adapter classes?

That sounds good. 
So I could use the flow:
```kotlin
if( state == LoadingState.ERROR ) {
     val ex = getLastError()
     if( ex is IOException )
[...]

PR #1603, opened 2019-03-20
Feature - Realtime Database Pagination

PatilShreyas:
Here, I have implemented **FirebaseRecyclerPagingAdapter** that supports Pagination of Firebase Database with RecyclerView as issue #1601


app/src/main/java/com/firebase/uidemo/database/realtime/FirebaseDbPagingActivity.java
```diff
@@ -0,0 +1,153 @@
...
+
+        //Initialize Paging Configurations
+        PagedList.Config config = new PagedList.Config.Builder()
+                .setEnablePlaceholders(false)
+                .setPrefetchDistance(5)
+                .setPageSize(30)
+                .build();
+
+        //Initialize Firebase Paging Options
+        FirebasePagingOptions<Post> options = new FirebasePagingOptions.Builder<Post>()
```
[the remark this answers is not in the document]
samtstern: Actually I am fine with leaving this as-is.  Since it's parallel to the other names in this package.

database/src/main/java/com/firebase/ui/database/paging/FirebaseDataSource.java
```diff
@@ -0,0 +1,182 @@
...
+
+/**
+ * Data source to power a {@link FirebaseRecyclerPagingAdapter}.
+ *
+ * Note: although loadInitial, loadBefore, and loadAfter are not called on the main thread by the
+ *       paging library, we treat them as if they were so that we can facilitate retry without
+ *       managing our own thread pool or requiring the user to pass us an executor.
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
+public class FirebaseDataSource<T> extends PageKeyedDataSource<String, T> {
```
samtstern: If you look at the Firestore implementation, we have the DataSource deal only in snapshots rather than putting in `Class<T>`.  I think it's simpler that way and also I would like for this code to be as similar to the Firestore code as possible to enable future maintenance.

database/src/main/java/com/firebase/ui/database/paging/FirebaseDataSource.java
```diff
@@ -0,0 +1,182 @@
...
+        Query mInitQuery = mQuery.limitToFirst(params.requestedLoadSize);
+        mInitQuery.addListenerForSingleValueEvent(new ValueEventListener() {
+            @Override
+            public void onDataChange(@NonNull DataSnapshot dataSnapshot) {
+                if (dataSnapshot.exists()) {
+                    ArrayList<T> mDataList = new ArrayList<T>();
+                    for (DataSnapshot ds : dataSnapshot.getChildren()) {
+                        T data = ds.getValue(mClass);
+                        String key = ds.getKey();
+                        mKeyList.add(key);
```
samtstern: See above comment.  If we were dealing only in `DataSnapshot` here we would not have to keep separate lists of the keys and the data, which will reduce bugs and complexity.

database/src/main/java/com/firebase/ui/database/paging/FirebaseDataSource.java
```diff
@@ -0,0 +1,182 @@
...
+                    for (DataSnapshot ds : dataSnapshot.getChildren()) {
+                        T data = ds.getValue(mClass);
+                        String key = ds.getKey();
+
+                    /*
+                      Check for first Item.
+                      Because in Firebase Database there is no query for startAfter(key).
+                      So we're ignoring first data item
+                    */
+                        if (!isFirstItem) {
```
samtstern: This loop seems over complicated.  Rather than checking `firstItem` in the loop over and over why not just iterate from child `1` to the end, skipping the first one?

database/src/main/java/com/firebase/ui/database/paging/FirebaseDataSource.java
```diff
@@ -0,0 +1,182 @@
...
+            public void onDataChange(@NonNull DataSnapshot dataSnapshot) {
+                if (dataSnapshot.exists()) {
+                    ArrayList<T> mList = new ArrayList<T>();
+                    boolean isFirstItem = true;
+
+                    for (DataSnapshot ds : dataSnapshot.getChildren()) {
+                        T data = ds.getValue(mClass);
+                        String key = ds.getKey();
+
+                    /*
```
samtstern: style nit: don't use block comments inside classes or methods.  Please add back the code style settings in this PR which should help.

database/src/main/java/com/firebase/ui/database/paging/FirebaseDataSource.java
```diff
@@ -0,0 +1,182 @@
...
+    private Class<T> mClass;
+    private ArrayList<String> mKeyList;
+
+    private static final String TAG = "FirebaseDataSource";
+
+    private final MutableLiveData<LoadingState> mLoadingState = new MutableLiveData<>();
+    private final MutableLiveData<DatabaseError> mError = new MutableLiveData<>();
+    private final MutableLiveData<ArrayList<String>> mKeyLiveData = new MutableLiveData<>();
+
+    private static final String WRONG_DATA_PATH_STATUS = "WRONG_PATH";
```
samtstern: What does "wrong" mean in this case?  The error should help the developer understand what they did wrong and what they should do to fix it.

firestore/proguard-rules.pro
```diff
@@ -1 +1,2 @@
 -dontwarn com.firebase.ui.firestore.paging.**
+-dontwarn com.firebase.ui.database.paging.**
```
samtstern: We should not need to mention the database adapter in the firestore package.  I think this is in the wrong file.

database/src/main/java/com/firebase/ui/database/paging/FirebaseDataSource.java
PatilShreyas: "wrong" means if developer given reference of database path that not exists.
For e.g.
 `FirebaseDatabase().getInstance().getReference().child("list");`
If 'list' is not there in database then path is invalid or wrong.
So for such mistake, I have given "wrong".

PatilShreyas:
Thank you @samtstern 
I will make changes soon as you suggested.

app/src/main/java/com/firebase/uidemo/database/realtime/FirebaseDbPagingActivity.java
```diff
@@ -0,0 +1,153 @@
...
+                                                    @NonNull Post model) {
+                        holder.bind(model);
+                    }
+                };
+
+        mRecycler.setLayoutManager(new LinearLayoutManager(this));
+        mRecycler.setAdapter(mAdapter);
+
+        mAdapter.setStateChangedListener(new StateChangedListener() {
+            @Override
```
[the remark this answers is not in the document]
PatilShreyas: But in that case we will not able to show `databaseError` in `onError`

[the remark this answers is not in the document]
PatilShreyas: Can I override two methods in adapter class like..
```
onLoadingStateChanged(LoadingState loadingState)
onError(DatabaseError databaseError)
```
So that it will display information about database errors ?

PatilShreyas:
@samtstern I have updated code as you suggested.
Issue #1601

app/src/main/java/com/firebase/uidemo/database/realtime/FirebaseDbPagingActivity.java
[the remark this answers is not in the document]
samtstern: Ah yes that seems the same as this feature request:
https://github.com/firebase/FirebaseUI-Android/issues/1592

I am fine with adding a second callback!  That will solve this issue at the same time.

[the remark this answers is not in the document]
PatilShreyas: Okay. I Have added it in a recent commit.

samtstern:
@PatilShreyas on first glance this looks great!  Thank you for such a well-formed PR.  I will hopefully have time to do a final thorough review later today.

PatilShreyas:
Thank You!

database/src/main/java/com/firebase/ui/database/paging/DatabasePagingOptions.java
```diff
@@ -0,0 +1,145 @@
...
+         * See {@link #setQuery(Query, PagedList.Config, SnapshotParser)}.
+         */
+        @NonNull
+        public Builder<T> setQuery(@NonNull Query query,
+                                   @NonNull PagedList.Config config,
+                                   @NonNull Class<T> modelClass) {
+            return setQuery(query, config, new ClassSnapshotParser<>(modelClass));
+        }
+        /**
+         * Sets the Firestore query to paginate.
```
samtstern: nit: this says Firestore

database/src/main/java/com/firebase/ui/database/paging/FirebaseDataSource.java
```diff
@@ -0,0 +1,239 @@
...
+                if (dataSnapshot.exists()) {
+
+                    //Make List of DataSnapshot
+                    List<DataSnapshot> data = new ArrayList<>();
+                    String lastKey = null;
+
+                    Iterator<DataSnapshot> iterator = dataSnapshot.getChildren().iterator();
+
+                    //Skip First Item
+                    if (iterator.hasNext())
```
samtstern: Style nit: please always use `{}` with `if` statements:

```
if (iterator.hasNext()) {
  iterator.next()
}

File: database/src/main/java/com/firebase/ui/database/paging/FirebaseDataSource.java [hunk 11]
```diff
@@ -0,0 +1,239 @@
...
+
+    @Override
+    public void loadAfter(@NonNull final LoadParams<String> params,
+                          @NonNull final LoadCallback<String, DataSnapshot> callback) {
+
+        // Set loading state
+        mLoadingState.postValue(LoadingState.LOADING_MORE);
+
+        //Load params.requestedLoadSize+1 because, first data item is getting ignored.
+        Query mNewQuery = mQuery.startAt(null, params.key).limitToFirst(params.requestedLoadSize + 1);
```
samtstern: Is there a reason we use the strategy of calling `startAt` and then skipping the first item instead of calling `startAfter`?

PatilShreyas: Actually, `Query` of *Firebase Realtime Database* is not having method like `startAfter()` to get data after skipping any value.
Also, `DataSnapshot` not have any method to convert it into `List<DataSnapshot>` so that we can pass its sublist from 1st index.
This is the reason here we used `startAt`

samtstern: D'oh you're right!  I was looking at Firestore.  My bad.

PatilShreyas:
@samtstern I have fixed the code as your suggestions.

samtstern:
@PatilShreyas thanks for this!  Here's my plan:

  * I am going to release version 4.3.2 today since it contains important bug fixes
  * Then I will create a new branch for version-4.4.0-dev
  * I'll merge your PR into that branch, and then we can put a few more things together and release 4.4.0

PatilShreyas:
@samtstern Okay!
Thank You!

Issue #1608, opened 2019-03-27
FirestorePagingAdapter - retry() failing


PatilShreyas:
Hi @samtstern 
I go through [FirestorePagingAdapter](https://github.com/firebase/FirebaseUI-Android/blob/master/firestore/src/main/java/com/firebase/ui/firestore/paging/FirestorePagingAdapter.java) and there is a method in it.

```java
public void retry() {
        FirestoreDataSource source = mDataSource.getValue();
        if (source == null) {
            Log.w(TAG, "Called retry() when FirestoreDataSource is null!");
            return;
        }

        source.retry();
    }
```
I observed that `source` always remains null. Every time it shows Log...
 _"Called retry() when FirestoreDataSource is null!"_. 

I also tested calling this method from [FirestorePagingActivity](https://github.com/firebase/FirebaseUI-Android/blob/master/app/src/main/java/com/firebase/uidemo/database/firestore/FirestorePagingActivity.java) after caught in error. Still, it failed.

I implemented this in our [FirebaseRecyclerPagingAdapter](https://github.com/firebase/FirebaseUI-Android/blob/version-4.4.0-dev/database/src/main/java/com/firebase/ui/database/paging/FirebaseRecyclerPagingAdapter.java) and it is running fine!
I did it by adding `Observer` for `DataSource` and it's running fine without null value.

samtstern:
I noticed that fix in the other adapter, thanks for that!  We should fix this in the Firestore one too, for `4.4.0`.

I'd also like to understand the underlying behavior here.  Does LiveData not hold values if nobody is observing?

PatilShreyas:
Yes.
I observed that while working on [`FirebaseRecyclerPagingAdaper`](https://github.com/firebase/FirebaseUI-Android/blob/version-4.4.0-dev/database/src/main/java/com/firebase/ui/database/paging/FirebaseRecyclerPagingAdapter.java).

I did an experiment:
- When I added this line 
 `mDataSource.observeForever(mDataSourceObserver)`
Then after calling `getValue()` on `mDataSource` doesn't returned null value

- When I commented the code of `observeForever`
Then `mDataSource.getValue()` always returned null value.

By observing this, may be LiveData not hold values if nobody is observing

samtstern:
Apparently if you're using `Transformations.map` or `Transformations.switchMap` then the expected behavior is for there to be no value until you add an `Observer` ... which is really tricky!

I just found this out by asking someone.

[...]

PR #1609, opened 2019-03-27
Fix Issue #1608 - retry() failing in FirestorePagingAdapter

PatilShreyas:
I have fixed `retry()` failing and added a new method `refresh()` to refresh the list from the beginning as issue #1608

### DIFF
--- a/firestore/src/main/java/com/firebase/ui/firestore/paging/FirestorePagingAdapter.java
+++ b/firestore/src/main/java/com/firebase/ui/firestore/paging/FirestorePagingAdapter.java
@@ -34,7 +34,10 @@ public abstract class FirestorePagingAdapter<T, VH extends RecyclerView.ViewHold
     private final LiveData<LoadingState> mLoadingState;
     private final LiveData<FirestoreDataSource> mDataSource;
 
-
+    /*
+        LiveData created via Transformation do not have a value until an Observer is attached.  
+        We attach this empty observer so that our getValue() calls return non-null later.
+    */
     private final Observer<FirestoreDataSource> mDataSourceObserver = new Observer<FirestoreDataSource>() {
         @Override
         public void onChanged(@Nullable FirestoreDataSource source) {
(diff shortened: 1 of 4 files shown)


firestore/src/main/java/com/firebase/ui/firestore/paging/FirestorePagingAdapter.java
```diff
@@ -34,6 +34,14 @@
     private final LiveData<LoadingState> mLoadingState;
     private final LiveData<FirestoreDataSource> mDataSource;
 
+
+    private final Observer<FirestoreDataSource> mDataSourceObserver = new Observer<FirestoreDataSource>() {
+        @Override
+        public void onChanged(@Nullable FirestoreDataSource source) {
+
```
samtstern: Can we add a comment inside here explaining why we need an empty observer so that future people who read this code know?

PatilShreyas: @samtstern yes.
What comment should be added ?
Please suggest me.

PatilShreyas: `//Observer to hold the value for FirestoreDataSource` ??

PatilShreyas:
Suggest me comment to write.

samtstern:
I'd say something like: "LiveData created via Transformation do not have a
value until an Observer is attached.  We attach this empty observer so that
our getValue() calls return non-null later."

PatilShreyas:
Fine! 
I'm adding this.

PatilShreyas:
@samtstern  Is it OK?

PatilShreyas:
@samtstern We should add this in `FirebaseRecyclerPagingLibrary` too.
Is it OK to make changes in `FirebaseRecyclerPagingAdapter` in this PR?

samtstern:
@PatilShreyas yep that's fine with me.  I'll wait to merge until those changes are made also

PR #1615, opened 2019-04-01
Ability to get occurred exception - Issue 1592

PatilShreyas:
As issue #1592,
I have updated code as `FirebaseRecyclerPagingAdapter` and added an ability to get the occurred `Exception` while using `FirestorePagingAdapter`.

To use this, the user can Override `onError()` method of Adapter class.

e.g.
```java
 @Override
 protected void onError(@NonNull Exception e) {
          mSwipeRefreshLayout.setRefreshing(false);
          Log.e(TAG, e.getMessage(), e);
 }
```


samtstern:
@PatilShreyas thank you again!

PatilShreyas:
Welcome!
Love Firebase !! 🔥🔥

Issue #1626, opened 2019-04-29
Remove TwitterSignInHandler because its dependency TwitterKit is deprecated.


efrohnhoefer:
See: https://blog.twitter.com/developer/en_us/topics/tools/2018/discontinuing-support-for-twitter-kit-sdk.html
See: https://github.com/firebase/FirebaseUI-iOS/pull/650

lsirac:
We can't remove this and replace it with Generic IDP just yet, some changes are needed. Generic IDP was focused on providers supporting OAuth2 while Twitter uses OAuth1 for user authentication.

We're going to be making some changes to support this, and then we'll replace it with Generic IDP. 

Releases work differently for Android, which is why it'll take longer :(

samtstern:
Version `6.2.0` has been released.

PR #1627, opened 2019-04-29
Remove TwitterSignInHandler because its dependency TwitterKit is deprecated

efrohnhoefer:
Closes: #1626


samtstern:
@efrohnhoefer thanks for sending this PR!  I appreciate the effort however this is not the route we are going to take with Twitter.  First we need to re-implement this on top of the "Generic IDP" flow (@lsirac can comment) and then we can remove this.

samtstern:
This was fixed by @lsirac in #1714

Issue #1614, opened 2019-03-29
FirestoreRecyclerAdapter method to replace Query/Snapshots


luaz:
FirestoreRecyclerAdapter currently only support passing query using FirestoreRecyclerOptions in constructor. If I want to change the query, I would need to recreate a new adapter.

It would be nice to have a FirestoreRecyclerAdapter method to replace Query/Snapshots, or don't make mSnapshots private.

In some cases, the adapter already hold some variables to be used at onBindViewHolder for data display. Currently in order to replace the query, I have to create a new adapter and copy the variables from the old adapter.

samtstern:
@luaz that's an interesting request, worth considering!  Can you show a brief example of some of the state you're keeping in your adapter?

PatilShreyas:
Hi @luaz and @samtstern, I tried to implement this feature and check below link for output. Do you want an exact feature like this?

[***Click here to see Video Output***](https://drive.google.com/file/d/1uc_9KLMmjNjs2BrwRrWcEUZB3341ojbX/view?usp=sharing)

I have done this by creating a new method in `FirebaseRecyclerAdapter` called `updateQuery()`.
I created two Query instances as
```java
        mQuery1 = mDatabase.getReference().child("items").limitToFirst(5);
        mQuery2 = mDatabase.getReference().child("items").limitToLast(5);
```
First, I passed `Query1` instance to an adapter using `setQuery()`.
Then on the menu click, I called the method of Adapter i.e. `updateQuery()` as below
```java
    @Override
    public boolean onOptionsItemSelected(@NonNull MenuItem item) {
        if (item.getItemId() == R.id.item_replace) {
            showToast("Replacing Query...");
            mAdapter.updateQuery(mQuery2);
            return true;
        }
        return super.onOptionsItemSelected(item);
    }
```

PatilShreyas:
After replacing query, what should be the behaviour of adapter? Previous snapshots should be there? or they should be cleaned and result of new query is loaded?

PatilShreyas:
@samtstern @luaz 
I tried to achieve it basically.
Please have a look at the code.

- First of all, I added an abstract method in **`ObservableSnapshotArray`** like...
```java
 public abstract void updateQuery(@NonNull Query newQuery);
```

- Then, In **`FirestoreRecyclerAdapter`**, I added the following method...
```java
    public void updateQuery(@NonNull Query newQuery) {
        mSnapshots.updateQuery(newQuery);
    }
```
[...]

PR #1659, opened 2019-07-20
Method to update Query in Firebase and Firestore Adapters.

PatilShreyas:
This PR includes :-
- Implements a method to change/update `Query` in both Firebase as well as Firestore Adapters. (Issue #1614)
- Fixes some `auth` lint checks failed during `gradle check` command.


PatilShreyas:
@samtstern I think the build is failing when the same values are present in `strings.xml` with the different names.

PR #1660, opened 2019-07-20
Method to update Query in Firebase and Firestore Adapters.

PatilShreyas:
This PR Implements a method to change/update `Query` in both Firebase as well as Firestore Adapters. (Issue #1614)


PatilShreyas:
@samtstern There are some Auth lint errors exists that's why Trivis CI failed.

PatilShreyas:
@samtstern These are errors.
```
  The first 3 errors (out of 84) were:
  /home/travis/build/firebase/FirebaseUI-Android/auth/src/main/res/layout/fui_auth_method_picker_layout.xml:11: Error: This view is not constrained horizontally: at runtime it will jump to the left unless you add a horizontal constraint [MissingConstraints]
      <me.zhanghai.android.materialprogressbar.MaterialProgressBar
       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  /home/travis/build/firebase/FirebaseUI-Android/auth/src/main/res/layout-land/fui_auth_method_picker_layout.xml:12: Error: This view is not constrained horizontally: at runtime it will jump to the left unless you add a horizontal constraint [MissingConstraints]
      <me.zhanghai.android.materialprogressbar.MaterialProgressBar
       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  /home/travis/build/firebase/FirebaseUI-Android/auth/src/main/res/layout/fui_confirmation_code_layout.xml:9: Error: This view is not constrained. It only has designtime positions, so it will jump to (0,0) at runtime unless you add the constraints [MissingConstraints]
      <me.zhanghai.android.materialprogressbar.MaterialProgressBar
       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

```
  The first 3 errors (out of 4) were:
  /home/travis/build/firebase/FirebaseUI-Android/app/src/main/res/layout/auth_method_picker_custom_layout.xml:81: Error: This namespace declaration is redundant [RedundantNamespace]
      <com.firebase.ui.auth.util.ui.SupportVectorDrawablesButton xmlns:android="http://schemas.android.com/apk/res/android"
                                                                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  /home/travis/build/firebase/FirebaseUI-Android/app/src/main/res/layout-land/auth_method_picker_custom_layout.xml:98: Error: This namespace declaration is redundant [RedundantNamespace]
      <com.firebase.ui.auth.util.ui.SupportVectorDrawablesButton xmlns:android="http://schemas.android.com/apk/res/android"
                                                                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  /home/travis/build/firebase/FirebaseUI-Android/app/src/main/res/values/strings.xml:26: Error: Duplicate string value Google, used in logo_google, providers_google and tos_pp_google [DuplicateStrings]
      <string name="providers_google">Google</string>
      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      /home/travis/build/firebase/FirebaseUI-Android/app/src/main/res/values/strings.xml:46: Duplicates value in providers_google
      /home/travis/build/firebase/FirebaseUI-Android/app/src/main/res/values/strings.xml:50: Duplicates value in providers_google
```

database/src/main/java/com/firebase/ui/database/ObservableSnapshotArray.java
```diff
@@ -24,4 +25,11 @@
     public ObservableSnapshotArray(@NonNull SnapshotParser<T> parser) {
         super(new CachingSnapshotParser<>(parser));
     }
+
+    /**
+     * Method for changing/updating the {@link Query} in existing adapter.
+     *
+     * @param newQuery is a new updated query.
+     */
+    public abstract void updateQuery(@NonNull Query newQuery);
```
samtstern: Should this be added in `BaseObservableSnapshotArray` instead? It seems a bit strange that this would be the only method defined on `ObservableSnapshotArray`

database/src/main/java/com/firebase/ui/database/paging/DefaultSnapshotDiffCallback.java
```diff
@@ -25,6 +26,7 @@ public boolean areItemsTheSame(@NonNull DataSnapshot oldItem,
         return oldItem.getKey().equals(newItem.getKey());
     }
 
+    @SuppressLint("DiffUtilEquals")
```
samtstern: Why is this lint suppression added in this PR when nothing else about this file changed?

firestore/src/main/java/com/firebase/ui/firestore/ObservableSnapshotArray.java
```diff
@@ -18,4 +19,11 @@
     public ObservableSnapshotArray(@NonNull SnapshotParser<T> parser) {
         super(new CachingSnapshotParser<>(parser));
     }
+
+    /**
+     * Method for changing/updating the {@link Query} in existing adapter.
+     *
+     * @param newQuery is a new updated query.
+     */
+    public abstract void updateQuery(@NonNull Query newQuery);
```
samtstern: Same question here as on the `database` version.

buildSrc/build.gradle.kts
```diff
@@ -1,7 +1,21 @@
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
 repositories {
     jcenter()
 }
 
 plugins {
     `kotlin-dsl`
+    kotlin("jvm") version "1.3.41"
```
samtstern: Can you explain why we need all of these kotlin changes in this PR?  Are they related?

database/src/main/java/com/firebase/ui/database/ObservableSnapshotArray.java
PatilShreyas: Because `BaseObservableSnapshotArray` is inherited in both `ObservableSnapshotArray` of Firestore as well as Firebase realtime database too. And most importantly, the `Query` class of both databases are not the same.
- Firestore Query Class: `com.google.firebase.firestore.Query`
- Firebase Query Class: `com.google.firebase.database.Query`

database/src/main/java/com/firebase/ui/database/paging/DefaultSnapshotDiffCallback.java
PatilShreyas: Because it caused an error when I run the command `gradlew assembleDebug proguard-tests:build check`

buildSrc/build.gradle.kts
PatilShreyas: No. I just updated it as it's on my system.

database/src/main/java/com/firebase/ui/database/ObservableSnapshotArray.java
PatilShreyas: If we add this method in `BaseObservableSnapshotArray` then `Query` parameter will be different for them. What can we do?

firestore/src/main/java/com/firebase/ui/firestore/ObservableSnapshotArray.java
PatilShreyas: The reason is also the same as I explained above.

buildSrc/build.gradle.kts
samtstern: Ok let's revert these changes since they are not related to the feature.

buildSrc/src/main/kotlin/Config.kt
```diff
@@ -2,7 +2,7 @@ object Config {
     const val version = "5.0.1-SNAPSHOT"
     val submodules = listOf("auth", "auth-github", "common", "firestore", "database", "storage")
 
-    private const val kotlinVersion = "1.2.61"
+    private const val kotlinVersion = "1.3.31"
```
samtstern: Let's also revert these changes.

PatilShreyas: Okay. Fine!

database/src/main/java/com/firebase/ui/database/ObservableSnapshotArray.java
samtstern: I would say let's just not make this part of the base class at all, it can just be something implemented in each of the `Array` classes.

That will also allow us to give the indexed version a better name: `updateKeyQuery()`

database/src/main/java/com/firebase/ui/database/paging/DefaultSnapshotDiffCallback.java
samtstern: Ok I am fine with this, although I bet if we revert the unrelated changes in this PR, this will go away.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.1.1-all.zip
```
samtstern: Let's also revert this, this could be a big change and I want to do it separately (when I have time to verify the release process with gradle 5)

database/src/main/java/com/firebase/ui/database/paging/DefaultSnapshotDiffCallback.java
PatilShreyas: Oh! Okay.

database/src/main/java/com/firebase/ui/database/ObservableSnapshotArray.java
PatilShreyas: We have instantiated `FirestoreArray` by its superclass references in `Adapter` classes. So, Will it be fine them to be instantiated as `Array` classes?

samtstern: Ah ok I see the issue.  Let me think about it for a minute (I'll clone the branch and play around)

samtstern: Ok yeah I think that's the best solution, instantiate it as an `Array` but have the `getSnapshots()` method still return `ObservableSnapshotArray` so that people only call `updateQuery` on the `*Adapter` classes.

PatilShreyas: Fine 😃. Okay then 👍 .

PatilShreyas: I have a question. How could we know that when to call `updateQuery()` or `updateKeyQuery()`?
In `FirebaseRecyclerAdapter` if I create an instance of `FirebaseArray` then we can't instantiate `FirebaseIndexArray`. How to do it?

samtstern: Oh no that is a problem ... any ideas?  I don't know why I am so opposed to adding `updateQuery` to the `ObservabeSnapshotArray` classes ... it just feels wrong.  But maybe that is the best way.

PatilShreyas: I also thought a lot about it before implementation in `ObservableSnapshotArray` class. But I can't find anything helpful.

PatilShreyas: I had one idea. But that idea was not much good to implement. Means, it's not good practice to implement. I had idea to use `instanceof` to check whether it's instance of Firebase array or index array. We will check it in our method and by casting we will call array class method. That was the idea.

samtstern:
@PatilShreyas ok I did some thinking and experimenting.  First I found better words to explain myself :-) the reason I am not comfortable with `updateQuery` on `ObservableSnapshotArray` is because that class previous did not know that `Query` existed at all.

Second I think it may be better to structure this as `setOptions` where instead of just replacing the query you replace the whole options object you used to start the Adapter.  Here's an example of how that might look (for RTDB only):
https://github.com/PatilShreyas/FirebaseUI-Android/pull/2

What do you think?

PatilShreyas:
Right @samtstern!

PatilShreyas:
@samtstern I have merged your PR and make changes as you suggested. Also implemented in Firestore adapter.

samtstern:
@PatilShreyas sorry I missed your last comment!  Thanks for working with me on this, LGTM.

PatilShreyas:
Anytime! Thank you!

Issue #1658, opened 2019-07-20
EmailLink && custom AuthmethodPickerLayout crash: "No button found for auth provider: emailLink"


Entreco:
### Step 1: Describe your environment

  * Android device: OnePlus 6
  * Android OS version: 9
  * Google Play Services version: 15.4.22
  * Firebase/Play Services SDK version: com.google.firebase:firebase-auth:18.0.0
  * FirebaseUI version: 5.0
  
### Step 2: Describe the problem:

Using EmailLink sign in, in combination with a custom ui crashes the app.

#### Steps to reproduce:

  1. Enable EmailLink in Firebase console
  2. Enable EmailLink in AuthProvider: AuthUI.IdpConfig.EmailBuilder().enableEmailLinkSignIn().setActionCodeSettings(email).build(),
  3. Set Custom AuthMethodPicker layout: .setAuthMethodPickerLayout(layout)
  4. Start startActivityForResult(intent, )
  
#### Observed Results:

  * What happened?  
App crashes with `Caused by: java.lang.IllegalStateException: No button found for auth provider: emailLink`
  
#### Expected Results:

  * What did you expect to happen?
AuthPickerActivity should be launched with my custom email button
  
#### Relevant Code:

  ```
private val providers =
        listOf(            AuthUI.IdpConfig.EmailBuilder().enableEmailLinkSignIn().setRequireName(true).setActionCodeSettings(ActionCodeSettings.newBuilder()
.setAndroidPackageName("package.name", true, null)
            .setHandleCodeInApp(true)
            .setUrl("https://google.com") // This URL needs to be whitelisted
.build()).build()
        )

private val picker = AuthMethodPickerLayout.Builder(_layout).setEmailButtonId(R.id.btn).build()
  
val intent = AuthUI.createSignInIntentBuilder()
            .setAuthMethodPickerLayout(picker)
            .setAvailableProviders(providers)
            .build()

startActivityForResult(intent, 1234)
  ```

The reason this error occurs, is that when you call `.enableEmailLinkSignIn` the provider Id changes from "password" to "emailLink" -> hence when `populateIdpListCustomLayout` is called, there is no matching button

samtstern:
This has been released in version `5.1.0`

PR #1661, opened 2019-07-24
#1658: Fix for using custom Layout icm Email Link

Entreco:
Fix for #1658 .

After this is merged, it is possible to use the EmailLink setting in combination with a custom layout.

This is a work-around at best, because ideally, someone would spend some time to refactor the EmailLink settings. I think it is confusing, that by calling setEmailLinkEnabled(), the ID of the provider is changed (without) any notice.

Let me know your thoughts.


samtstern:
@Entreco this works great for me!  I just tested it.  Thanks for contributing to FirebaseUI.  I am happy to merge this now if you're ready for it.

Entreco:
Thanks for the prompt response! I'm ready if you are ;)

PR #1667, opened 2019-08-30
Version 5.1.0

samtstern:
Releasing version 5.1.0


samtstern:
cc @lsirac this will be the last 5.x version so we can move onto 6.x which will support AndroidX and (maybe?) remove Twitter

PR #1674, opened 2019-09-12
Follow system dark mode by default

SUPERCILEX:
This works for both battery mode and the dark mode toggle in API 29.


Ches1422:
[firebase-monetization-tips-master (1).zip](https://github.com/firebase/FirebaseUI-Android/files/4330336/firebase-monetization-tips-master.1.zip)

Issue #1676, opened 2019-09-13
6.0.0 missing FirestoreRecyclerOptions?


achuinard:
Welcome to FirebaseUI and thanks for submitting an issue!

Please take a look at [open issues](https://github.com/firebase/FirebaseUI-Android/issues?q=is%3Aopen+is%3Aissue), as well as [resolved issues](https://github.com/firebase/FirebaseUI-Android/issues?q=is%3Aissue+is%3Aclosed), to see if your issue is either already being addressed, or has been solved by someone else.

If not, please feel free to fill in the following info so we can help faster!

### Step 1: Are you in the right place?

  * For issues or feature requests related to the code **in this repository** file a GitHub issue.
  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/) tagged appropriately.
  * For general Firebase discussion, use the [firebase-talk google group](https://groups.google.com/forum/#!forum/firebase-talk)
  * For help troubleshooting your application that does not fall under one of the above categories, reach out to the [personalized Firebase support channel](https://firebase.google.com/support/contact/troubleshooting/)

### Step 2: Describe your environment

  * FirebaseUI version: 6.0.0
  
### Step 3: Describe the problem:

App will not compile, says FirestoreRecyclerOptions doesn't exist as a class.

#### Steps to reproduce:

Upgrade to 6.0.0 in your build.gradle.

#### Observed Results:

It doesn't compile.
  
#### Expected Results:

I wanted the app to compile.

willgurht:
I see the same, some files are missing from the downloaded .aar
<img width="510" alt="missingfiles" src="https://user-images.githubusercontent.com/2734652/64879892-42a32e00-d657-11e9-9c8d-36528ed8abe0.png">

samtstern:
@achuinard @willgurht thank you for reporting!  This is ... not good.  I will have to take a look.

samtstern:
OK I used [dex-member-list](https://github.com/JakeWharton/dex-member-list) to analyze the members of firestore 5.1.0 vs 6.0.0 and this is the diff:
http://www.mergely.com/NlVBQrgZ/

There are a lot of missing classes (`FirestoreArray`, `FirestoreRecyclerOptions`) but also some very suspicious missing methods like `FirestoreRecyclerAdapter.updateOptions` ... I have no idea what's going on here but it's definitely a real issue!

samtstern:
[...]

PR #1678, opened 2019-09-13
Preserve unused classes

samtstern:
See issue #1676

For version 6.0.0 we went from AGP 3.2.1 to AGP 3.5.0, looks like the R8 enablement in 3.4.0 got us:
https://developer.android.com/studio/build/shrink-code#configuration-files

Here's how Firestore's dex methods list looks after this change (left is 5.1.0, right is this commit);
http://www.mergely.com/W2XBd87e/

Looks correct to me.

### DIFF
--- a/auth/build.gradle.kts
+++ b/auth/build.gradle.kts
@@ -4,6 +4,7 @@ android {
     buildTypes {
         named("release").configure {
             postprocessing {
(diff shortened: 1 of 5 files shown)


samtstern:
Talked to @SUPERCILEX on chat, this is his suggestion so I have approval.

Issue #1677, opened 2019-09-13
Method to update Options in Paging Adapters


PatilShreyas:
@samtstern Recently we developed and released method `updateOptions()` to update options in Adapters. But this isn't available in Paging adapters yet. If we again work on this again, it'll be helpful to update options at runtime without changing adapters. Should we work on this again?

If we work on this, `final` modifier from  `LiveData<>` fields should be removed. Because, while changing options `LiveData<>` fields should again re-initialised.

Is it Okay?

samtstern:
@PatilShreyas I think this is a good idea but a few things:

1) We'll have to see if the underlying paging library supports this
2) I am going to be super busy until the Firebase Summit on September 26th so let's hold off on any feature development until after that.

Thanks for always contributing your thoughts and code to this project!

PatilShreyas:
Yeah sure! We will discuss on it later. Thank you!

mateusdias:
Dears, sorry ask this, but how can i change the query object and "restart" all process to bind recyclerview etc? 
Let me explain: If you have an activity and filter dialog. So... when user interact on filter dialog i factory one custom query to perform on firestore (after activity load default query and setup options, adapter, etc). 
So, how can I change the query, update options and notify adapter to "refresh the new results of the new query?"

PatilShreyas:
@mateusdias Which adapter you're using? Are you using `FirestoreRecyclerAdapter` or `FirestorePagingAdapter`?
Please let me know about this.

mateusdias:
@PatilShreyas Thanks for reply! 
I am using FirestorePagingAdapter. By the way, I implemented exactly as in your example tutorial in the medium, here: https://medium.com/firebase-developers/firestore-pagination-in-android-using-firebaseui-library-1d7fe1a75704

As I said, in this activity I have a filter that processes user search preferences and dynamically builds new queries. So every time the user applies the filter it's like I literally need to reload everything from scratch, starting a new pagination with a new query. What I found most interesting is that even trying "RE INSTANTIATE everything... Adapter, RecyclerView etc" doesn't work. The result is blank. To make sure the problem wasn't with queries, I simply forced a query object to see what happened and didn't bring up the results.

II tried to control the state of the adapter manually too, but it didn't work.

Its important clarify, i'm using RecyclerView from androidx library: import androidx.recyclerview.widget.RecyclerView;

PatilShreyas:
[...]

PR #1692, opened 2019-10-09
Method to update options at runtime [Paging Adapters]

PatilShreyas:
I have added option to change options at runtime in `FirestorePagingAdapter` class.
All `LiveData<>` objects should be re-initialised after changing options. So, I have removed their `final` modifier and also added new `init()` method which initializes `LiveData<>` objects from constructor and from `updateOptions` method.

Closes: #1677


samtstern:
@PatilShreyas LGTM thank you!

PatilShreyas:
@samtstern Anytime!

PR #1694, opened 2019-10-10
Method to Update Options at runtime [RTDB Paging Adapter]

PatilShreyas:
I have added method `updateOptions()` in RTDB Paging Adapter. It's exactly same version as Firestore Paging Adapter.

Closes: #1677


samtstern:
@PatilShreyas LGTM!

PatilShreyas:
Always! 😍

PR #1714, opened 2019-11-11
[TwitterKit Deprecation] Replace TwitterKit use with Generic IDP

lsirac:


samtstern:
This PR is a thing of pure beauty.  I am going to just leave it here until `6.1.0` is out so I don't get too confused but looking over it once I see no major issues.

One thing: should we maybe try and detect if the developer has the Twitter SDK installed and log a warning saying something like "Hey with FirebaseUI x.x.x you don't need that anymore"?

samtstern:
cc @SUPERCILEX who will like this/

lsirac:
Will fix tests tomorrow :)

SUPERCILEX:
Yaaaaasssss, this is so nice! Bring on the red:
image

Can't wait for the rest to fall!

auth/src/main/java/com/firebase/ui/auth/util/data/ProviderAvailability.java
```diff
@@ -8,8 +8,6 @@
             exists("com.firebase.ui.auth.data.remote.GitHubSignInHandler");
     public static final boolean IS_FACEBOOK_AVAILABLE =
             exists("com.facebook.login.LoginManager");
-    public static final boolean IS_TWITTER_AVAILABLE =
```
samtstern: So could we still try and detect this and then log a warning saying "by the way you don't need twitter kit anymore"?  Maybe we should also do the same for the `R.string.twitter_` config values we used to have devs put in.  Then we can maybe get more people to actually benefit from this change.

lsirac: Done (TwitterKit check only).

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -201,6 +201,10 @@ public static AuthUI getInstance() {
      */
     @NonNull
     public static AuthUI getInstance(@NonNull FirebaseApp app) {
+        if (ProviderAvailability.IS_TWITTER_AVAILABLE) {
+            Log.w(TAG, "You no longer need TwitterKit to sign in with Twitter.");
```
samtstern: Let's flesh this out.  "Beginning with FirebaseUI 6.2.0 you no longer need to include the TwitterKit SDK to sign in with Twitter.  Go to https://{whatever the release page link} is for more information"

PR #1719, opened 2019-11-19
Add support for GitHub sign-in via Generic IDP

lsirac:


auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -201,11 +202,14 @@ public static AuthUI getInstance() {
...
+                "sign in with %s. Go to %s for more information";
         if (ProviderAvailability.IS_TWITTER_AVAILABLE) {
-            String releaseUrl = "https://github.com/firebase/FirebaseUI-Android/releases/tag/6.2.0";
-            Log.w(TAG, String.format("Beginning with FirebaseUI 6.2.0 you no longer need to " +
-                    "include the TwitterKit SDK to sign in with Twitter. " +
-                    "Go to %s for more information", releaseUrl));
+            Log.w(TAG, String.format(devWarning, "the TwitterKit SDK", "Twitter", releaseUrl));
+        }
+        if (ProviderAvailability.IS_GITHUB_AVAILABLE) {
+            Log.w(TAG, String.format(devWarning, "auth-github", "GitHub", releaseUrl));
```
samtstern: instead of "auth-github" say "com.firebaseui:firebase-ui-auth-github"

lsirac:
> Approved with one small comment. This is great! Double-checking but this isn't breaking at all right?

Not breaking as far as I can tell :smile:

Issue #1775, opened 2020-04-07
Handle dynamic list of identity providers


robapaul88:
1. have a dynamic list of identity providers in place (by user rights for example)
  2. define a custom layout that adds support for all possible idps
  
#### Observed Results:

  * the non-supported ones will still have their buttons shown but not-working
  
#### Expected Results:

  * the non-supported ones should be hidden (simplest fix)
  
#### Relevant Code:

com.firebase.ui.auth.ui.idp.AuthMethodPickerActivity#populateIdpListCustomLayout should have this content (it also has some safer checks than the existing one):
  ```java

 Map<String, Integer> providerButtonIds = customLayout.getProvidersButton();
        for (IdpConfig idpConfig : providerConfigs) {
            final String providerId = providerOrEmailLinkProvider(idpConfig.getProviderId());

            Integer buttonResId = providerButtonIds.get(providerId);
            if (buttonResId == null) {
                throw new IllegalStateException("No button found for auth provider: " + idpConfig.getProviderId());
            }

            @IdRes int buttonId = buttonResId;
            View loginButton = findViewById(buttonId);
            handleSignInOperation(idpConfig, loginButton);
        }
        //hide custom layout buttons that don't have their identity provider set
        for (String providerBtnId : providerButtonIds.keySet()) {
            if (providerBtnId == null) continue;
            boolean hasProvider = false;
            for (IdpConfig idpConfig : providerConfigs) {
                if (providerBtnId.equals(idpConfig.getProviderId())) {
                    hasProvider = true;
                    break;
                }
            }
            if (!hasProvider) {
                Integer resId = providerButtonIds.get(providerBtnId);
                if (resId == null) continue;
                @IdRes int buttonId = resId;
                findViewById(buttonId).setVisibility(View.GONE);
            }
        }
  ```

samtstern:
@robapaul88 thanks for this and sorry for my slow response.  I think this is a good idea.  Would you like to send a Pull Request adding this functionality?  It seems like you've already written most of the code.

robapaul88:
Not necessarily. You can just paste it there as it is, it's not doing anything out of the ordinary, so I don't mind. Thanks!

samtstern:
This issue is fixed and released in version `6.3.0`.

PR #1799, opened 2020-06-22
Handle dynamic list of identity providers

robapaul88:
This will take care of also hiding buttons from the custom layout when adding them for all possible identity providers, but the actual list is dynamic so some won't be available.

Issue: https://github.com/firebase/FirebaseUI-Android/issues/1775

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
@@ -51,7 +51,6 @@ import com.google.firebase.FirebaseApp;
 import com.google.firebase.auth.EmailAuthProvider;
 import com.google.firebase.auth.FacebookAuthProvider;
 import com.google.firebase.auth.FirebaseAuth;
-import com.google.firebase.auth.GithubAuthProvider;
 import com.google.firebase.auth.GoogleAuthProvider;
 import com.google.firebase.auth.PhoneAuthProvider;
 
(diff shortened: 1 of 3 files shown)


robapaul88:
@googlebot I signed it!

robapaul88:
@googlebot I fixed it.

samtstern:
@robapaul88 thanks for this contribution!  BTW I changed the target branch to `version-6.3.0-dev` since that's where we're working at the moment.

auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
```diff
@@ -229,14 +228,32 @@ private void populateIdpListCustomLayout(List<IdpConfig> providerConfigs) {
...
             }
 
-            @IdRes int buttonId = providerButtonIds.get(providerId);
+            @IdRes int buttonId = buttonResId;
             View loginButton = findViewById(buttonId);
             handleSignInOperation(idpConfig, loginButton);
         }
+        //hide custom layout buttons that don't have their identity provider set
+        for (String providerBtnId : providerButtonIds.keySet()) {
+            if (providerBtnId == null) continue;
```
samtstern, line 242: nit: we try not to use single-line if statements
```suggestion
            if (providerBtnId == null) {
                continue;
            }
```

auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
```diff
@@ -229,14 +228,32 @@ private void populateIdpListCustomLayout(List<IdpConfig> providerConfigs) {
...
+            boolean hasProvider = false;
+            for (IdpConfig idpConfig : providerConfigs) {
+                if (providerBtnId.equals(idpConfig.getProviderId())) {
+                    hasProvider = true;
+                    break;
+                }
+            }
+            if (!hasProvider) {
+                Integer resId = providerButtonIds.get(providerBtnId);
+                if (resId == null) continue;
```
samtstern, line 252:
```suggestion
                if (resId == null) {
                    continue;
                }
```

samtstern, line 252: Is this null condition even possible?  How would we get into this state?

robapaul88: Well, it's an unboxing, In theory I know that map should only have provider ids and layout res ids. But since that Map is passed around by reference, placing an Integer in it that's later nullified, you can end up with a null here. Better safe than sorry.

robapaul88: Done.

auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
robapaul88: Done.

robapaul88:
Created https://github.com/firebase/FirebaseUI-Android/pull/1800 to have this pushed to the right branch, redirecting this from master seems to also affect some other files (probably reverting some other commits from 6.3.0-dev).

PR #1800, opened 2020-06-23
Handle dynamic list of identity providers

robapaul88:
This will take care of also hiding buttons from the custom layout when adding them for all possible identity providers, but the actual list is dynamic so some won't be available.

Issue: #1775


samtstern:
@robapaul88 looks great, thank you for your contribution!  Once this is merged it should be available in the `6.3.0-SNAPSHOT` builds within about 30 minutes.

robapaul88:
Hi @samtstern . Any idea why https://oss.jfrog.org/webapp/#/artifacts/browse/tree/General/oss-snapshot-local/com/firebaseui/firebase-ui-auth/6.3.0-SNAPSHOT is not getting updated?

samtstern:
@robapaul88 no idea, the build job ran successfully:
https://github.com/firebase/FirebaseUI-Android/runs/799349580?check_suite_focus=true

But looks like it hasn't updated in 90 days!

samtstern:
Hmm it looks like `firebase-ui-common` is deploying to artifactory but the others are not ...

Issue #1825, opened 2020-08-23
Feature request: ability to always skip provider choice


ubragg:
I'm interested in using FirebaseUI for all of its many great features *except* that I'd like to not use the provider selector UI and instead trigger the experience using my own buttons for each provider on my own login screen, as my login screen has other functionality that can't be expressed in XML alone (as far as I have determined).

I was able to do this by wiring up each of my Google, Apple, and Facebook buttons to the following code:
```
   activity.startActivityForResult(
        AuthUI.getInstance()
            .createSignInIntentBuilder()
            .setAvailableProviders(Collections.singletonList(new AuthUI.IdpConfig.XXXXXBuilder().build()))
            .build(),
        requestId);
```

Because I only ever call this with a single provider, and set the (singleton) providers list separately for each button, it always skips the AuthMethodPickerActivity, and this works perfectly in the common case.

The problem is that when a user that has an account with Google tries to sign in with Facebook, and they run into the issue described in #301, rather than gracefully resolving it by having them sign in with Google, in my case, it fails with a developer error, because a given flow can't fall back on the other providers unless they're in the list that started the flow.

I have investigated many ways to address this, and am still trying to figure out a way to work around it, but it seems the easiest fix for this to be supported at the library level would be one of the following:
- Add a `setNeverShowSignInMethodScreen()` option akin to `setAlwaysShowSignInMethodScreen()` that simply uses the first one in the list to start, and falls back to the others only if necessary.
- Allow the ability to launch the flow with a full list of providers, but with one specified as the default to start with to bypass the select screen.

I'm even happy to submit a pull request if any of these suggestions would have a chance to be accepted.

As for short term fixes, I'd love suggestions for that too.  I have tried to override `FlowParameters` (which in the current state of things would simply require me to be able to override `shouldShowProviderChoice()` to always return true), but to get there I'd need to also override `AuthUI.SignInIntentBuilder`, which I can't seem to do because it has a private constructor.  I've even tried making a custom UI that's just a transparent or themed activity with invisible buttons that I click programmatically when it starts, but that has some issues when the user backs out of the flow, since they get to a blank page, plus it's quite a bit more hacky than I'd like.

samtstern:
@ubragg thanks for the feedback and feature request.  I think this would be best expressed as `setDefaultProvider(...)` to let us know to skip the `AuthMethodPickerActivity` and go straight into the flow for a certain provider.

ubragg:
[...]

PR #1829, opened 2020-09-02
add option to specify a default provider to skip the selection screen

ubragg:
Here is my attempt to address #1825 as discussed.

New tests added and all tests pass.

I don't believe this feature is necessary in the iOS version of the library, as you can already get this behavior due to the fact that the providers are not specified as part of the flow.


samtstern:
@ubragg wow this is a very thorough PR, thank you for thinking of the edge cases and adding test coverage!  Merging.

Issue #1737, opened 2020-02-14
isNewUser always return for false for Microsoft and Twitter Sign In


ian-sayles:
Android 
### Step 2: Describe your environment

  * Android device: Sony ZX3
  * Android OS version: 9
  * Google Play Services version: 20.1.04
  * Firebase/Play Services SDK version: 
        com.google.firebase:firebase-auth:19.2.0
        com.google.gms:google-services:4.3.3

  * FirebaseUI version: com.google.firebase:firebase-auth:19.2.0
  
### Step 3: Describe the problem:
After a new user signed into my app the "isNewUser"  is only true when they have signed in using email or Google providers, both Microsoft and Twitter it is always false.  
#### Steps to reproduce:
 
I have tried this on Samsung phone run from Android Studio and from an app submitted into Play Store on my Sony mobile device with the same result for both devices.

#### Relevant Code:

Start Sign In:

          startActivityForResult(
                AuthUI.getInstance()
                    .createSignInIntentBuilder()
                    .setAvailableProviders(
                        listOf(
                            GoogleBuilder().build(),
                            MicrosoftBuilder().build(),
                            EmailBuilder().build(),
                            TwitterBuilder().build()
                        )
                    )
                    .build(),
                SIGNIN
            )

 override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
        super.onActivityResult(requestCode, resultCode, data)

        if (requestCode == SIGNIN) {

            val response = IdpResponse.fromResultIntent(data)

            if (resultCode == Activity.RESULT_OK) {

                if (response == null) {
                    finish()
                    return
                }

                if (response.isNewUser) {

                    Toast.makeText(
                        this,
                        "You are a new user",
                        Toast.LENGTH_LONG
                    ).show()

                    registerNewUser()

                } else {

                    Toast.makeText(
                        this,
                        "You are an existing user",
                        Toast.LENGTH_LONG
                    ).show()

                    checkAdmin()
                }

samtstern:
This has been released in version `6.4.0`

PR #1832, opened 2020-09-07
Fixing isNewUser

laurentiu-git:
Hey there!
I want to share my fix for the issue #1737 , I was able to get the expected result for `isNewUser` when user signs in with Twitter, but I was not able to sign in with Microsoft, I'm not sure if it's because of my setup it's wrong or it's something else.


laurentiu-git:
@googlebot I signed it!

samtstern:
@laurentiu-git thanks for this!  Looks like a simple and safe change.

Issue #1834, opened 2020-09-14
Orientation lock on phones


dimipaun:
There are many times when you'd like to lock the activities to portrait on the phone, but not on a tablet. For ex:
https://stackoverflow.com/questions/50596476/disable-landscape-orientation-for-phones-but-allow-for-tablets/50596650

This has to be done through code it seems.

Would a flag controlling this behavior be acceptable? E.g. `setLockOrientationToPortraitOnSmallDevices(true)`

samtstern:
@dimipaun thanks for the feedback.  I'm open to something like this but I think it would just be `setLockOrientation(...)` and I'd leave it up to the developer to do conditional logic to only set that on small devices.  What do you think?

samtstern:
This has been released in version `6.4.0`

PR #1839, opened 2020-09-26
Added setLockOrientation for activities

laurentiu-git:
Hello, I have tried to implement the feature requested #1834 .


auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
```diff
@@ -101,6 +103,9 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
         mHandler = ViewModelProviders.of(this).get(SocialProviderResponseHandler.class);
         mHandler.init(params);
 
+       if (params.lockOrientation) {
```
samtstern, line 106: Can this be moved into `AppCompatBase` so that we don't have to repeat it in each subclass?

auth/src/main/res/values-sw480dp/bools.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
```
samtstern, line 1: I think adding a boolean resource to FirebaseUI for this is a step too far.  The developer should add this in their own app and then tell us to lock the orientation accordingly.

auth/src/main/java/com/firebase/ui/auth/ui/HelperActivityBase.java
```diff
@@ -58,6 +61,13 @@ public FlowParameters getFlowParams() {
         return mParams;
     }
 
+    @SuppressLint("all")
+    public void lockOrientation() {
+        if (getResources().getBoolean(R.bool.fui_portrait_orientation)) {
```
samtstern, line 66: We can remove the conditional here.  If the developer requests this through the `AuthUI` builder then we should always do it.

samtstern (commented):
Thank you for this!  Mostly it looks great, just a few comments.

auth/src/main/res/values-sw480dp/bools.xml
laurentiu-git: I understood the feature wrong, sorry for the inconvenience.
I hope it's alright now, thanks for the guidance.

auth/src/main/java/com/firebase/ui/auth/ui/AppCompatBase.java
```diff
@@ -53,4 +59,9 @@ protected void switchFragment(@NonNull Fragment fragment,
     protected void switchFragment(@NonNull Fragment fragment, int fragmentId, @NonNull String tag) {
         switchFragment(fragment, fragmentId, tag, false, false);
     }
+
+    @SuppressLint("all")
```
samtstern, line 63: Do we actually need this `@SuppressLint` anymore?  If so can we be more specific than `all`?

auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
```diff
@@ -14,8 +14,10 @@
 
 package com.firebase.ui.auth.ui.idp;
 
+import android.annotation.SuppressLint;
```
samtstern, line 17: There are no changes in this file anymore besides these two imports and some whitespace, so maybe we can revert the whole file?

samtstern (approved):
Looks great, thank you!  Just two small comments and then I will merge.

Oh and please add a line to `CHANGELOG.md` describing your change.

auth/src/main/java/com/firebase/ui/auth/ui/AppCompatBase.java
samtstern: So then could it be `@SuppressLint("SourceLockedOrientationActivity")`?

auth/src/main/java/com/firebase/ui/auth/ui/idp/AuthMethodPickerActivity.java
```diff
@@ -1,412 +0,0 @@
-/*
```
samtstern, line 1: I think you accidentally deleted the whole file!

samtstern:
@laurentiu-git LGTM!  If tests pass I am going to merge.

PR #1840, opened 2020-09-28
Fix to allow bolding of text that coincides with the end of body.

ElectricMagic:
`textToBold` that is a suffix of `text` would be rejected by the `end>=text.length` test.
`end > text.length()` is the correct indication that `text.indexof` gave us a bogus offset.


samtstern:
@ElectricMagic thanks for the PR!  Merged into the `6.4.0` branch

Issue #1746, opened 2020-03-12
Customize Password Reset Link


kmcnellis:
Firebase UI should allow devs to customize their password reset links, to allow them to create a custom handler using [email action links in the Admin SDK](https://firebase.google.com/docs/auth/admin/email-action-links).

This will allow them to customize the emails sent to give custom instructions and have branded styling.

We should add a builder setting for `.setPasswordResetLink(...)` or something similar

kmcnellis:
https://github.com/firebase/FirebaseUI-iOS/issues/843

samtstern:
This has been released in version `6.4.0`

PR #1850, opened 2020-10-30
Implemented custom password reset link

laurentiu-git:
Hey there!

I gave it a try to implement the request #1746. 
I'm not exactly sure if this is the feature expected.

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/ui/email/RecoverPasswordActivity.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/email/RecoverPasswordActivity.java
@@ -118,8 +118,19 @@ public class RecoverPasswordActivity extends AppCompatBase implements View.OnCli
     @Override
     public void onDonePressed() {
         if (mEmailFieldValidator.validate(mEmailEditText.getText())) {
-            mHandler.startReset(mEmailEditText.getText().toString());
+            if (getFlowParams().passwordResetSettings != null)
+                customResetPassword();
+            else
+                normalResetPassword();
         }
+    }
+
+    private void normalResetPassword() {
+        mHandler.startReset(mEmailEditText.getText().toString());
+    }
+
+    private void customResetPassword() {
+        mHandler.startCustomReset(mEmailEditText.getText().toString(), getFlowParams().passwordResetSettings);
     }
 
     private void showEmailSentDialog(String email) {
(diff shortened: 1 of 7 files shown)


auth/src/main/java/com/firebase/ui/auth/ui/email/RecoverPasswordActivity.java
```diff
@@ -118,10 +118,21 @@ public void onClick(View view) {
     @Override
     public void onDonePressed() {
         if (mEmailFieldValidator.validate(mEmailEditText.getText())) {
-            mHandler.startReset(mEmailEditText.getText().toString());
+            if (getFlowParams().passwordResetSettings != null)
```
samtstern, line 121: Just a style nit: please use `{` and `}` for all blocks, even one-line blocks.  This is more consistent with the rest of the codebase and reduces bugs.

auth/src/main/java/com/firebase/ui/auth/viewmodel/email/RecoverPasswordHandler.java
```diff
@@ -29,4 +30,18 @@ public void onComplete(@NonNull Task<Void> task) {
                     }
                 });
     }
+
+    public void startCustomReset(@NonNull final String email, @NonNull ActionCodeSettings actionCodeSettings) {
```
samtstern, line 34: This method is almost exactly the same as `startReset` could it be combined?  We could make the `actoinCodeSettings` argument `@Nullable` and then do something like this:

```java
Task<Void> reset = actionCodeSettings != null 
  ? getAuth().sendPasswordResetEmail(email, actionCodeSettings)
  : getAuth().sendPasswordResetEmail(email)

reset.addOnCompleteListener(...
```

samtstern (approved):
I only have some very small comments about style but otherwise this LGTM!  Thank you so much for contributing this feature.

auth/src/main/java/com/firebase/ui/auth/ui/email/RecoverPasswordActivity.java
laurentiu-git: Should I keep the functions `normalResetPassword` and `customResetPassword` since now there is only one function  `startReset` ?

samtstern: Good point, I think we can just have one function there too.

samtstern (approved):

samtstern:
@laurentiu-git thank you for this!

P.S. The reason I call my branches `ss-fix-1234` is because my initials are `ss` (Sam Stern).  So you can call yours `ld-fix-1234` if you want.

laurentiu-git:
> @laurentiu-git thank you for this!
> 
> P.S. The reason I call my branches `ss-fix-1234` is because my initials are `ss` (Sam Stern). So you can call yours `ld-fix-1234` if you want.

Oh, this makes more sense now 😄, I thought this was some internal thing going on.
Thank you!

Issue #1670, opened 2019-09-02
[Feature Request] Allow a default email address to be specified in AuthUI.IdpConfig.EmailBuilder()


avoirtech:
It can often be the case that the calling code will have a good idea what email address the user is going to use to sign in with (for example, when resuming after sending a verification email).

It would make things smoother for the end user if we could have a method to pre-populate the email address field in the email sign in page. Something like:
`AuthUI.IdpConfig.EmailBuilder().setEmail("someone@example.com").build()`

samtstern:
@avoirtech thanks for this!  Sounds like a nice improvement.

danielfaust:
Is there a way to get the password the user used to sign up? Because it is needed since `FirebaseAuth.getInstance().signInWithEmailAndPassword(strEmail, strPassword)` requires it, and FirebaseUI doesn't provide this functionality.

Adding the feature of this issue would be more than just a nice improvement.

samtstern:
@danielfaust no there is currently no way to get the password the user signed up with.  Can you file a new issue explaining what you're doing that would require this?

samtstern:
This was released in version `7.1.0`

PR #1866, opened 2020-11-17
Added setDefaultEmail for EmailBuilder

laurentiu-git:
Hey @samtstern !

I have implemented the feature for #1670 .

### DIFF
--- a/.opensource/project.json
+++ b/.opensource/project.json
@@ -18,7 +18,8 @@
         "docs/upgrade-to-3.0.md": "Upgrade to v3.0",
         "docs/upgrade-to-4.0.md": "Upgrade to v4.0",
         "docs/upgrade-to-5.0.md": "Upgrade to v5.0",
-        "docs/upgrade-to-6.0.md": "Upgrade to v6.0"
+        "docs/upgrade-to-6.0.md": "Upgrade to v6.0",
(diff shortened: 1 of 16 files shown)


samtstern (commented):
Two small changes:

1. Is there a test you could add for this to make sure that changing this setting results in the right flow?
2. Could you add an entry to `CHANGELOG.md`?

laurentiu-git:
I will think about a test, and about `CHANGELOG.md`  it came in my mind after I did the pull request :)

laurentiu-git:
Besides checking in `AuthUITest` that the value is stored inside `ExtraConstants.DEFAULT_EMAIL`, should I think of another way to test the flow? Perhaps inside the `EmailActivityTest` ?

samtstern:
@laurentiu-git the ideal test would be one that checks the effect of the value, not just that the value is set.  So for example checking when the value is set and the user exists, the flow heads to the password screen not the email screen.

samtstern:
@laurentiu-git these tests look great, thank you!  I'm going to resolve the merge conflicts (hope you don't mind) and then merge when CI is green.

laurentiu-git:
While you are at it, I saw this test `testOnTroubleSigningIn_expectTroubleSigningInFragment`  being commented out and it appears that now it's working.

Issue #1296, opened 2018-05-09
Reconsider an RTDB paging adapter


samtstern:
Based on the compromises we made for Firestore, I think we could make an RTDB paging adapter with equivalent functionality.  Previous discussions focused on the difficulty of managing realtime listeners but that's not a concern at the moment.

(Related #17)

rosariopfernandes:
Is this issue still relevant? It seems like our RTDB paging adapter already has the same functionality as the firestore one.

samtstern:
@rosariopfernandes good call!

rosariopfernandes:
@samtstern However, we could replace the current `singleValueEventListener` with the new `Task<DataSnapshot> get()` to avoid getting stale results from cache as [puf explains here](https://stackoverflow.com/a/66072472/5861618). I believe that would make it more consistent with Firestore?

samtstern:
@rosariopfernandes I think that's a good idea!

PR #1912, opened 2021-02-12
refactor(database): replace singleValueEventListener with Query.get()

rosariopfernandes:
As discussed in https://github.com/firebase/FirebaseUI-Android/issues/1296#issuecomment-778131774

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ google-services.json
 
 crashlytics-build.properties
 auth/src/main/res/values/com_crashlytics_export_strings.xml
+*.log
--- a/.opensource/project.json
+++ b/.opensource/project.json
@@ -18,7 +18,8 @@
         "docs/upgrade-to-3.0.md": "Upgrade to v3.0",
         "docs/upgrade-to-4.0.md": "Upgrade to v4.0",
         "docs/upgrade-to-5.0.md": "Upgrade to v5.0",
-        "docs/upgrade-to-6.0.md": "Upgrade to v6.0"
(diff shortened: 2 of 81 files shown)


samtstern (approved):

samtstern:
@rosariopfernandes very nice! Once the build is fixed I am happy to merge.  Although we should merge this into a dev branch.

samtstern:
@rosariopfernandes I changed this to target `version-7.2.0-dev`

Also cc @puf who will be happy to see this.

puf:
*Very* cool to see this. I somehow expected bigger code gains from this, so thanks for proving that I have rose colored glasses on when it comes to this change. But even without LOC gains, this means we get rid of a pesky race condition.

I indeed recommend holding off for a bit as we may have an issue still with `get()` on `Query` objects (as opposed to `DatabaseReference`). But the fix for that should be out Real Soon Now.

rosariopfernandes:
@puf FWIW there is a LOC gain in Kotlin as this:

```kotlin
postReference.addListenerForSingleValueEvent(object : ValueEventListener {
    override fun onDataChange(dataSnapshot: DataSnapshot) {
        // ...
    }

    override fun onCancelled(databaseError: DatabaseError) {
        // ...
    }
})
```

becomes this:
```kotlin
postReference.get().addOnSuccessListener {
    // ...
}.addOnFailureListener{
    // ...
}

```

---

I didn't know about that issue. Sure, we can hold this until the fix is out.

samtstern:
I assume the fix we were waiting for is out by now :-)

Issue #1898, opened 2021-01-05
Country code missing


markuval:
Welcome to FirebaseUI and thanks for submitting an issue!

Please take a look at [open issues](https://github.com/firebase/FirebaseUI-Android/issues?q=is%3Aopen+is%3Aissue), as well as [resolved issues](https://github.com/firebase/FirebaseUI-Android/issues?q=is%3Aissue+is%3Aclosed), to see if your issue is either already being addressed, or has been solved by someone else.

If not, please feel free to fill in the following info so we can help faster!

### Step 1: Are you in the right place?

 I believe so!

### Step 2: Describe your environment

  * Android device: Samsung
  * Android OS version: 9, 10
  * Google Play Services version:Latest Verison
  * Firebase/Play Services SDK version: 26.2.0
  * FirebaseUI version: 6.3.0
  
### Step 3: Describe the problem:

My team uses Firebase for user app verification with SMS.  Our app is a taxi booking app when users can open an account with their phone number and order taxis. Kosovo country code which is +383 we can't find on the list. 

#### Steps to reproduce:

  1. I open my app
  2. Go to country list for verification
  3. Country code +383 is not on the list
  
#### Observed Results:

  * Users can't register from that country if that country code is not on the list.
  
#### Expected Results:

  * We expect this country code to be added somehow in this list or is another way that we can do it?
  
#### Relevant Code:

map.put(380, singletonList("UA"));
        map.put(381, singletonList("RS"));
        map.put(382, singletonList("ME"));
        map.put(385, singletonList("HR"));
        map.put(386, singletonList("SI"));

---There is no +383 ('XK')--
  ```
  // TODO(you): code here to reproduce the problem
  ```

samtstern:
@markuval thank you!  This seems like a bug, we'll get Kosovo added in `7.1.2`

samtstern:
This was fixed and released in version `7.2.0`

PR #1913, opened 2021-02-13
Added Kosovo prefix

laurentiu-git:
Hello, I have added the Kosovo prefix code( #1898 ). Before, it was added as Serbia prefix code.


samtstern:
@laurentiu-git thank you!

PR #1918, opened 2021-03-03
AuthUI First-party support for Activity Result APIs

### LINKED ISSUE #1916: [Feature Request] Support Activity Result APIs for AuthUI
The Android Developers documentation [recommends using the Activity Result APIs](https://developer.android.com/training/basics/intents/result) instead of relying on intents for launching activities for results. Using the new APIs would enable a nicer development experience. Take the following as an example:

```kotlin
class FirebaseAuthUIActivityResult: ActivityResultContract<AuthUI.SignInIntentBuilder, IdpResponse>() { ... }
```

This API could be used like this:

```kotlin
val firebaseLogin = registerForActivityResult(FirebaseAuthUIActivityResult()) { result ->
    if (result.resultCode == Activity.RESULT_OK) {
        ...
    }
}
firebaseLogin.launch(
    AuthUI.getInstance().createSignInIntentBuilder()
        .setAvailableProviders(authenticationMethods)
        .setTheme(R.style.SignInTheme)
        .setAuthMethodPickerLayout(signInLayout)
)
```

By using a contract with types, this is safer for the end-user. More importantly, it enables future usage from Jetpack Compose.
I'd be happy to contribute here if we can scope it a bit :)

Let me know what you think!

jossiwolf:
See #1916.

@samtstern, there are is a looot of internal code that could/should be migrated. Do you want that as part of this PR? It'd probably take a bit but it's up to you!

I'm guessing additional work in the snippets repo is needed, right?

I haven't written Java in quiiiite a bit so any feedback on Java code style is appreciated!

Cheers!

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUIActivityResultContract.java
+++ b/auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUIActivityResultContract.java
@@ -26,6 +26,7 @@ public class FirebaseAuthUIActivityResultContract extends
     }
 
     @Override
+    @NonNull
     public FirebaseAuthUIAuthenticationResult parseResult(int resultCode, @Nullable Intent intent) {
         return new FirebaseAuthUIAuthenticationResult(resultCode, IdpResponse.fromResultIntent(intent));
     }
--- a/auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUIActivityResultContract.java
+++ b/auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUIActivityResultContract.java
@@ -10,17 +10,19 @@ import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
 /**
- * A {@link ActivityResultContract} describing that the caller can launch authentication with a
- * {@link com.firebase.ui.auth.AuthUI.SignInIntentBuilder} and is guaranteed to receive a
- * {@link FirebaseAuthUIAuthenticationResult} as result.
+ * A {@link ActivityResultContract} describing that the caller can launch authentication flow with a
+ * {@link Intent} and is guaranteed to receive a {@link FirebaseAuthUIAuthenticationResult} as
+ * result. The given input intent <b>must</b> be created using a
+ * {@link com.firebase.ui.auth.AuthUI.SignInIntentBuilder} in order to guarantee a successful
+ * launch of the authentication flow.
  */
 public class FirebaseAuthUIActivityResultContract extends
-        ActivityResultContract<AuthUI.SignInIntentBuilder, FirebaseAuthUIAuthenticationResult> {
+        ActivityResultContract<Intent, FirebaseAuthUIAuthenticationResult> {
 
     @NonNull
     @Override
-    public Intent createIntent(@NonNull Context context, AuthUI.SignInIntentBuilder input) {
-        return input.build();
+    public Intent createIntent(@NonNull Context context, Intent input) {
+        return input;
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUIActivityResultContract.java
```diff
@@ -0,0 +1,31 @@
...
+ * {@link com.firebase.ui.auth.AuthUI.SignInIntentBuilder} and is guaranteed to receive a
+ * {@link FirebaseAuthUIAuthenticationResult} as result.
+ */
+public class FirebaseAuthUIActivityResultContract extends
+        ActivityResultContract<AuthUI.SignInIntentBuilder, FirebaseAuthUIAuthenticationResult> {
+
+    @NonNull
+    @Override
+    public Intent createIntent(@NonNull Context context, AuthUI.SignInIntentBuilder input) {
+        return input.build();
```
rosariopfernandes, line 23: I'm curious about why we're passing the `IntentBuilder` to the `Contract` instead of the actual `Intent`?

rosariopfernandes:
@jossiwolf Our GH Actions workflow is failing due to lint errors. You can run `gradlew checkstyle` locally to detect those before pushing a commit.

auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUIActivityResultContract.java
jossiwolf: Thanks for highlighting this, I wanted to come back to it! I actually don't really like this as input type here. It would be great to use an immutable data structure (`FlowParams`?) instead. I prefer that over an intent because it represents the input more accurately and is more intuitive. Say you're looking at how to use this the first time, an intent can contain anything so it's not necessarily clear how to construct it/what it should contain.

What are your thoughts on that? Do you have any model in mind that we could re-use as input type here? Thanks for looking at it!:)

jossiwolf:
I swear checkstyle ran on my machine🤔😂 Thanks I'll upload a patch later!

auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUIActivityResultContract.java
samtstern: Thanks @rosariopfernandes for flagging this! @jossiwolf I totally get where you're coming from, `Intent` allows too much room for error.

Unfortunately right now we don't have a good internal type to cleanly represent the state of the `IntentBuilder` ... there was never much use for that before this PR.

I think for `v7.2.0` we should take `Intent` as the param here but then in `v8.0.0` we can look at refactoring this to take an immutable input type representing all of the options.

samtstern (approved):
I love it! Thanks @jossiwolf.  Just see my comments on the one thread about the `IntentBuilder` thing and let me know what you you think.

Also as far as I can tell we can launch this in `7.2.0` right?

auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUIActivityResultContract.java
jossiwolf: Cool, sounds good! This would be great to have in any case I think (I can see this coming in handy for Compose support) so that work could be done separately too.

I'll refactor this to accept an Intent as input, in that case the samples change a bit less too which is great!

jossiwolf:
@samtstern Lovely, thanks for the quick review! Yup this should be ready to be launched with `7.2.0` :) There might be some additional work needed in the snippets repo to make sure the docs are up to date then, right? How do you want to go about that?

samtstern:
@jossiwolf I just changed this to target the `7.2.0-dev` branch. As soon as you rebase or merge in the base then we can merge this!

As for snippets we can handle that when `7.2.0` is actually released. No need to worry about it for now.

jossiwolf:
Oh perfect, I'll rebase that once CI went through successfully!

PR #1921, opened 2021-03-05
Jfk material fixes

joknu1:
Hi @samtstern,

I've looked into some issues I had with the ss-material-components branch. This is my first PR here, so feel free to give harsh feedback. E.g. Travis did not pass, since I don't have a json for google-services.

Cheers,
Jonas

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/util/ui/PreambleHandler.java
+++ b/auth/src/main/java/com/firebase/ui/auth/util/ui/PreambleHandler.java
@@ -5,14 +5,13 @@ import android.net.Uri;
 import android.text.SpannableStringBuilder;
 import android.text.TextUtils;
 import android.text.method.LinkMovementMethod;
-import android.text.style.ForegroundColorSpan;
 import android.text.style.URLSpan;
-import android.util.TypedValue;
 import android.view.View;
 import android.widget.TextView;
 
 import com.firebase.ui.auth.R;
 import com.firebase.ui.auth.data.model.FlowParameters;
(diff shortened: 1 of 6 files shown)


auth/src/main/java/com/firebase/ui/auth/util/ui/PreambleHandler.java
```diff
@@ -33,16 +32,13 @@
...
     private final int mButtonText;
-    private final ForegroundColorSpan mLinkSpan;
 
     private SpannableStringBuilder mBuilder;
 
     private PreambleHandler(Context context, FlowParameters parameters, @StringRes int buttonText) {
         mContext = context;
         mFlowParameters = parameters;
         mButtonText = buttonText;
-        mLinkSpan = new ForegroundColorSpan(ContextCompat.getColor(mContext,
```
samtstern, line 44: What happened to this span?

samtstern:
@joknu1 thank you for this contribution! Happy to have someone new onboard.

auth/src/main/java/com/firebase/ui/auth/util/ui/PreambleHandler.java
joknu1: To my understanding - this span was to get the correct color on the URL links. This seem to be handled by `URLSpan` on it's own - given that `<item name="android:textColorLink">@color/fui_linkColor</item>` is set in the theme.

samtstern: 👍 yep!

Issue #1907, opened 2021-01-28
Progress Indication - Replace third party library with material components LinearProgressIndicator


postfixNotation:
* FirebaseUI version: 7.1.1
  
With the most recent releases of Material Components for Android there's a [ProgressIndicator](https://material.io/components/progress-indicators/android) component.

In my opinion this component should be used to replace [me.zhanghai.android.materialprogressbar:library:1.6.1](https://github.com/firebase/FirebaseUI-Android/blob/master/buildSrc/src/main/kotlin/Config.kt) which is used in [the auth module](https://github.com/firebase/FirebaseUI-Android/blob/master/auth/build.gradle.kts).

The intention of this issue is to reduce the number of unofficial third party libs.

samtstern:
@postfixNotation thanks for pointing this out!  We're looking at moving to all Material Components as part of an `8.0.0` release (it's a breaking change because of how themes work) so I'll add this work to that milestone.

samtstern:
This has been implemented and released in version `8.0.0`

PR #1924, opened 2021-03-12
Jfk additional material migrations

joknu1:
A continuation of the migration to Material Components

Material lib updated, third party lib for progress bar removed (#1907), and Dialogs updated.

I've tried this head on various brands and android version from 5 to 10 (I have no android 4), and I can no longer find any theme related problems that I can't override in my own extension of FirebaseUI theme.

I'm unsure about my `Merge pull request`commit in this PR, maybe that should be removed?

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -32,7 +32,7 @@ android {
 }
 
 dependencies {
-    implementation(Config.Libs.Androidx.design)
+    implementation(Config.Libs.Androidx.materialDesign)
     implementation(Config.Libs.Androidx.multidex)
 
     implementation(project(":auth"))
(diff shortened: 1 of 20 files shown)


rosariopfernandes (approved):
@joknu1 This looks great! 😃 



> I'm unsure about my Merge pull requestcommit in this PR, maybe that should be removed?

Don't worry about it, commits are usually squashed when a PR is merged into the Firebase repos.

samtstern:
@joknu1 thank you, this is great stuff! I'm going to merge this, what other changes do you think we need to make to complete the transition to Material?

joknu1:
Haha @samtstern, I was going to ask you the same question!

I think maybe the theme related `readMe` file need to be reviewed (not completely sure though). Also test that the example themes from the demo app work as expected on all supported android versions. Even though Material should be backwards compatible, there's a lot of unexpected behavior on some API versions. However, I could not find any unsolvable problems for my app, that uses Firebase UI for Google/Facebook/Email login on Android 5 and higher.

Let me know what else I can do to make material branch ready for merge!

samtstern:
@joknu1 thanks!  I think you're right that the README and the Sample App themes are probably up next. Even though it should be backwards compatible we'll want to make sure we document the latest.

I'll have to find some time next week to do more testing of my own

PR #1930, opened 2021-04-07
refactor(app): replace ButterKnife with View Binding

rosariopfernandes:
This should replace the deprecated ButterKnife library with View Binding in our sample app.

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/viewmodel/AuthViewModelBase.java
+++ b/auth/src/main/java/com/firebase/ui/auth/viewmodel/AuthViewModelBase.java
@@ -2,7 +2,6 @@ package com.firebase.ui.auth.viewmodel;
 
 import android.app.Application;
 
-import com.firebase.ui.auth.AuthUI;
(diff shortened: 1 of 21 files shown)


auth/src/main/java/com/firebase/ui/auth/viewmodel/AuthViewModelBase.java
```diff
@@ -2,7 +2,6 @@
 
 import android.app.Application;
 
-import com.firebase.ui.auth.AuthUI;
```
rosariopfernandes, line 5: This change is unrelated to this PR.
I'm not sure how it ended up here but it is harmless, so I guess it's ok to keep it?

samtstern: Yep not a problem.

samtstern (approved):

samtstern:
🙏  thank you so much for this @rosariopfernandes!  Goodbye ButterKnife 👋

Issue #1966, opened 2021-06-23
[Internal] Move to Java 8 and AGP 7


samtstern:
Version 8.0 of FirebaseUI can/should require Java 8 to allow us to take on newer dependencies. This also partially unblocks a future move to Kotlin.

cc @rosariopfernandes

samtstern:
Looks like the latest version of AGP is `7.0.0-beta05`:
https://maven.google.com/web/index.html?q=com.android.tools.#com.android.tools.build:gradle:7.0.0-beta05

@rosariopfernandes have you heard anything about when `7.0.0` stable will be out?

rosariopfernandes:
@samtstern Not really, but I know Jetpack Compose is going stable this month ([rc01 is out](https://developer.android.com/jetpack/androidx/releases/compose)). I'm guessing AGP7 will be stable alongside Compose and AS Arctic Fox.

samtstern:
This has been implemented and released in version `8.0.0`

PR #1979, opened 2021-07-18
upgrade to AGP7

rosariopfernandes:
See #1966 

This PR should:
- [x] update AGP to 7.0.0 stable
- [x] update Gradle to 7.0.2
- [x] update the GitHub Actions workflow to use Java 11 - this is required by Gradle 7.
- [x] remove unused strings, colors and ids to avoid `UnusedIds` and `UnusedResources` lint errors.
- [x] suppress the `UnusedIds` lint error thrown in the `app` module - seems like AGP7 can't yet detect when an id is being used by ViewBinding.
- [x] suppress the `MediaCapabilities` lint error thrown in the `app` module - I couldn't figure out where this error is coming from, but I'm guessing Glide is responsible for it. 
- [x] fix many other lint issues that were thrown by AGP7

### DIFF
--- a/auth/src/main/res/values-ar/strings.xml
+++ b/auth/src/main/res/values-ar/strings.xml
@@ -1,7 +1,6 @@
 <resources tools:ignore="ExtraTranslation" xmlns:tools="http://schemas.android.com/tools" xmlns:xliff="urn:oasis:names:tc:xliff:document:1.2">
     <string name="fui_progress_dialog_loading">جارٍ التحميل…</string>
     <string name="fui_sign_in_default">تسجيل الدخول</string>
-    <string name="fui_cancel">إلغاء</string>
     <string name="fui_continue">متابعة</string>
     <string name="fui_tos_and_pp">تشير المتابعة إلى موافقتك على %1$s و%2$s.</string>
     <string name="fui_tos_and_pp_footer">%1$s \u00A0 \u00A0 %2$s</string>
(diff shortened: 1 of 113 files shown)


auth/src/main/res/values-ar/strings.xml
```diff
@@ -1,7 +1,6 @@
 <resources tools:ignore="ExtraTranslation" xmlns:tools="http://schemas.android.com/tools" xmlns:xliff="urn:oasis:names:tc:xliff:document:1.2">
     <string name="fui_progress_dialog_loading">جارٍ التحميل…</string>
     <string name="fui_sign_in_default">تسجيل الدخول</string>
-    <string name="fui_cancel">إلغاء</string>
```
samtstern, line 4: I am surprised we don't use "Cancel" anywhere!

database/src/main/java/com/firebase/ui/database/FirebaseRecyclerAdapter.java
```diff
@@ -57,7 +57,6 @@ public void startListening() {
     @OnLifecycleEvent(Lifecycle.Event.ON_STOP)
     public void stopListening() {
         mSnapshots.removeChangeEventListener(this);
-        notifyDataSetChanged();
```
samtstern, line 60: Is this change related to AGP7?

samtstern:
@rosariopfernandes you are a legend! I can't believe you got this all working so quickly.

database/src/main/java/com/firebase/ui/database/FirebaseRecyclerAdapter.java
rosariopfernandes: @samtstern Kind of. Actually, lint got mad at this method call. It suggests calling a more specific one such as `notifyItemInserted`, `notifyItemDeleted`, `notifyItemMoved` etc. 

But I figured that since we're only removing the listener and not adding/removing any item, maybe there's nothing to notify the adapter about?

rosariopfernandes:
@samtstern I was just about to push that, OMG :laughing:

samtstern:
@rosariopfernandes you were right about July 28th! 7.0 is out and I just updated your branch (I wanted to be faster than you for once in my life haha)

database/src/main/java/com/firebase/ui/database/FirebaseRecyclerAdapter.java
samtstern: SGTM

rosariopfernandes:
@samtstern You did it, I'm impressed. I bow down to you :laughing:

samtstern:
I'll release 8.0 tomorrow!

rosariopfernandes:
I was updating rtdb to use Paging 3, but there were some runtime errors, I
think I got something wrong.
I'll send a PR with the changes tomorrow so we can have a look at it
together.

samtstern:
@rosariopfernandes oh wow I almost forgot about that change! Thanks for the reminder. No rush on that PR, 8.0 will go out when that's ready.

Issue #1982, opened 2021-07-28
Convert RTDB to Paging3


samtstern:
Tracking bug for @rosariopfernandes

samtstern:
This has been implemented and released in version `8.0.0`

PR #1983, opened 2021-07-29
refactor(database): migrate to paging 3

rosariopfernandes:
This PR should:
- update the database module to use Android Jetpack's Paging 3
- update the sample app to use Paging 3
- update the docs

See #1982


database/README.md
```diff
@@ -245,13 +245,13 @@ FirebaseRecyclerAdapter adapter = new FirebaseRecyclerAdapter<Chat, ChatHolder>(
 
 The `FirebaseRecyclerPagingAdapter` binds a `Query` to a `RecyclerView` by loading documents in pages.
 This results in a time and memory efficient binding, however it gives up the real-time events
-afforded by the `FirestoreRecyclerAdapter`.
```
rosariopfernandes, line 248: I wonder how long we been living with this copy-paste error :grin:

samtstern: Haha oh no!

database/src/main/java/com/firebase/ui/database/paging/DatabasePagingSource.java
```diff
@@ -0,0 +1,120 @@
...
+    public Single<LoadResult<String, DataSnapshot>> loadSingle(@NonNull LoadParams<String> params) {
+        Task<DataSnapshot> task;
+        if (params.getKey() == null) {
+            task = mQuery.limitToFirst(params.getLoadSize()).get();
+        } else {
+            task = mQuery.startAt(null, params.getKey()).limitToFirst(params.getLoadSize() + 1).get();
+        }
+
+        return Single.fromCallable(() -> {
+            Tasks.await(task);
```
samtstern, line 49: I think this will throw if the Tasks fails:
https://developers.google.com/android/reference/com/google/android/gms/tasks/Tasks#await(com.google.android.gms.tasks.Task%3CTResult%3E)

So we can't rely on getting to the next line. Although I think it's fine since you re-throw at the bottom, but I just don't think you'll ever get that far.

Also we can probably remove the `task.isSuccessful` nesting since we know if we even get to the next line, the task is successful.

samtstern (approved):
Approve, just one comment you may want to address before I merge. Thank you for this!

database/src/main/java/com/firebase/ui/database/paging/DatabasePagingSource.java
rosariopfernandes: @samtstern Interesting, I didn't know that.

I have wrapped it around a try/catch so that developers using FirebaseUI can get the original cause  instead of `await`'s `ExecutionException`.

PR #2025, opened 2022-02-24
fix: Set selected contry spinner

### LINKED ISSUE #2005: CountryListSpinner does not set the selected CountryInfo
Firebase Auth-UI sign-in with phone number has an issue as the CountryListSpinner does not provider the actual selected CountryInfo via method getSelectedCountryInfo().
I've checked the code an there is a small change required:

        mListPopupWindow.setOnItemClickListener(new AdapterView.OnItemClickListener() {
            @Override
            public void onItemClick(AdapterView<?> parent, View view, int position, long id) {
                CountryInfo info = mCountryListAdapter.getItem(position);
                if (info != null) {
                    mSelectedCountryInfo = info; // This needs to be added in order to set the selected country info 
                    setText(info.toShortString());
                }

                onUnfocus();
            }
        });

Please check, "mSelectedCountryInfo = info;" needs to be set as commented above.
Please update, as phone number authentication is not working in version 8.0.0 due to missing country code.

kroikie:
fixes #2005

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/ui/phone/CountryListSpinner.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/phone/CountryListSpinner.java
@@ -86,6 +86,7 @@ public final class CountryListSpinner extends TextInputEditText implements View.
             public void onItemClick(AdapterView<?> parent, View view, int position, long id) {
                 CountryInfo info = mCountryListAdapter.getItem(position);
                 if (info != null) {
+                    mSelectedCountryInfo = info;
                     setText(info.toShortString());
                 }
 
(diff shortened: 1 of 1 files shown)


auth/src/main/java/com/firebase/ui/auth/ui/phone/CountryListSpinner.java
```diff
@@ -86,6 +86,7 @@ public CountryListSpinner(Context context, AttributeSet attrs, int defStyle) {
             public void onItemClick(AdapterView<?> parent, View view, int position, long id) {
                 CountryInfo info = mCountryListAdapter.getItem(position);
                 if (info != null) {
+                    mSelectedCountryInfo = info;
                     setText(info.toShortString());
```
rosalyntan, line 90: nit: Maybe replace these two lines with a call to the `setSelectedForCountry` helper method?

rosalyntan (approved):

PR #2055, opened 2022-08-18
refactor(auth): log an exception if the device has no browser installed

thatfiredev:
Adding a workaround to prevent app from crashing when there is no browser installed.
See firebase/firebase-android-sdk#4006

Checks if the device has a browser. If not, log an exception.

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/ui/phone/PhoneNumberVerificationHandler.java
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/phone/PhoneNumberVerificationHandler.java
@@ -2,8 +2,10 @@ package com.firebase.ui.auth.ui.phone;
 
 import android.app.Activity;
 import android.app.Application;
+import android.content.ActivityNotFoundException;
+import android.content.Intent;
+import android.net.Uri;
 import android.os.Bundle;
-
 import com.firebase.ui.auth.data.model.PhoneNumberVerificationRequiredException;
 import com.firebase.ui.auth.data.model.Resource;
 import com.firebase.ui.auth.viewmodel.AuthViewModelBase;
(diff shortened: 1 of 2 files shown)


auth/src/main/java/com/firebase/ui/auth/ui/phone/PhoneNumberVerificationHandler.java
```diff
@@ -58,7 +60,11 @@ public void onCodeSent(@NonNull String verificationId,
         if (force) {
             optionsBuilder.setForceResendingToken(mForceResendingToken);
         }
-        PhoneAuthProvider.verifyPhoneNumber(optionsBuilder.build());
+        if (isBrowserAvailable(activity)) {
+            PhoneAuthProvider.verifyPhoneNumber(optionsBuilder.build());
+        } else {
+            setResult(Resource.forFailure(new ActivityNotFoundException("No browser was found in this device")));
```
marinacoelho, line 66: Is it only logging the message, or is it also displaying something to the user?

thatfiredev: User sees a "Unknown error occurred" red message on the screen.
I know it's not a very explanatory message, but unfortunately the library was designed like that. We have lots of non-explanatory messages...
My favorite one is when the logcat shows `DEVELOPER_ERROR`:
https://github.com/firebase/FirebaseUI-Android/blob/4455f411e3605743bb4511d1a75cda7f398bc399/auth/src/main/java/com/firebase/ui/auth/ErrorCodes.java#L27-L29

marinacoelho: Indeed, this one might be the best non-explanatory message I've seen...

marinacoelho (approved):

PR #2064, opened 2022-09-26
Release version 8.0.2

thatfiredev:


achuinard:
Looking forward to the page keys comparison fix. Thank you.

gradle.properties
```diff
@@ -4,9 +4,15 @@ org.gradle.caching=true
...
+# Kotlin code style for this project: "official" or "obsolete":
+kotlin.code.style=official
+# Enables namespacing of each library's R class so that its R class includes only the
+# resources declared in the library itself and none from the library's dependencies,
+# thereby reducing the size of the R class for that library
+android.nonTransitiveRClass=false
 
 GROUP=com.firebaseui
-VERSION_NAME=8.0.1
+VERSION_NAME=8.0.2-SNAPSHOT
```
thatfiredev, line 15:
```suggestion
VERSION_NAME=8.0.2
```

buildSrc/src/main/kotlin/Config.kt
```diff
@@ -1,8 +1,8 @@
 object Config {
-    const val version = "8.0.1"
+    const val version = "8.0.2-SNAPSHOT"
```
thatfiredev, line 2:
```suggestion
    const val version = "8.0.2"
```

gradle.properties
thatfiredev, line 15:
```suggestion
VERSION_NAME=8.0.2
```

PR #2180, opened 2025-01-28
feat: move from SmartLock to CredentialManager

Lyokone:
Hey there! So you want to contribute to FirebaseUI? Before you file this pull request, follow these steps:

  * Read [the contribution guidelines](CONTRIBUTING.md).
  * If this has been discussed in an issue, make sure to mention the issue number here. If not, go file an issue about this to make sure this is a desirable change.


.github/workflows/android.yml
```diff
@@ -8,13 +8,13 @@ jobs:
...
-      with:
-        java-version: 11
-    - name: Build with Gradle
-      run: ./scripts/build.sh
-    - name: Print Logs
-      if: failure()
-      run: ./scripts/print_build_logs.sh
\ No newline at end of file
+      - uses: actions/checkout@v2
+      - name: set up JDK 11
```
thatfiredev, line 12:
```suggestion
      - name: set up JDK 17
```

auth/build.gradle.kts
```diff
@@ -72,8 +76,12 @@ dependencies {
     implementation(Config.Libs.Androidx.fragment)
     implementation(Config.Libs.Androidx.customTabs)
     implementation(Config.Libs.Androidx.constraint)
+    implementation("androidx.credentials:credentials:1.3.0")
+    implementation("androidx.core:core-ktx:1.10.1")
 
     implementation(Config.Libs.Androidx.lifecycleExtensions)
+    implementation("androidx.core:core-ktx:1.13.1")
```
thatfiredev, line 83: We seem to have the same `core-ktx` dependency listed twice with different versions?
Can we remove the older version?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.3.3-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.5-all.zip
```
thatfiredev, line 3: Do we need `all`? I tend to use `bin` for the small download size.

```suggestion
distributionUrl=https\://services.gradle.org/distributions/gradle-7.5-bin.zip
```

settings.gradle
```diff
@@ -1,6 +1,15 @@
...
+    repositories {
+        gradlePluginPortal()
+        google()
+        mavenCentral()
+    }
+}
+
 plugins {
     id "com.gradle.enterprise" version "3.3.3"
+    id "org.jetbrains.dokka" version "2.0.0" apply false
```
thatfiredev, line 12: Are we using dokka now, or is this for later? :)

buildSrc/src/main/kotlin/Config.kt
```diff
@@ -5,9 +5,9 @@ object Config {
     private const val kotlinVersion = "1.7.10"
 
     object SdkVersions {
-        const val compile = 29
-        const val target = 29
-        const val min = 16
+        const val compile = 34
+        const val target = 34
+        const val min = 21
```
thatfiredev, line 10: The latest version of Auth requires minSdk 23, so we might have to bump this one a little higher 😅

build.gradle.kts
```diff
@@ -5,6 +5,7 @@ buildscript {
         google()
         mavenCentral()
         mavenLocal()
+        maven("https://oss.jfrog.org/artifactory/oss-release-local")
```
thatfiredev, line 8: I assume we use this repo for `build-info-extractor-gradle` ?
I have no idea why we use that plugin, can we maybe add a comment here to say that this line is needed for that plugin? That way we removed them altogether if we find out that it is not needed.

auth/src/main/java/com/firebase/ui/auth/viewmodel/credentialmanager/CredentialManagerHandler.kt
```diff
@@ -0,0 +1,87 @@
...
+
+        // Example: Password credential with the user's email as the identifier
+        val request = CreatePasswordRequest(
+            id = firebaseUser.email!!,
+            password = password
+        )
+
+        viewModelScope.launch {
+            try {
+                credentialManager.createCredential(activity, request)
```
thatfiredev, line 62: Looks like [`createCredential()`](https://developer.android.com/reference/android/credentials/CredentialManager#createCredential(android.content.Context,%20android.credentials.CreateCredentialRequest,%20android.os.CancellationSignal,%20java.util.concurrent.Executor,%20android.os.OutcomeReceiver%3Candroid.credentials.CreateCredentialResponse,android.credentials.CreateCredentialException%3E)) takes `Context` as argument. Maybe we can pass that to `saveCredentials()` instead of `activity: androidx.activity.ComponentActivity` ? 

Also, do we care about the result of `createCredential()` ? it seems like we're ignoring it right now.

thatfiredev (commented):
Sorry for the delayed review!

auth/src/main/java/com/firebase/ui/auth/ui/credentials/CredentialSaveActivity.kt
```diff
@@ -0,0 +1,89 @@
...
+import com.firebase.ui.auth.util.ExtraConstants
+import com.firebase.ui.auth.viewmodel.ResourceObserver
+import com.firebase.ui.auth.viewmodel.credentialmanager.CredentialManagerHandler
+import com.google.android.gms.auth.api.credentials.Credential
+import androidx.lifecycle.ViewModelProvider
+import com.google.firebase.auth.FirebaseAuth
+
+class CredentialSaveActivity : InvisibleActivityBase() {
+
+    companion object {
```
marinacoelho, line 20: Can we leave companions objects at the end of the class, conforming to [Kotlin's coding conventions](https://kotlinlang.org/docs/coding-conventions.html#class-layout)?

auth/src/main/java/com/firebase/ui/auth/viewmodel/credentialmanager/CredentialManagerHandler.kt
marinacoelho: From what I understood, we don't need to get the result - if `createCredential()` succeeds, we've safely saved the user's password. But that brings me to another question: (out of curiosity) what's _in_ the `result` of this call?

auth/src/main/java/com/firebase/ui/auth/AuthUI.java
```diff
@@ -393,41 +281,22 @@ public Task<Void> signOut(@NonNull Context context) {
...
-                            && ((ApiException) e).getStatusCode() == CommonStatusCodes
-                            .CANCELED) {
-                        Log.w(TAG, "Could not disable auto-sign in, maybe there are no " +
-                                "SmartLock accounts available?", e);
-                        return null;
-                    }
-
-                    return task.getResult();
-                });
+        Task<Void> maybeDisableAutoSignIn = Tasks.forResult((Void) null);
```
thatfiredev, line 284: We can remove this task completely. And in the return statement we use:

```diff
- return Tasks.whenAll(
-                 signOutIdps(context),
-                 maybeDisableAutoSignIn
-        ).continueWith(task -> {
        
+ return signOutIdps(context).continueWith(task -> {
```

auth/src/main/java/com/firebase/ui/auth/data/remote/SignInKickstarter.java
```diff
@@ -50,74 +49,14 @@
...
-                                startAuthMethodChoice();
-                            }
-                        } catch (ApiException e) {
-                            startAuthMethodChoice();
-                        }
-                    });
-        } else {
-            startAuthMethodChoice();
-        }
+        startAuthMethodChoice();
```
thatfiredev, line 59: Some of the code we deleted here seems to be related to `emailLink` - do we want to keep that?

auth/src/main/java/com/firebase/ui/auth/viewmodel/SignInViewModelBase.java
```diff
@@ -19,11 +19,11 @@ protected SignInViewModelBase(Application application) {
     }
 
     @Override
-    protected void setResult(Resource<IdpResponse> output) {
+    public void setResult(Resource<IdpResponse> output) {
         super.setResult(output);
     }
 
-    protected void handleSuccess(@NonNull IdpResponse response, @NonNull AuthResult result) {
+    public void handleSuccess(@NonNull IdpResponse response, @NonNull AuthResult result) {
```
thatfiredev, line 26: Out of curiosity: why are these public now?

auth/src/main/java/com/firebase/ui/auth/KickoffActivity.java
```diff
@@ -25,7 +25,7 @@
 
 @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
 public class KickoffActivity extends InvisibleActivityBase {
-    private SignInKickstarter mKickstarter;
+    public static SignInKickstarter mKickstarter;
```
thatfiredev, line 28: Keeping a reference to a static object in Android is a really bad idea, but I see what you meant when you said that state management is hard when we move stuff from KickoffActivity to AuthMethodPicker - I'll let this be for now, and we can change in the future if we re-architect this flow.

auth/src/main/java/com/firebase/ui/auth/viewmodel/SignInViewModelBase.java
Lyokone: I'm calling those from AuthMethodPicking activity now :/

settings.gradle
thatfiredev: Yeah, I also noticed this when I was trying to fix CI. I don't think we use dokka, I'm guessing one of our dependencies is incorrectly pulling dokka. We can proceed with this for now and figure it out later.

thatfiredev (approved):

marinacoelho (approved):

PR #2188, opened 2025-03-03
feat: rename setIsSmartLockEnabled to setCredentialManagerEnabled

Lyokone:
Rename the function and remove hints since it's not available with CredentialManager

### DIFF
--- a/app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
+++ b/app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
@@ -194,8 +194,7 @@ public class AuthUiActivity extends AppCompatActivity
                 .setTheme(getSelectedTheme())
                 .setLogo(getSelectedLogo())
                 .setAvailableProviders(getSelectedProviders())
-                .setIsSmartLockEnabled(mBinding.credentialSelectorEnabled.isChecked(),
-                        mBinding.hintSelectorEnabled.isChecked());
+                .setCredentialManagerEnabled(mBinding.credentialSelectorEnabled.isChecked());
 
         if (mBinding.customLayout.isChecked()) {
             AuthMethodPickerLayout customLayout = new AuthMethodPickerLayout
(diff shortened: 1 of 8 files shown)


marinacoelho (approved):

app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
```diff
@@ -194,8 +194,7 @@ private Intent getSignInIntent(@Nullable String link) {
                 .setTheme(getSelectedTheme())
                 .setLogo(getSelectedLogo())
                 .setAvailableProviders(getSelectedProviders())
-                .setIsSmartLockEnabled(mBinding.credentialSelectorEnabled.isChecked(),
-                        mBinding.hintSelectorEnabled.isChecked());
+                .setCredentialManagerEnabled(mBinding.credentialSelectorEnabled.isChecked());
```
thatfiredev, line 197: This `hintSelectorEnabled` view that we just deleted is no longer needed, we can probably delete it from the xml file too:

https://github.com/firebase/FirebaseUI-Android/blob/5b6c9f01f5a674d754d8e9b1242197fe040086e7/app/src/main/res/layout/auth_ui_layout.xml#L323-L328

thatfiredev: Also: can we change the app's `options_enable_credential_selector` String to say "Enable Credential Manager" instead? The second String below can be removed.

https://github.com/firebase/FirebaseUI-Android/blob/5b6c9f01f5a674d754d8e9b1242197fe040086e7/app/src/main/res/values/strings.xml#L66-L67

thatfiredev (changes requested):

app/src/main/java/com/firebase/uidemo/auth/AuthUiActivity.java
marinacoelho: Resolving conversation as comments were addressed.

PR #2216, opened 2025-09-15
feat(AuthUIConfiguration): implement configuration model, DSL builder and tests

demolaf:
- Add AuthUIConfiguration data class with all properties
- Implement authUIConfiguration DSL builder function
- Add AuthProvider, AuthUITheme, AuthUIStringProvider, and PasswordRule models
- Create comprehensive tests for default and custom configurations


auth/src/main/java/com/firebase/ui/auth/compose/configuration/AuthProvider.kt
```diff
@@ -0,0 +1,270 @@
...
+
+        /**
+         * The tenant ID for Azure Active Directory.
+         */
+        val tenant: String?,
+
+        /**
+         * A map of custom OAuth parameters.
+         */
+        val customParameters: Map<String, String>
```
Lyokone, line 193: Would be good to extend a base class to prevent duplicate declaration

auth/src/main/java/com/firebase/ui/auth/compose/configuration/AuthUITheme.kt
```diff
@@ -0,0 +1,134 @@
...
+            colorScheme = lightColorScheme(),
+            typography = Typography(),
+            shapes = Shapes(),
+            // TODO(demolaf): do we provide default styles for each provider?
+            providerStyles = mapOf<String, ProviderStyle>(
+                "google.com" to ProviderStyle(
+                    backgroundColor = Color.White,
+                    contentColor = Color.Black
+                )
+            )
```
Lyokone, line 105: Yes that would be good so we can match what is currently done in the library (each provider already has some default)

auth/src/test/java/com/firebase/ui/auth/compose/configuration/AuthUIConfigurationTest.kt
```diff
@@ -0,0 +1,233 @@
...
+import androidx.compose.material.icons.filled.AccountCircle
+import com.google.common.truth.Truth.assertThat
+import com.google.firebase.auth.actionCodeSettings
+import org.junit.Assert.assertThrows
+import org.junit.Test
+import java.util.Locale
+import kotlin.reflect.KMutableProperty
+import kotlin.reflect.full.memberProperties
+
+class AuthUIConfigurationTest {
```
Lyokone, line 27: Would like to see a test for the Builder pattern as well

demolaf: not sure if i understand, doesn't the authUIConfiguration dsl function tests handle the builder pattern test also?

auth/build.gradle.kts
```diff
@@ -4,6 +4,7 @@ plugins {
     id("com.android.library")
     id("com.vanniktech.maven.publish")
     id("org.jetbrains.kotlin.android")
+    id("org.jetbrains.kotlin.plugin.compose") version "2.0.0"
```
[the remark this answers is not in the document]
demolaf: > Why do we need to add it now?

without it we won't be able to use compose, and we need access to ImageVector too.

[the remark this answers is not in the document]
demolaf: > or using the base toml file

i thought since we're using an existing library we should follow how dependencies were added in this case that is using the Config class hence why i did not use a toml file but i could replace it and also take out the hardcoded version `id("org.jetbrains.kotlin.plugin.compose") version "2.0.0"`, wdyt?

auth/src/main/java/com/firebase/ui/auth/compose/configuration/AuthProvider.kt
```diff
@@ -0,0 +1,342 @@
...
+
+        /**
+         * A list of allowed country codes.
+         */
+        val allowedCountries: List<String>?,
+
+        /**
+         * The expected length of the SMS verification code. Defaults to 6.
+         */
+        val smsCodeLength: Int = 6,
```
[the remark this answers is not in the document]
demolaf: I got this from the API Design docs, is there somewhere else I could check this from?

PR #2217, opened 2025-09-18
feat: Core: Auth State & User Accessors

Lyokone:
Implement isSignedIn(), getCurrentUser(), and authStateFlow(): Flow<AuthState> to expose auth state and user.

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/compose/FirebaseAuthUI.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/compose/FirebaseAuthUI.kt
@@ -17,8 +17,14 @@ package com.firebase.ui.auth.compose
 import androidx.annotation.RestrictTo
 import com.google.firebase.FirebaseApp
 import com.google.firebase.auth.FirebaseAuth
+import com.google.firebase.auth.FirebaseAuth.AuthStateListener
+import com.google.firebase.auth.FirebaseUser
 import com.google.firebase.auth.ktx.auth
 import com.google.firebase.ktx.Firebase
+import kotlinx.coroutines.channels.awaitClose
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.callbackFlow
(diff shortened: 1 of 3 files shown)


auth/src/main/java/com/firebase/ui/auth/compose/FirebaseAuthUI.kt
```diff
@@ -56,6 +62,154 @@ class FirebaseAuthUI private constructor(
...
+        trySend(initialState)
+
+        // Create auth state listener
+        val authStateListener = AuthStateListener { firebaseAuth ->
+            val currentUser = firebaseAuth.currentUser
+            val state = if (currentUser != null) {
+                // Check if email verification is required
+                if (!currentUser.isEmailVerified &&
+                    currentUser.email != null &&
+                    currentUser.providerData.any { it.providerId == "password" }) {
```
demolaf, line 168: should we use `it.providerId == EmailAuthProvider.EMAIL_PASSWORD_SIGN_IN_METHOD` here instead, so it's clearer what is being compared?

Lyokone: I think with the comments, it's pretty clear so no need to add extra checks

auth/src/main/java/com/firebase/ui/auth/compose/FirebaseAuthUI.kt
```diff
@@ -56,6 +62,154 @@ class FirebaseAuthUI private constructor(
...
+        // Set initial state based on current auth state
+        val initialState = auth.currentUser?.let { user ->
+            AuthState.Success(result = null, user = user, isNewUser = false)
+        } ?: AuthState.Idle
+
+        trySend(initialState)
+
+        // Create auth state listener
+        val authStateListener = AuthStateListener { firebaseAuth ->
+            val currentUser = firebaseAuth.currentUser
```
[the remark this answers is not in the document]
Lyokone: Hum, we are determining states here depending on conditions, not sure it's required.

PR #2225, opened 2025-09-25
feat: Error Recovery & Dialog

Lyokone:
ErrorRecoveryDialog composable with onRetry, onDismiss, optional onRecover; uses AuthUIStringProvider.


auth/src/main/java/com/firebase/ui/auth/compose/ErrorRecoveryDialog.kt
```diff
@@ -0,0 +1,199 @@
...
+ * @since 10.0.0
+ */
+@Composable
+fun ErrorRecoveryDialog(
+    error: AuthException,
+    stringProvider: AuthUIStringProvider,
+    onRetry: () -> Unit,
+    onDismiss: () -> Unit,
+    onRecover: ((AuthException) -> Unit)? = null,
+    modifier: Modifier = Modifier,
```
demolaf, line 70: there's a lint error stating modifier should come first.

PR #2226, opened 2025-09-26
feat: Core: Sign-out & Delete

Lyokone:
Sign-out clears user; flow reflects change.

Delete removes current account or surfaces reauth requirement.

Cancellations map to AuthCancelledException.


demolaf (approved):
LGTM

PR #2231, opened 2025-09-30
feat: AuthTextField

demolaf:
- Works with EmailValidator/PasswordValidator.
- Supports leading/trailing icons and transformations.
- Comprehensive test covering text input states and validations.


auth/build.gradle.kts
```diff
@@ -74,16 +74,17 @@ android {
...
 
 dependencies {
-    implementation(platform(Config.Libs.Androidx.Compose.bom))
-    implementation(Config.Libs.Androidx.Compose.ui)
-    implementation(Config.Libs.Androidx.Compose.uiGraphics)
-    implementation(Config.Libs.Androidx.Compose.material3)
-    implementation(Config.Libs.Androidx.Compose.foundation)
-    implementation(Config.Libs.Androidx.Compose.tooling)
-    implementation(Config.Libs.Androidx.Compose.toolingPreview)
-    implementation(Config.Libs.Androidx.Compose.activityCompose)
```
Lyokone, line 84: Let's keep everything in the Config file, so it can be shared later by other librairies

PR #2235, opened 2025-10-07
feat: CountryData & Selector Utilities

Lyokone:
Define CountryData list (name, dialCode, countryCode, flagEmoji) and helper APIs for filtering/formatting.


demolaf (approved):
LGTM

PR #2244, opened 2025-10-17
feat: MFA Screens

Lyokone:
Composable screens for MFA enrollment and challenge with slot states.

https://github.com/user-attachments/assets/8a17805b-abcf-46a3-8063-b36d6aceb2cd


composeapp/src/main/java/com/firebase/composeapp/MainActivity.kt
```diff
@@ -210,4 +222,57 @@ class MainActivity : ComponentActivity() {
             authUI = authUI,
         )
     }
+
+    @Composable
+    private fun LaunchMfaEnrollment(
+        authUI: FirebaseAuthUI,
+        backStack: androidx.compose.runtime.snapshots.SnapshotStateList<androidx.navigation3.runtime.NavKey>
```
demolaf: this can be NavBackStack instead. i.e. backStack: NavBackStack

PR #2246, opened 2025-10-21
feat: extract routes from example app to main library

Lyokone:
Currently some of the routing was only done in the example app.
We have now properly moved the logic inside the main library.

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/compose/ui/screens/FirebaseAuthScreen.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/compose/ui/screens/FirebaseAuthScreen.kt
@@ -52,6 +52,7 @@ import com.firebase.ui.auth.compose.configuration.MfaConfiguration
 import com.firebase.ui.auth.compose.configuration.auth_provider.AuthProvider
 import com.firebase.ui.auth.compose.configuration.auth_provider.rememberSignInWithFacebookLauncher
 import com.firebase.ui.auth.compose.configuration.auth_provider.signInWithEmailLink
+import com.firebase.ui.auth.compose.configuration.string_provider.AuthUIStringProvider
 import com.firebase.ui.auth.compose.configuration.string_provider.DefaultAuthUIStringProvider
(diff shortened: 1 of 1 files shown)


auth/src/main/java/com/firebase/ui/auth/compose/ui/screens/FirebaseAuthScreen.kt
```diff
@@ -0,0 +1,603 @@
...
+                    }
+                }
+                onSignInCancelled()
+            }
+
+            is AuthState.Idle -> {
+                pendingResolver.value = null
+                pendingLinkingCredential.value = null
+                lastSuccessfulUserId.value = null
+                if (currentRoute != AuthRoute.MethodPicker.route) {
```
demolaf: i caught a bug while running e2e tests, if auth state is Idle this throws `Cannot navigate to auth_method_picker. Navigation graph has not been set for
 NavController`.

moving this to after NavHost has been initialized fixed it.

> NavHost(navController = navController, startDestination = AuthRoute.MethodPicker.route)
> 
> // Handle email link sign-in (deep links)
> LaunchedEffect(emailLink) {}
> 
> // Synchronise auth state changes with navigation stack.
> LaunchedEffect(authState) {}
>

auth/src/main/java/com/firebase/ui/auth/compose/ui/screens/FirebaseAuthScreen.kt
```diff
@@ -0,0 +1,603 @@
...
+    object Email : AuthRoute("auth_email")
+    object Phone : AuthRoute("auth_phone")
+    object Success : AuthRoute("auth_success")
+    object MfaEnrollment : AuthRoute("auth_mfa_enrollment")
+    object MfaChallenge : AuthRoute("auth_mfa_challenge")
+}
+
+data class AuthSuccessUiContext(
+    val authUI: FirebaseAuthUI,
+    val stringProvider: com.firebase.ui.auth.compose.configuration.string_provider.AuthUIStringProvider,
```
demolaf: this can be `val stringProvider: AuthUIStringProvider`

PR #2247, opened 2025-10-22
feat: Anonymous Auth & Upgrade

demolaf:
- Anonymous sign-in.
- Anonymous auto upgrade for account linking.
- e2e tests covering anonymous sign in and isAnonymousUpgradeEnabled configuration.


auth/src/main/java/com/firebase/ui/auth/compose/configuration/auth_provider/AuthProvider.kt
```diff
@@ -285,7 +285,8 @@ abstract class AuthProvider(open val providerId: String, open val name: String)
          * An interface to wrap the static `EmailAuthProvider.getCredential` method to make it testable.
          * @suppress
          */
-        internal interface CredentialProvider {
+        // TODO(demolaf): make this internal
```
Lyokone: Can we do it in this PR?

PR #2263, opened 2025-10-28
fix: Email link sign-in

demolaf:


auth/src/main/java/com/firebase/ui/auth/compose/AuthException.kt
```diff
@@ -212,38 +212,84 @@ abstract class AuthException(
...
+     *
+     * This exception is thrown when the link is not a valid Firebase email link,
+     * has incorrect format, or is missing required parameters.
+     *
+     * @property cause The underlying [Throwable] that caused this exception
+     */
     class InvalidEmailLinkException(
         cause: Throwable? = null
-    ) : AuthException("You are are attempting to sign in with an invalid email link", cause)
+    ) : AuthException("", cause)
```
Lyokone: Why are we removing the text reason for everything in this file?

PR #2271, opened 2025-11-13
replace google services json file for sample app

demolaf:


waisus996-del:
[Samsung account Privacy Notice.html](https://github.com/user-attachments/files/26202424/Samsung.account.Privacy.Notice.html)

Issue #2310, opened 2026-03-20
[Auth-UI] [v10.0.0-beta02] setProviderChoiceAlwaysShown(false) does not work


nerocristallo:
I am updating the library from version 8.x.x to version 10.x.x and have encountered the following problem:

In the old implementation, I requested sign-in via email directly, without any intermediate screen for selecting the authentication provider.
This makes sense because my app only uses email authentication, and the intermediate view is pretty confusing for the user.

<img width="540" height="1212" alt="Image" src="https://github.com/user-attachments/assets/61ae5aad-b8e8-4556-8d63-71281a015d7b" />

For version 8, I can do this as follows:

```
Use the email authentication provider
List<AuthUI.IdpConfig> providers = List.of( new AuthUI.IdpConfig.EmailBuilder() .setAllowNewAccounts(false) .setDefaultEmail(defaultEmail) .setRequireName(false) .build());

Create and launch sign-in intent
Intent signInIntent = AuthUI.getInstance(AuthManager.getSharedInstance().getFirebaseApp()) .createSignInIntentBuilder() .setTheme(R.style.LoginTheme) .setCredentialManagerEnabled(true) .setAvailableProviders(providers) .build(); signInLauncher.launch(signInIntent);
```

I am trying to do the same for version 10, adding only the email provider, but it is not working. The intermediate view for provider selection is always shown, regardless of the value of the setProviderChoiceAlwaysShown parameter.

Is this the intended behaviour? What is the purpose of setProviderChoiceAlwaysShown?

russellwheatley:
Hey @nerocristallo - thanks for raising the issue. I'll take a look and let you know what I think 👍

russellwheatley:
This issue is closed out with this PR: https://github.com/firebase/FirebaseUI-Android/pull/2311

PR #2311, opened 2026-03-27
fix: ensure that when selecting phone or email, it routes straight to that screen

russellwheatley:
- Demonstrates fix, when only email is selected as provider (and `isProviderChoiceAlwaysShown` is set to false which is the default), it should open that screen immediately. 
- Updated `FirebaseAuthUI.kt` to reuse existing logic to get initial state.
- I've pointed this to `version-10.0.0-beta03` which is the latest of master so that we can version another beta release.

fixes https://github.com/firebase/FirebaseUI-Android/issues/2310

### Updated tests for the following reasons:
- EmailAuthScreenTest.kt Updated for the new single-email-provider behavior: direct email screen by default, picker only when isProviderChoiceAlwaysShown = true.
- AnonymousAuthScreenTest.kt Switched to unique test email data to avoid upgrade collisions exposed while updating this flow.
- GoogleAuthScreenTest.kt Switched to unique mock Google identity data to avoid account-linking collisions during upgrade tests.
- FirebaseAuthUI.kt Added IdTokenListener to make upgrade/link auth state updates observable and keep tests stable.

### Screen recording showing moving straight to email screen when only email selected and `isProviderChoiceAlwaysShown` is false

[Screen_recording_20260327_115854.webm](https://github.com/user-attachments/assets/22d1d4d7-4d77-4c29-8ecf-f44e7a40e9f0)

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUI.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUI.kt
@@ -25,6 +25,7 @@ import com.google.firebase.Firebase
 import com.google.firebase.FirebaseApp
 import com.google.firebase.auth.FirebaseAuth
 import com.google.firebase.auth.FirebaseAuth.AuthStateListener
+import com.google.firebase.auth.FirebaseAuth.IdTokenListener
 import com.google.firebase.auth.FirebaseUser
 import com.google.firebase.auth.auth
 import kotlinx.coroutines.CancellationException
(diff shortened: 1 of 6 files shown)


demolaf:
@russellwheatley looks good, does this work for non-phone/email providers? e.g. google etc.

auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUI.kt
```diff
@@ -296,15 +276,31 @@ class FirebaseAuthUI private constructor(
...
+
+            trySend(initialState)
+
+            // Create auth state listener
+            val authStateListener = AuthStateListener { firebaseAuth ->
+                trySend(buildState(firebaseAuth.currentUser))
+            }
+
+            // AuthStateListener does not reliably fire for account linking, but IdTokenListener does.
+            val idTokenListener = IdTokenListener { firebaseAuth ->
```
demolaf, line 292: does this mean `IdTokenListener` can replace `AuthStateListener`? the only problem would be refactoring tests 😅.

demolaf (approved):

russellwheatley:
> @russellwheatley looks good, does this work for non-phone/email providers? e.g. google etc.

@demolaf - no, the other auth providers just show the button (i.e. google button) because there is no additional auth flow for them, email and phone go straight to their auth flow. I've added a test to show if google the only provider, it should still show method picker.

auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUI.kt
russellwheatley: No, it's fine to keep them both, we could refactor in a later PR if we wished.

thatfiredev (approved):

PR #2320, opened 2026-05-11
fix(auth): make AuthException messages customisable via AuthUIStringProvider

### LINKED ISSUE #1815: Can't customize 'account disabled' string in auth UI
When the user is blocked in the console, the UI show hard coded string: `The user account has been disabled by an administrator`

I need to customize this string.

I have tried adding this `<string name="account_disabled">Account is disabled by the administrator</string>` from (https://github.com/firebase/FirebaseUI-Android/blob/master/app/src/main/res/values/strings.xml#L78) to my`strings.xml`, but this didn't works.

`<string name="account_disabled" ...` is not available in the customizable strings for auth (https://github.com/firebase/FirebaseUI-Android/blob/master/auth/src/main/res/values/strings.xml)

demolaf:
Fixes #1815 — `AuthException` messages were hardcoded in English with no
way for apps to customise or translate them.

### From strings.xml in consumers app
Screenshot 2026-05-11 at 18 27 09

### Expected Behavior
Screenshot_1778520434

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/AuthException.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/AuthException.kt
@@ -14,7 +14,10 @@
 
 package com.firebase.ui.auth
 
+import android.content.Context
 import com.firebase.ui.auth.AuthException.Companion.from
+import com.firebase.ui.auth.configuration.string_provider.AuthUIStringProvider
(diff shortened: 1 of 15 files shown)


russellwheatley (approved):
Nice one 👍

PR #2322, opened 2026-05-15
fix(auth): log out before Facebook sign-in to clear stale cached token

### LINKED ISSUE #879: Not handling error: "FacebookProvider: Error logging in with Facebook. User logged in as different Facebook user"
* Android device: HTC A9
  * Android OS version: 6.0
  * Google Play Services version: 3.1.0 
  * Firebase/Play Services SDK version: 11.0.4
  * FirebaseUI version: 2.3.0
  * Facebook-android-sdk: 4.25.0
  
Firebase UI does not handle "E/FacebookProvider: Error logging in with Facebook. User logged in as different Facebook user."

#### Steps to reproduce:

  1. Use native Facebook app to log out current user A
  2. Launch firebase-ui and try log in in with another facebook user B

Expected: firebase-ui should display any errors, or onActivityResult should be called with some error

Actual: Nothing happens. Logcat prints "E/FacebookProvider: Error logging in with Facebook. User logged in as different Facebook user."

This use case might be quite rare, but it is confusing for customer to not get any feedback.

demolaf:
Closes #879 

When a user switches accounts in the native Facebook app without signing out of FirebaseUI, the Facebook SDK retains a stale cached token. On the next sign-in attempt this caused a silent failure with "User logged in as different Facebook user".

Calling logOut() at the start of signInWithFacebook clears the local token cache before each sign-in, forcing a fresh authentication.

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/configuration/auth_provider/FacebookAuthProvider+FirebaseAuthUI.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/configuration/auth_provider/FacebookAuthProvider+FirebaseAuthUI.kt
@@ -56,6 +56,7 @@ internal fun FirebaseAuthUI.rememberSignInWithFacebookLauncher(
     context: Context,
     config: AuthUIConfiguration,
     provider: AuthProvider.Facebook,
+    loginManagerProvider: AuthProvider.Facebook.LoginManagerProvider = AuthProvider.Facebook.DefaultLoginManagerProvider(),
 ): () -> Unit {
     val coroutineScope = rememberCoroutineScope()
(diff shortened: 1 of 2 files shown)


russellwheatley:
@demolaf - I've made some updates:

- Wrapped pre-sign-in Facebook logOut() in try/catch so a logout failure cannot strand the user in Loading or block the sign-in launcher.
- Logged that swallowed logout failure as a warning because sign-in still proceeds.
- Added a regression test covering a throwing stale-token logout path, ensuring that specific logout exception is not propagated.
- Updated KDoc for the new loginManagerProvider parameter so its purpose is clear.

russellwheatley (approved):

PR #2326, opened 2026-05-19
fix(auth): emit AuthResult on sign-in success and fix MFA tooltip auto-open

### LINKED ISSUE #2318: onSignInSuccess never called — AuthState.Success always emits result=null
Library version: firebase-ui-auth:10.0.0-beta02

Since View based solution is no longer maintained and my number one crash in production is listed among the other issues here, I thought I'd give the compose beta a shot.

I've let AI handle the description below - it seem to be clear enough:

###   What happens                                              
                                                                                                                                                                                                                                                                                                          
  `onSignInSuccess` is never called, even on a first-time fresh email/password sign-in with no prior session. Instead, the user sees the default "Signed in as / Manage two-factor auth / Sign out" screen with no way to proceed.
                                                                                                                                                                                                                                                                                                          
  ###   Root cause                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                          
  In `FirebaseAuthScreen.kt` line 523:                                                                                                                                                                                                                                                                      
                                                            
```
  state.result?.let { result ->
      if (state.user.uid != lastSuccessfulUserId.value) {                                                                                                                                                                                                                                                 
          onSignInSuccess(result)                                                                                                                                                                                                                                                                         
      }                                                                                                                                                                                                                                                                                                   
  } 
```                                                                                                                                                                                                                                                                                                      
                                                            
  `onSignInSuccess` is gated on `result != null`. However, `authUI.authStateFlow()` emits `AuthState.Success(result=null, ...) ` even after a fresh sign-in completes. This is confirmed by the library's own debug log immediately after a successful email/password login:                                       
   
  Current state: `AuthState.Success(result=null, user=..., isNewUser=false) `                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                          
  Navigation to `AuthRoute.Success.route` happens unconditionally (line 536–541) regardless of whether result is null, so the success screen always shows — but the caller's completion callback never fires.  


###   How to reproduce                                                                                                                                                                                                                                                                                        
                                                            
  1. Integrate FirebaseAuthScreen with only `onSignInSuccess` as the completion signal.                                                                                                                                                                                                                     
  2. Launch the app with no prior Firebase session.
  3. Sign in with email and password.                                                                                                                                                                                                                                                                     
  4. `onSignInSuccess` is never called. The "Signed in as" screen is shown with no way to proceed.
                                                                                                                                                                                                                                                                                                          
  Additional issue: tooltip auto-shows on the success screen
                                                                                                                                                                                                                                                                                                          
  With `isMfaEnabled = false` in the configuration, the tooltip on the disabled "Manage two-factor auth" button auto-opens on every render because of:                                                                                                                                                      
   
```
  // FirebaseAuthScreen.kt line 760                                                                                                                                                                                                                                                                       
  state = rememberTooltipState(                             
      initialIsVisible = !configuration.isMfaEnabled
  )
```

  !false = true, so the tooltip is immediately visible. This is unexpected — a disabled feature should not announce itself to the user unprompted.                                                                                                                                                        
   
  Suggested fixes                                                                                                                                                                                                                                                                                         
                                                            
  1. Primary: Call `onSignInSuccess` when `AuthState.Success` is reached regardless of whether result is null. Either pass a synthetic wrapper or add a separate `onAlreadyAuthenticated: (FirebaseUser) -> Unit` callback for the null-result case so callers can handle both paths with named, distinct  callbacks.
  2. Secondary: Change `initialIsVisible = !configuration.isMfaEnabled` to `initialIsVisible = false` — or only show the tooltip on explicit user interaction, not on initial render.                                                                                                                         
                                                                                                                                                                                                                                                                                                          
###   Workaround used in the meantime                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                          
  Providing `authenticatedContent` and calling the app's own sign-in completion logic from there when `authState` is `AuthState.Success`.

demolaf:
Closes #2318 .

All providers called `updateAuthState(Idle)` after sign-in, so the only `AuthState.Success` came from Firebase's `AuthStateListener` which never carries an `AuthResult`. The `result?.let` guard in `FirebaseAuthScreen` then silently blocked `onSignInSuccess` every time.

Fix: emit `AuthState.Success` with the real `AuthResult` via a shared `updateAuthStateWithResult()` helper. Also fixes MFA tooltip auto-opening when disabled.

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUI.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUI.kt
@@ -259,21 +259,7 @@ class FirebaseAuthUI private constructor(
         val firebaseAuthFlow = callbackFlow {
             fun buildState(currentUser: FirebaseUser?): AuthState {
                 return if (currentUser != null) {
-                    if (!currentUser.isEmailVerified &&
-                        currentUser.email != null &&
-                        currentUser.providerData.any { it.providerId == "password" }
-                    ) {
-                        AuthState.RequiresEmailVerification(
-                            user = currentUser,
-                            email = currentUser.email!!
-                        )
-                    } else {
-                        AuthState.Success(
-                            result = null,
-                            user = currentUser,
-                            isNewUser = false
-                        )
-                    }
+                    stateForUser(currentUser, result = null, isNewUser = false)
                 } else {
                     AuthState.Idle
                 }
@@ -286,6 +272,17 @@ class FirebaseAuthUI private constructor(
 
             // Create auth state listener
             val authStateListener = AuthStateListener { firebaseAuth ->
+                // When user signs out, clear stale user-presence internal states so the combine
+                // doesn't return Success/RequiresEmailVerification after the user is gone.
+                if (firebaseAuth.currentUser == null) {
+                    val current = _authStateFlow.value
+                    if (current is AuthState.Success ||
+                        current is AuthState.RequiresEmailVerification ||
+                        current is AuthState.RequiresProfileCompletion
+                    ) {
+                        _authStateFlow.value = AuthState.Idle
+                    }
+                }
                 trySend(buildState(firebaseAuth.currentUser))
             }
 
@@ -326,13 +323,27 @@ class FirebaseAuthUI private constructor(
         _authStateFlow.value = state
     }
 
+    private fun stateForUser(user: FirebaseUser, result: AuthResult?, isNewUser: Boolean): AuthState {
+        return if (!user.isEmailVerified &&
+            user.email != null &&
+            user.providerData.any { it.providerId == "password" }
+        ) {
+            AuthState.RequiresEmailVerification(user = user, email = user.email!!)
+        } else {
+            AuthState.Success(result = result, user = user, isNewUser = isNewUser)
+        }
+    }
+
     internal fun updateAuthStateWithResult(result: AuthResult?, defaultIsNewUser: Boolean = false) {
-        if (result?.user != null) {
-            updateAuthState(AuthState.Success(
-                result = result,
-                user = result.user!!,
-                isNewUser = result.additionalUserInfo?.isNewUser ?: defaultIsNewUser
-            ))
+        val user = result?.user
+        if (user != null) {
+            updateAuthState(
+                stateForUser(
+                    user = user,
+                    result = result,
+                    isNewUser = result.additionalUserInfo?.isNewUser ?: defaultIsNewUser
+                )
+            )
         } else {
             updateAuthState(AuthState.Idle)
         }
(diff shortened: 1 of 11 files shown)


auth/src/main/java/com/firebase/ui/auth/configuration/auth_provider/OAuthProvider+FirebaseAuthUI.kt
```diff
@@ -162,7 +162,6 @@ internal suspend fun FirebaseAuthUI.signInWithProvider(
                     photoUrl = authResult.user?.photoUrl,
                 )
             }
-            updateAuthState(AuthState.Idle)
```
mikehardy, line 165: this case just goes away and does not need `updateAuthStateWithResult` ?

auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUI.kt
```diff
@@ -325,6 +323,32 @@ class FirebaseAuthUI private constructor(
         _authStateFlow.value = state
     }
 
+    private fun stateForUser(user: FirebaseUser, result: AuthResult?, isNewUser: Boolean): AuthState {
```
mikehardy: name is a bit generic for a focused helper, on read it looks more like `successOrRequiresEmailVerification` or similar since it will always just check and return one of those two things

auth/src/main/java/com/firebase/ui/auth/configuration/auth_provider/OAuthProvider+FirebaseAuthUI.kt
demolaf: yes, that was actually a bug. We've emitted AuthState.Success and then emitting AuthState.Idle which would reset the Success state.

Also signInAndLinkWithCredential handles the pending auth result condition and calls updateAuthStateWithResult.

auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUI.kt
demolaf: yeah this makes sense, i'll rename it

demolaf: how about `handleAuthUserState`?

auth/src/main/java/com/firebase/ui/auth/configuration/auth_provider/OAuthProvider+FirebaseAuthUI.kt
mikehardy: great - my analysis was partial + incorrect then, good fix given the extra context

mikehardy (approved):
I don't appear to have the ability to resolve my comment threads but I consider them resolved now. I'm also not sure if I have repo-authority such that my approval counts but it was a real review :-). +1, looks good

dackers86 (approved):

PR #2330, opened 2026-06-04
feat(auth): add customMethodPickerTermsContent slot to FirebaseAuthScreen

### LINKED ISSUE #1338: Checkbox to accept Terms

demolaf:
Closes #1338

Adds a `customMethodPickerTermsContent` slot to `FirebaseAuthScreen` (and `termsContent` to `AuthMethodPicker`) so developers can replace the default "By continuing..." footer with their own UI — e.g. an explicit acceptance checkbox required by some jurisdictions (GDPR, French law).

## Usage
On `FirebaseAuthScreen` composable:

```dart
customMethodPickerTermsConfiguration = MethodPickerTermsConfiguration(
    content = {
        Row(
            modifier = Modifier.padding(horizontal = 16.dp, vertical = 8.dp),
            verticalAlignment = Alignment.CenterVertically
        ) {
            Checkbox(
                checked = termsAccepted,
                onCheckedChange = { termsAccepted = it }
            )
            Text(
                text = "I have read and accept the Terms of Service and Privacy Policy",
                style = MaterialTheme.typography.bodySmall,
                modifier = Modifier.padding(start = 8.dp)
            )
        }
    },
    accepted = termsAccepted,
    disableProvidersUntilAccepted = true,
),
```

### Preview
[TOS Acceptance Checkbox & Disabled State (Opt-in)](https://github.com/user-attachments/assets/08220324-5244-480e-9b59-a906dd6c2a9b)

### DIFF
--- a/app/src/main/java/com/firebaseui/android/demo/CustomMethodPickerDemoActivity.kt
+++ b/app/src/main/java/com/firebaseui/android/demo/CustomMethodPickerDemoActivity.kt
@@ -284,7 +284,7 @@ fun SpotlightMethodPicker(
         anonymous?.let {
             item {
                 Spacer(modifier = Modifier.height(8.dp))
-                TextButton(onClick = { onProviderSelected(it) }) {
+                TextButton(onClick = { onProviderSelected(it) }, enabled = enabled) {
                     Text("Continue as guest")
                 }
             }
--- a/app/src/main/java/com/firebaseui/android/demo/CustomMethodPickerDemoActivity.kt
+++ b/app/src/main/java/com/firebaseui/android/demo/CustomMethodPickerDemoActivity.kt
@@ -247,6 +247,7 @@ fun SpotlightMethodPicker(
                         ProviderIconButton(
                             style = style,
                             contentDescription = provider.providerId,
+                            enabled = enabled,
                             onClick = { onProviderSelected(provider) }
                         )
                     }
@@ -292,9 +293,11 @@ private fun ProviderIconButton(
     style: AuthUITheme.ProviderStyle,
     contentDescription: String,
     onClick: () -> Unit,
+    enabled: Boolean = true,
 ) {
     Button(
         onClick = onClick,
+        enabled = enabled,
         modifier = Modifier.size(52.dp),
         shape = CircleShape,
         colors = ButtonDefaults.buttonColors(containerColor = style.backgroundColor),
--- a/app/src/main/java/com/firebaseui/android/demo/CustomMethodPickerDemoActivity.kt
+++ b/app/src/main/java/com/firebaseui/android/demo/CustomMethodPickerDemoActivity.kt
@@ -51,6 +51,7 @@ import com.firebase.ui.auth.configuration.theme.AuthUIAsset
 import com.firebase.ui.auth.configuration.theme.AuthUITheme
 import com.firebase.ui.auth.configuration.theme.ProviderStyleDefaults
 import com.firebase.ui.auth.ui.components.AuthProviderButton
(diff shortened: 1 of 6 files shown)


just1and0 (commented):
When `disableProvidersUntilAccepted` is true, this only disables the default provider buttons. Custom layouts still receive the raw `onProviderSelected` callback, so they can start auth before terms are accepted. Since this is meant to enforce explicit consent, should we wrap the callback and pass the gated version to both the default layout and custom layouts or is this expected behaviour ?

app/src/main/java/com/firebaseui/android/demo/CustomSlotsThemingDemoActivity.kt
```diff
@@ -102,8 +102,8 @@ fun CustomSlotsDemoChooser(
         )
 
         DemoCard(
-            title = "Custom Method Picker Layout",
-            description = "Replace the default vertical provider list with a 2-column grid using customMethodPickerLayout on FirebaseAuthScreen.",
+            title = "Custom Method Picker Layout & Terms",
+            description = "Replace the default provider list with a custom layout, and swap the 'By continuing...' footer with a checkbox using customMethodPickerLayout and customMethodPickerTermsContent on FirebaseAuthScreen.",
```
just1and0: Small typo: this should say `customMethodPickerTermsConfiguration`, not `customMethodPickerTermsContent`.

app/src/main/java/com/firebaseui/android/demo/CustomMethodPickerDemoActivity.kt
just1and0, line 290: The anonymous provider path does not use the `enabled` flag, so this demo still lets users continue as guest before accepting the terms checkbox. Can we pass `enabled = enabled` to this `TextButton` too?

just1and0 (changes requested):

demolaf:
> When `disableProvidersUntilAccepted` is true, this only disables the default provider buttons. Custom layouts still receive the raw `onProviderSelected` callback, so they can start auth before terms are accepted. Since this is meant to enforce explicit consent, should we wrap the callback and pass the gated version to both the default layout and custom layouts or is this expected behaviour ?

If the user has chosen to provide their own layout then they would be responsible for handling the behaviour of `onProviderSelected`, if we gate this then this takes away control from the user.

This is the intended behaviour.

just1and0 (approved):

PR #2331, opened 2026-06-05
fix(auth): allow SAML providers in AuthUIConfiguration

demolaf:
The provider ID allowlist only permitted oidc.-prefixed IDs for custom providers via GenericOAuth. This adds `saml. `to the same check so Firebase SAML providers (e.g. `saml.corp-sso`) pass validation rather than being rejected as unknown.


just1and0 (approved):
LGTM

PR #2336, opened 2026-06-17
fix(database): fix DatabasePagingSource pagination when using orderByChild

### LINKED ISSUE #2000: DatabasePagingSource  not working when firebase query contains orderByChild clause.

  * Android device: Samsung S21+5G
  * Android OS version: 11
  * Google Play Services version: August 1 2021
  * Firebase/Play Services SDK version: com.google.firebase:firebase-bom:28.3.0
  * FirebaseUI version: 8.0.0
  
DatabasePagingSource (and therefore also FirebaseRecyclerPagingAdapter) not working when firebase query contains ``` orderByChild``` clause. 
I have verified that the problem is not in FirebaseUI itself but in method 

```Java
private Query startAt(Node node, String key)
```
in ```Query``` class in package ```com.google.firebase.database```;
according code javadocs for the method:

```Java
   * Creates a query constrained to only return child nodes with a value greater than or equal to
   * the given value, using the given {@code orderBy} directive or priority as default, and
   * additionally only child nodes with a key greater than or equal to the given key.
```
but I have verified that if the query is using an ```orderByChild``` clause, the ```key``` parameter in ```startAt``` is ignored
Because of this I am getting crashes cause by exception
```
    java.lang.IllegalStateException: The same value, VWGuwwgOg3tg_15vtqkJf32bhICi-zI_yt---, was passed as the nextKey in two
     sequential Pages loaded from a PagingSource. Re-using load keys in
     PagingSource is often an error, and must be explicitly enabled by
     overriding PagingSource.keyReuseSupported.
```
because, since the ```key``` parameter is ignored, we reread each time the same data, each time the method 
```
 public Single<LoadResult<String, DataSnapshot>> loadSingle(@NonNull LoadParams<String> params)
```
is called
#### Steps to reproduce:

  1. see above
  
#### Proposed Fix

I have written code to fix the issue. note that I am working with a fork of the original FirebaseUI-Android, so I have the original code converted to Kotlin, but you can get the idea if you look at it. In brief what I do is keep not only the key of the last read document but also the value of the childkey, and use them together when I need to query for the next batch of documents

  
####  Code With Bug Fix:
```kotlin
class DatabasePagingSource(private val mQuery: Query) : PagingSource<Any, DataSnapshot>() {

    @SuppressLint("RestrictedApi")
    fun getChildValue(snapshot: DataSnapshot, index:PathIndex):Any? {
        val keypath=index.queryDefinition
        val data=snapshot.child(keypath)
        if(!data.exists()) return null
        return data.value
    }
    fun Query.startAt_childvalue(startvalue:Any?,keyvalue:String?):Query {
        return when(startvalue) {
            is String -> startAt(startvalue,keyvalue)
            is Boolean -> startAt(startvalue,keyvalue)
            is Double -> startAt(startvalue,keyvalue)
            is Long -> startAt(startvalue.toDouble(),keyvalue)
            else -> this
        }
    }
    /**
     * reason of SuppressLint: DatabaseError.fromStatus() is not meant to be public.
     */
    @SuppressLint("RestrictedApi")
    override suspend fun load(params: LoadParams<Any>): LoadResult<Any, DataSnapshot> {
        //change mQuery.startAt at value  if child index

        //if not null then what we have here is orderByChild query
        var querychildpathindex:PathIndex? = mQuery.spec.index as? PathIndex
        val pkey=params.key as Pair<Any?,String>?

        val task: Task<DataSnapshot> =

            if (params.key == null) {
                mQuery.limitToFirst(params.loadSize).get()
            } else {
                if (querychildpathindex != null)  //orderByChild query mode
                    mQuery.startAt_childvalue(pkey?.first, pkey?.second).limitToFirst(params.loadSize + 1).get()
                else
                    mQuery.startAt(null,pkey?.second).limitToFirst(params.loadSize + 1).get()
            }
        try {
            val dataSnapshot = task.await()
            if (dataSnapshot.exists()) {
                //Make List of DataSnapshot
                val data: MutableList<DataSnapshot> = ArrayList()
                var lastKey: Pair<Any?,String>? = null
                if (params.key == null) {
                    for (snapshot in dataSnapshot.children) {
                        data.add(snapshot)
                    }
                } else {
                    val iterator: Iterator<DataSnapshot> = dataSnapshot.children.iterator()

                    //Skip First Item that corresponds to lastKey read in previous batch
                    if (iterator.hasNext()) {
                        iterator.next()
                    }
                    while (iterator.hasNext()) {
                        val snapshot = iterator.next()
                        data.add(snapshot)
                    }
                }

                //Detect End of Data
                if (!data.isEmpty()) {
                    //Get Last Key
                    val lastkey_c = getLastPageChildKey(data,querychildpathindex)
                    val lastkey_k = getLastPageKey(data)
                    lastKey = if (lastkey_c == null && lastkey_k == null)
                        null
                    else
                        if (lastkey_k == null) Pair(lastkey_c, "") else Pair(lastkey_c, lastkey_k)
                }
                return toLoadResult(data, lastKey)
            } else {
                val details = DETAILS_DATABASE_NOT_FOUND + mQuery.toString()
                throw DatabaseError.fromStatus(
                    STATUS_DATABASE_NOT_FOUND,
                    MESSAGE_DATABASE_NOT_FOUND,
                    details
                ).toException()
            }
        } catch (e: ExecutionException) {
            return LoadResult.Error<Any, DataSnapshot>(e)
        }
    }


    private fun toLoadResult(
        snapshots: List<DataSnapshot>,
        nextPage: Pair<Any?,String>?
    ): LoadResult<Any, DataSnapshot> {
        return LoadResult.Page(
            snapshots,
            null,  // Only paging forward.
            nextPage,
            LoadResult.Page.COUNT_UNDEFINED,
            LoadResult.Page.COUNT_UNDEFINED
        )
    }
    private fun getLastPageChildKey(data: List<DataSnapshot>,index: PathIndex?): Any? {
        if(index==null) return null
        return if (data.isEmpty()) {
            null
        } else {
            getChildValue(data[data.size - 1],index)
        }
    }

    private fun getLastPageKey(data: List<DataSnapshot>): String? {
        return if (data.isEmpty()) {
            null
        } else {
            data[data.size - 1].key
        }
    }

    override fun getRefreshKey(state: PagingState<Any, DataSnapshot>): Pair<Any?, String>? {
        return null
    }

    companion object {
        private const val STATUS_DATABASE_NOT_FOUND = "DATA_NOT_FOUND"
        private const val MESSAGE_DATABASE_NOT_FOUND = "Data not found at given child path!"
        private const val DETAILS_DATABASE_NOT_FOUND = "No data was returned for the given query: "
    }
}
```

demolaf:
Fixes #2000.

`DatabasePagingSource` was calling `startAt(null, key)` for all queries, but Firebase Database ignores the key parameter when the query uses `orderByChild()`, causing every subsequent page load to return the same first page of results and throwing `IllegalStateException: The same value was passed as the nextKey in two sequential Pages`.

The fix detects `orderByChild()` queries via the internal `PathIndex` and calls `startAt(childValue, key)` instead, storing both the child value and node key as the page cursor in the new `DatabasePagingKey` type.

**Changes:**
- `DatabasePagingKey` — made `public` (required for cross-package Kotlin access) with `equals()`/`hashCode()` (required by the Paging library for key diffing and state restoration)
- `DatabasePagingSource` — extends cursor fix to `orderByValue()` queries via `ValueIndex` in addition to `orderByChild()` via `PathIndex`; surfaces load errors to logcat via `onErrorReturn`
- `FirebaseRecyclerPagingAdapter` — replaces deprecated `@OnLifecycleEvent` with `LifecycleEventObserver` (removed in lifecycle 2.7.0)
- Adds instrumented test verifying no duplicate keys across pages when using `orderByChild()`

### DIFF
--- a/database/src/androidTest/java/com/firebase/ui/database/paging/DatabasePagingSourceTest.java
+++ b/database/src/androidTest/java/com/firebase/ui/database/paging/DatabasePagingSourceTest.java
@@ -1,0 +1,117 @@
+package com.firebase.ui.database.paging;
+
+import android.content.Context;
+
+import androidx.paging.PagingSource;
+import androidx.test.core.app.ApplicationProvider;
+import androidx.test.ext.junit.runners.AndroidJUnit4;
+
+import com.firebase.ui.database.Bean;
+import com.google.firebase.FirebaseApp;
+import com.google.firebase.FirebaseOptions;
+import com.google.firebase.database.DataSnapshot;
+import com.google.firebase.database.DatabaseError;
+import com.google.firebase.database.DatabaseReference;
+import com.google.firebase.database.FirebaseDatabase;
+import com.google.firebase.database.ValueEventListener;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.util.HashSet;
+import java.util.Set;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+import androidx.annotation.NonNull;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+@RunWith(AndroidJUnit4.class)
+public class DatabasePagingSourceTest {
+    private static final String APP_NAME = "firebaseui-paging-tests";
+    private static final int PAGE_SIZE = 3;
+    private static final int TOTAL_ITEMS = 6;
+
+    private DatabaseReference mRef;
+
+    @Before
+    public void setUp() throws InterruptedException {
+        FirebaseApp app;
+        try {
+            app = FirebaseApp.getInstance(APP_NAME);
+        } catch (IllegalStateException e) {
+            Context context = ApplicationProvider.getApplicationContext();
+            app = FirebaseApp.initializeApp(context, new FirebaseOptions.Builder()
+                    .setApplicationId("fir-ui-tests")
+                    .setDatabaseUrl("https://fir-ui-tests.firebaseio.com/")
+                    .build(), APP_NAME);
+        }
+
+        mRef = FirebaseDatabase.getInstance(app).getReference().child("paging_test");
+
+        mRef.removeValue();
+        for (int i = 1; i <= TOTAL_ITEMS; i++) {
+            mRef.push().setValue(new Bean(i));
+        }
+
+        CountDownLatch latch = new CountDownLatch(1);
+        mRef.addValueEventListener(new ValueEventListener() {
+            @Override
+            public void onDataChange(@NonNull DataSnapshot snapshot) {
+                if (snapshot.getChildrenCount() >= TOTAL_ITEMS) {
+                    mRef.removeEventListener(this);
+                    latch.countDown();
+                }
+            }
+
+            @Override
+            public void onCancelled(@NonNull DatabaseError error) {}
+        });
+        assertTrue("Timed out seeding test data", latch.await(30, TimeUnit.SECONDS));
+    }
+
+    @After
+    public void tearDown() {
+        mRef.removeValue();
+    }
+
+    @Test
+    public void testOrderByChild_noDuplicatesAcrossPages() {
+        DatabasePagingSource source = new DatabasePagingSource(mRef.orderByChild("number"));
+
+        PagingSource.LoadResult<DatabasePagingKey, DataSnapshot> result1 =
+                source.loadSingle(new PagingSource.LoadParams.Refresh<>(null, PAGE_SIZE, false))
+                        .timeout(30, TimeUnit.SECONDS)
+                        .blockingGet();
+
+        assertTrue(result1 instanceof PagingSource.LoadResult.Page);
+        PagingSource.LoadResult.Page<DatabasePagingKey, DataSnapshot> page1 =
+                (PagingSource.LoadResult.Page<DatabasePagingKey, DataSnapshot>) result1;
+        assertEquals(PAGE_SIZE, page1.getData().size());
+
+        DatabasePagingKey nextKey = page1.getNextKey();
+
+        PagingSource.LoadResult<DatabasePagingKey, DataSnapshot> result2 =
+                source.loadSingle(new PagingSource.LoadParams.Append<>(nextKey, PAGE_SIZE, false))
+                        .timeout(30, TimeUnit.SECONDS)
+                        .blockingGet();
+
+        assertTrue(result2 instanceof PagingSource.LoadResult.Page);
+        PagingSource.LoadResult.Page<DatabasePagingKey, DataSnapshot> page2 =
+                (PagingSource.LoadResult.Page<DatabasePagingKey, DataSnapshot>) result2;
+
+        Set<String> allKeys = new HashSet<>();
+        for (DataSnapshot snapshot : page1.getData()) {
+            allKeys.add(snapshot.getKey());
+        }
+        for (DataSnapshot snapshot : page2.getData()) {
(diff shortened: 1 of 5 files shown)


database/src/main/java/com/firebase/ui/database/paging/DatabasePagingKey.java
```diff
@@ -0,0 +1,19 @@
...
+    }
+
+    Object getChildValue() {
+        return mChildValue;
+    }
+
+    String getNodeKey() {
+        return mNodeKey;
+    }
+}
```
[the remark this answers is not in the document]
mikehardy: The PR description wasn't updated to match the visibility change in the commit addressing this

database/src/androidTest/java/com/firebase/ui/database/paging/DatabasePagingSourceTest.java
```diff
@@ -0,0 +1,117 @@
...
+    private static final int PAGE_SIZE = 3;
+    private static final int TOTAL_ITEMS = 6;
+
+    private DatabaseReference mRef;
+
+    @Before
+    public void setUp() throws InterruptedException {
+        FirebaseApp app;
+        try {
+            app = FirebaseApp.getInstance(APP_NAME);
```
mikehardy: curious if using s separate app is a deliberate choice, if not TestUtils has app init code in `TestUtils.getAppInstance()`

database/src/main/java/com/firebase/ui/database/paging/DatabasePagingSource.java
```diff
@@ -33,28 +35,36 @@ public DatabasePagingSource(Query query) {
...
+
         if (params.getKey() == null) {
             task = mQuery.limitToFirst(params.getLoadSize()).get();
         } else {
-            task = mQuery.startAt(null, params.getKey()).limitToFirst(params.getLoadSize() + 1).get();
+            DatabasePagingKey key = params.getKey();
+            if (pathIndex != null) {
+                task = startAtChildValue(key.getChildValue(), key.getNodeKey())
+                        .limitToFirst(params.getLoadSize() + 1).get();
+            } else {
```
mikehardy, line 61: For non-`PathIndex` queries this falls back to `startAt(null, nodeKey)`, which works for `orderByKey()` but likely not for `orderByValue()` — same cursor issue as #2000. Worth handling here (and in the `lastKey` construction ~L89) or updating the `setQuery()` docs to drop `orderByValue()` until supported.

database/src/main/java/com/firebase/ui/database/paging/DatabasePagingSource.java
```diff
@@ -89,19 +100,39 @@ public Single<LoadResult<String, DataSnapshot>> loadSingle(@NonNull LoadParams<S
...
+    @SuppressLint("RestrictedApi")
+    private Query startAtChildValue(Object childValue, String nodeKey) {
+        if (childValue instanceof String) {
+            return mQuery.startAt((String) childValue, nodeKey);
+        } else if (childValue instanceof Boolean) {
+            return mQuery.startAt((Boolean) childValue, nodeKey);
+        } else if (childValue instanceof Number) {
+            return mQuery.startAt(((Number) childValue).doubleValue(), nodeKey);
+        }
+        return mQuery.startAt(null, nodeKey);
```
mikehardy, line 143: hmm - looks like this fallback to `startAt(null, nodeKey)` is the same broken pattern as #2000 for `orderByChild()` — probably fine for missing child fields, but worth a one-line comment so it's not mistaken as the general case

mikehardy (commented):
looks like a perfect fix for the existing issue but may have a not-too-hard extension possibility to be even more general and fix the general case that 2000 was just a specific instance of

PR #2337, opened 2026-06-19
chore: add dependabot configuration for automated dependency updates

just1and0:
Adds dependabot.yml to keep Gradle dependencies and GitHub Actions up to date on a daily schedule with 7-day cooldown. PRs target the dependency-updates branch.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,7 +9,6 @@ updates:
       semver-major-days: 7
       semver-minor-days: 7
       semver-patch-days: 7
-    target-branch: dependency-updates
     labels:
       - 'dependencies'
 
@@ -22,6 +21,5 @@ updates:
       semver-major-days: 7
       semver-minor-days: 7
       semver-patch-days: 7
-    target-branch: dependency-updates
     labels:
       - 'dependencies'
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,11 +4,6 @@ updates:
     directory: '/'
     schedule:
       interval: 'daily'
-    cooldown:
-      default-days: 7
-      semver-major-days: 7
-      semver-minor-days: 7
-      semver-patch-days: 7
     labels:
       - 'dependencies'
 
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,25 @@
...
+
+  - package-ecosystem: 'github-actions'
+    directory: '/'
+    schedule:
+      interval: 'daily'
+    cooldown:
+      default-days: 7
+      semver-major-days: 7
+      semver-minor-days: 7
+      semver-patch-days: 7
```
mikehardy: this appears problematic, I wasn't aware github-actions ecosystem diverged in support for cooldown params, but

```
The property '#/updates/1/cooldown/semver-major-days' is not supported for the package ecosystem 'github-actions'.
The property '#/updates/1/cooldown/semver-minor-days' is not supported for the package ecosystem 'github-actions'.
The property '#/updates/1/cooldown/semver-patch-days' is not supported for the package ecosystem 'github-actions'.
```

https://github.com/firebase/FirebaseUI-Android/pull/2337/checks?check_run_id=82360741954

mikehardy (changes requested):
apparently some issue with what params are supported in what ecosystems

.github/dependabot.yml
```diff
@@ -0,0 +1,15 @@
+version: 2
+updates:
+  - package-ecosystem: 'gradle'
+    directory: '/'
+    schedule:
+      interval: 'daily'
+    labels:
+      - 'dependencies'
```
mikehardy, line 8: github-actions ecosystem doesn't support cooldown semver params, but gradle does - it should have maintained all cooldown params here, the problem was with github-actions ecosystem

.github/dependabot.yml
```diff
@@ -0,0 +1,15 @@
...
+
+  - package-ecosystem: 'github-actions'
+    directory: '/'
+    schedule:
+      interval: 'daily'
+    labels:
+      - 'dependencies'
```
mikehardy, line 15: github-actions does support cooldown days, it just doesn't support the semver params for them, the main cooldown days should have stayed for supply-chain attack hardening

mikehardy (commented):
sorry I didn't get to this before merge - but the cooldown params here were fully reverted vs just the semver params for github-actions

the param support table https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown- has the full list of which ecosystems support which parts

gemini gets this wrong because it has a training data cutoff that's pretty stale combined with no access to a web search tool to fetch updated docs

PR #2339, opened 2026-06-24
fix(auth): replace hardcoded loading state strings with translatable string resources

### LINKED ISSUE #2338: [Auth-UI] [v10.0.0-beta03] Untranslatable strings
Methods in the `com.firebase.ui.auth.configuration.auth_provider` package emit hardcoded English strings, like `updateAuthState(AuthState.Loading("Signing in user...")).` This would be reasonable for internal messages; however, `FirebaseAuthScreen` shows them to the user in `LoadingDialog` and `dialogController.CurrentDialog()`, which are not configurable.

demolaf:
Closes #2338 .

All `AuthState.Loading` messages emitted during auth operations (sign-in, sign-out, delete, phone, email, etc.) were hardcoded English strings. This replaces every instance with a string resource lookup via `Context.getString(R.string.fui_loading_*)`, making them overridable and translatable.

## Usage
You can either override using `strings.xml` or through a custom [AuthUIStringProvider](https://github.com/firebase/FirebaseUI-Android/blob/ac26fac620c7471921a213c1c12762d52aa8bd85/auth/src/main/java/com/firebase/ui/auth/configuration/string_provider/AuthUIStringProvider.kt#L48), see below examples:

**Using strings.xml:**
```xml
<resources>
    <string name="fui_loading_signing_in_anonymously" translatable="false">some random string here</string>
    <string name="fui_loading_deleting_account" translatable="false">some random string here</string>
</resources>
```

**Using AuthUIStringProvider:**

```kotlin
class CustomAuthUIStringProvider(
    private val defaultProvider: AuthUIStringProvider
) : AuthUIStringProvider by defaultProvider {
    override val loadingSigningInAnonymously: String
        get() = "Overriding signing in anonymously loading message..."
}

val customStringProvider = CustomAuthUIStringProvider(DefaultAuthUIStringProvider(applicationContext))

val configuration = authUIConfiguration {
    // ... (existing params)
    stringProvider = customStringProvider
}
```

## Preview
[override-loading-state-preview.webm](https://github.com/user-attachments/assets/e8d52304-f58c-4150-883f-4a3bf31f12b0)

### DIFF
--- a/auth/src/main/res/values/strings.xml
+++ b/auth/src/main/res/values/strings.xml
@@ -3,25 +3,25 @@
 
     <!-- Common -->
     <string name="fui_default_toolbar_title" translatable="false">@string/app_name</string>
-    <string name="fui_progress_dialog_loading" translation_description="Loading text in dialog">Loading…</string>
+    <string name="fui_progress_dialog_loading" translation_description="Loading text in dialog">Loading...</string>
     <string name="fui_initializing" translation_description="Text shown while the auth UI is initializing">Initializing</string>
 
     <!-- Loading state messages shown in the progress dialog during auth operations -->
-    <string name="fui_loading_signing_in_anonymously" translation_description="Progress message while signing in as a guest">Signing in as guest…</string>
-    <string name="fui_loading_signing_in_with_google" translation_description="Progress message while signing in with Google">Signing in with Google…</string>
-    <string name="fui_loading_signing_in_with_facebook" translation_description="Progress message while signing in with Facebook">Signing in with Facebook…</string>
-    <string name="fui_loading_signing_in_with_provider" translation_description="Progress message while signing in with a named provider. %1$s is the provider name.">Signing in with <xliff:g example="Twitter" id="provider_name">%1$s</xliff:g>…</string>
-    <string name="fui_loading_verifying_phone_number" translation_description="Progress message while verifying the user\'s phone number">Verifying phone number…</string>
-    <string name="fui_loading_submitting_verification_code" translation_description="Progress message while submitting the SMS verification code">Submitting verification code…</string>
-    <string name="fui_loading_signing_in_with_phone" translation_description="Progress message while completing phone number sign-in">Signing in with phone…</string>
-    <string name="fui_loading_creating_user" translation_description="Progress message while creating a new user account">Creating account…</string>
-    <string name="fui_loading_signing_in" translation_description="Progress message while signing in with email and password">Signing in…</string>
-    <string name="fui_loading_linking_credential" translation_description="Progress message while linking a credential to the current account">Signing in…</string>
-    <string name="fui_loading_sending_email_link" translation_description="Progress message while sending the sign-in email link">Sending sign-in link…</string>
-    <string name="fui_loading_signing_in_with_email_link" translation_description="Progress message while completing email link sign-in">Signing in with email link…</string>
-    <string name="fui_loading_sending_password_reset" translation_description="Progress message while sending the password reset email">Sending password reset email…</string>
-    <string name="fui_loading_signing_out" translation_description="Progress message while signing the user out">Signing out…</string>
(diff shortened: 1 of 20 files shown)


auth/src/main/res/values/strings.xml
```diff
@@ -4,6 +4,24 @@
     <!-- Common -->
     <string name="fui_default_toolbar_title" translatable="false">@string/app_name</string>
     <string name="fui_progress_dialog_loading" translation_description="Loading text in dialog">Loading…</string>
+    <string name="fui_initializing" translation_description="Text shown while the auth UI is initializing">Initializing</string>
+
+    <!-- Loading state messages shown in the progress dialog during auth operations -->
+    <string name="fui_loading_signing_in_anonymously" translation_description="Progress message while signing in as a guest">Signing in as guest…</string>
```
mikehardy: for unrelated reasons I've been reading the google style guide a lot lately and I remembered it had a rule on ellipses, but I didn't recall what it was

turns out the preference is for three periods in a row, not the ellipses character?

https://developers.google.com/style/ellipses#punctuation-and-spacing-of-ellipses

may apply here

auth/src/main/res/values/strings.xml
```diff
@@ -69,6 +87,7 @@
     <string name="fui_mfa_label_totp_authentication" translation_description="Label for the authenticator app MFA factor.">Authenticator app</string>
     <string name="fui_mfa_label_unknown_method" translation_description="Label for an unknown MFA factor.">Unknown method</string>
     <string name="fui_mfa_enrolled_on" translation_description="Label indicating the enrollment date of an MFA factor.">Enrolled on %1$s</string>
+    <string name="fui_enter_totp_code" translation_description="Prompt asking the user to enter a TOTP verification code">Enter TOTP code</string>
```
mikehardy, line 90: similarly, the style guide has opinions on abbreviations. This one is likely a grey area. Is TOTP familiar to the audience? To us yes, but to mass consumers...possibly not? This abbreviation my be better as something like "Enter one time code" or something?

https://developers.google.com/style/abbreviations#spelling-out

auth/src/test/java/com/firebase/ui/auth/FirebaseAuthUITest.kt
mikehardy, line 59: fun trick, you can ask robolectric to run a test in a different locale to verify that things *actually* work to load different translations

https://robolectric.org/using-qualifiers/

and then there is a ParameterizedRobolectricTestRunner you could use on an array of locales to probe default / implemented / not-implemented-fallback behavior sort of like 

https://github.com/ankidroid/Anki-Android/blob/c5bc834e72d1633268690f4e2900ffa7495bcf8b/AnkiDroid/src/test/java/com/ichi2/anki/browser/CardBrowserColumnTest.kt#L22-L47

Not a blocker by any means but maybe interesting to know about, perhaps interesting to try

mikehardy (approved):
implementation looks really sound
my comments are informational as I don't know how strongly google style guide applies here and how deep the testing contract is - certainly the tests prove the general paths as is

auth/src/main/res/values/strings.xml
demolaf: how it's able to pickup which is a character and three periods is interesting 😅, i'll remove those it serves no purpose anyways.

auth/src/main/res/values/strings.xml
demolaf: hmmm that's a valid point

auth/src/test/java/com/firebase/ui/auth/FirebaseAuthUITest.kt
demolaf: yup I actually used it in a separate test, best to apply it here too.

auth/src/main/res/values/strings.xml
mikehardy: I saw them visually because I read the guide recently and tried to automate it - in AnkiDroid we have a lint rule for it and after knowing the difference you can't unsee it 😅

PR #2349, opened 2026-06-30
chore: migrate to version catalog, drop Jetifier, fix Gradle 9.0 deprecations

demolaf:
Migrates all dependency and plugin versions from `buildSrc/Config.kt` to `gradle/libs.versions.toml`, enabling Dependabot to track and update them automatically. 

Removes the legacy `buildscript {}` block in favour of the modern `plugins {}` approach, and disables Jetifier after confirming all modules are clean via `checkJetifier`. 

Also upgrades `com.gradle.enterprise` to `com.gradle.develocity` 3.19.2, removes deprecated global `buildConfig` and `targetSdk` settings, and migrates `kotlinOptions.jvmTarget` to the `kotlin { compilerOptions {} }` DSL — fixing all actionable Gradle 9.0 deprecation warnings.


auth/build.gradle.kts
```diff
@@ -60,83 +61,86 @@ android {
     }
 
     testOptions {
+        targetSdk = Config.SdkVersions.target
         unitTests {
             isIncludeAndroidResources = true
         }
     }
```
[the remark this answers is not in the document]
demolaf: testOptions.targetSdk is valid AGP DSL for library modules

gradle/libs.versions.toml
```diff
@@ -0,0 +1,121 @@
...
+
+[libraries]
+# Kotlin
+kotlin-stdlib = { module = "org.jetbrains.kotlin:kotlin-stdlib-jdk8", version.ref = "kotlin" }
+kotlin-reflect = { module = "org.jetbrains.kotlin:kotlin-reflect", version.ref = "kotlin" }
+kotlinx-coroutines-core = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-core", version.ref = "coroutines" }
+kotlinx-coroutines-android = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-android", version.ref = "coroutines" }
+
+# AndroidX
+androidx-annotation = { module = "androidx.annotation:annotation", version = "1.2.0" }
```
mikehardy: most things in `[libraries]` now using reference to `[versions]` block, but some aren't (yet?)
appears to be an inconsistent style - should all versions be in `[versions]` block and used by reference in `[libraries]` ? lifecycle - for example - has the version in 3 spots here vs 1-spot-use-by-reference

mikehardy (changes requested):
this is nice - modern gradle setup +1!

one comment on full consolidation vs partial for version catalog - seems like one more pass to put all versions in versions block would make it completely clean

`checkJetifier` clean locally after a temporary local test-only jetifier re-enable to run the check

gradle/libs.versions.toml
demolaf: should be good now

mikehardy (approved):
a thing of beauty 🚢

PR #2354, opened 2026-07-01
fix(auth): remember authStateFlow(), add e2e coverage for phone linking

demolaf:
Firebase authentication screens (`PhoneAuthScreen`, `FirebaseAuthScreen`, `EmailAuthScreen`) call `authUI.authStateFlow().collectAsState(...)` directly in the composable body without `remember` which causes unnecessary recompositions.

## Changes
- Wrap `authUI.authStateFlow()` in `remember(authUI) { ... }` in `PhoneAuthScreen.kt`, `FirebaseAuthScreen.kt`, and `EmailAuthScreen.kt`, so the flow is subscribed once per screen instance instead of on every recomposition.
- Added e2e test to `PhoneAuthScreenTest.kt` for a returning, already-signed-in user linking a phone number.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ google-services.json
 crashlytics-build.properties
 auth/src/main/res/values/com_crashlytics_export_strings.xml
 *.log
+.kotlin/
--- a/README.md
+++ b/README.md
@@ -197,7 +197,7 @@ Then you can depend on snapshot versions:
 implementation 'com.firebaseui:firebase-ui-auth:$X.Y.Z-SNAPSHOT'
 ```
 
-You can see which `SNAPSHOT` builds are avaiable here:
+You can see which `SNAPSHOT` builds are available here:
 https://oss.jfrog.org/webapp/#/artifacts/browse/tree/General/oss-snapshot-local/com/firebaseui
 
 Snapshot builds come with absolutely no guarantees and we will close any issues asking to troubleshoot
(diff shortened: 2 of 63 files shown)


mikehardy (approved):
looks good - out of curiosity I probed whether `authUI` was stable, it's a plain class with no equals/hashCode override so it'll get reference identity and be stable ✅ 

there was a subtle interaction with multi-tenancy or custom stuff - if people use FirebaseAuthUI.create() inside a composable body `remember` won't have a stable key and the optimization here won't take effect. May be worth adding to the method note with a cross link to this PR for context so people can avoid a footgun https://github.com/firebase/FirebaseUI-Android/blob/version-10.0.0-beta03/auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUI.kt#L667-L668

not a code problem though for sure, no reason not to +1 on my review

demolaf:
> looks good - out of curiosity I probed whether `authUI` was stable, it's a plain class with no equals/hashCode override so it'll get reference identity and be stable ✅
> 
> there was a subtle interaction with multi-tenancy or custom stuff - if people use FirebaseAuthUI.create() inside a composable body `remember` won't have a stable key and the optimization here won't take effect. May be worth adding to the method note with a cross link to this PR for context so people can avoid a footgun https://github.com/firebase/FirebaseUI-Android/blob/version-10.0.0-beta03/auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUI.kt#L667-L668
> 
> not a code problem though for sure, no reason not to +1 on my review

Solid point, i'll add it to the doc comments as a "NOTE". thanks.

PR #2391, opened 2026-07-15
fix(auth): allow customizing top app bar colors via AuthUITheme

### LINKED ISSUE #2390: Allow customizing `AuthUITheme.topAppBarColors` from outside AuthUITheme
## Describe the feature request

I would like firebase-ui-auth to allow apps to customize the TopAppBar colors used by the built-in Compose auth screens.
At the moment, AuthUITheme.topAppBarColors is hardcoded internally and cannot be overridden from outside.
Current logic in AuthUITheme.kt:

```kotlin
  @get:Composable
  val topAppBarColors
      get() = TopAppBarDefaults.topAppBarColors(
          containerColor = MaterialTheme.colorScheme.primary,
          titleContentColor = MaterialTheme.colorScheme.onPrimary,
          navigationIconContentColor = MaterialTheme.colorScheme.onPrimary,
      )
```

Because of this, apps that integrate FirebaseUI Compose auth screens cannot align the app bar colors with their own theme system unless they replace the whole screen content themselves.

## Suggested API options

Any of these would solve the problem:

### Option 1: Add topAppBarColors to AuthUITheme

Example:

```kotlin
  data class AuthUITheme(
      val colorScheme: ColorScheme = ...,
      val typography: Typography = ...,
      val shapes: Shapes = ...,
      val topAppBarColors: TopAppBarColors? = null,
  )
```

And internally:
```kotlin
  val resolvedTopAppBarColors =
      authUITheme.topAppBarColors ?: TopAppBarDefaults.topAppBarColors(...)
```

### Option 2: Add a topAppBarColors parameter to screen composables

Example:
```kotlin
  PhoneAuthScreen(
      ...,
      topAppBarColors = TopAppBarDefaults.topAppBarColors(...)
  )
```

demolaf:
Fixes #2390.

`AuthUITheme.topAppBarColors` was a companion-object getter derived from the ambient `MaterialTheme.colorScheme`, so there was no way to override it per `AuthUITheme` instance — apps couldn't align the auth screens' top app bar with their own branding.

## Usage
```kotlin
val customTheme = AuthUITheme.Default.copy(
    topAppBarColors = TopAppBarColors(
        containerColor = Color(0xFFFFA000), // firebase primary color - amber
        scrolledContainerColor = Color(0xFFFFA000),
        navigationIconContentColor = Color.White,
        titleContentColor = Color.White,
        actionIconContentColor = Color.White,
        subtitleContentColor = Color.White,
    )
)

val configuration = authUIConfiguration {
    context = applicationContext
    theme = customTheme
    // ...
}
```

## Preview
Screenshot_1784126283

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/configuration/theme/AuthUITheme.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/configuration/theme/AuthUITheme.kt
@@ -19,6 +19,7 @@ import androidx.compose.material3.ColorScheme
 import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Shapes
+import androidx.compose.material3.TopAppBarColors
 import androidx.compose.material3.TopAppBarDefaults
 import androidx.compose.material3.Typography
 import androidx.compose.material3.darkColorScheme
@@ -81,6 +82,12 @@ class AuthUITheme(
      * ```
      */
     val providerButtonShape: Shape? = null,
+
+    /**
+     * Custom colors for the top app bar shown on auth screens. If null, falls back to
+     * colors derived from [colorScheme] (see [AuthUITheme.topAppBarColors]).
+     */
+    val topAppBarColors: TopAppBarColors? = null,
 ) {
 
     /**
@@ -91,6 +98,7 @@ class AuthUITheme(
      * @param shapes The shapes to use. Defaults to this theme's shapes.
      * @param providerStyles Custom styling for individual providers. Defaults to this theme's provider styles.
      * @param providerButtonShape Default shape for provider buttons. Defaults to this theme's provider button shape.
+     * @param topAppBarColors Custom top app bar colors. Defaults to this theme's top app bar colors.
      * @return A new AuthUITheme instance with the specified properties.
      */
     fun copy(
@@ -99,13 +107,15 @@ class AuthUITheme(
         shapes: Shapes = this.shapes,
         providerStyles: Map<String, ProviderStyle> = this.providerStyles,
         providerButtonShape: Shape? = this.providerButtonShape,
+        topAppBarColors: TopAppBarColors? = this.topAppBarColors,
     ): AuthUITheme {
         return AuthUITheme(
             colorScheme = colorScheme,
             typography = typography,
             shapes = shapes,
             providerStyles = providerStyles,
-            providerButtonShape = providerButtonShape
+            providerButtonShape = providerButtonShape,
+            topAppBarColors = topAppBarColors
         )
     }
 
@@ -118,6 +128,7 @@ class AuthUITheme(
         if (shapes != other.shapes) return false
         if (providerStyles != other.providerStyles) return false
         if (providerButtonShape != other.providerButtonShape) return false
+        if (topAppBarColors != other.topAppBarColors) return false
 
         return true
     }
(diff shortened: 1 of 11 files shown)


auth/src/main/java/com/firebase/ui/auth/configuration/theme/AuthUITheme.kt
```diff
@@ -81,6 +82,12 @@ class AuthUITheme(
      * ```
      */
     val providerButtonShape: Shape? = null,
+
+    /**
+     * Custom colors for the top app bar shown on auth screens. If null, falls back to
+     * colors derived from [colorScheme] (see [AuthUITheme.topAppBarColors]).
+     */
+    val topAppBarColors: TopAppBarColors? = null,
```
russellwheatley, line 89: LocalAuthUITheme.current.topAppBarColors ?: AuthUITheme.topAppBarColors is copied into all 6 screen files - can we pull that into one resolved property on AuthUITheme instead? Saves updating 6 files next time this logic changes.

auth/src/test/java/com/firebase/ui/auth/configuration/theme/AuthUIThemeTest.kt
```diff
@@ -313,6 +316,59 @@ class AuthUIThemeTest {
...
+    // ========================================================================
+
+    @Test
+    fun `Default theme has null topAppBarColors`() {
+        assertThat(AuthUITheme.Default.topAppBarColors).isNull()
+    }
+
+    @OptIn(ExperimentalMaterial3Api::class)
+    @Test
+    fun `Copy with custom topAppBarColors applies correctly`() {
```
russellwheatley, line 328: The new tests just re-check the theme object, not the actual screens - since the fallback's duplicated everywhere, this won't catch a screen missing it. Worth testing off the shared helper once it's extracted (see point 1)

russellwheatley (changes requested):
A few bits of feedback 

SelectFactorUI's TopAppBar in MfaEnrollmentDefaults.kt doesn't take colors at all, so it won't pick up topAppBarColors (or even the existing default). Can you wire that one up too, otherwise the 2FA screen won't match once someone sets a custom color.

see auth/src/main/java/com/firebase/ui/auth/ui/screens/MfaEnrollmentDefaults.kt:255 (the TopAppBar call inside SelectFactorUI)

russellwheatley (changes requested):
A few bits of feedback

`topAppBarColors` on `AuthUITheme` (and the matching `copy()` param) is typed as `TopAppBarColors`, which is still `@ExperimentalMaterial3Api`, but neither is annotated. Anyone who actually sets it - the whole point of this PR -gets a compile error with no hint why unless they already have `@OptIn` in scope elsewhere. Can you propagate the opt-in onto the property and the `copy()` param?

see auth/src/main/java/com/firebase/ui/auth/configuration/theme/AuthUITheme.kt:90 and :110

Same issue bites the demo itself — `TopAppBarColors(...)` is constructed inside `onCreate`/`setContent`, but the only `@OptIn(ExperimentalMaterial3Api::class)` in the file is on the unrelated composable further down. As it stands this won't compile.

see app/src/main/java/com/firebaseui/android/demo/HighLevelApiDemoActivity.kt:88 (missing `@OptIn` on `onCreate`, existing one is at line 252)

Nit: README and demo both construct `TopAppBarColors(...)` with all six params instead of going through `TopAppBarDefaults.topAppBarColors(...)` like the rest of the codebase does (including our own default at AuthUITheme.kt:247). Worth switching for consistency and so we don't break if Material3 adds params later.

demolaf:
> topAppBarColors on AuthUITheme (and the matching copy() param) is typed as TopAppBarColors, which is still @ExperimentalMaterial3Api, but neither is annotated. Anyone who actually sets it - the whole point of this PR -gets a compile error with no hint why unless they already have @OptIn in scope elsewhere. Can you propagate the opt-in onto the property and the copy() param?

I think it compiles fine. In newer versions of compose bom the APIs are now stable, before the compose bom was updated this (i.e. `@OptIn(ExperimentalMaterial3Api::class)`) was needed but i've taken it out now.

demolaf:
> Nit: README and demo both construct TopAppBarColors(...) with all six params instead of going through TopAppBarDefaults.topAppBarColors(...) like the rest of the codebase does (including our own default at AuthUITheme.kt:247). Worth switching for consistency and so we don't break if Material3 adds params later.

Valid point, i'll make the necessary changes to the demo and README.

demolaf:
@russellwheatley anything else blocking this from being merged?

russellwheatley:
@demolaf - just the merge conflict and I think we're good to go 👍

russellwheatley (approved):

PR #2397, opened 2026-07-17
refactor(auth)!: hide sign-up button instead of showing disabled state with tooltip. BREAKING - Removed unused `newAccountsDisabledTooltip` from public interface `AuthUIStringProvider`

### LINKED ISSUE #2382: Inconsistent behavior when AllowNewAccounts = false in the AuthProvider.Email UI between the old and new library versions
There is inconsistent behavior when AllowNewAccounts = false in the AuthProvider.Email UI between the old and new library versions.

While updating the library from version 8.x.x to 10.0.0-beta03, I encountered the following problem:

It seems that the new version of the library behaves quite differently regarding the "setAllowNewAccounts" option in AuthProvider.Email.

When using the old version, the sign-up button disappears when AllowNewAccounts is set to false.

When AllowNewAccounts is set to false using the new version of the library, the sign-up button remains visible but is always disabled. An ugly pop-up message appears as soon as the sign-in UI is displayed: "This button is currently disabled because new accounts are not allowed." 

This message is not relevant to my app, which does not allow sign-up at all. It is especially problematic compared to the cleaner UI obtained with version 8.x.x of the library.

Image

demolaf:
Fixes #2382.

When `AuthProvider.Email.isNewAccountsAllowed` is false, the sign-up button on the email sign-in screen was shown disabled with a tooltip explaining why — a UX regression from the pre-Compose UI, where the button was hidden entirely. This restores the hidden-button behavior.

- `SignInUI.kt`: sign-up button is now only emitted when `provider.isNewAccountsAllowed` is true; removed the TooltipBox/tooltip wrapper.
- Removed the now-unused `newAccountsDisabledTooltip` property from `AuthUIStringProvider`/`DefaultAuthUIStringProvider` and its string resource from all locale files (breaking change for anyone who overrode this property in a custom string provider).
- Added `SignInUITest.kt `covering both the hidden and enabled states of the sign-up button.

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/configuration/string_provider/AuthUIStringProvider.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/configuration/string_provider/AuthUIStringProvider.kt
@@ -582,9 +582,6 @@ interface AuthUIStringProvider {
     /** ToS and Privacy Policy combined message with placeholders for links */
     fun tosAndPrivacyPolicy(termsOfServiceLabel: String, privacyPolicyLabel: String): String
 
-    /** Tooltip message shown when new account sign-up is disabled */
-    val newAccountsDisabledTooltip: String
-
     /** Tooltip message shown when MFA is disabled */
     val mfaDisabledTooltip: String
 
--- a/auth/src/main/java/com/firebase/ui/auth/configuration/string_provider/DefaultAuthUIStringProvider.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/configuration/string_provider/DefaultAuthUIStringProvider.kt
@@ -523,9 +523,6 @@ class DefaultAuthUIStringProvider(
     override fun tosAndPrivacyPolicy(termsOfServiceLabel: String, privacyPolicyLabel: String): String =
         localizedContext.getString(R.string.fui_tos_and_pp, termsOfServiceLabel, privacyPolicyLabel)
 
-    override val newAccountsDisabledTooltip: String
-        get() = localizedContext.getString(R.string.fui_new_accounts_disabled_tooltip)
-
     override val mfaDisabledTooltip: String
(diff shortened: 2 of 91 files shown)


demolaf:
> ## Code Review
> This pull request hides the sign-up button entirely when new accounts are not allowed, rather than displaying a disabled button with a tooltip. This change simplifies the UI in `SignInUI.kt` and removes the now-unused `newAccountsDisabledTooltip` property and its localized string resources across multiple languages. Additionally, a new unit test class `SignInUITest.kt` has been introduced to verify this behavior. The reviewer noted that removing `newAccountsDisabledTooltip` from the public `AuthUIStringProvider` interface is a breaking change for external implementers and suggested deprecating it with a default implementation instead to preserve backward compatibility.
> 
> I am having trouble creating individual review comments. Click here to see my feedback.
> ### auth/src/main/java/com/firebase/ui/auth/configuration/string_provider/AuthUIStringProvider.kt (585-586)
> medium
> 
> Removing `newAccountsDisabledTooltip` from the public `AuthUIStringProvider` interface is a breaking change for any library consumers who implement this interface. To maintain backward compatibility, consider deprecating the property and providing a default implementation (getter) instead of removing it entirely.
> 
> ```
>     /** Tooltip message shown when new account sign-up is disabled */
>     @Deprecated(
>         message = "No longer used as the sign-up button is hidden when new accounts are disabled.",
>         level = DeprecationLevel.WARNING
>     )
>     val newAccountsDisabledTooltip: String
>         get() = ""
> ```

is this necessary if we are still in beta?

russellwheatley (commented):
LGTM - I suspect we don't have to deprecate the API and make backwards compatible, but I think we should ask Rosario if we need to. What do you think, @thatfiredev?

demolaf:
@russellwheatley I think it's fine since we are still in beta, if there are no blockers can we get this merged?

russellwheatley (approved):

PR #2413, opened 2026-07-22
fix(auth): handle edge-to-edge insets in MFA challenge and auth screens

### LINKED ISSUE #2403: The default UI of the MFA does not handle edge-to-edge properly.
The default UI of the MFA does not handle edge-to-edge properly.

'Verify your code' interferes with the camera hole, and the status bar icons are not visible due to the white-on-white issue.

Image

demolaf:
Fixes #2403.

The MFA challenge screen had no inset handling, so its title collided with the status bar/camera cutout on edge-to-edge devices. Also found a related, opposite bug in six other auth screens combining Scaffold's innerPadding with a redundant, double-counted safeDrawingPadding() call.

- `MfaChallengeDefaults.kt`: wrapped `DefaultMfaChallengeContent` in a `Scaffold`
- `SignInUI`, `SignUpUI`, `SignInEmailLinkUI`, `ResetPasswordUI`, `EnterPhoneNumberUI`, `EnterVerificationCodeUI`: removed the redundant `.safeDrawingPadding()`, added a uniform `16.dp` margin
- `FirebaseAuthScreen.customMethodPickerLayout` now renders full-screen (was Scaffold-confined) so fully custom layouts can go edge-to-edge; direct behavior change since library is pre-1.0

## Preview
Screenshot_1784708851

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/ui/screens/FirebaseAuthScreen.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/screens/FirebaseAuthScreen.kt
@@ -20,6 +20,7 @@ import androidx.compose.animation.core.tween
 import androidx.compose.animation.fadeIn
 import androidx.compose.animation.fadeOut
 import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxSize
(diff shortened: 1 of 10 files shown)


auth/src/main/java/com/firebase/ui/auth/ui/screens/FirebaseAuthScreen.kt
```diff
@@ -104,6 +105,14 @@ import kotlinx.coroutines.tasks.await
  * @param authenticatedContent Optional slot that allows callers to render the authenticated
  * state themselves. When provided, it receives the current [AuthState] alongside an
  * [AuthSuccessUiContext] containing common callbacks (sign out, manage MFA, reload user).
+ * @param customMethodPickerLayout Optional slot that fully replaces the method-picker screen.
+ * When provided, it renders as the *entire* screen content — edge-to-edge, with no logo, no
```
russellwheatley, line 109: This says customMethodPickerLayout always renders full-screen with no ToS footer, but ReauthSheetContent's call site doesn't match, it's still Scaffold-wrapped and still shows the footer. Can you scope this KDoc to the primary flow only, or make ReauthSheetContent consistent with the new behavior?

auth/src/main/java/com/firebase/ui/auth/ui/screens/FirebaseAuthScreen.kt
```diff
@@ -207,19 +216,24 @@ fun FirebaseAuthScreen(
...
-                            logo = logoAsset,
-                            termsOfServiceUrl = configuration.tosUrl,
-                            privacyPolicyUrl = configuration.privacyPolicyUrl,
-                            lastSignInPreference = lastSignInPreference.value,
-                            customLayout = customMethodPickerLayout,
-                            termsConfiguration = customMethodPickerTermsConfiguration,
-                            onProviderSelected = onProviderSelected,
-                        )
+                    if (customMethodPickerLayout != null) {
+                        Box(modifier = modifier.fillMaxSize()) {
```
russellwheatley, line 220: No test covers the new full-screen/no-chrome behavior here. Worth adding one that asserts the logo/footer aren't rendered when customMethodPickerLayout is set.

russellwheatley (changes requested):
A couple of suggestions on the customMethodPickerLayout changes below.

Also, this branch looks like it's stacked on top of the unmerged display-name-validation fix (d382bd5b) rather than rebased onto version-10.0.0-beta04, so the diff here currently pulls in an unrelated SignUpUI/SignUpUITest.kt change. Can you rebase onto the base branch before merge so the diff only has your two commits?

russellwheatley (approved):
LGTM

PR #2414, opened 2026-07-22
sample(auth): full customization demo

demolaf:
Extends the full customization demo to supply a custom UI for every `FirebaseAuthScreen` content slot.

## Changes

- Each email mode becomes its own Navigation 3 route instead of a mode swap inside one screen, so the steps get real transitions, a working predictive-back gesture, and an address that survives moving between them.
- Reauthentication renders its own compact confirmation step, because the library composes that step inside a modal bottom sheet where a full sign-in page does not fit. Declining it now reads as a cancellation rather than a failure.
- A returning email link routes back to whichever demo sent it, so several auth demos can sit behind one deep link.

## Preview

- Full Customization Flow

https://github.com/user-attachments/assets/72fb5081-fa4c-4586-8588-a992f67eac25

- Email Link Sign In Flow

https://github.com/user-attachments/assets/be1002c7-3eb0-40e7-964c-faed7b94e90b

- Reauthentication Flow

https://github.com/user-attachments/assets/72d98d6b-889a-4ba2-8ff6-067b17c3bd81

- MFA Flow (Enrollment)

https://github.com/user-attachments/assets/1059065f-88e4-4a7c-b5a6-977e71d3524e

- MFA Flow (Challenge)

https://github.com/user-attachments/assets/7cb5a7e7-1b1b-4e74-ad1c-17da60ca7a35

- Error Dialog (Inherits font change)
Screenshot_1784723834


russellwheatley (approved):
LGTM. Two small non-blocking things worth a look:

- `MainActivity.USE_AUTH_EMULATOR` flips from `true` to `false` — unrelated to this PR's scope, might be intentional but flagging in case it's a leftover from local testing.
- `PhoneEntryStep`'s CTA reads "Sign Up" but the screen is the general phone sign-in entry (title says "Login by phone number"); looks copy-pasted from the email entry step's create-account button.

PR #2415, opened 2026-07-23
fix(auth): stale one-off AuthState no longer leaks across screen instances

demolaf:
A one-off AuthState (Error, Cancelled, SMSAutoVerified, PasswordResetLinkSent, EmailSignInLinkSent) written to FirebaseAuthUI's singleton state flow was never reset back to Idle after being consumed, so it got redelivered to a completely new screen/Activity that had nothing to do with the original event — reproduced by triggering a phone-auth error, backing out of the Activity, and re-entering: a stale "Authentication Error" dialog appeared immediately, before touching anything.

FirebaseAuthScreen, EmailAuthScreen, and PhoneAuthScreen now reset auth state back to Idle immediately after consuming Error/Cancelled/SMSAutoVerified in the same effect that reacts to them. TopLevelDialogController no longer keys its remember on the live auth state (or on an unmemoized stringProvider), since either caused the controller — and the dialog it was showing — to be torn down on the very next recomposition.

- `EmailAuthScreen.kt`: `resetLinkSent`/`emailSignInLinkSent` are now latched into local state instead of derived directly from `authState`, since their confirmation dialogs latch on the boolean's transition and would otherwise close immediately when the reset above fires.

Added regression coverage in `FirebaseAuthUIAuthStateTest` and a new `TopLevelDialogControllerTest` proving stale state doesn't leak to a fresh collector and the dialog survives auth state changes.

## Behavior

- Before

[bug-demo.webm](https://github.com/user-attachments/assets/cd050a85-c7bd-472a-8d20-ca88fba87325)

- After

[fix-dialog.webm](https://github.com/user-attachments/assets/bb02fec9-7c74-41a6-8a05-e4f92f163457)

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/ui/screens/FirebaseAuthScreen.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/screens/FirebaseAuthScreen.kt
@@ -458,24 +458,30 @@ fun FirebaseAuthScreen(
                         pendingResolver.value = null
                         pendingLinkingCredential.value = null
 
-                        // If reauth just completed, execute the pending retry and skip normal success handling
-                        pendingReauthOperation.value?.let { retry ->
-                            pendingReauthOperation.value = null
-                            pendingReauthConfig.value = null
-                            pendingReauthState.value = null
-                            // Lock the state to Loading before launching the retry so no
-                            // intermediate Success emission can navigate to AuthRoute.Success.
-                            authUI.updateAuthState(AuthState.Loading())
-                            coroutineScope.launch {
-                                try {
-                                    retry(context)
-                                } catch (e: kotlinx.coroutines.CancellationException) {
-                                    throw e
-                                } catch (e: Exception) {
-                                    authUI.updateAuthState(AuthState.Error(e))
+                        // If reauth just completed, execute the pending retry and skip normal success handling.
+                        // Guard on !previous.isNotification: authStateFlow() masks a notification's
+                        // self-consumed Idle (e.g. a failed reauth's Error) back into Success while the
+                        // user stays signed in — that masked Success must not be mistaken for a
+                        // genuinely completed reauth.
+                        if (!previous.isNotification) {
+                            pendingReauthOperation.value?.let { retry ->
+                                pendingReauthOperation.value = null
+                                pendingReauthConfig.value = null
+                                pendingReauthState.value = null
+                                // Lock the state to Loading before launching the retry so no
+                                // intermediate Success emission can navigate to AuthRoute.Success.
+                                authUI.updateAuthState(AuthState.Loading())
+                                coroutineScope.launch {
+                                    try {
+                                        retry(context)
+                                    } catch (e: kotlinx.coroutines.CancellationException) {
+                                        throw e
+                                    } catch (e: Exception) {
+                                        authUI.updateAuthState(AuthState.Error(e))
+                                    }
                                 }
+                                return@LaunchedEffect
                             }
-                            return@LaunchedEffect
                         }
 
                         state.result?.let { result ->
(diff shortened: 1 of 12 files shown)


auth/src/main/java/com/firebase/ui/auth/ui/screens/FirebaseAuthScreen.kt
```diff
@@ -646,6 +649,8 @@ fun FirebaseAuthScreen(
                             // Dialog dismissed
                         }
                     )
+                    // Consumed immediately so this doesn't leak to a freshly created screen.
+                    authUI.updateAuthState(AuthState.Idle)
```
russellwheatley, line 666: This reset also flows through the nav-sync effect's `is AuthState.Idle` branch above, which navigates back to startRoute whenever we're not already there. With multiple providers (or isProviderChoiceAlwaysShown), that means a plain error on Email/Phone would kick the user back to the method picker screen right after the dialog shows, since currentRoute != startRoute in that case. Can you add a test with a multi-provider config that triggers an error on Email/Phone and confirms we stay on that screen after the dialog is dismissed?

e2eTest/src/test/java/com/firebase/ui/auth/ui/screens/EmailAuthScreenTest.kt
```diff
@@ -585,22 +592,24 @@ class EmailAuthScreenTest {
         shadowOf(Looper.getMainLooper()).idle()
         composeAndroidTestRule.waitForIdle()
 
-        // Wait for auth state to transition to EmailSignInLinkSent
-        println("TEST: Waiting for auth state change... Current state: $currentAuthState")
+        // Wait for the "email link sent" dialog to appear, rather than polling currentAuthState:
```
russellwheatley, line 592: Nice fix here. The sibling test above, `trouble signing in emits PasswordResetLinkSent auth state and shows dialog`, still polls a second authStateFlow collector and asserts on the transient PasswordResetLinkSent value, so it has the same race this test just fixed. Can you apply the same dialog-based wait there too?

russellwheatley (changes requested):
LGTM overall, nice fix for the stale auth state leak. Just a couple of concerns before merging, left inline.

auth/src/main/java/com/firebase/ui/auth/ui/screens/FirebaseAuthScreen.kt
demolaf: Nice catch — reproduced it with a new multi-provider test, then fixed by having the nav-sync effect skip its `startRoute` navigation when the previous state was a notification — via a new `AuthState.isNotification` property enforced on every subclass, so it can't be silently missed for a future one-off state.

auth/src/main/java/com/firebase/ui/auth/ui/screens/FirebaseAuthScreen.kt
```diff
@@ -565,7 +572,9 @@ fun FirebaseAuthScreen(
                         pendingResolver.value = null
                         pendingLinkingCredential.value = null
                         lastSuccessfulUserId.value = null
-                        if (currentRoute != startRoute.route) {
+                        // A notification resets to Idle purely to avoid leaking to a freshly
+                        // created screen — that's not a request to leave the current one.
+                        if (!previous.isNotification && currentRoute != startRoute.route) {
```
russellwheatley: Nav is correctly gated on `!previous.isNotification`, but the pending clears just above still run unconditionally on every Idle- including Idle that only exists because we just consumed a notification (`Error`/`Cancelled`/etc.).

Before this PR, `Error` stayed `Error`, so that teardown didn't fire. Now `Error → Idle` will wipe `pendingReauthConfig` / `pendingLinkingCredential` / `pendingResolver` mid-flow. Concrete case: reauth sheet shares the same `authUI`, so a wrong-password `Error` during reauth resets to Idle here and dismisses the sheet.

Please gate those clears the same way as navigation (`if (!previous.isNotification) { ... }`), or otherwise treat notification-consume Idle as "don't tear down session". A regression test for reauth (and ideally linking) would help lock this in.

russellwheatley (changes requested):
Thanks for addressing the nav-bounce and e2e race. One more concern on the Idle path before merging.

auth/src/main/java/com/firebase/ui/auth/ui/screens/FirebaseAuthScreen.kt
demolaf: i wrote an e2e test that validates this bug actually surfaces, this should be resolved now.

russellwheatley (approved):

PR #2416, opened 2026-07-23
fix(auth): mitigate tapjacking on the auth activity window

### LINKED ISSUE #2041: Security issue with tapjacking
### Step 1: Are you in the right place?

  Yes

  * Android device: N/A
  * Android OS version: 21 up to 32
  * Google Play Services version: 4.3.10
  * Firebase/Play Services SDK version: 21.0.5
  * FirebaseUI version: 8.0.1
  
image

image

#### Steps to reproduce:

Use FirebaseUI Auth UI
  
#### Observed Results:

Security risk level is medium
  
#### Expected Results:

Possible use of `filterTouchesWhenObscure` set to true?
  
#### Relevant Code:

  ```
  // TODO(you): code here to reproduce the problem
  ```

just1and0:
## Summary
- Adds `window.decorView.filterTouchesWhenObscured = true` in `FirebaseAuthActivity.onCreate()` to mitigate tapjacking/clickjacking attacks where an overlay intercepts user touches on the auth UI.
- `FirebaseAuthActivity` is the only activity in the auth module (v10 is Compose-based), so this single-line change covers the entire auth flow.

Fixes #2041

## Test plan
- [ ] Launch the auth flow and verify touch input works normally without overlays
- [ ] Verify that touches are ignored when another app draws an overlay on top of the auth UI

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/FirebaseAuthActivity.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/FirebaseAuthActivity.kt
@@ -77,6 +77,7 @@ class FirebaseAuthActivity : ComponentActivity() {
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+        window.decorView.filterTouchesWhenObscured = true
         enableEdgeToEdge()
 
         // Extract configuration and auth instance from cache using UUID key
--- a/auth/src/main/java/com/firebase/ui/auth/ui/components/ErrorRecoveryDialog.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/components/ErrorRecoveryDialog.kt
@@ -19,7 +19,9 @@ import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
 import androidx.compose.material3.TextButton
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.SideEffect
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.LocalView
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.window.DialogProperties
 import com.firebase.ui.auth.AuthException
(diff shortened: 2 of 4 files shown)


demolaf (approved):
Looks good, the target branch should be `version-10.0.0-beta04`

auth/src/main/java/com/firebase/ui/auth/FirebaseAuthActivity.kt
```diff
@@ -77,6 +77,7 @@ class FirebaseAuthActivity : ComponentActivity() {
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+        window.decorView.filterTouchesWhenObscured = true
```
russellwheatley, line 80: This covers the activity's own window, but the Compose Dialogs in this module (e.g. ReauthenticationDialog for password re-entry) create their own separate native Window that this flag doesn't propagate to. Could you also set filterTouchesWhenObscured on those dialog windows, or note in the description that they're out of scope for this fix?

russellwheatley, line 80: FirebaseAuthActivityTest.kt already has a Robolectric harness for this activity. Worth adding a small test asserting filterTouchesWhenObscured is true after onCreate so this can't silently regress.

russellwheatley (changes requested):
Nice one 👍 One gap worth flagging on scope, plus a small test suggestion

auth/src/main/java/com/firebase/ui/auth/FirebaseAuthActivity.kt
just1and0: Fixed in 850d011 — added filterTouchesWhenObscured to both ReauthenticationDialog and ErrorRecoveryDialog via a SideEffect that sets the flag on the dialog's root view.

just1and0: Added in 850d011 — new Robolectric test asserting filterTouchesWhenObscured is true on the activity window after onCreate.

russellwheatley (approved):
LGTM

PR #2417, opened 2026-07-23
fix(internal): append CHAR_LIMIT to translation descriptions in export script

### LINKED ISSUE #1613: Translation script needs to be updated
We need to append "[CHAR_LIMIT=xxx]" at the end of each translation description. We can probably default this to ~1.5x the English length.

just1and0:
## Summary
- Updates `export_translations.py` to automatically append `[CHAR_LIMIT=xxx]` to each `translation_description` that doesn't already have one
- Limit defaults to ~1.5x the English string length, rounded up to the nearest 5
- Handles plurals `<item>` tags individually (each item gets its own limit)
- Decodes XML entities (`&amp;` → `&`) and Android escape sequences (`\'`, `\n`, ` `) before calculating visible text length
- Preserves existing manually-set `[CHAR_LIMIT=...]` values
- Fixes Python 2 `print` statement for Python 3 compatibility

Fixes #1613

## Test plan
- [ ] Run `python3 scripts/translations/export_translations.py` and verify CHAR_LIMIT is appended to all descriptions with visible text
- [ ] Verify strings with existing `[CHAR_LIMIT=...]` are not modified
- [ ] Verify empty override strings (e.g. `fui_error_invalid_credentials`) are correctly skipped
- [ ] Verify plurals items each get their own CHAR_LIMIT

### DIFF
--- a/scripts/translations/export_translations.py
+++ b/scripts/translations/export_translations.py
@@ -14,11 +14,24 @@ CHAR_LIMIT_PATTERN = re.compile(r'\[CHAR_LIMIT=\d+\]')
 TRANSLATION_DESC_PATTERN = re.compile(r'(translation_description="[^"]*?)(")')
 XML_TAG_PATTERN = re.compile(r'<[^>]+>')
 STRING_VALUE_PATTERN = re.compile(r'>([^<]*(?:<xliff:g[^>]*>[^<]*</xliff:g>[^<]*)*)</string>')
+ITEM_VALUE_PATTERN = re.compile(r'>([^<]*(?:<xliff:g[^>]*>[^<]*</xliff:g>[^<]*)*)</item>')
+
+XML_ENTITIES = {'&amp;': '&', '&lt;': '<', '&gt;': '>', '&quot;': '"', '&apos;': "'"}
+ESCAPE_PATTERN = re.compile(r'\\(u[0-9A-Fa-f]{4}|n|t|\'|")')
+
+
+def _decode_entities(text):
+    """Decode XML entities and Android escape sequences to get true visible length."""
+    for entity, char in XML_ENTITIES.items():
+        text = text.replace(entity, char)
+    text = ESCAPE_PATTERN.sub('X', text)
+    return text
 
 
 def _extract_visible_text(value):
-    """Strip XML tags to get the visible text length."""
-    return XML_TAG_PATTERN.sub('', value).strip()
+    """Strip XML tags and decode entities to get the visible text length."""
+    text = XML_TAG_PATTERN.sub('', value).strip()
+    return _decode_entities(text)
 
 
 def _calculate_char_limit(english_text):
@@ -26,6 +39,23 @@ def _calculate_char_limit(english_text):
     length = len(english_text)
     limit = math.ceil(length * 1.5)
     return int(math.ceil(limit / 5.0) * 5)
+
+
+def _add_char_limit(text, value_pattern):
+    """Add [CHAR_LIMIT=xxx] to translation_description if missing."""
+    if 'translation_description=' not in text:
+        return text
+    if CHAR_LIMIT_PATTERN.search(text):
+        return text
+    match = value_pattern.search(text)
+    if not match:
+        return text
+    visible = _extract_visible_text(match.group(1))
+    if not visible:
+        return text
+    limit = _calculate_char_limit(visible)
+    return TRANSLATION_DESC_PATTERN.sub(
+        r'\1 [CHAR_LIMIT=%d]\2' % limit, text, count=1)
 
 
 class ExportTranslationsScript(BaseStringScript):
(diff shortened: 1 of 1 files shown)


scripts/translations/export_translations.py
```diff
@@ -9,20 +10,76 @@
...
+
+        if type == self.TYPE_STR:
+            joined = _add_char_limit(joined, STRING_VALUE_PATTERN)
+        elif type == self.TYPE_PLUR:
+            result_lines = []
+            for l in joined.split('\n'):
+                if '<item ' in l and '</item>' in l:
+                    l = _add_char_limit(l, ITEM_VALUE_PATTERN)
+                result_lines.append(l)
+            joined = '\n'.join(result_lines)
```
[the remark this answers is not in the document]
just1and0: Good catch — applied this suggestion in b8eaae98. The `re.sub` with `re.DOTALL` correctly handles multiline `<item>` tags without the fragile line-by-line split.

scripts/translations/export_translations.py
```diff
@@ -9,20 +10,76 @@
 PREFIXED_NAME_START = 'name="fui_'
 UNPREFIXED_NAME_START = 'name="'
 
+CHAR_LIMIT_PATTERN = re.compile(r'\[CHAR_LIMIT=\d+\]')
+TRANSLATION_DESC_PATTERN = re.compile(r'(translation_description="[^"]*?)(")')
+XML_TAG_PATTERN = re.compile(r'<[^>]+>')
+STRING_VALUE_PATTERN = re.compile(r'>([^<]*(?:<xliff:g[^>]*>[^<]*</xliff:g>[^<]*)*)</string>')
+ITEM_VALUE_PATTERN = re.compile(r'>([^<]*(?:<xliff:g[^>]*>[^<]*</xliff:g>[^<]*)*)</item>')
```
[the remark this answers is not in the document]
just1and0: Declining this one — the simplified pattern `>(.*)</string>` with `re.DOTALL` would greedily match from the first `>` in the tag (inside the attributes), not the one closing the opening tag. For example, with:

```xml
<string name="x" translation_description="Label">Phone number</string>
```

the captured group would be `" translation_description="Label">Phone number` instead of just `Phone number`, which would inflate the CHAR_LIMIT calculation with attribute text.

The concern about non-`<xliff:g>` formatting tags (`<b>`, `<i>`, etc.) is valid in general, but none of the strings in this resource file use those tags, so the current pattern handles all actual data correctly.

scripts/translations/export_translations.py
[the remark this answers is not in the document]
demolaf: seems like after the rebase you might have lost this commit (although rebasing rewrites history), the changes don't show up in the diff.

worth re-checking.

demolaf:
Did a second pass with an agent, maybe worth checking?

Comment:

`_extract_visible_text` only trims the outer edges, so internal newlines/indentation from wrapped multi-line `<string>` tags get counted as visible characters — e.g. `fui_welcome_back_idp_prompt` comes out to `CHAR_LIMIT=100` when the actual rendered text is ~56 chars (limit should be ~85). Worth collapsing internal whitespace (e.g. `re.sub(r'\s+', ' ', text).strip()`) before measuring length, since inflated limits defeat the purpose of this PR for exactly the long/wrapped strings where an accurate limit matters most.

scripts/translations/export_translations.py
```diff
@@ -9,20 +10,76 @@
...
+    """Decode XML entities and Android escape sequences to get true visible length."""
+    for entity, char in XML_ENTITIES.items():
+        text = text.replace(entity, char)
+    text = ESCAPE_PATTERN.sub('X', text)
+    return text
+
+
+def _extract_visible_text(value):
+    """Strip XML tags and decode entities to get the visible text length."""
+    text = XML_TAG_PATTERN.sub('', value).strip()
```
russellwheatley, line 35: This strips tags but only trims the outer edges, so multi-line strings still count their internal newlines/indentation as visible text. Android collapses that whitespace to a single space when rendering, so the limit ends up inflated for anything that spans multiple lines in the XML (e.g. fui_password_recovery_body). Can you collapse runs of whitespace to a single space before measuring?

scripts/translations/export_translations.py
```diff
@@ -9,20 +10,76 @@
...
+    length = len(english_text)
+    limit = math.ceil(length * 1.5)
+    return int(math.ceil(limit / 5.0) * 5)
+
+
+def _add_char_limit(text, value_pattern):
+    """Add [CHAR_LIMIT=xxx] to translation_description if missing."""
+    if 'translation_description=' not in text:
+        return text
+    if CHAR_LIMIT_PATTERN.search(text):
```
russellwheatley: This checks for an existing CHAR_LIMIT against the whole tag text, including any nested xliff:g description, not just the outer translation_description being edited. Worth scoping the check to just the matched attribute so it doesn't skip the outer description if a nested one ever ends up with its own limit.

scripts/translations/export_translations.py
```diff
@@ -9,20 +10,76 @@
...
+STRING_VALUE_PATTERN = re.compile(r'>([^<]*(?:<xliff:g[^>]*>[^<]*</xliff:g>[^<]*)*)</string>')
+ITEM_VALUE_PATTERN = re.compile(r'>([^<]*(?:<xliff:g[^>]*>[^<]*</xliff:g>[^<]*)*)</item>')
+
+XML_ENTITIES = {'&amp;': '&', '&lt;': '<', '&gt;': '>', '&quot;': '"', '&apos;': "'"}
+ESCAPE_PATTERN = re.compile(r'\\(u[0-9A-Fa-f]{4}|n|t|\'|")')
+
+
+def _decode_entities(text):
+    """Decode XML entities and Android escape sequences to get true visible length."""
+    for entity, char in XML_ENTITIES.items():
```
russellwheatley: Decoding entities one at a time in dict order means a genuinely double-encoded entity like &amp;lt; would get mangled by the subsequent &lt; pass. Might be worth doing this in a single regex pass instead, though nothing in strings.xml hits this today.

russellwheatley (changes requested):
A couple of bits of feedback on the char limit calc

scripts/translations/export_translations.py
just1and0: Fixed in a112142 — now collapsing whitespace runs to a single space after stripping tags and decoding entities, so multi-line XML strings get measured the way Android renders them.

scripts/translations/export_translations.py
just1and0: Fixed in a112142 — the CHAR_LIMIT check is now scoped to just the translation_description attribute value instead of the whole tag text.

scripts/translations/export_translations.py
just1and0: Fixed in a112142 — switched to a single-pass re.sub with a lambda callback, so double-encoded entities like &amp;amp;lt; decode correctly to &amp;lt; instead of being mangled by a second pass.

russellwheatley (approved):
All review feedback has been addressed. CI is green across the board. LGTM.

PR #2418, opened 2026-07-23
fix(auth): handle system back press in phone auth flow

### LINKED ISSUE #2058: Disable back press in Firebase Auth Phone
Welcome to FirebaseUI and thanks for submitting an issue!

Please take a look at [open issues](https://github.com/firebase/FirebaseUI-Android/issues?q=is%3Aopen+is%3Aissue), as well as [resolved issues](https://github.com/firebase/FirebaseUI-Android/issues?q=is%3Aissue+is%3Aclosed), to see if your issue is either already being addressed, or has been solved by someone else.

If not, please feel free to fill in the following info so we can help faster!

### Step 1: Are you in the right place?

Yes

  * Android device: Any
  * Android OS version: Any
  * Google Play Services version: Any
  * Firebase/Play Services SDK version: Any
  * FirebaseUI version: Any
  
Using Firebase UI for Phone Auth only, how can we disable back press?

#### Steps to reproduce:

  1. Create a sample project
  2. Integrate Firebase UI SDK Android
  3. Use Phone auth provider
  4. Press back

#### Observed Results:

 No way of controlling the back press with Phone Auth Activity
  
#### Expected Results:

Should allow to add `OnBackPressedCallback` to allow consumer to control the back press behavior.
  
#### Relevant Code:

Something like this 

  ```
AuthUI.IdpConfig.PhoneBuilder().addBackPressCallback(backPressedCallback)

private val backPressedCallback = object : OnBackPressedCallback(true) {

        override fun handleOnBackPressed() {

           

        }

    }
  ```

just1and0:
## Summary
- Adds `BackHandler` to `PhoneAuthScreen` so the system back button is handled consistently for both default and custom phone content
- The base handler delegates to `onCancel()`, which navigates to the method picker or prevents premature activity finish in phone-only flows
- The default UI adds a step-specific override on the verification code screen: system back returns to phone number entry instead of cancelling
- Custom `phoneContent` slots inherit the base behavior and can override with their own `BackHandler`

Fixes #2058

## Test plan
- [ ] Launch phone auth as the only provider — verify system back does not finish the activity
- [ ] Launch phone auth with multiple providers — verify system back returns to method picker
- [ ] On verification code screen — verify system back returns to phone number entry
- [ ] With custom `phoneContent` slot — verify consumer `BackHandler` overrides the default

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/ui/screens/phone/PhoneAuthScreen.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/screens/phone/PhoneAuthScreen.kt
@@ -16,6 +16,7 @@ package com.firebase.ui.auth.ui.screens.phone
 
 import android.content.Context
 import android.util.Log
+import androidx.activity.compose.BackHandler
 import androidx.activity.compose.LocalActivity
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
@@ -359,6 +360,7 @@ private fun DefaultPhoneAuthContent(
 ) {
     when (state.step) {
         PhoneAuthStep.EnterPhoneNumber -> {
+            BackHandler { onCancel() }
             EnterPhoneNumberUI(
                 configuration = configuration,
                 isLoading = state.isLoading,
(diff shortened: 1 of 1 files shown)


just1and0:
Closing — no user interest on the original issue (no thumbs up), and confirmed by team that this isn't worth pursuing.

PR #2421, opened 2026-07-24
fix(auth): prevent MFA code input from triggering password manager save (#2405)

### LINKED ISSUE #2405: The default UI of the MFA interfers with the password manager
https://github.com/user-attachments/assets/34338661-f7ac-40a8-ba31-7ae0bf409da2

The MFA default UI (in v10 beta3) interferes with the password manager in some way, meaning that one of the code numbers is saved instead of the password itself.

The problem is connected to the fifth digit of the code, which is also stored instead of the password.

just1and0:
## Summary

- The `VerificationCodeInputField` component used `KeyboardType.NumberPassword` for each digit's `BasicTextField`. This told Android's autofill framework the field was a password input, causing the password manager to capture an MFA code digit instead of the user's actual account password.
- Changed to `KeyboardType.Number` which continues requesting a numeric keyboard without triggering password-manager save behavior.

Fixes #2405

https://github.com/user-attachments/assets/0e2b0e76-a1d5-4c90-bdab-2c6c2afea89f

## Test plan

- [ ] Enter email/password on the sign-in screen, then proceed to MFA code entry
- [ ] Complete the 6-digit MFA code — verify the password manager does **not** prompt to save/update with a single digit
- [ ] Verify the numeric keyboard still appears when each digit field is focused
- [ ] Verify the TOTP enrollment flow (`VerifyTotpUI`) is unaffected (it already uses `KeyboardType.Number`)
- [ ] Verify phone number verification code entry still works correctly (`EnterVerificationCodeUI` shares `VerificationCodeInputField`)

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/ui/components/VerificationCodeInputField.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/components/VerificationCodeInputField.kt
@@ -292,7 +292,8 @@ private fun SingleDigitField(
                 lineHeight = 24.sp,
             ),
             keyboardOptions = KeyboardOptions(
-                keyboardType = KeyboardType.NumberPassword
+                keyboardType = KeyboardType.Number,
+                autoCorrect = false
             ),
             decorationBox = { innerTextField ->
                 Box(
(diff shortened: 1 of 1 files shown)


demolaf (approved):
Looks good, can we get a screen recording attached to the PR desc too?

russellwheatley (approved):
LGTM - pending Ademola's screen recording.

PR #2425, opened 2026-07-27
fix(auth): pre-fill email when "Continue as" button is tapped (#2423)

### LINKED ISSUE #2423: "Continue as…" button ignores the saved identifier — behaves identically to the plain provider button
## Environment
  
  * FirebaseUI version: `com.firebaseui:firebase-ui-auth:10.0.0-beta03` (Compose `FirebaseAuthScreen`)
  * Device: OnePlus 10 Pro (NE2213), Android 16 (API 36)
  * Google Play services: 26.26.34

  ## Steps to reproduce

  1. Configure `FirebaseAuthScreen` with an Email provider (plus others).
  2. Sign in with email once, then sign out and return to the auth screen.
  3. The method picker now shows a prominent "Continue with email" button with the
     previously used address as subtitle (from `SignInPreferenceManager`).
  4. Tap it.

  ## Observed
  
  The email entry screen opens completely blank — identical to tapping the regular
  "Sign in with email" button below the divider. The saved identifier shown on the
  button is not used anywhere in the flow.

  ## Expected

  A "Continue as <identifier>" affordance implies the identifier is carried into the flow:
  either skip straight to the password step for that address, or at minimum pre-fill the
  email field. Otherwise the button is a duplicate of the provider button and confuses
  users (two buttons, same destination).

  ## Analysis
  
  In `AuthMethodPicker.kt`, `ContinueAsButton`'s click handler is just
  `onProviderSelected(lastProvider)` — the same callback and provider object as the plain
  provider button. The `preference.identifier` is passed to the button only as display
  `subtitle`, and there is no mechanism to hand it to the provider flow —
  `AuthProvider.Email` has no `defaultEmail`-style parameter that the picker could populate.

  The same applies to the other "Continue with X" variants, though for federated providers
  the provider's own account picker masks it; for email the button is pure decoration.

just1and0:
## Summary

- The "Continue as…" button in `AuthMethodPicker` displayed the saved email/identifier but discarded it on click — the user landed on a blank email form identical to tapping the regular provider button.
- Added an `onContinueAsSelected` callback to `AuthMethodPicker` that carries the saved identifier through to `EmailAuthScreen`, which now pre-fills the email field with the saved address.
- Regular provider button taps clear the pre-fill so the two paths remain distinct.

Fixes #2423

https://github.com/user-attachments/assets/07467e54-b7e5-4a98-8488-0e8c5323310a
 
## Test plan

- [ ] Sign in with email, sign out, return to the auth method picker
- [ ] Tap the "Continue as <email>" button — verify the email field is pre-filled with the saved address
- [ ] Tap the regular "Sign in with email" button below the divider — verify the email field is empty
- [ ] Verify federated providers (Google, Facebook, etc.) are unaffected by the change
- [ ] Verify custom layouts via `customLayout` still work (they don't receive `onContinueAsSelected`)

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/ui/method_picker/AuthMethodPicker.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/method_picker/AuthMethodPicker.kt
@@ -82,6 +82,9 @@ class MethodPickerTermsConfiguration(
  * @param providers The list of providers to display.
  * @param logo An optional logo to display.
  * @param onProviderSelected A callback when a provider is selected.
+ * @param onContinueAsSelected A callback when the "Continue as..." button is selected, with the
+ * provider and saved identifier (email, phone number, etc.). Falls back to [onProviderSelected]
+ * if not provided.
  * @param customLayout An optional custom layout composable for the provider buttons.
  * @param termsOfServiceUrl The URL for the Terms of Service.
  * @param privacyPolicyUrl The URL for the Privacy Policy.
@@ -97,12 +100,15 @@ fun AuthMethodPicker(
     providers: List<AuthProvider>,
     logo: AuthUIAsset? = null,
     onProviderSelected: (AuthProvider) -> Unit,
+    onContinueAsSelected: ((AuthProvider, String?) -> Unit)? = null,
     termsOfServiceUrl: String? = null,
     privacyPolicyUrl: String? = null,
     lastSignInPreference: SignInPreferenceManager.SignInPreference? = null,
     customLayout: (@Composable (List<AuthProvider>, (AuthProvider) -> Unit) -> Unit)? = null,
     termsConfiguration: MethodPickerTermsConfiguration? = null,
 ) {
+    val continueAsHandler: (AuthProvider, String?) -> Unit =
+        onContinueAsSelected ?: { provider, _ -> onProviderSelected(provider) }
     val context = LocalContext.current
     val inPreview = LocalInspectionMode.current
     val stringProvider = LocalAuthUIStringProvider.current
@@ -148,7 +154,7 @@ fun AuthMethodPicker(
                                     provider = lastProvider,
                                     identifier = preference.identifier,
                                     enabled = providerButtonsEnabled,
-                                    onClick = { onProviderSelected(lastProvider) }
+                                    onClick = { continueAsHandler(lastProvider, preference.identifier) }
                                 )
                                 Spacer(modifier = Modifier.height(24.dp))
 
--- a/auth/src/main/java/com/firebase/ui/auth/ui/method_picker/AuthMethodPicker.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/method_picker/AuthMethodPicker.kt
@@ -89,7 +89,7 @@ class MethodPickerTermsConfiguration(
  * @param termsConfiguration Optional configuration for a custom ToS/Privacy Policy footer.
  * When provided, replaces the default "By continuing..." text. See [MethodPickerTermsConfiguration].
  * @param onContinueAsSelected A callback when the "Continue as..." button is selected, with the
- * provider and saved identifier (email, phone number, etc.). Falls back to [onProviderSelected]
+ * provider and saved identifier (email address). Falls back to [onProviderSelected]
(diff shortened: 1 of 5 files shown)


auth/src/main/java/com/firebase/ui/auth/ui/screens/FirebaseAuthScreen.kt
```diff
@@ -218,7 +219,14 @@ fun FirebaseAuthScreen(
...
                             termsConfiguration = customMethodPickerTermsConfiguration,
-                            onProviderSelected = onProviderSelected,
+                            onProviderSelected = { provider ->
+                                prefillEmail.value = null
+                                onProviderSelected(provider)
+                            },
+                            onContinueAsSelected = { provider, identifier ->
+                                prefillEmail.value = identifier
+                                onProviderSelected(provider)
+                            },
```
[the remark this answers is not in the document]
just1and0: Good catch — fixed. We now guard with `if (provider is AuthProvider.Email) identifier else null`, so only email identifiers pre-fill the field. Also moved `onContinueAsSelected` and `prefillEmail` to the end of their respective signatures to preserve positional API compatibility for existing callers.

auth/src/main/java/com/firebase/ui/auth/ui/method_picker/AuthMethodPicker.kt
```diff
@@ -88,6 +88,9 @@ class MethodPickerTermsConfiguration(
  * @param lastSignInPreference The last sign-in preference to show a "Continue as..." button.
  * @param termsConfiguration Optional configuration for a custom ToS/Privacy Policy footer.
  * When provided, replaces the default "By continuing..." text. See [MethodPickerTermsConfiguration].
+ * @param onContinueAsSelected A callback when the "Continue as..." button is selected, with the
+ * provider and saved identifier (email, phone number, etc.). Falls back to [onProviderSelected]
+ * if not provided.
```
demolaf, line 93: the kDoc comment says "phone number" but that isn't wired up yet in this PR, either include it in the PR or adjust the doc comments.

auth/src/main/java/com/firebase/ui/auth/ui/screens/email/EmailAuthScreen.kt
```diff
@@ -150,7 +151,7 @@ fun EmailAuthScreen(
     }
     val mode = rememberSaveable { mutableStateOf(initialMode) }
     val displayNameValue = rememberSaveable { mutableStateOf("") }
-    val emailTextValue = rememberSaveable { mutableStateOf("") }
+    val emailTextValue = rememberSaveable { mutableStateOf(prefillEmail ?: "") }
```
demolaf, line 155: we can add a unit test that asserts the email is pre-filled. there's already a `SignUpUITest.kt`.

demolaf (changes requested):
LGTM, just a couple of changes.

auth/src/main/java/com/firebase/ui/auth/ui/method_picker/AuthMethodPicker.kt
just1and0: Updated — changed both kDoc comments to say "email address" instead of "phone number" since phone pre-fill isn't part of this PR.

auth/src/main/java/com/firebase/ui/auth/ui/screens/email/EmailAuthScreen.kt
just1and0: Added `SignInUITest.kt` with two tests: one asserts the email field is pre-filled when an initial value is provided, and another asserts it's empty when no value is given. Followed the same pattern as `SignUpUITest.kt`.

demolaf: @just1and0 this doesn't test from `AuthMethodPicker`, I was thinking a test that passes `lastSignInPreference` (via `SignInPreferenceManager.SignInPreference`) to render the button, clicks "ContinueAsButton", and asserts `onContinueAsSelected` fires with the right (provider, identifier)?

demolaf (approved):

PR #2427, opened 2026-07-27
fix(auth): stop showing error dialog when Google sign-in is cancelled

### LINKED ISSUE #2422: Cancelling the Google sign-in sheet shows an error dialog with the raw exception message as button text (expected AuthState.Cancelled, no dialog)
## Environment

  * FirebaseUI version: `com.firebaseui:firebase-ui-auth:10.0.0-beta03` (Compose `FirebaseAuthScreen`)
  * Device: OnePlus 10 Pro (NE2213), Android 16 (API 36)
  * Google Play services: 26.26.34

  ## Steps to reproduce

  1. Configure `FirebaseAuthScreen` with a Google provider:

  ```kotlin
  FirebaseAuthScreen(
      configuration = authUIConfiguration {
          this.context = context
          isCredentialManagerEnabled = false
          isMfaEnabled = false
          providers {
              provider(AuthProvider.Google(scopes = emptyList(), serverClientId = webClientId))
              provider(AuthProvider.Facebook())
              provider(AuthProvider.Email(emailLinkActionCodeSettings = null, passwordValidationRules = emptyList()))
          }
      },
      onSignInSuccess = { ... },
      onSignInFailure = { ... },
      onSignInCancelled = { ... },
  )
  ```

  2. Tap "Sign in with Google".
  3. Dismiss the Credential Manager account sheet (swipe back / tap outside).

  ## Observed

  An "Authentication Error" dialog appears:

  * Message: "Authentication was cancelled. Please try again when ready."
  * **Confirm button labeled "User cancelled the selector"** — the raw
    `GetCredentialCancellationException` message rendered as button text.
  * `onSignInCancelled` is never invoked.

Image

## Expected

  Cancelling a sign-in is a normal user action, not an error. Expected the flow to return
  to the method picker silently, with `onSignInCancelled` invoked — i.e. the existing
  `AuthState.Cancelled` path, which already does exactly this and shows no dialog.

  ## Analysis

  Two separate problems compound here:

  1. **Cancellation is routed to the error state.** In
     `GoogleAuthProvider+FirebaseAuthUI.kt` (~line 217), the `CancellationException` from the
     credential flow is caught and re-emitted as `updateAuthState(AuthState.Error(AuthCancelledException(...)))`
     rather than `AuthState.Cancelled`. The top-level dialog controller in `FirebaseAuthScreen`
     then unconditionally shows `ErrorRecoveryDialog` for any `AuthState.Error`.

  2. **The raw exception message is used as the button label.** In `ErrorRecoveryDialog.kt`,
     `getRecoveryActionText()` maps `AuthCancelledException -> error.message ?: stringProvider.continueText`,
     so whatever message the underlying exception carries ("User cancelled the selector")
     becomes the confirm-button text. Even for genuine error cases this should presumably be
     `stringProvider.continueText` (or another localized action label), never `error.message`.

  Suggested fix: emit `AuthState.Cancelled` for user-initiated cancellations in the Google
  provider (matching the documented behavior of `AuthCancelledException`: "thrown when the
  user cancels an authentication flow"), and stop using `error.message` as an action label in
  `ErrorRecoveryDialog`.

demolaf:
Fixes #2422.

Cancelling the Google sign-in Credential Manager sheet showed an "Authentication Error" dialog with the raw `GetCredentialCancellationException` message ("User cancelled the selector") as the button text, instead of returning silently via `AuthState.Cancelled`. `signInWithGoogle` only caught `kotlinx.coroutines.CancellationException`, not the actual exception Credential Manager throws on dismissal, so it fell through to the generic error path.

- `GoogleAuthProvider+FirebaseAuthUI.kt`: catches `GetCredentialCancellationException` explicitly and updates `AuthState.Cancelled` without rethrowing, so the sign-in flow returns to the picker silently and `onSignInCancelled` fires.
- `ErrorRecoveryDialog.kt`: `getRecoveryActionText` no longer uses the raw exception message as the button label for `AuthCancelledException`.

Added a test covering the Credential Manager cancellation path, and fixed `ErrorRecoveryDialogLogicTest` to exercise the real (now-`internal`) dialog logic functions instead of a stale duplicate that had drifted from the actual code.

## Preview
[no-cancel-dialog.webm](https://github.com/user-attachments/assets/df91860d-74e9-4c4d-bb53-0071c8673700)

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/configuration/auth_provider/GoogleAuthProvider+FirebaseAuthUI.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/configuration/auth_provider/GoogleAuthProvider+FirebaseAuthUI.kt
@@ -6,6 +6,7 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
 import androidx.credentials.CredentialManager
+import androidx.credentials.exceptions.GetCredentialCancellationException
 import androidx.credentials.exceptions.GetCredentialException
 import androidx.credentials.exceptions.NoCredentialException
 import com.firebase.ui.auth.AuthException
@@ -94,7 +95,9 @@ internal fun FirebaseAuthUI.rememberGoogleSignInHandler(
  * **Error Handling:**
  * - [GoogleIdTokenParsingException]: Library version mismatch
  * - [NoCredentialException]: No Google accounts on device
- * - [GetCredentialException]: User cancellation, configuration errors, or no credentials
+ * - [GetCredentialCancellationException]: User dismissed the Credential Manager sheet -
+ *   updates [AuthState.Cancelled] and does not throw
+ * - [GetCredentialException]: Configuration errors or no credentials
  * - Configuration errors trigger detailed developer guidance logs
  *
  * @param context Android context for Credential Manager
@@ -214,6 +217,13 @@ internal suspend fun FirebaseAuthUI.signInWithGoogle(
         // Re-throw to let UI handle the account linking flow
         updateAuthState(AuthState.Error(e))
         throw e
+    } catch (e: GetCredentialCancellationException) {
+        // User dismissed the Credential Manager sheet - this is a normal user action,
+        // not an error, so it goes to AuthState.Cancelled instead of AuthState.Error.
+        // Swallow (don't rethrow) so rememberGoogleSignInHandler's catch block doesn't
+        // overwrite this state with AuthState.Error.
+        updateAuthState(AuthState.Cancelled)
+
     } catch (e: CancellationException) {
         val cancelledException = AuthException.AuthCancelledException(
             message = "Sign in with google was cancelled",
--- a/auth/src/main/java/com/firebase/ui/auth/ui/components/ErrorRecoveryDialog.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/components/ErrorRecoveryDialog.kt
@@ -130,7 +130,7 @@ fun ErrorRecoveryDialog(
  * @param stringProvider The [AuthUIStringProvider] for localized strings
  * @return The localized recovery message
  */
-private fun getRecoveryMessage(
+internal fun getRecoveryMessage(
     error: AuthException,
     stringProvider: AuthUIStringProvider
 ): String {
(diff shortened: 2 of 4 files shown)


auth/src/main/java/com/firebase/ui/auth/configuration/auth_provider/GoogleAuthProvider+FirebaseAuthUI.kt
```diff
@@ -214,6 +217,13 @@ internal suspend fun FirebaseAuthUI.signInWithGoogle(
         // Re-throw to let UI handle the account linking flow
         updateAuthState(AuthState.Error(e))
         throw e
+    } catch (e: GetCredentialCancellationException) {
+        // User dismissed the Credential Manager sheet - this is a normal user action,
+        // not an error, so it goes to AuthState.Cancelled instead of AuthState.Error.
+        // Swallow (don't rethrow) so rememberGoogleSignInHandler's catch block doesn't
+        // overwrite this state with AuthState.Error.
+        updateAuthState(AuthState.Cancelled)
```
russellwheatley, line 225: Routing sheet dismiss to `AuthState.Cancelled` fixes the Compose path (no dialog, `onSignInCancelled` fires), but `FirebaseAuthActivity` finishes with `RESULT_CANCELED` on that state. So Activity / `AuthFlowController` consumers would leave the whole auth UI when the user just dismisses the Google sheet, rather than staying on the method picker.

Is that intentional? If those consumers should stay on the picker, this probably wants `Idle` (or similar) instead of `Cancelled`, and `onSignInCancelled` would need a different trigger for provider-level dismiss.

russellwheatley (changes requested):
One concern on where sheet dismiss lands in the state machine.

auth/src/main/java/com/firebase/ui/auth/configuration/auth_provider/GoogleAuthProvider+FirebaseAuthUI.kt
demolaf: there's a gap in the implementation, the `AuthFlowController.cancel` triggers `AuthState.Cancelled` and terminates the entire activity which means `AuthState.Cancelled` is doing multiple things.

Using `AuthState.Idle` doesn't convey that an operation was cancelled, so that's ruled out.

We can leave `AuthState.Cancelled` for when an operation is cancelled and introduce a new state e.g. `AuthState.Finished` for when `AuthFlowController.cancel` is invoked?

This would be a breaking change.

russellwheatley: Yes, it would be a breaking change in behaviour, but `Cancelled` is doing two jobs. So I think we have to split it out. Here is what it should look like:

`Cancelled` cancels the current step within the flow (provider sheet dismissed, MFA challenge backed out of) and returns to the picker.
`Aborted` (new) aborts the entire flow, triggered by `AuthFlowController.cancel()`, and finishes the activity.
Went with `Aborted` over `Finished`, reads too close to `Success` in my opinion.

Another thing: `Cancelled` isn't just consumed in `FirebaseAuthActivity`, it's also handled in `FirebaseAuthScreen`, `PhoneAuthScreen`, and `EmailAuthScreen`. `AuthState` isn't `sealed`, so none of those will fail to compile if we forget to add the matching `Aborted` branch, they'll just silently no-op. Worth a test per call site rather than relying on the compiler to catch it.

Also, MFA-challenge-cancel currently routes through `Cancelled` too, so today it already finishes the whole activity via that same listener. After the split it'll correctly stay in the flow instead, that's a behavior change that should have its own test, not just a side effect.

demolaf:
Waiting on this PR https://github.com/firebase/FirebaseUI-Android/pull/2440 after which we can merge this.

russellwheatley (approved):

PR #2433, opened 2026-07-30
fix(auth): report Google/Facebook/OAuth/Anonymous sign-in failures via onSignInFailure

demolaf:
`FirebaseAuthScreen`'s `onSignInFailure` callback was only ever invoked from `EmailAuthScreen` and `PhoneAuthScreen`'s own effects. Google, Facebook, OAuth (Apple/Github/Microsoft/Yahoo/Twitter/generic), and Anonymous sign-in failures only updated the shared `AuthState.Error` - which the top-level dialog controller shows - but never reached `onSignInFailure`, so a consuming app's own error logging/analytics wired through that callback silently never fired for those providers.

Threads an `onSignInFailure` callback into all four `remember*SignInHandler` functions, calling it alongside each existing `updateAuthState(AuthState.Error(...))` so each provider reports its own failure exactly once, at the source - symmetric with how Email/Phone already report theirs.

- `GoogleAuthProvider+FirebaseAuthUI.kt`, `FacebookAuthProvider+FirebaseAuthUI.kt`, `OAuthProvider+FirebaseAuthUI.kt`, `AnonymousAuthProvider+FirebaseAuthUI.kt`: added an `onSignInFailure` parameter to each `remember*Handler`, called in their catch blocks.
- `FirebaseAuthScreen.kt`: threads `onSignInFailure` through `rememberOnProviderSelected` and its two call sites - the main flow passes the real callback; the reauth sheet keeps its existing no-op default, matching how it already no-ops Email/Phone's `onError` there.

Added direct tests forcing a failure (`testCredentialManagerProvider` for Google, a failing `signInAnonymously()` task for Anonymous) confirming `onSignInFailure` fires immediately without navigating anywhere - verified both are load-bearing by temporarily reverting the fix and confirming they fail.

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/configuration/auth_provider/FacebookAuthProvider+FirebaseAuthUI.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/configuration/auth_provider/FacebookAuthProvider+FirebaseAuthUI.kt
@@ -116,7 +116,7 @@ internal fun FirebaseAuthUI.rememberSignInWithFacebookLauncher(
                             authException
                         )
                     )
-                    currentOnSignInFailure(authException)
+                    if (authException !is AuthException.AuthCancelledException) currentOnSignInFailure(authException)
                 }
             })
 
--- a/auth/src/main/java/com/firebase/ui/auth/configuration/auth_provider/FacebookAuthProvider+FirebaseAuthUI.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/configuration/auth_provider/FacebookAuthProvider+FirebaseAuthUI.kt
@@ -19,8 +19,10 @@ import android.util.Log
 import androidx.activity.compose.rememberLauncherForActivityResult
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.DisposableEffect
+import androidx.compose.runtime.getValue
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
+import androidx.compose.runtime.rememberUpdatedState
 import com.facebook.AccessToken
 import com.facebook.CallbackManager
 import com.facebook.FacebookCallback
@@ -64,6 +66,10 @@ internal fun FirebaseAuthUI.rememberSignInWithFacebookLauncher(
     val coroutineScope = rememberCoroutineScope()
     val callbackManager = remember { CallbackManager.Factory.create() }
     val loginManager = LoginManager.getInstance()
+    val currentContext by rememberUpdatedState(context)
+    val currentConfig by rememberUpdatedState(config)
+    val currentProvider by rememberUpdatedState(provider)
+    val currentOnSignInFailure by rememberUpdatedState(onSignInFailure)
 
     val launcher = rememberLauncherForActivityResult(
         loginManager.createLogInActivityResultContract(
@@ -73,7 +79,7 @@ internal fun FirebaseAuthUI.rememberSignInWithFacebookLauncher(
         onResult = {},
     )
 
-    DisposableEffect(config) {
+    DisposableEffect(Unit) {
         loginManager.registerCallback(
             callbackManager,
             object : FacebookCallback<LoginResult> {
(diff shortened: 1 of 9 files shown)


auth/src/main/java/com/firebase/ui/auth/configuration/auth_provider/OAuthProvider+FirebaseAuthUI.kt
```diff
@@ -54,29 +54,30 @@ internal fun FirebaseAuthUI.rememberOAuthSignInHandler(
...
+                    activity = activity,
+                    provider = provider
+                )
+            } catch (e: AuthException) {
+                updateAuthState(AuthState.Error(e))
+                if (e !is AuthException.AuthCancelledException) onSignInFailure(e)
+            } catch (e: Exception) {
+                val authException = AuthException.from(e, context)
+                updateAuthState(AuthState.Error(authException))
+                if (authException !is AuthException.AuthCancelledException) onSignInFailure(authException)
```
russellwheatley, line 81: This guard does not catch real OAuth user cancels.

Backing out of the custom tab throws `FirebaseAuthWebException("ERROR_WEB_CONTEXT_CANCELED", ...)`, which extends `FirebaseAuthException`. `AuthException.from` maps that branch to `UnknownException` (only special-cases `ERROR_TOO_MANY_REQUESTS`) before the message-based `canceled` check in the final `else`, so this still calls `onSignInFailure`.

Before this PR that cancel never reached the host callback. After it, analytics wired to `onSignInFailure` get cancel noise for Apple/GitHub/Microsoft/Yahoo/Twitter/generic.

Map `ERROR_WEB_CONTEXT_CANCELED` to `AuthCancelledException` in `AuthException.from` (or special-case it in `signInWithProvider`), and cover it with a test that uses a real `FirebaseAuthWebException`, not a synthetic `CancellationException`.

auth/src/main/java/com/firebase/ui/auth/configuration/auth_provider/FacebookAuthProvider+FirebaseAuthUI.kt
```diff
@@ -100,12 +110,13 @@ internal fun FirebaseAuthUI.rememberSignInWithFacebookLauncher(
...
                 override fun onError(error: FacebookException) {
                     Log.e("FacebookAuthProvider", "Error during Facebook sign in", error)
-                    val authException = AuthException.from(error, context)
+                    val authException = AuthException.from(error, currentContext)
                     updateAuthState(
                         AuthState.Error(
                             authException
                         )
                     )
+                    currentOnSignInFailure(authException)
```
russellwheatley: `onSuccess` catch blocks filter `AuthCancelledException`, but `onError` always reports.

Real Facebook cancels go through `onCancel()`, so this is unlikely in practice. Still worth the same one-line guard for consistency with the other call sites in this PR:

```kotlin
if (authException !is AuthException.AuthCancelledException) currentOnSignInFailure(authException)
```

auth/src/main/java/com/firebase/ui/auth/configuration/auth_provider/GoogleAuthProvider+FirebaseAuthUI.kt
```diff
@@ -57,19 +57,20 @@ internal fun FirebaseAuthUI.rememberGoogleSignInHandler(
...
-            coroutineScope.launch {
-                try {
-                    signInWithGoogle(context, config, provider)
-                } catch (e: AuthException) {
-                    updateAuthState(AuthState.Error(e))
-                } catch (e: Exception) {
-                    val authException = AuthException.from(e, context)
-                    updateAuthState(AuthState.Error(authException))
-                }
+    return {
```
russellwheatley, line 63: Removing `remember` is fine for freshness, but the KDoc above still says the callback is "stable across recompositions". Update that (same for Anonymous/OAuth if they still claim it), or use `rememberUpdatedState` for the captured inputs if you want a stable returned lambda again.

auth/src/test/java/com/firebase/ui/auth/configuration/auth_provider/GoogleAuthProviderFirebaseAuthUITest.kt
```diff
@@ -920,4 +925,59 @@ class GoogleAuthProviderFirebaseAuthUITest {
...
+                onSignInFailure = { reportedFailures.add(it) },
+            )
+        }
+
+        composeTestRule.runOnIdle { launcher?.invoke() }
+        composeTestRule.waitForIdle()
+
+        assertThat(reportedFailures).hasSize(1)
+        assertThat(reportedFailures.single()).isInstanceOf(AuthException.UnknownException::class.java)
+    }
```
russellwheatley, line 982: No negative test that cancel does not call `onSignInFailure`.

Add one next to this (and ideally the same for OAuth with `FirebaseAuthWebException(ERROR_WEB_CONTEXT_CANCELED)`). Right now every new test only asserts the positive "failure reaches the callback" case, so commit 3's filter has no coverage.

russellwheatley (changes requested):
Cancel filtering looks incomplete for OAuth web cancels, plus a couple of smaller consistency/coverage notes

auth/src/main/java/com/firebase/ui/auth/configuration/auth_provider/OAuthProvider+FirebaseAuthUI.kt
demolaf: this makes sense, `AuthException.from()` now maps `FirebaseAuthWebException`'s `ERROR_WEB_CONTEXT_CANCELED` to `AuthCancelledException`.

russellwheatley (approved):

PR #2435, opened 2026-07-31
fix(auth): log diagnostic warning when Google sign-in NoCredentialException fallback exhausts

demolaf:
`NoCredentialException` from Credential Manager is ambiguous: it means either the device has no Google account or the app is misconfigured (missing Android OAuth client / SHA-1). Developers currently get no signal to tell them apart.

Logs a `Log.w` listing the possible causes when both credential attempts are exhausted. The user-facing message is unchanged and deliberately stays generic — most users hitting this genuinely have no account, and Firebase Console guidance would only confuse them.


russellwheatley (approved):
LGTM

PR #2440, opened 2026-08-06
fix(auth)!: only exit auth flow on AuthState.Aborted, not per-provider Cancelled. BREAKING: Splitting two different behaviours via AuthState into `Cancelled` and `Aborted`

demolaf:
Splits `AuthState.Cancelled` into two states: it was overloaded to mean both "user backed out of one provider's sheet" (should stay in the flow) and "exit the whole auth Activity" - but `FirebaseAuthActivity` only implemented the latter, so any per-provider cancel would exit the whole flow instead of returning to the method picker.

- `AuthState.kt`: adds `AuthState.Aborted` for "flow ended programmatically, exit entirely"; `Cancelled` now means "this sign-in attempt was cancelled, stay in the flow."
- `FirebaseAuthActivity.kt`: only finishes the Activity on `Aborted`, not `Cancelled`.
- `AuthFlowController.kt`: `cancel()` now emits `AuthState.Aborted` instead of `AuthState.Cancelled` — **breaking change** for any consumer pattern-matching on `is AuthState.Cancelled` to detect a host-initiated `cancel()`.
- `FirebaseAuthScreen.kt`: adds an `Aborted` branch mirroring `Cancelled`'s reset/self-consume handling.
- `README.md`: documents `AuthState.Aborted` alongside `Cancelled` in the `AuthFlowController` and low-level API examples.

Added a test confirming the Activity does *not* finish on `Cancelled`, and updated the existing cancel-flow tests to assert `Aborted` instead.

### DIFF
--- a/app/src/main/java/com/firebaseui/android/demo/auth/AuthFlowControllerDemoActivity.kt
+++ b/app/src/main/java/com/firebaseui/android/demo/auth/AuthFlowControllerDemoActivity.kt
@@ -227,6 +227,7 @@ fun AuthFlowDemo(
                         is AuthState.Success -> "Success - User: ${(authState as AuthState.Success).user.email}"
                         is AuthState.Error -> "Error: ${(authState as AuthState.Error).exception.message}"
                         is AuthState.Cancelled -> "Cancelled"
+                        is AuthState.Aborted -> "Aborted"
                         is AuthState.RequiresMfa -> "MFA Required"
                         is AuthState.RequiresEmailVerification -> "Email Verification Required"
                         else -> "Unknown"
--- a/auth/README.md
+++ b/auth/README.md
@@ -343,7 +343,12 @@ lifecycleScope.launch {
             Log.e(TAG, "Auth failed", state.exception)
         }
         is AuthState.Cancelled -> {
-            // User cancelled
+            // User cancelled a single sign-in attempt (e.g. dismissed the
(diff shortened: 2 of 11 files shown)


russellwheatley (changes requested):
The `Cancelled`/`Aborted` split itself works and is well tested in isolation. One thing from the earlier `gemini-code-assist` threads didn't fully land though.

Both threads (on the Activity's and Screen's `Aborted` branches) got marked resolved, but only half of what was suggested shipped: the Activity now resets to `Idle` before `finish()`, which fixes the stuck-state-across-relaunches bug. The other half, having `FirebaseAuthScreen` skip its own `Aborted` handling when hosted by `FirebaseAuthActivity`, never went in. Both collectors still react independently to the same `_authStateFlow`.

It seems to hold up today because `cancel()` runs synchronously and the Activity's `Main.immediate` collector wins the race before Compose's `LaunchedEffect` gets a dispatch. But that's an implementation detail of how `combine()`/`StateFlow` propagate the caller's dispatcher, not a guarantee, and the comment explaining that reasoning got trimmed in 8fbe9cbf. Nothing in the code tells the next person this ordering is load-bearing.

Could you either add the `activity !is FirebaseAuthActivity` guard gemini suggested, or, if you're confident the current ordering is safe, restore a short comment saying why and add a test that composes the real Activity + Screen together so a future coroutines change that flips the ordering actually fails a test?

russellwheatley (approved):
LGTM

PR #2443, opened 2026-08-07
fix(e2e): retry and fail loudly when clearing Firebase Auth emulator data

demolaf:
Fixes flaky `CredentialLinkingScreenTest` Google-link test on CI by making `EmulatorAuthApi.clearEmulatorData()` retry with backoff and throw on failure instead of silently leaving stale emulator state.


russellwheatley (approved):
LGTM

PR #2447, opened 2026-08-13
fix(auth): fix "Already resumed" crash in phone auth SMS auto-verification

### LINKED ISSUE #2446: Phone Auth crashes with "Already resumed" in DefaultVerifier.onVerificationCompleted
### Environment

- Devices: multiple devices
- Android versions: 11–17
- FirebaseUI Auth: `10.0.0-beta03`
- Firebase Auth: `24.0.1`

### Description

Our Android app crashes during Phone Authentication when Firebase automatically completes verification after detecting the SMS verification code.

The crash is fatal on the main thread and prevents affected users from signing in with SMS.

### Steps to reproduce

1. Configure FirebaseUI Phone Auth:

   ```kotlin
   AuthProvider.Phone(
       defaultNumber = null,
       defaultCountryCode = "<country>",
       allowedCountries = listOf("<country>"),
   )
   ```

2. Start phone number verification.
3. Receive the verification SMS.
4. In affected cases, Firebase automatically detects the SMS code and completes verification.
5. The app crashes.

### Observed result

```text
Fatal Exception: java.lang.IllegalStateException: Already resumed
    at kotlin.coroutines.SafeContinuation.resumeWith(SafeContinuationJvm.kt:44)
    at com.firebase.ui.auth.configuration.auth_provider.AuthProvider$Phone$DefaultVerifier$verifyPhoneNumber$2$options$1.onVerificationCompleted(AuthProvider.kt:418)
    at com.google.firebase.auth.zzl.onVerificationCompleted(com.google.firebase:firebase-auth@@24.0.1:6)
    at com.google.android.gms.internal.firebase-auth-api.zzaga.onVerificationCompleted(com.google.firebase:firebase-auth@@24.0.1:8)
    at com.google.android.gms.internal.firebase-auth-api.zzafi.zza(com.google.firebase:firebase-auth@@24.0.1:2)
    at com.google.android.gms.internal.firebase-auth-api.zzafn.run(com.google.firebase:firebase-auth@@24.0.1:4)
```

`DefaultVerifier` wraps Phone Auth callbacks with a single `suspendCoroutine` continuation. It resumes that continuation from `onVerificationCompleted`, `onVerificationFailed`, and `onCodeSent`.

The exception indicates that `onVerificationCompleted` attempted to resume a continuation that had already been completed by another callback. This may occur when `onCodeSent` is followed by SMS auto-retrieval.

### Expected result

Phone verification should either complete automatically or transition to manual code entry without crashing. Any callback received after a terminal result should be safely ignored.

demolaf:
Fixes #2446.

`DefaultVerifier` adapted Firebase's `OnVerificationStateChangedCallbacks` — which can fire more than once for a single request — onto a single-shot `suspendCoroutine`, resuming the same continuation from `onVerificationCompleted`, `onVerificationFailed` and `onCodeSent`. On the SMS auto-retrieval path Firebase delivers `onCodeSent` and then `onVerificationCompleted`, so the second resume threw `IllegalStateException: Already resumed` on the main thread inside Firebase's own dispatcher, where nothing in the library can catch it.

`Verifier` now returns a `Flow` built with `callbackFlow`, so several callbacks are simply several emissions and there is no continuation left to double-resume. A credential auto-retrieved after the code was sent now reaches `AuthState.SMSAutoVerified` rather than being discarded, so the user is no longer asked to type a code Firebase has already verified.

Because the collection stays open for the verification window instead of returning at `onCodeSent`, `PhoneAuthScreen` now cancels superseded attempts on send, resend, change-number and success, and skips auto sign-in while a manually submitted code is still signing in. Cancellation is not reported to the host as an error.

Added `PhoneAuthDefaultVerifierTest` and `PhoneAuthScreenVerificationLifecycleTest`, covering every callback ordering, the auto-retrieval timeout terminal, and the screen's cancellation and guard lifecycle — the multi-callback cases fail against the old code.

## Preview
- Before

https://github.com/user-attachments/assets/cea05c5b-43e3-4844-ad0d-452006146343

- After

https://github.com/user-attachments/assets/61c43a8f-368e-4cf9-9965-dc66f30039be

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUI.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUI.kt
@@ -41,6 +41,7 @@ import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.tasks.await
 import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.atomic.AtomicLong
 
 /**
  * The central class that coordinates all authentication operations for Firebase Auth UI Compose.
@@ -79,6 +80,7 @@ class FirebaseAuthUI private constructor(
 ) {
 
     private val _authStateFlow = MutableStateFlow<AuthState>(AuthState.Idle)
+    private val authStateRevision = AtomicLong(0)
 
     @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
     var testCredentialManagerProvider: AuthProvider.Google.CredentialManagerProvider? = null
@@ -363,8 +365,28 @@ class FirebaseAuthUI private constructor(
      */
     @MainThread
     fun updateAuthState(state: AuthState) {
+        authStateRevision.incrementAndGet()
         _authStateFlow.value = state
     }
+
+    /**
+     * Retracts a pending [AuthState.Loading] by resetting to [AuthState.Idle], but only while
+     * [revision] is still the most recent write. Any state emitted since is left untouched.
+     *
+     * The revision is what makes this precise: [AuthState.Loading] compares equal whenever the
+     * message matches, and [MutableStateFlow] drops a write equal to the current value without
+     * replacing the stored reference - so neither equality nor identity can tell a concurrent
+     * operation's Loading apart from the caller's.
+     *
+     * @param revision The value [currentAuthStateRevision] returned right after the caller emitted
+     * the [AuthState.Loading] it now wants to retract
+     */
+    internal fun clearLoadingState(revision: Long) {
+        if (authStateRevision.get() == revision) updateAuthState(AuthState.Idle)
+    }
+
+    /** Identifies the most recent [updateAuthState] write. See [clearLoadingState]. */
+    internal fun currentAuthStateRevision(): Long = authStateRevision.get()
 
     internal fun updateAuthStateWithResult(result: AuthResult?, defaultIsNewUser: Boolean = false) {
         val user = result?.user
--- a/auth/src/main/java/com/firebase/ui/auth/configuration/auth_provider/AuthProvider.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/configuration/auth_provider/AuthProvider.kt
@@ -55,12 +55,13 @@ import com.google.firebase.auth.PhoneAuthProvider
 import com.google.firebase.auth.TwitterAuthProvider
 import com.google.firebase.auth.UserProfileChangeRequest
 import com.google.firebase.auth.actionCodeSettings
+import kotlinx.coroutines.channels.awaitClose
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.callbackFlow
 import kotlinx.coroutines.suspendCancellableCoroutine
 import kotlinx.coroutines.tasks.await
 import java.util.concurrent.TimeUnit
 import kotlin.coroutines.resume
-import kotlin.coroutines.resumeWithException
-import kotlin.coroutines.suspendCoroutine
 
 @AuthUIConfigurationDsl
 class AuthProvidersBuilder {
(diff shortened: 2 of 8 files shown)


auth/src/main/java/com/firebase/ui/auth/configuration/auth_provider/AuthProvider.kt
```diff
@@ -407,25 +407,56 @@ abstract class AuthProvider(open val providerId: String, open val providerName:
...
+                    // keep any later callback off the continuation entirely.
+                    continuation.invokeOnCancellation { isResolved.set(true) }
                     val options = PhoneAuthOptions.newBuilder(auth)
                         .setPhoneNumber(phoneNumber)
                         .requireSmsValidation(!isInstantVerificationEnabled)
                         .setTimeout(timeout, TimeUnit.SECONDS)
                         .setCallbacks(object :
                             PhoneAuthProvider.OnVerificationStateChangedCallbacks() {
                             override fun onVerificationCompleted(credential: PhoneAuthCredential) {
+                                if (!isResolved.compareAndSet(false, true)) {
```
russellwheatley: This latch stops the crash, but it also drops the auto-retrieved credential on the exact path this PR is fixing. Firebase's SMS auto-retrieval order is onCodeSent then onVerificationCompleted, so the user lands on the code-entry screen after Firebase already verified them. The Verifier interface is single-shot, so you'll need a side channel (PhoneAuthScreen already signs in on SMSAutoVerified). Can you keep the latch for the crash, but still emit SMSAutoVerified when a late onVerificationCompleted arrives after onCodeSent?

auth/src/main/java/com/firebase/ui/auth/configuration/auth_provider/PhoneAuthProvider+FirebaseAuthUI.kt
```diff
@@ -136,12 +136,10 @@ internal suspend fun FirebaseAuthUI.verifyPhoneNumber(
             }
         }
     } catch (e: CancellationException) {
```
russellwheatley, line 150: This is now the only provider that rethrows a raw CancellationException and resets to Idle. Everything else wraps it as AuthCancelledException and emits AuthState.Error. Rethrowing is closer to structured concurrency, but it's a one-off. Can you either match the other providers, or call out why phone verify is special?

auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUI.kt
```diff
@@ -366,6 +367,16 @@ class FirebaseAuthUI private constructor(
         _authStateFlow.value = state
     }
 
+    /**
+     * Clears a pending [AuthState.Loading] by resetting to [AuthState.Idle]. States written by
+     * another operation in the meantime are left untouched.
+     */
+    internal fun clearLoadingState() {
```
russellwheatley: This only skips non-Loading states, so a newer Loading from a resend or another in-flight call would get reset to Idle. The new test covers PasswordResetLinkSent, not a second Loading. Can you key this on the Loading this call actually wrote, or drop the "left untouched" claim in the doc comment if that's not actually guaranteed?

russellwheatley (changes requested):
Crash is gone, but first-callback-wins still drops auto-retrieval on the path this PR is fixing.

auth/src/main/java/com/firebase/ui/auth/configuration/auth_provider/PhoneAuthProvider+FirebaseAuthUI.kt
demolaf: I've done to match the others. I'll look into rethrowing in a separate PR for improvements.

auth/src/main/java/com/firebase/ui/auth/configuration/auth_provider/AuthProvider.kt
demolaf: Firebase's callbacks are multi-shot, so a one-shot `suspendCoroutine` was the actual defect rather than the double-resume being a symptom to guard. Emitting them as a `Flow` means a late auto-verification is just a second emission — no latch, and the credential reaches `SMSAutoVerified` as you asked.

auth/src/main/java/com/firebase/ui/auth/configuration/auth_provider/PhoneAuthProvider+FirebaseAuthUI.kt
russellwheatley: Matching the others now bites because this collect stays open. PhoneAuthScreen cancels it on success, resend and change-number as its own bookkeeping, which writes `AuthState.Error(AuthCancelledException)` into the shared flow. The screen filters that for `onError`, but `FirebaseAuthScreen`'s error `LaunchedEffect` does not, so a successful sign-in or a resend can pop the cancelled recovery dialog. Can you skip publishing that Error when the screen tore the collection down itself? Phone verify is special here: the collect has to stay open for the SMS window, unlike the other providers.

russellwheatley (approved):
LGTM.

PR #2454, opened 2026-08-25
refactor(auth): scope phone verification loading state to the composition that owns it

demolaf:
`FirebaseAuthUI` carried a revision counter plus `clearLoadingState` and `currentAuthStateRevision` purely so `verifyPhoneNumber` could retract its own `AuthState.Loading` on cancellation without clobbering whatever superseded it.

`verifyPhoneNumber` now writes no state on cancellation and just rethrows, and `PhoneAuthScreen` retracts its own `Loading` instead — in `onChangeNumberClick`, and from a `DisposableEffect` when the composition goes away mid-verification. Cancelling an attempt was already the screen's own bookkeeping, so the state it leaves behind belongs to the screen too.

Dropped three tests covering the removed API and the old callee-retracts contract, and added one to `PhoneAuthScreenVerificationLifecycleTest` for the change-number case — verified it fails without the change.

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUI.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUI.kt
@@ -41,7 +41,6 @@ import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.tasks.await
 import java.util.concurrent.ConcurrentHashMap
-import java.util.concurrent.atomic.AtomicLong
 
 /**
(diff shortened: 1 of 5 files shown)


russellwheatley (approved):
LGTM. The core simplification is sound, traced through all the cancellation call sites and the `DisposableEffect(authUI)` cleanup correctly fixes the per-instance state capture bug flagged in review.

One thing worth a follow-up look: `submitVerificationCode` and `signInWithPhoneAuthCredential` still write `AuthState.Error` on `CancellationException`, so the same stale-state-after-disposal issue this PR fixes for `verifyPhoneNumber` might still apply if the screen disposes mid-flight on those. Pre-existing and out of scope here, not blocking.

PR #2463, opened 2026-09-01
fix(auth): clear MFA enrollment flow state on every entry

demolaf:
- Related to #2467

After a successful MFA enrollment the flow state was never cleared, so re-entering landed the user on a form pre-filled from the previous attempt — phone number, verification code and a consumed `smsSession` all still live. #2467 moved that state above the `NavHost` and made most of it `rememberSaveable`, so it now outlives both step navigation and process death: longer-lived rather than cleaner. Across a sign-out, that is the previous user's data.

`MfaEnrollmentFlowState.reset()` returns all nine fields to the values `rememberMfaEnrollmentFlowState()` starts them at, and the host calls it on flow **entry** rather than on completion, through a new `enterMfaEnrollment` extension. Resetting on completion would mutate state a leaving step still reads while it is composed for the exit transition; on the way in there is no such step.

Every entry now goes through that helper, including one that previously did not. `AuthRoute.MfaEnrollment` and `AuthRoute.MfaEnrollment.SelectFactor` are different objects reporting the same `route` string, so `onNavigate(AuthRoute.MfaEnrollment.SelectFactor)` looked equivalent to `onNavigate(AuthRoute.MfaEnrollment)` but fell through to a plain `navigate` with no clear. A step named that way now clears and still lands on the step the host asked for, rather than being redirected through the resolved start step — a host deliberately showing the picker under a single-factor configuration keeps that behaviour.

This replaces the earlier approach on this branch, which reset inside `MfaEnrollmentScreen` and retained the pre-reset state through the pop transition. @russellwheatley found that the frozen snapshot's callbacks still closed over the cleared state, so Verify stayed tappable through the fade and a tap surfaced `onError` after enrollment had already succeeded. Per-step navigation removes the reset window altogether, so all ~117 lines of retention machinery are dropped rather than patched.

Adds 8 tests. Each clause of the widened entry condition is independently pinned by mutation, as is the step-resolution path and every field of `reset()`: three mutants, each killing exactly the test that names the behaviour it breaks. Note the one-clause "simplification" compiles cleanly, so only a test catches it.

A single-pop bug in this area — steps push, but completion popped one entry and landed on `ConfigureSms` instead of leaving the flow — was found while writing these tests and is being fixed in #2467 instead, since it propagates up the nav stack and a fix landing only here would be silently reintroduced by #2469.

---
Maintainer note: Fixes internal CPRN-384

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/ui/screens/FirebaseAuthScreen.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/screens/FirebaseAuthScreen.kt
@@ -98,9 +98,9 @@ import com.firebase.ui.auth.ui.screens.email.isEmailLinkSignInOffered
 import com.firebase.ui.auth.ui.screens.email.isEmailSignUpOffered
 import com.firebase.ui.auth.ui.screens.email.navigateToEmailStep
 import com.firebase.ui.auth.ui.screens.mfa.MfaChallengeScreen
+import com.firebase.ui.auth.ui.screens.mfa.enterMfaEnrollment
 import com.firebase.ui.auth.ui.screens.mfa.exitMfaEnrollment
 import com.firebase.ui.auth.ui.screens.mfa.mfaEnrollmentDestinations
-import com.firebase.ui.auth.ui.screens.mfa.mfaEnrollmentStartStep
 import com.firebase.ui.auth.ui.screens.mfa.rememberMfaEnrollmentFlowState
 import com.firebase.ui.auth.ui.screens.phone.PhoneAuthContentState
 import com.firebase.ui.auth.ui.screens.phone.PhoneAuthScreen
@@ -410,8 +410,9 @@ fun FirebaseAuthScreen(
                                 // Inert while armed: this content stays composed beneath the slot.
                                 if (reauthState == null) {
                                     if (configuration.isMfaEnabled) {
-                                        navController.navigate(
-                                            mfaEnrollmentStartStep(mfaConfiguration).route
+                                        navController.enterMfaEnrollment(
+                                            mfaConfiguration,
+                                            mfaEnrollmentFlowState,
                                         )
                                     } else {
                                         val exception = AuthException.AuthCancelledException(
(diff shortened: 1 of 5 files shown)


auth/src/main/java/com/firebase/ui/auth/ui/screens/FirebaseAuthScreen.kt
```diff
@@ -997,6 +1017,99 @@ fun FirebaseAuthScreen(
...
+        return state
+    }
+
+    /**
+     * Latches the most recently rendered state so later updates are ignored. A no-op when nothing
+     * has been rendered yet, which leaves the screen following the live state.
+     */
+    fun freeze() {
+        frozen.value = latest
+    }
```
russellwheatley: The frozen snapshot keeps the last composed callbacks, and those close over the live `mutableState` holders that `resetAfterEnrollment()` has already cleared.

If enroll returns without a frame in between (the unit test does this), `latest` still has `isLoading = false`, so Verify stays enabled through the pop fade. A tap then hits `selectedFactor == null` and throws `"No factor selected"`, which `FirebaseAuthScreen` routes through `onError` -> `onSignInFailure` after enrollment already succeeded.

Freeze a copy that cannot submit. `isLoading = true` disables the default Verify button; no-op clicks cover a custom slot that ignores loading.

```suggestion
    fun freeze() {
        frozen.value = latest?.copy(
            isLoading = true,
            onVerifyClick = {},
            onBackClick = {},
            onResendCodeClick = null,
            onVerificationCodeChange = {},
        )
    }
```

auth/src/test/java/com/firebase/ui/auth/ui/screens/FirebaseAuthScreenMfaEnrollmentRetentionTest.kt
```diff
@@ -0,0 +1,184 @@
...
+        }
+        composeTestRule.waitForIdle()
+
+        assertEquals(1, completeCount)
+        // The screen itself is back at a clean SelectFactor (asserted in MfaEnrollmentScreenTest),
+        // but the destination still shows what the user finished on, so the pop exit transition
+        // fades out the completed verification step instead of the factor list.
+        assertEquals(MfaEnrollmentStep.VerifyFactor, rendered?.step)
+        assertEquals(VERIFICATION_CODE, rendered?.verificationCode)
+        assertEquals(PHONE_NUMBER, rendered?.phoneNumber)
```
russellwheatley: This is the more dangerous frozen snapshot: with an immediate enroll stub there is no frame between `isLoading = true` and `freeze()`, so `latest` still has `isLoading = false` and Verify stays enabled.

Pin that freeze disables the form, and that a tap afterwards does not fire `onComplete` again.

```suggestion
        assertEquals(MfaEnrollmentStep.VerifyFactor, rendered?.step)
        assertEquals(VERIFICATION_CODE, rendered?.verificationCode)
        assertEquals(PHONE_NUMBER, rendered?.phoneNumber)
        assertEquals(true, rendered?.isLoading)

        rendered?.onVerifyClick?.invoke()
        composeTestRule.waitForIdle()
        assertEquals(1, completeCount)
```

russellwheatley (changes requested):
Frozen Verify stays tappable through the pop fade. Those callbacks close over the already-reset live state, so a tap can surface `onError` after enrollment already succeeded.

Suggested a `copy` in `freeze()` that disables the form, and an `isLoading` pin on the retention test.

russellwheatley (approved):

PR #2474, opened 2026-09-03
fix(auth): don't strand phone-only users on email verification after reload

demolaf:
`onReloadUser` on `AuthSuccessUiContext` re-derived "does this user need email verification?" from `isEmailVerified` alone, dropping the `email != null` and `password`-provider checks `FirebaseAuthUI` already applied. A phone-only user is unverified and has no email, so any `authenticatedContent` slot calling `uiContext.onReloadUser()` published `RequiresEmailVerification(email = "")` — and because `authStateFlow()` prefers any non-Idle internal state, it stuck there with signing out the only exit. Latent in the stock UI, which wires the callback only to its own verify-email screen.

Deriving auth state isn't the screen's job, so the reload moves to `FirebaseAuthUI.reloadUser()` and reuses the existing `handleAuthUserState` rule rather than duplicating it. It also revalidates the current user after its two suspension points: signing out mid-reload previously published `Success` for the departed user, made permanent by that same precedence. The screen keeps only the error logging, and now rethrows `CancellationException` instead of logging it as a refresh failure.

Tests go in `FirebaseAuthUIAuthStateTest` — phone-only, federated, unverified password, and sign-out mid-reload. Verified by mutation rather than assumed: reinstating the old derivation reds the first two, and dropping the revalidation reds the last.

---
Maintainer note: Fixes internal CPRN-407

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUI.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUI.kt
@@ -535,6 +535,24 @@ class FirebaseAuthUI private constructor(
         }
     }
 
+    /**
+     * Re-reads the signed-in user from the server and republishes the resulting auth state.
+     * No-op when nobody is signed in.
+     */
+    internal suspend fun reloadUser() {
+        val user = auth.currentUser ?: return
+        user.reload().await()
+        user.getIdToken(true).await()
+        // Signing out (or switching account) mid-reload must win: publishing here would pin the
+        // combine in authStateFlow() to a Success for a user who is already gone.
+        if (auth.currentUser?.uid != user.uid) return
+        updateAuthState(handleAuthUserState(user, result = null, isNewUser = false))
+    }
+
+    /**
+     * Single source of truth for whether a signed-in user still owes email verification.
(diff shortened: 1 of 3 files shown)


russellwheatley (approved):
LGTM. The split derivation was the bug, and putting reload behind `handleAuthUserState` is the right fix.

Two nits, neither blocking:

The unverified-password test does not pin republish the way the other two do. Phone-only and federated call `pinToEmailVerification()` first, so they fail unless `reloadUser()` overwrites the stuck internal state. The password test starts from `Idle`, and `authStateFlow()` would already emit `RequiresEmailVerification` from the Firebase listener even if `reloadUser()` never called `updateAuthState`. Pinning with `email = ""` first, then asserting it becomes `"test@example.com"`, would make that test do the same job.

The sign-out-mid-reload test stubs `getIdToken(true)` to succeed, so the uid check is what saves it. In production, signing out during `reload()` is more likely to make `getIdToken` throw, which the screen logs as `"Failed to refresh user"` before that check runs. I don't think that re-strands anyone: the `AuthStateListener` already clears stale `RequiresEmailVerification` / `Success` on sign-out. An extra uid check between the two awaits would just avoid the misleading log.

PR #2475, opened 2026-09-03
docs(auth): update Screen Transitions samples to the Navigation 3 AuthUITransitions

demolaf:
The Screen Transitions samples still use the Navigation 2 constructor — `enterTransition` / `exitTransition` / `popEnterTransition` / `popExitTransition` — which no longer exists, so none of the four compile. `AuthUITransitions` now takes one `ContentTransform` per direction: `transitionSpec`, `popTransitionSpec` and `predictivePopTransitionSpec`.

Rewrites the four samples against the new constructor, pairing each enter/exit with `togetherWith`, and adds a per-destination sample using the public `Scene<NavKey>.authRoute()`. The closing note now says each spec falls back to the default 700ms cross-fade independently — `predictivePopTransitionSpec` does not fall back to `popTransitionSpec`, and it fires when the gesture starts rather than when a back completes.

Compiled all five samples verbatim against the module to check they build.

---
Maintainer note: Fixes internal CPRN-404

### DIFF
--- a/auth/README.md
+++ b/auth/README.md
@@ -1675,7 +1675,7 @@ val configuration = authUIConfiguration {
         provider(AuthProvider.Email())
     }
     transitions = AuthUITransitions(
-        // Slide up on the way in, slide down on the way out
+        // A vertical push: the new step rises from the bottom as the old one leaves via the top
         transitionSpec = { slideInVertically { it } togetherWith slideOutVertically { -it } }
     )
 }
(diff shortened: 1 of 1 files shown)


demolaf:
Superseded by #2496, which absorbs both commits from this branch with authorship intact and adds the `context` line the rest of that PR sets everywhere.

Closing here so the two stop conflicting: #2496 rewrites §AuthFlowController and §Low-Level API in the same file, and both PRs edited §Screen Transitions from the same base.

Nothing from this branch is lost - the five samples, the per-destination `authRoute()` example and the fallback note all carry over unchanged, including the vertical-slide exit-direction fix in a86f6937.

PR #2459, opened 2026-08-28
Version 10.0.0-beta05

russellwheatley:
## ⚠️ Breaking Changes
- feat(auth): reshape reauthContent into a ReauthContentState content slot- #2452
- feat(auth)!: make each email sign-in mode a real navigation destination - #2466
- feat(auth)!: migrate auth navigation to Navigation 3 - #2469
- build(deps)!: upgrade AGP to 9.3.0 and bump compileSdk to 37- #2464

## Changes
- fix(auth)!: remove non-functional MFA recovery codes - #2457
- refactor(auth): scope phone verification loading state to the composition that owns it - #2454
- fix(e2e): fix flaky Google credential-linking test- #2455
- fix(auth): complete MFA challenge sign-in navigation and success callback- #2458
- feat(auth): add stable resource ids for auth input fields- #2449
- test(auth): pin onComplete and factor refresh on successful MFA enrollment - #2462
- feat(auth): make each MFA enrollment step a real navigation destination - #2467
- feat(auth): make reauthentication a navigation destination - #2471
- fix(auth): clear MFA enrollment flow state on every entry- #2473
- feat(auth): make each phone auth step a real navigation destination- #2472
- fix(auth): don't strand phone-only users on email verification after reload- #2474
- test(auth): cover the write-through recoveries and the custom reauth slot's sheet chrome- #2476

PR #2478, opened 2026-09-04
refactor(auth)!: decouple reauthentication from the FirebaseAuthUI shared state flow

demolaf:
`FirebaseAuthUI` was the SDK entry point, the process-wide `AuthState` channel and the reauthentication session store all at once. Reauthentication requests now travel on their own channel, provider code emits to an `AuthFlowScope` sink instead of the singleton's flow, and the phase machine lives in a composition-scoped holder — nine members deleted from `FirebaseAuthUI`, none added.

`withReauth` no longer retains the caller's lambda on a process-scoped singleton: it raises a request, suspends on a resolver, and retries on the caller's own coroutine. `delete()` goes through the same path instead of publishing `Required` and throwing.

- `AuthFlowScope.kt`: provider extensions move off `FirebaseAuthUI`, so provider code can no longer reach the public state flow at all — the isolation is compiler-held rather than review-held.
- `ReauthFlowState.kt`: the phase machine, scoped to one `FirebaseAuthScreen`'s composition. `Aborted` now clears the phase and resolves the caller under both hosts.
- Added `SignInStateSequenceTest` and `ReauthFlowStateTest`, plus emulator-backed e2e coverage for a password change and an account deletion through reauthentication.

## ⚠️ Breaking Changes

- `AuthState.Reauthentication.Required.retryOperation` is removed outright — pre-GA, zero external usage, no shim.
- `AuthState.Reauthentication.Required`'s `(user, reason)` constructor is now `internal`. Raising a request is `withReauth`/`delete()`'s job; constructing one yourself no longer did anything.
- A declined reauthentication throws `AuthException.AuthCancelledException` instead of returning quietly, and `delete()` reports that rather than `InvalidCredentialsException`.
- Publishing `AuthState.Reauthentication.Required` to `authStateFlow()` yourself no longer raises a request; only `withReauth` and `delete()` do.
- Rotation during the reauth sheet abandons the pending operation unless the caller launched from a scope that survives it.

---
Maintainer note: Fixes internal CPRN-413

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUI.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUI.kt
@@ -480,9 +480,9 @@ class FirebaseAuthUI private constructor(
      * and presents a reauthentication sheet; once credentials are accepted the [operation] runs
      * again on this same coroutine, so nothing about the caller is retained by the library.
      *
-     * If the user backs out, or this coroutine's scope is cancelled while the sheet is up, the
-     * operation is not retried. A caller that must survive Activity recreation should launch from
-     * a scope that does too.
+     * If the user backs out, this throws [AuthException.AuthCancelledException] and the operation
+     * is not retried — so a caller can always tell a decline from a completed operation. A caller
+     * that must survive Activity recreation should launch from a scope that does too.
      *
      * All other exceptions propagate normally.
      *
@@ -498,6 +498,7 @@ class FirebaseAuthUI private constructor(
      * @param context Android [Context]
      * @param reason Optional message shown to the user explaining why reauthentication is needed
      * @param operation The sensitive operation to attempt
+     * @throws AuthException.AuthCancelledException if the user declines reauthentication
      * @since 10.0.0
      */
     suspend fun withReauth(
@@ -524,7 +525,15 @@ class FirebaseAuthUI private constructor(
                     )
                 )
             )
-            if (resolver.await()) operation()
+            // Thrown from this frame rather than out of the resolver: the resolver is parented to
+            // the caller's job, so failing it would cancel the caller's whole scope instead of
+            // just this call. A caller always learns whether its operation ran.
+            if (!resolver.await()) {
+                throw AuthException.AuthCancelledException(
+                    message = "Reauthentication was cancelled"
+                )
+            }
+            operation()
         }
     }
 
@@ -554,6 +563,11 @@ class FirebaseAuthUI private constructor(
                 // The user is deleted and therefore signed out.
                 updateAuthState(AuthState.Idle)
             }
+        } catch (e: AuthException.AuthCancelledException) {
+            // The user declined the reauthentication. The screen already published the terminal
+            // state for that, so republishing it as an Error would put a dialog over a flow the
+            // user deliberately left.
+            throw e
         } catch (e: CancellationException) {
             // Handle coroutine cancellation
             val cancelledException = AuthException.AuthCancelledException(
--- a/auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUI.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUI.kt
@@ -20,10 +20,10 @@ import androidx.annotation.MainThread
 import androidx.annotation.RestrictTo
 import com.firebase.ui.auth.configuration.AuthUIConfiguration
 import com.firebase.ui.auth.configuration.auth_provider.AuthProvider
-import com.firebase.ui.auth.configuration.auth_provider.filterToLinkedProviders
 import com.google.firebase.auth.FirebaseAuthRecentLoginRequiredException
 import com.firebase.ui.auth.configuration.auth_provider.signOutFromFacebook
 import com.firebase.ui.auth.configuration.auth_provider.signOutFromGoogle
+import com.firebase.ui.auth.ui.screens.reauth.toReauthConfiguration
 import com.google.firebase.Firebase
 import com.google.firebase.FirebaseApp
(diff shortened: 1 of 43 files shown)


auth/src/main/java/com/firebase/ui/auth/ui/screens/FirebaseAuthScreen.kt
```diff
@@ -833,10 +717,111 @@ fun FirebaseAuthScreen(
...
+            }
+
+            // Takes on the request waiting on pendingReauth; a recreated screen re-runs it.
+            LaunchedEffect(pendingReauth) {
+                val required = pendingReauth ?: return@LaunchedEffect
+                if (reauthFlowState.phase?.requestId == required.requestId) return@LaunchedEffect
+
+                val reauthConfiguration = configuration.toReauthConfiguration(required.user)
+                if (reauthConfiguration == null) {
+                    finishReauth(
```
russellwheatley: I think this leaves the `withReauth` caller suspended forever when there are no linked providers.

`finishReauth(_, false)` calls `reauthFlowState.finish(false)`, which does `phaseState.value?.request?.decline()`. But `reauthFlowState.accept(required)` doesn't run until line 763, below this early return, so `phaseState.value` is still null here — `decline()` becomes a no-op on `null` and the `CompletableDeferred` behind `required.request` never completes. The coroutine parked on `resolver.await()` in `FirebaseAuthUI.withReauth()` just hangs.

The `no linked providers leaves no request outstanding` test in `FirebaseAuthScreenReauthContentStateTest.kt` only asserts `retryRan == false`, which is true whether the resolver was declined or just stuck, so it wouldn't catch this.

Might be worth either calling `reauthFlowState.accept(required)` before `finishReauth` here, or declining `required.request` directly instead of routing through `reauthFlowState.finish()`.

auth/src/main/java/com/firebase/ui/auth/FirebaseAuthUI.kt
```diff
@@ -697,60 +507,68 @@ class FirebaseAuthUI private constructor(
...
+                )
+            }
+            // The screen handed over on a loading state, so it goes however the retry ends.
+            try {
+                operation()
+            } finally {
+                updateAuthState(
+                    auth.currentUser?.let { authUserState(it, result = null, isNewUser = false) }
+                        ?: AuthState.Idle
+                )
```
russellwheatley: If the retried `operation()` throws here (not a reauth exception, just a normal failure), this `finally` still runs and publishes `Success`/`RequiresEmailVerification` for the still-signed-in user, before the real `Error` gets published by `delete()`'s outer catch a moment later. Anything collecting every emission off `authStateFlow()` sees a flash of success on a failed retry.

Not sure but publishing `AuthState.Idle` unconditionally instead of re-deriving the user's state might avoid it, or just not touching `_authStateFlow` here at all and letting the outer catch own the terminal state.

auth/src/test/java/com/firebase/ui/auth/ui/screens/reauth/ReauthSurfaceGateTest.kt
```diff
@@ -68,7 +70,7 @@ import org.robolectric.annotation.Config
 /**
  * The reauthentication surface exists on exactly one condition: a resolved [ReauthSurface].
  *
- * A saved back stack can carry an [AuthRoute.Reauth] entry into a process with no armed request —
+ * A saved back stack can carry an [AuthRoute.Reauth] entry into a process with no outstanding request —
  * the state machine publishes `Interrupted` and pops it, but the entry composes first. This pins
```
russellwheatley: nit: `Interrupted` doesn't exist anymore, looks like a leftover from before this PR renamed things.

russellwheatley (commented):
Went through this end to end (checked out the branch, read the actual files rather than just the diff, since a phase-machine/coroutine refactor like this is easy to get wrong on diff alone).

The decoupling itself looks solid: `AuthFlowScope` genuinely walls provider code off from the old singleton flow, all six provider files migrated consistently, and the nine deleted `FirebaseAuthUI` members have no leftover references anywhere.

Found one hang I think is real (see inline) plus a smaller state-coherence issue on failed retries, and a couple of stale doc comments left over from the refactor. Details inline.

auth/src/test/java/com/firebase/ui/auth/ui/screens/reauth/ReauthSurfaceGateTest.kt
```diff
@@ -243,11 +245,11 @@ class ReauthSurfaceGateTest {
...
      *
-     * @param entryRequestId The id the reauthentication entry is keyed to. Defaults to the armed
-     * request's, which is what the host's steady state looks like.
-     * @param armedRequest When given, the stack is re-armed from it in a `LaunchedEffect`, the way
+     * @param entryRequestId The id the reauthentication entry is keyed to. Defaults to the
+     * outstanding request's, which is what the host's steady state looks like.
+     * @param presentedRequest When given, the stack is rebuilt from it in a `LaunchedEffect`, the way
      * the host does — which is what puts a composition between a new request and its entry.
      * @param useSlot Installs a `reauthContent` slot that records what the entry hands it. The
      * slot is the only path to the entry's `updateReauthentication` writes, so nothing recorded
```
russellwheatley: nit: same here, `updateReauthentication` was removed from `FirebaseAuthUI` in this PR.

russellwheatley (approved):
LGTM

demolaf:
This was not merged into `master` deliberately. GitHub retargeted it from `version-10.0.0-beta05` to `master` when #2459 was merged by mistake, so it landed on top of that merge. Reverted from `master` in #2487 — these commits now sit on `version-10.0.0-beta05` and ship as part of #2486.

PR #2480, opened 2026-09-05
feat(auth)!: require the multi-step auth screens to be given navigation

demolaf:
`EmailAuthScreen`, `PhoneAuthScreen` and `MfaEnrollmentScreen` took their navigation parameters as nullable and fell back to internal `localMode`/`localStep` state when they were omitted. That fallback had no back stack and no `BackHandler`, so system back did nothing and the in-UI switch controls were the only way to move between steps — while the KDoc described those same parameters as a supported way to drive the screens from outside.

The parameters are now required and the fallback is gone, so every caller brings real navigation. Nothing new was published to support that: both slot demos were rewritten to host the screens on their own Navigation 3 back stacks using only existing public API, which is what makes the documented capability real rather than implied.

- `EmailAuthSlotDemoActivity.kt` / `PhoneAuthSlotDemoActivity.kt`: each defines its own `NavKey` and back stack, so system back steps between modes/steps instead of leaving the flow. Verified on device.
- Un-ignored and de-flaked all four `@Ignore`d e2e tests, including `change phone number navigates back to EnterPhoneNumber step` — the back path this change reworks, which had no coverage before.

## ⚠️ Breaking Changes

- `EmailAuthScreen`: `mode` and `onNavigateToMode` are now required.
- `PhoneAuthScreen` and `MfaEnrollmentScreen`: `step`, `onNavigateToStep`, `onNavigateBack` and `flowState` are now required.
- `EmailAuthScreen` no longer shows an error dialog of its own. This only affected a caller that installed `LocalTopLevelDialogController` itself; every caller still receives `onError`.

---
Maintainer note: Fixes internal CPRN-415

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/mfa/MfaEnrollmentContentState.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/mfa/MfaEnrollmentContentState.kt
@@ -28,7 +28,20 @@ import com.google.firebase.auth.MultiFactorInfo
  * Use a `when` expression on [step] to determine which UI to render:
  *
  * ```kotlin
- * MfaEnrollmentScreen(user, config, onComplete, onSkip) { state ->
+ * MfaEnrollmentScreen(
+ *     user = user,
+ *     auth = auth,
+ *     configuration = config,
+ *     onComplete = onComplete,
+ *     onSkip = onSkip,
+ *     // The host owns the step: every one of them is its own navigation destination. Both writes
+ *     // are guarded — a step already on top must not be pushed twice, and the first step must not
+ *     // be popped, because NavDisplay throws on an empty back stack from recomposition.
+ *     step = step,
+ *     onNavigateToStep = { if (backStack.lastOrNull() != it) backStack.add(it) },
+ *     onNavigateBack = { if (backStack.size > 1) backStack.removeLastOrNull() },
+ *     flowState = flowState,
+ * ) { state ->
  *     when (state.step) {
  *         MfaEnrollmentStep.SelectFactor -> {
  *             // Render factor selection UI using state.availableFactors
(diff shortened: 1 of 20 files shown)


auth/src/main/java/com/firebase/ui/auth/mfa/MfaEnrollmentContentState.kt
```diff
@@ -28,7 +28,18 @@ import com.google.firebase.auth.MultiFactorInfo
...
+ * MfaEnrollmentScreen(
+ *     user = user,
+ *     auth = auth,
+ *     configuration = config,
+ *     onComplete = onComplete,
+ *     onSkip = onSkip,
+ *     // The host owns the step: every one of them is its own navigation destination.
+ *     step = step,
+ *     onNavigateToStep = { backStack.add(it) },
+ *     onNavigateBack = { backStack.removeLastOrNull() },
```
russellwheatley: This is the sample people will copy, and it uses the write `popOrNull` exists to forbid. `NavDisplay` throws on an empty stack from recomposition, so in-UI back on the first step is a crash. Unguarded `add` is also the resend-stacks-a-second-code-entry bug the phone demo already hit.

Can we match the library guard here? Don't add if it's already on top, and don't pop the last entry. Same pattern in both slot demos' `NavDisplay.onBack`.

russellwheatley (commented):
The core change looks right to me, and CI is green. A couple of things before I'd approve.

`docs/upgrade-to-10.0.md` still shows `EmailAuthScreen` without `mode` / `onNavigateToMode`. That sample compiled before this PR and will not after. `auth/README.md` MFA samples were already stale (`onEnrollmentComplete`, missing `auth`) and now also miss the required nav params.

I'd want those samples (and the MFA KDoc one inline) to match the new contract before this goes in. After that this is an approve from me.

russellwheatley (approved):

PR #2483, opened 2026-09-06
feat(auth)!: decouple SMS multi-factor configuration from the phone sign-in provider

demolaf:
The SMS enrolment step read its country restriction off the hosting configuration's `AuthProvider.Phone`, and `EnterPhoneNumberUI` took `.first()` of that list, so opening SMS enrolment from a configuration without a phone provider crashed with `NoSuchElementException`. That is the ordinary case rather than an edge case — Firebase enables SMS second factors separately from phone sign-in, and phone sign-in cannot carry a second factor at all.

The restriction now comes from `MfaConfiguration.allowedCountries` and `EnterPhoneNumberUI` takes it as a parameter, so the step no longer resolves a provider at all. That option takes ISO 3166-1 alpha-2 codes and validates them — a dial code throws rather than silently restricting the selector to nothing — and the step opens on a permitted country instead of the device's own when the two disagree. `AuthProvider.Phone.allowedCountries` is unchanged and still drives phone sign-in. `AuthProvider.Phone.smsCodeLength` is removed — nothing read it, both flows validate against a hardcoded 6, and Firebase exposes no SMS code length setting for it to mirror.

Also adds the first genuine end-to-end MFA coverage: `MfaSmsFlowTest` enrols an SMS factor and completes an MFA challenge against the Auth emulator, asserting on `user.multiFactor.enrolledFactors` and `AuthState.Success` rather than on screen state. The emulator supports SMS MFA in full — the two mocked MFA test files claimed it could not, and are corrected, TOTP's exact failure included. `MfaEnrollmentSmsStepTest` covers the crash, verified failing on the old code.

## ⚠️ Breaking Changes

- `AuthProvider.Phone.smsCodeLength` is removed. Nothing read it, so nothing behaves differently without it.
- `EnterPhoneNumberUI` gained a **required** `allowedCountries` parameter and no longer resolves a phone provider from the configuration. Callers stop compiling until they supply it, deliberately: with a default, a host driving this step from a custom `phoneAuthContent` would have silently lost the restriction it used to get implicitly. Derive it from your own provider — `configuration.providers.filterIsInstance<AuthProvider.Phone>().firstOrNull()?.allowedCountries?.toSet()` — or pass `null` for no restriction.

## Usage

```kotlin
FirebaseAuthScreen(
    configuration = configuration,
    mfaConfiguration = MfaConfiguration(
        allowedFactors = listOf(MfaFactor.Sms),
        allowedCountries = listOf("GB", "IE"),
    ),
    onSignInSuccess = { },
    onSignInFailure = { },
    onSignInCancelled = { },
)
```

---
Maintainer note: Fixes internal CPRN-423

### DIFF
--- a/auth/README.md
+++ b/auth/README.md
@@ -1114,10 +1114,14 @@ fun MfaEnrollmentFlow() {
 
     if (currentUser != null) {
         val mfaConfig = MfaConfiguration(
-            allowedFactors = listOf(MfaFactor.Sms, MfaFactor.Totp)
+            allowedFactors = listOf(MfaFactor.Sms, MfaFactor.Totp),
+            allowedCountries = listOf("US", "CA", "GB")
         )
         val backStack = rememberNavBackStack(MfaStepKey(MfaEnrollmentStep.SelectFactor))
-        val flowState = rememberMfaEnrollmentFlowState()
+        // Pass the restriction so the SMS step opens on a country the selector will offer. The
+        // screen also reconciles this itself, so a host that forgets cannot end up sending to an
+        // unpermitted dial code.
+        val flowState = rememberMfaEnrollmentFlowState(mfaConfig.allowedCountries)
 
         NavDisplay(
             backStack = backStack,
--- a/auth/README.md
+++ b/auth/README.md
@@ -455,7 +455,6 @@ val phoneProvider = AuthProvider.Phone(
     allowedCountries = listOf("US", "CA", "GB"),
 
     // Optional: SMS code length (default: 6)
-    smsCodeLength = 6,
 
     // Optional: Timeout for SMS delivery in seconds (default: 60)
     timeout = 60L,
--- a/auth/README.md
+++ b/auth/README.md
@@ -454,8 +454,6 @@ val phoneProvider = AuthProvider.Phone(
     // Optional: Allowed countries
     allowedCountries = listOf("US", "CA", "GB"),
 
-    // Optional: SMS code length (default: 6)
-
     // Optional: Timeout for SMS delivery in seconds (default: 60)
     timeout = 60L,
 
@@ -1081,7 +1079,12 @@ val mfaConfig = MfaConfiguration(
     allowedFactors = listOf(MfaFactor.Sms, MfaFactor.Totp),
 
     // Optional: Require MFA enrollment (default: false)
-    requireEnrollment = false
+    requireEnrollment = false,
+
+    // Optional: restrict the SMS enrollment step's country selector, as ISO 3166-1 alpha-2
+    // codes (default: null, no restriction). Independent of the phone sign-in provider's own
+    // allowedCountries — an SMS second factor is configured separately from phone sign-in.
+    allowedCountries = listOf("US", "CA", "GB")
 )
 
 val configuration = authUIConfiguration {
--- a/auth/src/main/java/com/firebase/ui/auth/ui/screens/FirebaseAuthScreen.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/ui/screens/FirebaseAuthScreen.kt
@@ -197,7 +197,7 @@ fun FirebaseAuthScreen(
     val lastSuccessfulUserId = remember { mutableStateOf<String?>(null) }
     val pendingLinkingCredential = remember { mutableStateOf<AuthCredential?>(null) }
     val pendingResolver = remember { mutableStateOf<MultiFactorResolver?>(null) }
-    val mfaEnrollmentFlowState = rememberMfaEnrollmentFlowState()
+    val mfaEnrollmentFlowState = rememberMfaEnrollmentFlowState(mfaConfiguration.allowedCountries)
     val phoneAuthFlowState = rememberPhoneAuthFlowState(configuration)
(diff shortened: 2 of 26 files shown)


auth/src/main/java/com/firebase/ui/auth/ui/screens/mfa/MfaEnrollmentDestinations.kt
```diff
@@ -117,8 +118,25 @@ class MfaEnrollmentFlowState internal constructor(
...
+    val permitted = CountryUtils.filterByAllowedCountries(allowedCountries.toSet())
+    return permitted.firstOrNull { it.countryCode == deviceCountry.countryCode }
+        ?: permitted.firstOrNull()
+        ?: deviceCountry
+}
+
 @Composable
-fun rememberMfaEnrollmentFlowState(): MfaEnrollmentFlowState {
+fun rememberMfaEnrollmentFlowState(
+    allowedCountries: List<String>? = null,
```
russellwheatley, line 137: Defaulting this to `null` is what lets a host set `MfaConfiguration.allowedCountries` and still open on the device country. `FirebaseAuthScreen` passes the list. The README custom-host example does not.

I'd rather `MfaEnrollmentScreen` snap `selectedCountry` to a permitted country on entry, so forgetting this argument cannot bring the mismatch back. Either way, `initialEnrollmentCountry` needs a test that the pre-selected dial code is one of the allowed countries.

auth/src/main/java/com/firebase/ui/auth/ui/screens/mfa/MfaEnrollmentDestinations.kt
```diff
@@ -117,8 +118,25 @@ class MfaEnrollmentFlowState internal constructor(
  * with. Called once, above the `NavDisplay`, so the same instance is handed to every step — see
  * [MfaEnrollmentFlowState] for which of its fields actually survive Activity recreation.
  */
+/**
```
russellwheatley, line 116: This KDoc is now stacked under the `rememberMfaEnrollmentFlowState` block, which no longer sits on that function. Move the first block back onto `rememberMfaEnrollmentFlowState`.

auth/src/main/java/com/firebase/ui/auth/ui/screens/FirebaseAuthScreen.kt
```diff
@@ -196,7 +196,7 @@ fun FirebaseAuthScreen(
     val lastSuccessfulUserId = remember { mutableStateOf<String?>(null) }
     val pendingLinkingCredential = remember { mutableStateOf<AuthCredential?>(null) }
     val pendingResolver = remember { mutableStateOf<MultiFactorResolver?>(null) }
-    val mfaEnrollmentFlowState = rememberMfaEnrollmentFlowState()
+    val mfaEnrollmentFlowState = rememberMfaEnrollmentFlowState(mfaConfiguration.allowedCountries)
```
russellwheatley, line 200: This is the right wiring. The documented `MfaEnrollmentScreen` host path still calls `rememberMfaEnrollmentFlowState()` with no args, so the same mismatch is live there.

auth/README.md
```diff
@@ -1082,7 +1079,12 @@ val mfaConfig = MfaConfiguration(
     allowedFactors = listOf(MfaFactor.Sms, MfaFactor.Totp),
 
     // Optional: Require MFA enrollment (default: false)
-    requireEnrollment = false
+    requireEnrollment = false,
+
+    // Optional: restrict the SMS enrollment step's country selector, as ISO 3166-1 alpha-2
+    // codes (default: null, no restriction). Independent of the phone sign-in provider's own
+    // allowedCountries — an SMS second factor is configured separately from phone sign-in.
+    allowedCountries = listOf("US", "CA", "GB")
```
russellwheatley, line 1087: This documents the restriction on `MfaConfiguration`. The custom navigation example a few lines down still does `rememberMfaEnrollmentFlowState()` with no args, so a host who copies both snippets gets a filtered picker and a device-country dial code.

russellwheatley (changes requested):
The crash fix and the MFA/phone split are right. Requesting changes because the custom-host path can still open on a country the selector will not offer.

`FirebaseAuthScreen` now passes `mfaConfiguration.allowedCountries` into `rememberMfaEnrollmentFlowState`. Anyone driving `MfaEnrollmentScreen` themselves, including the README example, still calls `rememberMfaEnrollmentFlowState()` with the default `null`. The picker is filtered. Send uses `selectedCountry.value.dialCode` with no check. Same mismatch the fifth commit fixed for the high-level screen.

I'd reconcile inside `MfaEnrollmentScreen` so forgetting the second argument cannot bring this back. At minimum, pass the same list in the README example, and add a test that the pre-selected dial code is one of the allowed countries. `initialEnrollmentCountry` has none.

The old `rememberMfaEnrollmentFlowState` KDoc is now sitting above `initialEnrollmentCountry`. Move it back.

Android CI is green on this HEAD. e2e last ran on an older SHA, so worth re-triggering that job. Gemini's `remember` comments are already done.

russellwheatley (approved):

PR #2493, opened 2026-09-08
docs(auth): correct the auth KDoc against the current API

demolaf:
The auth KDoc had drifted from the API it documents. Thirteen internal functions still documented a `config` parameter that moved onto `AuthFlowScope`, `AuthState.PhoneNumberVerificationRequired` pointed at two `FirebaseAuthUI` methods that don't exist and an `isForceResendingTokenEnabled` flag that appears nowhere in the repo, `AuthFlowController.dispose()` claimed an `IllegalStateException` it can't throw, and three `authUIConfiguration {}` examples assigned `providers = listOf(...)`, which the builder's private setter rejects.

Fourteen internal helpers also carried 30-114 line KDocs whose examples called them on the wrong receiver, with the removed `config` parameter, and named types like `GoogleAuthCredential` that don't exist. Those are cut back to the claim. The one non-obvious behaviour worth keeping survives: `verifyPhoneNumber` doesn't return when the code is sent, so a host already on code entry still has to handle a late `SMSAutoVerified`.

Comments only - 91 insertions, 927 deletions, nothing changed outside a KDoc block. `:auth` unit suite green (78 classes, 985 tests) and checkstyle clean.

---

### DIFF
--- a/auth/src/main/java/com/firebase/ui/auth/AuthFlowController.kt
+++ b/auth/src/main/java/com/firebase/ui/auth/AuthFlowController.kt
@@ -52,10 +52,11 @@ import java.util.concurrent.atomic.AtomicBoolean
  *
  *         val authUI = FirebaseAuthUI.getInstance()
  *         val configuration = authUIConfiguration {
- *             providers = listOf(
- *                 AuthProvider.Email(),
- *                 AuthProvider.Google(...)
- *             )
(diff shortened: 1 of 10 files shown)


russellwheatley (approved):
LGTM. dispose() no longer claims an exception it can't throw, and the provider examples use the real DSL.

PR #2494, opened 2026-09-08
sample(app): drop the comments that restate the code in the auth demos

demolaf:
Every `//` comment in `AuthFlowControllerDemoActivity` and `AuthChooserActivity` was a label restating the line beneath it - `// Initialize FirebaseAuthUI` above `FirebaseAuthUI.getInstance()`, `// Clean up resources` above `dispose()`, `// Header` above a header. Eighteen of them, and not one explaining why anything was done. The newer demos (`EmailAuthSlotDemoActivity`, `PhoneAuthSlotDemoActivity`, `HighLevelApiDemoActivity`) all carry a reason instead, so these two were the odd ones out.

The class KDoc also used a `This demonstrates:` bullet list where every other demo uses prose. It now says what is actually distinctive about the Activity route: the flow runs in its own Activity, so the result arrives through an `ActivityResultLauncher` rather than a callback, and the controller owns a coroutine scope nothing in the composition will clean up.

Comments only. `:app:assembleDebug` and `checkstyle` both clean.

---

### DIFF
--- a/app/src/main/java/com/firebaseui/android/demo/auth/AuthChooserActivity.kt
+++ b/app/src/main/java/com/firebaseui/android/demo/auth/AuthChooserActivity.kt
@@ -78,6 +78,7 @@ fun AuthChooserScreen(
         verticalArrangement = Arrangement.spacedBy(24.dp)
     ) {
         Spacer(modifier = Modifier.height(16.dp))
+
         Text(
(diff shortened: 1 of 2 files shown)


russellwheatley (approved):
LGTM. The controller demo KDoc actually explains why that Activity looks the way it does.

PR #2495, opened 2026-09-08
docs(auth): correct the 10.0 upgrade guide against the current API

demolaf:
The 9.x-to-10.x guide had drifted from the API it documents, and the Low-Level API section was the worst of it: it showed `val state = controller.start()` returning an `AuthState` from a coroutine. No such method has ever existed - the real `start()` takes an Activity and a request code, returns `Unit`, and is deprecated. That section now registers an `ActivityResultLauncher` and launches `controller.createIntent(this)`, which is what `AuthFlowControllerDemoActivity` already does.

Every `authUIConfiguration {}` example also assigned `providers = listOf(...)`, which the builder rejects - its setter is private and the only entry point is the `providers { }` block. And none of the six config examples set `context`, which `build()` requires via `requireNotNull`, so each one threw at runtime as written.

Versions were inconsistent three ways: the dependency snippet said `beta02`, its own checklist said `beta01`, and the published version is `beta04`. The Compose BOM was pinned to `2024.01.00` against `2026.06.01` in `gradle/libs.versions.toml`.

Markdown only.

---

### DIFF
--- a/docs/upgrade-to-10.0.md
+++ b/docs/upgrade-to-10.0.md
@@ -50,10 +50,10 @@ dependencies {
     // FirebaseUI Auth
     // Check Maven Central for the latest version:
     // https://central.sonatype.com/artifact/com.firebaseui/firebase-ui-auth/versions
-    implementation("com.firebaseui:firebase-ui-auth:10.0.0-beta02")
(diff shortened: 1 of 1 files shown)


russellwheatley (approved):
LGTM. The start()/createIntent path matches AuthFlowController, and the builder snippets now set context and use the providers block.

PR #2496, opened 2026-09-08
docs(auth): correct the auth README against the current API

demolaf:
The auth README documented an API in several places that never existed. `§AuthFlowController` and `§Low-Level API` both showed `val state = controller.start()` returning an `AuthState` from a coroutine - the real `start()` takes an Activity and a request code, returns `Unit`, and is deprecated. Both now register an `ActivityResultLauncher` and launch `controller.createIntent(this)`, matching `AuthFlowControllerDemoActivity`. `state.result.user` became `state.user`, since `AuthState.Success.result` is nullable and `user` is the field being reached for.

Smaller corrections in the same vein: `MfaChallengeScreen` was passing a non-existent `onChallengeComplete` while omitting the required `auth` and `onSuccess`; `ErrorRecoveryDialog` was missing the required `stringProvider`; `FirebaseAuthUI.create` was missing the required `app`; two `providers = listOf(...)` assignments became the `providers { }` block, whose setter is private; and `context` is now set in all 40 `authUIConfiguration {}` examples - the builder requires it and throws without it, and only 3 set it.

`isCredentialLinkingEnabled`, `passwordResetActionCodeSettings` and `legacyFetchSignInWithEmail` are all read by the library but were documented nowhere, so they are now in the configuration reference. Minimum requirements said SDK 21 and Kotlin 1.9 against `minSdk = 23` and Kotlin 2.4, and the version checklist and Compose BOM were stale.

`§Migration Guide` is replaced by a link to `docs/upgrade-to-10.0.md`. It restated that file's own key-changes list and checklist, the two had drifted apart, and the OKF documentation policy asks for one owning doc per topic.

`§Screen Transitions` is #2475 absorbed into this PR rather than rewritten here - its two commits are cherry-picked with authorship intact, so the samples are the ones already compiled verbatim against the module. #2475 is closed in favour of this one. The only addition is the `context` line, for consistency with the rest of the file.

---
Maintainer note: Fixes internal CPRN-404

### DIFF
--- a/auth/README.md
+++ b/auth/README.md
@@ -1703,6 +1703,7 @@ import androidx.compose.animation.togetherWith
 import com.firebase.ui.auth.configuration.AuthUITransitions
 
 val configuration = authUIConfiguration {
+    context = applicationContext
     providers {
         provider(AuthProvider.Email())
         provider(AuthProvider.Google())
@@ -1729,6 +1730,7 @@ import androidx.compose.animation.togetherWith
 import com.firebase.ui.auth.configuration.AuthUITransitions
 
 val configuration = authUIConfiguration {
+    context = applicationContext
     providers {
         provider(AuthProvider.Phone())
     }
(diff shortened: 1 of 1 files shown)


Chewypewy (approved):

russellwheatley (approved):
LGTM. The fake controller.start() return value is gone, and the migration section now points at the upgrade guide instead of drifting from it.

